PR #4, opened 2016-08-11
DS-923: As a LU I don't want to see activities regarding comments on content in group multiple times in stream (aggregation)

ribel:
# Changes
- Added "Aggregate activity" setting for message templates
- Updated comment message templates with aggregated activity settings
- Added published status filter to activity views
- Added aggregation of comments
- Added behat test for comments aggregation in activity stream

# HTT

- [x] Checkout the branch and reinstall the site
- [x] Go to comments message templates and check that aggregation is enabled and there are two values for output text.
- [x] Post comments on post in the group from two different accounts (two comments from each account)
- [x] Check that on your profile you see only activity with latest comment and message _User commented on post in the group_
- [x] Check that on home and group stream you see only activity with latest comment and message _User and 1 others commented on post in the group_
- [x] Repeat same steps for content in group


modules/custom/activity_basics/config/install/message.template.create_comment_community_node.yml
```diff
@@ -12,11 +12,13 @@ third_party_settings:
...
       stream_profile: stream_profile
     activity_create_direct: 1
+    activity_aggregate: 1
 template: create_comment_community_node
 label: 'Create comment on node in the community'
 description: 'A user add a comment to content in the community'
 text:
-  - "<p><a href=\"[message:author:url:relative]\">[message:author:display-name]</a> commented on content</p>"
+  - "<p><a href=\"relative]\">[message:author:display-name]</a> commented on content</p>\r\n"
+  - "<p><a href=\"relative]\">[message:author:display-name]</a> and @count others commented on content</p>\r\n"
```
ronaldtebrake: Shouldn't this also be: <a href=\"[message:author:url:relative]\">

modules/custom/activity_creator/src/ActivityFactory.php
```diff
@@ -103,11 +118,19 @@ private function getFieldMessage($data) {
...
    */
-  private function getFieldOutputText(Message $message) {
+  private function getFieldOutputText(Message $message, $arguments = []) {
     $value = NULL;
     if (isset($message)) {
       $value = $message->getText(NULL);
-      $text = reset($value);
+      // Text for aggregated activities.
+      if (!empty($value[1])) {
+        $text = t($value[1], $arguments);
```
ronaldtebrake: will this t() actually work? 

https://www.drupal.org/developing/api/8/localization 
Chapter: Variables in user interface text

Seems like we need literal strings there coming from yml files, instead of message::load and the count parsed in it.

ronaldtebrake:
Hi Taras,

Couple of remarks:
1. See code remarks above
2. When I comment on content with willem wilmink, than arturo and than with willem again I see on the stream: "
Willem Wilmink and @count others commented on content in the group Local group Enschede"

But on the front stream it works correctly.
3. Also for demo content it seems to have "
Herman Brood and @count others commented on content in the group Open Social Developers"

Could you have a look at a fresh install and see if u can fix it?

modules/custom/activity_basics/config/install/message.template.create_comment_community_node.yml
ribel: Good catch! I will fix it.

ribel:
Hi @ronaldtebrake,
I have fixed issues you mentioned. About t() function - in this case it was used as better alternative of str_replace(). 
Can you check it again?

PR #6, opened 2016-08-15
Feature/ds 1737 main nav update

MaikelGG:
- [x] Explore has submenu items "all events, all groups, all topics, all users and community" 
 -- Should collapse in hamburger menu on mobile 
- [x] Move the link "home" from the user menu to the main-menu (first position)
 -- Should collapse in hamburger menu on mobile 
- [x] Add plus icon with link to add content (dialog page events/topics) to user menu (LU only)
- [x] check if search box has enough space on all screen widths.
- [x] On mobile subnav items must be expanded by default
- [x] search needs to be above explore on mobile
- [x] Hamburger needs be "menu" (text)


ribel:
Hi @MaikelGG,
Everything is ok, but i noticed two issues. Not sure that they are connected to this task, but they related to hamburger menu.
1) Explore not collapse correctly in hambrger menu. http://g.recordit.co/heYv8jNDeB.gif
2) Search is not work after switching to desktop mode http://g.recordit.co/8LnftPJba1.gif

MaikelGG:
@ribel Thanks for the feedback. 1). this should be fixed with the last commit, did you have this code when testing? The css property pointer-events:none should prevent you from clicking explore on mobile. 2). This is true, there is no event listener on a resize event. I am not sure if this is needed, if users will resize there screen that often. Adding such event listeners will slow down the site. If you can check 1. once more and if that works it's okay to merge for me.

PR #7, opened 2016-08-15
Feature/ds 969 theme attachments

MaikelGG:
- [x] Do a reinstall
- [x] Create an event and upload files
- [x] Upload an incorrect file type
- [x] Repeat steps for creating a topic
- [ ] View the event / topic and see that it looks themed.


MaikelGG:
@ribel Can you check why the feedback is printed twice when you try an upload which is not allowed?

PR #25, opened 2016-08-29
DS-1724 by nielsvandermolen: Added distinct option to view, probably …

NielsGG:
…could improve this by adding group by's by the joins in the visibility access filter

There was still an issue with multiple results in the activity stream. This is due the join on the node_access table that will return duplicate results when there are multiple matches in the join. There are ways to create a subquery in all the joins that groups them by id but I have chosen to go for a simpler solution for now and enable the distinct in the view. This could potentially lead to performance degradation which we should monitor when we have more data.

htt: Check that there are no double items in the activity stream


NielsGG:
Hi @jaapjan,

This story has a wrong story number it is a follow up from DS-1724 instead of DS-1847.

PR #28, opened 2016-08-29
DS-1718 by nielsvandermolen: Fixed behat test

NielsGG:


NielsGG:
merged without review

PR #43, opened 2016-09-01
DS-1496: Maintenance

ribel:


ronaldtebrake:
HI @ribel ,

Assigned it back to you for checking the following:
1. Views (search users) had broken/missing handlers. Due to the update in search api, they removed default filters for taxonomy search and placed it in a submodule. This way they do not have a dependency with Taxonomy unless people enabled the new submodule (but it broke ours :D )

PR #57, opened 2016-09-06
DS-1945: Make download_count module optional, override file fields config in social_download_count module

ribel:
# Changes
- Removed dependencies from topic, event, book, page features, set default file widget and public FS by default.
- Created new module `social_download_count` with just config overrides.

# HTT
- [x] Reinstall the site 
- [x] Log in as admin and enable `social_download_count` module
- [x] Log in as normal user
- [x] Create node with files
- [x] Download files and see that counter works


jaapjan:
This is pretty cool and works like a charm!

PR #64, opened 2016-09-09
DS-991: When a LU posts on my stream, I want to be notified via email

ribel:


modules/custom/activity_send/modules/activity_send_email/src/Plugin/ActivityDestination/EmailActivityDestination.php
```diff
@@ -40,4 +41,24 @@ public static function setSendEmailUserSettings($account, $values) {
...
 
+  /**
+   * Get field value for 'output_text' field from data array.
+   */
+  public static function getSendEmailOutputText(Message $message) {
+    $value = NULL;
+    if (isset($message)) {
+      $value = $message->getText(NULL);
+      // Text for email.
+      if (!empty($value[2])) {
```
jaapjan: I'm not sure if it will be clear to everyone that the 3rd value is used for e-mail templates. This might be something to improve upon later, or add description in form_alter in message_template form.

modules/custom/activity_send/modules/activity_send_email/src/Plugin/QueueWorker/ActivitySendEmailWorker.php
```diff
@@ -48,8 +48,7 @@ public function processItem($data) {
         if (!empty($user_email_settings[$message_template_id]) && isset($activity->field_activity_output_text)) {
           // Send Email
           $langcode = \Drupal::currentUser()->getPreferredLangcode();
-          $params['body'] = $activity->field_activity_output_text->value;
-          $params['target_account'] = $target_account;
+          $params['body'] = EmailActivityDestination::getSendEmailOutputText($message);
```
jaapjan: I think this is much cleaner than the previous solution!

modules/custom/activity_send/modules/activity_send_email/src/Plugin/ActivityDestination/EmailActivityDestination.php
ribel: Agree, that's a good point! I will add some description in next task about email notifications.

PR #71, opened 2016-09-13
Drupal 8.2.x

gnuget:
Drupal 8.2 is almost out, and I think open social can start taking advantage of the new features of this version.


jaapjan:
Hi @gnuget we should definitely update to 8.2 when this is available, but it should be a stable version. So closing the pullrequest for now, but once there is a stable release feel free to create a new pullrequest by then.

PR #79, opened 2016-09-16
DS-2070 by jaapjan: add any profile permission for sm, cm

jaapjan:
The permission is necessary to prevent wsod when creating a users profile as sm, cm.

### DIFF
--- a/modules/social_features/social_profile/social_profile.install
+++ b/modules/social_features/social_profile/social_profile.install
@@ -80,6 +80,7 @@ function _social_profile_get_permissions($role) {
 
   // Content manager.
   $permissions['contentmanager'] = array_merge($permissions['authenticated'], array(
+    'add any profile profile',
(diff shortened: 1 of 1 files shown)


jaapjan:
@ronaldtebrake I see the push build was failing: https://travis-ci.org/goalgorilla/open_social/builds/160403387 but this is the group-create-open.feature again. And the pr (most important) build passed. I think it can be safely merged.

ronaldtebrake:
Yo, moeten we deze niet nog even toevoegen in een update hookie?

jaapjan:
I'm still doubting what the exact procedure should be for this. I think we can set these permissions on install in the distro, but I dont like the idea of setting permissions for these user roles in an update hook since the roles may be removed by developers installing this distro or they may have adjusted the permissions for these roles already.

PR #76, opened 2016-09-15
DS-1034 by ribel: When a LU comments on a post, event or a topic created by me, I want to be notified via email

ribel:
# HTT
- [x] Pull code from this branch to active site installation.
- [x] Run updates, Clear cache and check that nothing is broken.
- [x] Find some user with generated contents and posts, go to his account edit, and enable _"A person commented on a post, topic or event created or organised by me"_.
- [x] Add some comments to nodes and posts of this user.
- [x] Check that activity streams (community, group, profile) works as before.
- [x] Run cron and check Mail, you should receive new email for each comment.
- [x] Reinstall site and do the same steps.
- [x] Check the code (see most important changes below)

# Changes
- Add ActivityEntityCondition plugin system
- Add ActivityEntityCondition plugins for Comments
- Add entity condition field to message template edit form
- Add titles to message template output text fields for clarity
- Move getDestination and getRecipient methods from ActivityAccessControlHandler to Activity class
- Add method isValidEntityCondition for ActivityEntityCondition plugins
- Rename method getActivityRelatedEntity in ActivityFactory class
- Add OwnerActivityContext plugin
- Add check for email destination when new activity is inserted
- Replace message template single value field activity_bundle_entity with multiple values activity_bundle_entities
- Add message template 'Create comment on authors node or post'
- Move config files from activity_basics module to social_activity feature

PR #91, opened 2016-09-20
DS-1038 by ribel: When a LU comments on a post posted on my stream, I want to be notified via e-mail

ribel:
# HTT
- [x] PR #76 should be merged first
- [x] Pull code from this branch to active site installation.
- [x] Clear cache, revert activity stream feature and check that nothing is broken.
- [x] Go to some account edit, and enable _'A person commented on the post posted on my profile'_.
- [x] Add post and comments to post of this user.
- [x] Check that activity streams (community, group, profile) works as before.
- [x] Run cron and check Mail, you should receive new email for each comment.
- [x] Check the code (see most important changes below)

# Changes
- Added message template 'Create comment on post on my profile'

### DIFF
--- a/modules/custom/activity_basics/src/Plugin/ActivityContext/ProfileActivityContext.php
+++ b/modules/custom/activity_basics/src/Plugin/ActivityContext/ProfileActivityContext.php
@@ -40,6 +40,11 @@ class ProfileActivityContext extends ActivityContextBase {
   }
 
   public function isValidEntity($entity) {
+    // Special cases for comments.
+    if ($entity->getEntityTypeId() === 'comment') {
+      // Returns the entity to which the comment is attached.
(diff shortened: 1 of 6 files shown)


NielsGG:
Hi @ribel,

Some questions:
- Should the email notifications be turned off by default?
- Now some notifications have only the email destination set. That will result in that they do not get notifications when someone 'comments on a post on the profile' when he is logged in. Is this intended or should we add the onscreen notifications to the message type?

ribel:
Hi @NielsGG ,
1) DS-1233 says that "By default it's turned off for new message types"
2) Good point, but we didn't have nothing mentioned about on screen notifications in acceptance criteria. And i am not sure if they work correctly, it also, could require aggregation, so i thin we can skip it for now.

PR #95, opened 2016-09-21
DS-1981 by ribel: As a LU, I want to search for events and users near a location

ribel:
# HTT
- [x] Reinstall your site first on 8.x-1.x branch, then checkout this branch and do a `drush updb -y; drush fra -y; drush cr` (to make sure it works for update path :) )
- [x] Login as normal user
- [x] Check that Search functionality works like before and nothing is broken
- [x] Login as admin
- [x] Enable `Social Geolocation` module
- [ ] Check that Search functionality works, but filters are missing
- [ ] Set Google Maps API key to Geolocation settings
- [ ] Add address for some event, profile and group.
- [ ] Check that Search functionality works and search filters works.
- [ ] Reinstall site in this branch and check that nothing is broken.
- [ ] Check the code

# Changes
- Move group proximity search to separate module - `social_geolocation`
- Move `navbar-search` script from `social_search` to `social_core`
- Add proximity search for events
- Add proximity search for user profiles


ronaldtebrake:
Hi Taras,

nice work :) couple of questions.

1. I like the warning message for users to uninstall the social_search, but i'm not entirely sure if we shouldn't just enforce it. When people install and enable social_geolocation social_search does nothing anymore and if they don't uninstall it, and we add logic in the future to the search feature.. it will keep on running on their installation, which could cause performance issues etc.
2. When a user enables social_geolocation, all the filters are removed form the search when there is no api key. Might be nicer to just remove the location filter (or was it discussed that everything should be removed?) now the user doesn't have access to the entire block, while the normal filters should be still accessible.  
3. Feature status doesn't look entirely clean, could you check that out?
4. social_geolocation.module has some commented out code, would you remove it?

ribel:
Hi @ronaldtebrake, thanks for review!

1. Yes, i also was thinking that when people install and enable `social_geolocation` module `social_search` does nothing anymore, and it can have performance impacts (because Search API indexing is still enabled), and that is the reason why i added this message. I tried to uninstall `social_search` in install hook of `social_geolocation`, but there were errors during installation/uninstallation. And i am not sure how to enforce it, maybe you know other ways?

2. Yes, we can do it, but i though that when user enable `social_geolocation` module he should set API key anyway, and without all filters we can enforce him to do it. And when he wants to use just normal filters he should use `social_search` module. What you think?

3. Yeah, i see it works well on clean install, but features marked as changed after upgrade, even after reverting. Will try to fix it.

4. I've tried to improve filters and add #states to content type filter - to show geolocation filters only when users select Events, but for some reason they didn't work, so i will remove them.

PR #102, opened 2016-09-26
DS-2089 by nielsvandermolen: Create a hook to alter the topic type view title

NielsGG:
This code is useful for enterprise projects.

HTT:

1. Implement the hook e.g. in social_core of the Social Topics API.

2. Check that the titles are altered on the newest topics view.

3. Check code

### DIFF
--- a/modules/social_features/social_topic/social_topic.api.php
+++ b/modules/social_features/social_topic/social_topic.api.php
@@ -1,0 +1,36 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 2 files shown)


ronaldtebrake:
Hi Niels, 

Works like a charm. Nice :)
Will you add the hook yourself for Y4P? 

Note that most likely they want the plural form of the topic term. Shouldn't be a problem though with the hook :).

PR #104, opened 2016-09-27
DS-320 Imagecropping

jochemvn:


drupal-org.make
```diff
@@ -28,6 +28,8 @@ projects[geocoder][version] = 2.0-alpha4
 projects[geolocation][type] = module
 projects[geolocation][version] = 1.7
 projects[geolocation][patch][] = "https://www.drupal.org/files/issues/geolocation-8.x-1.7-grouped-views-filters-2778227-5_0.patch"
+projects[image_widget_crop][type] = module
+projects[image_widget_crop][version] = 1.3
 projects[message][type] = module
```
ronaldtebrake: 1. Don't forget to add the crop api module (dependency of image_widget_crop). Also in composer.json
2. Don't forget to add the libraries here too (for packaging on drupal.org) can be done like this:

`projects[r4032login][type] = module
 projects[r4032login][download][type] = git
 projects[r4032login][download][revision] = 4b2077aa70e3f7b00b8a9cba25af5b948ba2e3b9
 projects[r4032login][download][branch] = 8.x-1.x`

modules/social_features/social_core/config/install/image.style.social_xx_large.yml
```diff
@@ -1,13 +1,21 @@
 langcode: en
 status: true
-dependencies: {  }
+dependencies:
+  module:
+    - crop
```
ronaldtebrake: I see this as a dependency, but it's not added to the composer.json / drupal-org.make (nor in the social_core.install) maybe add this.. Seems to be the API for the image_widget_crop

modules/social_features/social_core/social_core.install
```diff
@@ -198,3 +198,16 @@ function social_core_update_8002() {
...
 }
+
+/**
+ * Install image_widget_crop module.
+ */
+function social_core_update_8003() {
+
+  $modules = array(
+    'libraries',
+    'image_widget_crop',
```
ronaldtebrake: + 'crop',

PR #122, opened 2016-10-07
DS-2175 by ribel: As a LU I can't select if Comments should be open or closed anymore 

ribel:
# Changes:
- Add new permission `administer own comments`
- Enable `administer own comments` permission for authenticated users.
- Alter the default `list_class` for fields of type `comment`

# HTT
- [x] Log in as normal user
- [x] Check that you can't set "Comment status" when you add or edit content
- [x] Checkout this branch and do a `drush updb -y; drush cr` (to make sure it works for update path)
- [x] Check that now you are able to set "Comment status" when you add or edit content
- [x] Reinstall site in this branch and check that nothing is broken.
- [x] Check the code

### DIFF
--- a/modules/social_features/social_comment/social_comment.install
+++ b/modules/social_features/social_comment/social_comment.install
@@ -49,6 +49,7 @@ function _social_comment_get_permissions($role) {
     'skip comment approval',
     'edit own comments',
     'delete own comments',
+    'administer own comments',
   ));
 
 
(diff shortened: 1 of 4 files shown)


ronaldtebrake:
Hi @ribel,

I thought there was already an existing permission:
"Administer comments and comment settings" that should work, or does this come with a lot of extra settings and permissions the user doesn't need to have? I think it will only be used in the case a user can actually create/edit content.

PR #124, opened 2016-10-11
DS-2206 by nielsvandermolen: Disable views caching for search views

NielsGG:
There was an issue with changes not being displayed correctly in the teaser. The only place I could reproduce it was in the search pages.

The fix was to disable caching for the views page. the render cache for the individual teasers should still apply. Changing the views caching to tag based also did not solve the issue. 

HTT:

- [x] Enable caching if it is disabled
- [x] Checkout 8.1.x
- [x] Create an topic / event / book page.
- [x] Go to the search page
- [x] Change some fields
- [x] Changes should not be reflected in the teaser.
- [x] Checkout this branch and feature revert and cache clear
- [x] Go to the search page
- [x] Change some fields
- [x] Changes should now be reflected 
- [x] Check teasers on other places as well

### DIFF
--- a/modules/social_features/social_search/config/install/search_api.index.social_content.yml
+++ b/modules/social_features/social_search/config/install/search_api.index.social_content.yml
@@ -20,7 +20,6 @@ field_settings:
     boost: !!float 8
   rendered_item:
     label: 'Rendered HTML output'
-    datasource_id: null
     property_path: rendered_item
     type: text
     configuration:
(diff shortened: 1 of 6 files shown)


ribel:
Hi @NielsGG!
I also have tried to decrease caching to 1min for search api caching, but it didn't work. So the only possible solution is to disable cache for search at all. Also i have disabled cache for Group Search.

PR #130, opened 2016-10-12
DS-2207 by nielsvandermolen: Removed the block config for the stream

NielsGG:
Another PR for DS-2207. Because we now have the update hook we should remove the block from the config or the block will be reset to the social_base theme and not the subtheme.


NielsGG:
Hi @ribel, can you take a look at this PR?

ribel:
Hi @NielsGG,
I am not sure about what update hook you are talking about? 
I see that you removed block config, and for fresh distro reinstall it means that this block is not displayed on homepage, neither for LU or AN. See: http://prntscr.com/csw4sj
That's why tests are failing.

NielsGG:
Hi @ribel,

Yes, added an update hook without review to use in Y4P. see social_activity.install. Changed some code so the block is also added during installation.

NielsGG:
Weird now it seems to show up on the login page after installation and not on the front page...
screen shot 2016-10-12 at 14 14 03

ribel:
I think issue is here:
image

PR #148, opened 2016-10-24
DS-2304 - Remove 'in the group' from activity stream items

jochemvn:
# HTT

## Part 1:
- [x] revert the social_activity feature

- [x] create a topic/event/post in a group and empty the queue's

- [x] see that on the activity stream there's no longer the string 'created in the group ', but only 'in '

- [x] Do a re-install and notice the same for the demo content

## Part 2:

Check that on the pages below, the mentioned blocks in the complementary area have the correct subtitle and the links are also correct

### Subtitles should be:
Home stream >> 'in the community'
Group stream >> 'in the group'
Profile stream >> 'for this user'

- [x] Home stream
- Upcoming events
- Newest topics
- Newest groups
- Newest members

- [x] Group stream
- Upcoming events
- Newest topics
- Newest members

- [x] Profile stream
- Upcoming events
- Recently created topic
- Recently joined groups

### DIFF
--- a/modules/social_features/social_activity/config/install/message.template.create_comment_community_post.yml
+++ b/modules/social_features/social_activity/config/install/message.template.create_comment_community_post.yml
@@ -18,8 +18,8 @@ template: create_comment_community_post
 label: 'Create comment on post in the community'
 description: 'A user add a comment to post in the community'
 text:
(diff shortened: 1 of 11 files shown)


jochemvn:
Reviewed this with @ronaldtebrake.

PR #155, opened 2016-10-26
DS-2157: patch address module - fix installation error due to access level

jaapjan:
See: https://www.drupal.org/node/2822309

I'm not sure what would be the side effect of using this patch with library version 0.7.3. But I guess if you would run composer update on your project it should automatically update to the latest version of commerceguys/intl as well.


ronaldtebrake:
Nice catch! 

I've reviewed this because I need it for SSGT. Will add a task for maintenance tomorrow to update address to RC3 when it lands.

PR #160, opened 2016-10-28
DS-2292 by alex.ksis: Add patch to fix fatal error when uploading files

ksis:
# HTT
- [x] Reinstall site on this branch.
- [x] Upload image for new group content.
- [x] Save new group content. 
- [x] Check the code.
# Changes
- Fixed calling the cache in Group module.

Here the patch: https://www.drupal.org/node/2793621#comment-11753511


ronaldtebrake:
Hi @ksis, when applying patches or module updates don't forget to also apply it to the corresponding drupal-org.make or drupal-org-core.make files so the packaging of drupal will go correctly.

I will do it for this patch :)

PR #173, opened 2016-11-04
Feature/ds 2300

MaikelGG:
- [x] Node author name and avatar and creation date are shown at top of card
- [x] Group name is shown and linked at top of card
- [x] Amount of comments is shown at top of card and links to comments section, but only when comments are enabled for this content type
- [x] For event the event date and event location are shown before the body text.
- [x] All functions in modules should be cleaned up (social_core)
- [x] When content is featured, it is shown as a label in the node
- [x] The "show all notifications" link in the notifications dropdown is visible again on mobile (bug)
- [x] The amount of comments is shown in the comment title
- [x] The navbar links have title attributes which display on mouse-over. This is to clarify what icons mean.


ribel:
Hi @MaikelGG,
Overall changes looks good, but i have one question about item:
- When content is featured, it is shown as a label in the node

Currently we show label only in full view, maybe we should also reflect it in teaser? What do you think?

MaikelGG:
@ribel It would be a nice to have feature to show featured content in the teaser, but we do not have a good design for it yet. So we can create a follow for it. However, in a subtheme is quite easy now to implement it with a few lines of css. we are not planning to add a label in the teaser because of the lack of space.

PR #179, opened 2016-11-10
DS-2196 by vlad-nosykov: Increase max dimensions for images.

vladnosykov:


jaapjan:
Works fine for Topic, Events and profile. I think this dimension should work for most HD images!

Could you also implement the same change for 
- field.field.node.page.field_page_image.yml
- field.field.node.book.field_book_image.yml

After that (if tests pass) feel free to merge the PR.

PR #192, opened 2016-11-23
DS-2184 by alex.ksis: Remove ability of selecting text format if full_html is available.

ksis:
# Changes
- Removed ability of selecting text format if full_html is available.

# HTT
- [x] Clear Drupal cache.
- [x] Sign in as user which has access to the "Full HTML" text format.
- [x] Go to any form where the textarea with WYSIWYG editor.
- [x] Check that dropdown list of selecting text format doesn't appearing.
- [ ] Check the code.


jochemvn:
Hi @ksis, 
Can you test the following use case to make sure we don't break anything?

1. Login as a user that does NOT have access to full_html
2. Create a topic 
3. Logout and login as admin
4. Edit the topic you just created
5. Logout and login as the user that created the topic
6. You should still be able to edit the content. If not, we should consider fixing this in a different way.

peterpolman:
@jochemvn Hey, tested your use case and nothing wrong there.
- Editor formats for both users as expected
- Content stays editable for authenticated user after admin updated the node.

PR #198, opened 2016-11-28
DS-2074 by ribel: Add Views sort plugin for passed and upcoming events

ribel:
# HTT
- [x] Go to `/community-events`
- [x] When you see upcoming events you should see events ordered by ASC from current date
- [x] When you see passed events you should see events ordered by DESC from current date
- [x] Check the same for events in Group
- [x] Check the same for events on User Profile


NielsGG:
Hi @ribel,

It works well but I think we should create or enhance existing Behat tests for this because this is not something we will test often manually and it is fairly easy to write steps for it.

Greetings,
Niels

PR #199, opened 2016-11-29
SSGT-221 by vlad-nosykov: Add 'delete all comments' permission, adjus…

vladnosykov:
# HTT
- [x] Create event, topic via admin menu (not in group)
- [x] Add comments to created content
- [x] Login as SM
- [x] Should be able to delete any comment

### DIFF
--- a/modules/social_features/social_comment/social_comment.install
+++ b/modules/social_features/social_comment/social_comment.install
@@ -60,7 +60,7 @@ function _social_comment_get_permissions($role) {
 
   // Site manager.
   $permissions['sitemanager'] = array_merge($permissions['contentmanager'], array(
-    'delete all comments'
+
   ));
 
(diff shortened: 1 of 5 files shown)


modules/social_features/social_comment/social_comment.install
```diff
@@ -83,3 +83,21 @@ function social_comment_update_8001(&$sandbox) {
     }
   }
 }
+
+/**
+ * Enable 'delete all comments' permission for site manager users.
+ */
+function social_comment_update_8002(&$sandbox) {
```
ribel: @vladnosykov Please add this also to `hook_instal()` so we will have this permission by default in new installations.

vladnosykov: @ribel Done, please check latest commit

PR #206, opened 2016-12-08
ECI-122: Image cropping in not working correctly anymore

ksis:


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "goalgorilla/open_social",
+    "name": "open-social-eci/open_social",
```
NielsGG: Hi @ksis,

We should not commit this line to the repo (because it is an override for the Fork). Not sure what the best way is to fix it for now, maybe do another commit in the branch that sets the line back

jaapjan:
Not necessary anymore since I merged in: https://github.com/goalgorilla/open_social/pull/208 which does the same thing. Thanks!

PR #253, opened 2017-01-21
Issue 2845556 by Chipie: Group Image is not shown

luxio:
This Issue is mirrored on drupal.org by: https://www.drupal.org/node/2845556


ronaldtebrake:
Thanks for the patch & fix, updated the PR to here https://github.com/goalgorilla/open_social/pull/267 so tests will run :) 

Will make sure you get the correct contribution!

PR #269, opened 2017-02-14
Update Address

bojanz:
RC4 removes the need for a patch.

Note that if you had any alters targeting the address form element, you'll need to update them, since there's now an address form element used by the widget.


jochemvn:
Hi @bojanz,
Not sure what's exactly going on here, but I think one of our developers already updated address to rc4. He must have overlooked your PR.  I think this one can be declined now

bojanz:
Sure.

PR #274, opened 2017-02-21
Feature/ds 2387

peterpolman:
Currently this behaviour is applied in the block--local-tasks-block.html.twig template file by wrapping the contents in a .tabs-scrollable.


MaikelGG:
@peterpolman I have updated some of the styles and tested it. It works the same as "google.com" on mobile now. http://styleguide.getopensocial.com/section-molecules.html#kssref-molecules-navigation-nav-tabs. I think you can merge it.

PR #276, opened 2017-02-22
#2767295 Order of comments in posts

ribel:
See: https://www.drupal.org/node/2767295


NielsGG:
Hi @ribel 

Is there a reason the type of this field is changed?

screen shot 2017-02-27 at 13 28 45

ribel:
Hi @NielsGG,
This was autogenerated from active config. And it should be actually `text_default` (see `core.entity_view_display.post.post.default.yml` for example).

PR #279, opened 2017-02-23
 Issue #2835788 by ribel: Enable administrative area for Event and Profile address fields

ribel:
See: https://www.drupal.org/node/2835788

I have enabled administrative area for Event and Profile address fields.
Also i have updated behat tests in order to check this new field. (For Netherlands in address module  admin areas are not provided, so i changed it to some other country ;)


jaapjan:
😆  Lviv slowly taking over I see!

PR #291, opened 2017-03-13
ECI-339 by alex.ksis: Improve Social Demo with plugin manager

ksis:
Command to create demo content:
`drush demo-content-add file user group event_type event topic post like comment event_enrollment`

Command to remove demo content:
`drush demo-content-remove file user group event_type event topic post like comment event_enrollment`


modules/custom/social_demo/social_demo.drush.inc
```diff
@@ -5,72 +5,72 @@
...
  */
 
+use Drush\Log\LogLevel;
+use Drupal\user\Entity\User;
+
 /**
  * Implements hook_drush_command().
  */
 function social_demo_drush_command() {
-  $items['social-demo-add'] = array(
```
ronaldtebrake: Is there any specific reason why this got changed? This means all of our scripts that run demo content will now not work anymore.. would be nice to make it backwards compatible in the future :)

PR #319, opened 2017-04-04
#2866717 Change inline svg to external sprite

MaikelGG:
for all icons in theme

**How to test:**
- Check all pages the icons are loaded. The old inline svg is removed, so it will not fall back to that.
- Review the code changes


peterpolman:
@ronaldtebrake Don't you think we can set some sort of global variable for this path_to_basetheme that we can reuse in the templates? Now we are using the same logic in a bunch of preprocessors. 

If the improvement is complex and this approach is also fine with you I'm happy to merge it of course:)

PR #335, opened 2017-04-21
Sprint 25 - Maintenance

ronaldtebrake:


ronaldtebrake:
HTT.

1. Check the admin toolbar still works, also keep in mind it should work for SaaS
2. Check the search works, all the search tab should work and also make sure the filtering works.

PR #346, opened 2017-05-01
DS-3427 - Update core

jochemvn:
## HTT

Tested the following two scenario's:

1. Fresh install (also done by the tests below)
2. Install on master and then:
    - switch to this branch (in composer)
    - composer update
    - ds cc drush
    - ds updb --yes
    - ds cc drush
    - ds fra --yes

This last method seems to be the most interesting case and I can recommend trying that over a fresh install. Both is better obviously


ronaldtebrake:
Nice :)!

PR #354, opened 2017-05-03
Issue #2875275 by ToxaViking: Wrong redirect

AntonMartyniuk:
https://www.drupal.org/node/2875275


jaapjan:
Thanks! I've added a new pullrequest so we'll be able to run all the automated tests. 
https://github.com/goalgorilla/open_social/pull/368

I've moved your code to the PostDeleteForm class and I also made sure that when there is already a destination provided in the URL we will use that URL instead.

PR #365, opened 2017-05-10
Issue #2875187 by peterpolman: Fixed z-index of first stream-item.

peterpolman:
# HTT

- [] Go the the homepage as LU in IE11+
- [] Click the visibility button and see if it appears above all cards
- [] When clicking another visibility setting the button should be updated


jaapjan:
Thanks Maikel and Peter!

PR #366, opened 2017-05-10
Issue #2816397 by t0r: Add support for post bundles

semkulich:
# Changes

- Add support for post bundles
- Fix demo content for posts
- Fix menu items
- Remove composer.json


jochemvn:
Hi @semkulich,

Thanks for the great work done already. Looks cool and we really want to add it to our distro. I checked it out a bit yesterday (before your last commits) and noticed the upgrade path was still breaking. 
Normally when we create a pull request here, we assume that development is done. Can you say anything about when it and if will be fixed?

kind reagards,

Jochem

semkulich:
Hi @jochemvn,

I have added upgrade path and fix some other issues. Now it works for both new and old installations.
Please check and let me know if it works.

Best regards,
Semkulich

PR #372, opened 2017-05-15
ISSUE-2877309 by nielsvandermolen: CRON stuck because of infinite loo…

NielsGG:
…p for nodes authored on a created date

HTT:
Found a way to reproduce it. As a CM it is possible to set an authored on date which is the created date.

When this date is set in the future the CRON task will be stuck (and the process will not abort).

There is a potential issue that an incorrect notification is being sent for nodes that are created in a group with a authored on date set in the future. But that issue is less critical than the CRON breaking.


jaapjan:
@NielsGG What do you mean with "potential issue that an incorrect notification is being sent for nodes"? That users receive potentially a wrong notification?

NielsGG:
@jaapjan Yes, it would be handled as an ordinary node and not appear as a node in a group. It is sort of edge case because it is only when they change the authored on creation date. There are likely similar problems already with the existing code

PR #379, opened 2017-05-19
Issue #2879696 by alex.ksis: Add UserRoute context. Add cache tags an…

ksis:
There are no steps how to test because this bug reproduces randomly.


jaapjan:
@ksis remember you also need to create an update hook when changing block settings, see:
https://github.com/goalgorilla/open_social/pull/437

ksis:
@jaapjan I added an update hook

PR #380, opened 2017-05-22
DS-1136 - Add image to posts

jochemvn:
## HTT

- [x] Check the code
- [x] Check the prototype
- [x] Login as LU on OS
- [x] Notice the "add image" button on the post form
- [x] Add an image
- [x] Notice the placeholder text changes to "Say something about this photo"
- [x] Well... say somehing about the photo
- [x] POST!
- [x] See the photo in the activity stream
- [x] Try the same on profile and group

## Other stuff
- [x] Notice you cannot add multiple photos (architecture for multiple is there though)
- [x] ...


modules/social_features/social_post/social_post.install
```diff
@@ -4,9 +4,6 @@
  * Install, update and uninstall functions for the social_comment module.
  */
 
-use Drupal\Core\Database\Database;
-use Symfony\Component\Yaml\Yaml;
-
 /**
```
jaapjan: Perhaps add some information here that you should not add update hooks in this file. Your future self will thank you!

modules/social_features/social_post/modules/social_post_photo/social_post_photo.install
```diff
@@ -0,0 +1,31 @@
+<?php
+/**
+ * @file
+ * The Social post photo installation.
+ */
+
+
+function social_post_photo_uninstall() {
+  $configs = [
```
jaapjan: It is best to change the existing posts of bundle 'photo' to the old 'post' bundle again on uninstall. On install: the other way around.

PR #414, opened 2017-06-11
Issue #2862424 by alex.ksis: Hide entity id from entity autocomplete …

ksis:
Related issue: https://www.drupal.org/node/2862424


jochemvn:
Hi @ksis ,
Looks good. However since this is regression, I'm wondering if we can remove the solution that we added before: https://github.com/goalgorilla/open_social/pull/111/commits/1a560b6a251e7b44ea717fb7dcdac2fad2eef0fa

What do you think? Is that code still needed?

PR #424, opened 2017-06-14
Analytics KPI theming changes

NielsGG:


NielsGG:
Hi @MaikelGG, do you think we could merge this PR without enabling analytics or could it cause regression?

PR #433, opened 2017-06-19
DS-3623 - Redirect LU Registering with Social Login To Edit Profile

jaapjan:
**Description:**
Users signing up to the platform with one of the social authentication methods don't need to set their password after registering. Instead we will redirect these users to the edit profile page. In addition to this the social login buttons on the top of the register page were still shown after selecting one of the social authentication methods. This could be confusing to the user.

**Solution:**
Redirect users registering with social authentication to the user edit profile form. Hide the icons after the user is redirected from the selected authenticated method. The behaviour did not change for normal registrations.

I've used an unset in the form_alter. I think this is save since most of the logic is in separate Services anyways.

**How to test:**

- [x] Setup social authentication of your choice: https://github.com/goalgorilla/drupal_social/wiki/Social-Login
- [x] Register and verify that after selecting the authentication method your e-mail and username are filled in and the header with the icons is not shown. 
- [x] Click on continue and notice you are redirected to "Edit Profile".
- [x] Verify the normal registration flow
- [x] Check the code

### DIFF
--- a/modules/custom/social_auth_extra/social_auth_extra.module
+++ b/modules/custom/social_auth_extra/social_auth_extra.module
@@ -57,6 +57,11 @@ function social_auth_extra_form_user_register_form_alter(&$form, FormStateInterf
           $name = preg_replace('/[^0-9a-z]/si', '_', $name);
         }
 
+        // The user already selected an authentication method and there
+        // will be a drupal_set_message informing the user.
+        unset($form['social_auth_extra']);
+        unset($form['account']['title']);
+
         $form['account']['mail']['#default_value'] = $data_handler->get('mail');
         $form['account']['name']['#default_value'] = $name;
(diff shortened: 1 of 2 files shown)


modules/custom/social_auth_extra/social_auth_extra.module
```diff
@@ -197,9 +202,11 @@ function social_auth_extra_form_user_register_form_submit($form, FormStateInterf
         '@community_name' => \Drupal::config('system.site')->get('name'),
       ]));
 
-      $form_state->setRedirect('entity.user.edit_form', [
-        'user' => $account->id(),
-      ]);
+      // Redirect to profile.
```
jaapjan: @frankgraave I'm still discussing with Taras to move this to social_profile or social_user instead since the idea is to make this a separate module at one point. I'll make the change tomorrow or later today!

PR #438, opened 2017-06-21
Issue #2888016 by alex.ksis: Add additional cache context for SearchH…

ksis:
Related issue: https://www.drupal.org/node/2888016

## HTT

- [ ] Logout
- [ ] Go to **/search/content/test1**
- [ ] Go to **/search/content/test2**
- [ ] Make sure that value in search input is updated

### DIFF
--- a/modules/social_features/social_search/src/Plugin/Block/SearchHeroBlock.php
+++ b/modules/social_features/social_search/src/Plugin/Block/SearchHeroBlock.php
@@ -3,13 +3,14 @@
 namespace Drupal\social_search\Plugin\Block;
 
 use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Cache\Cache;
 
 /**
  * Provides a 'SearchHeroBlock' block.
  *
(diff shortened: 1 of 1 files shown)


jaapjan:
See also: https://github.com/goalgorilla/open_social/pull/431 where I already added a cache context to the Form Build array (which also solves this problem)

jaapjan:
@ksis  cc @ribel 

What do you think of the solution in #431 where I add cache tags in the form array instead. That should also solve the bug you have, correct?  Could you verify and if you agree close this pull request?

ksis:
Hi @jaapjan 
I tried to check it, but I did not find this fix in the SearchHeroForm that you have commit [here](https://github.com/goalgorilla/open_social/pull/431/files#diff-f57840bdb73b28e3a43d1ca46b48e500). I checked it on 8.x-1.x

ksis:
@jaapjan I have checked it, you're right. The issue was solved with your solution. So we can close this PR.

PR #455, opened 2017-06-28
DS-3773 - Add image_effects module for providing correct photo orientations

frankgraave:
# Background
When uploading a photo from mobile, the orientation sometimes is not correct. After some research we found out that with this module we can simply add an image effect to our image styles. This effect reads EXIF information from the picture and to determine the image orientation.

# HTT

- [ ] Check the code
- [ ] Check the PR in BB for the link to test it on mobile
- [ ] Take a picture with your mobile phone and do a photo post
- [ ] Also test it for any other places where pictures can be uploaded.
- [ ] See that your picture is always in the correct orientation!


jochemvn:
@frankgraave ,

You have a conflict that must be resolved.
Please do so and then you can merge it!

PR #472, opened 2017-07-12
#2893858 - Profile picture access

jochemvn:
## Description
Since we implemented the private file system, anonymous users no longer have access to view profile pictures. So on pages where we show profile pictures, such as the stream. We need to check if the current user has access to the related file and if not, replace it with the default image, or hide the image entirely.

## HTT
- [x] Make sure the private file system module is enabled
- [x] Create a new account and a profile WITH an image.
- [x] Post something PUBLIC in the stream
- [x] Logout
- [x] Notice you get a broken link to an image in the stream
- [x] Checkout this branch
- [x] Check the code
- [x] Rebuild caches and refresh the page
- [x] Notice you see the default image
- [x] Login as a normal LU
- [x] See that you see normal profile images

### DIFF
--- a/modules/social_features/social_profile/social_profile.module
+++ b/modules/social_features/social_profile/social_profile.module
@@ -339,7 +339,16 @@ function social_profile_profile_view_alter(array &$build, Drupal\Core\Entity\Ent
   $current_user = \Drupal::currentUser();
   // If the current user has no access to viewing user profiles, he might not have access to
   // the users profile.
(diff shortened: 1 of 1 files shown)


jochemvn:
@frankgraave although it's still a WIP, you might want to take a look at the code already. I'm having some issues replacing the profile image with the default image. However it currently work so, that no image is shown at all, which is already a bit better then a broken image.

jochemvn:
Should be fixed now. HTT still valid!

PR #473, opened 2017-07-13
Issue #2894571 by alex.ksis: Add AccessDeniedHttpException to prevent…

ksis:
Related issue: https://www.drupal.org/node/2894571

## HTT
- [ ] Sign in.
- [ ] Go to the About page of a closed group (make sure that you are not a member of this group).
- [ ] Make sure that you are not redirected too many times and caught **Access denied** page.


jochemvn:
Hi @ksis, Can you maybe explain in what case this error occurs? I can't reproduce it

ksis:
Hi @jochemvn!
I have attached simple module that could help to reproduce this bug.
I just changed canonical URL of a group from **/stream** to **/about**.

[test_cyclic_redirect.zip](https://github.com/goalgorilla/open_social/files/1214461/test_cyclic_redirect.zip)

## Steps to reproduce
1. Create closed group.
2. Sign in as a LU.
3. Go to **About** page of the created group.

frankgraave:
Hi @ksis 
I've tested this with the test_cyclic_redirect. It does work, but I don't think we should throw an Access Denied when this happens. 

From the homepage stream, the user can see that there is a (new) group in the blocks in the sidebar. When the user clicks on the group, with /group/x/stream, the user is redirected to the about from this group to provide some information about the closed group saying he can not join this group, but only on invitation (or request a membership in the future). 

With this fix in place, the user can not see this information about the group any more, and thus can not see what the group is about, who the manager of the group is, or request membership from this closed group. Please let me know what you think!

ksis:
Hi @frankgraave,
I agree with you. I would propose to add a new permission like **View stream page** to make access configuration more flexible. In this case we do not have to check if user is a member of a group. We just can check if user has access to view stream page, otherwise redirect him to **About** page if he has access to view it. But if user does not have access to view **About** page, we just throw AccessDeniedException. What do you think about it?

frankgraave:
Hi @ksis,
After looking further into this, and your second suggestion, I think this fix should not be in the distro. I think this fix should be done in the project where the altering of the group canonical is done. Adding a permission for the group stream would probably solve your issue, but won't be a proper solution with closed groups in my opinion. I think that altering the group canonical is not a very common use case.  So adding the permission would require a lot of other architectural changes as well for the closed groups I think. So I suggest this should be a custom fix for the use care itself.

PR #490, opened 2017-08-07
#2894160 by jochemvn: Change 'comment on a photo' into 'comment on a post'

jochemvn:
## Description
When someone comments on a post that has no photo attached to it, the author gets a notification that someone commented on his/her photo. To make sure it fits both cases with and without photo, we've changed it back to 'post'

## Drupal 
https://www.drupal.org/node/2894160

## HTT
- [x] Check the code
- [x] Checkout this branch and clear caches
- [x] Login as user A and create two posts (on with and one without photo). Also create a topic and an event
- [x] Login as user B and place comments on all created content types
- [x] Login as user A again.
- [x] Notice that you have four notifications (and/or emails)
- [x] Two of those say "User B commented on your post"
- [x] One says "User B commented on your topic"
- [x] One says "User B commented on your event"
- [x] Check both notifications AND emails


modules/social_features/social_comment/social_comment.tokens.inc
```diff
@@ -83,17 +83,24 @@ function social_comment_tokens($type, $tokens, array $data, array $options, Bubb
...
+              /** @var \Drupal\comment\Entity\Comment $comment */
               $comment = \Drupal::entityTypeManager()
                 ->getStorage($target_type)
                 ->load($target_id);
 
               if (!empty($comment)) {
                 if ($comment->getEntityTypeId() == 'comment') {
                   if (!empty($comment->getCommentedEntity())) {
+                    /** @var \Drupal\Core\Entity\Entity $node */
                     $node = $comment->getCommentedEntity();
```
Kingdutch: We should probably rename the occurrences of `$node` to `$entity` since this piece of code also works for the "Post" entity which is not a node.

We should ensure that the naming doesn't bite us in the future.

Kingdutch:
An alternative solution would be to implement a `getDisplayName` or `getNotificationName` for the Entities. Then in Nodes the code could be `return $this->bundle` and in posts it can be `return t('Post')`.

That means that no if-statement are needed and if a new type is added in the future that requires different handling, it can be done in that type instead of in this piece of code about comments.

Kingdutch:
@frankgraave I'm not entirely sure how that's possible. As far as I know there's nothing in the changes here that affect that. Unless there's some code somewhere that specifically needs a certain value for this token field.

I've triggered a rebuild on Travis to see if the errors related to log-in were fixed. The build seemed to fail unrelated to this PR.

Kingdutch:
Rebased this pull request on the latest version of 8.x-1.x

This should get rid of the errors that were in the CI which weren't related to this PR.

Kingdutch:
Because merging 8.x-1.x into this feature branch does not solve the issues with feature state. A new branch has been created and is ready for review in #522

PR #492, opened 2017-08-08
Issue #2900607 by Makishima: Fix activity stream tokens for group content

VolodymyrMartsynenko:
**Related issue**: https://www.drupal.org/node/2900607

**HTT:**
- [ ] Login as admin
- [ ] Go to Structure > Message templates
- [ ] Edit "Create topic in group" message template or create new message template with group content entity
- [ ] Use [gtitle] and [gurl] tokens in the field "Output text for normal activities in activity stream"
- [ ] Save message template. 
- [ ] Create a new topic in any group.
- [ ] Log in as a user which is a member of this group. 
- [ ] Check new message in activity stream with the title of group and link to the group.


bramtenhove:
@frankgraave you requested yourself as reviewer. Could you take a look at it?

bramtenhove:
@VolodymyrMartsynenko, @frankgraave, I tried testing this using the steps above but I wasn't able to get it working. Any clue?

Also I'm unsure why we need this? We already have a reference to the group via the message entity right?

frankgraave:
@bramtenhove I also did not succeed with the HTT above. Bu I was busy with the group notification story and in that context I tried this fix, which works nicely! Thanks @VolodymyrMartsynenko

PR #494, opened 2017-08-09
Issue #2896186 by frankgraave: Closed Groups should show up in search results

frankgraave:
# Background
In the search, closed groups did not show up. Also, groups in general only show up with some keywords of a group. When the full group title was filled in, the group did not show in the search.
Related issue: https://www.drupal.org/node/2896186

# HTT

- [ ] Check the code
- [ ] Checkout this branch `git checkout feature/2896186-group-search`
- [ ] Update with the `/var/www/scripts/social/install/update.sh` script
- [ ] See the search is being re-indexed
- [ ] Log in as any user and create a closed group
- [ ] Search for the group
- [ ] See it's shown in the "All" and "Groups" tab


jochemvn:
Solution works, but the update hook does not. The function you are calling does the following: 

> Force reindexing of one or all search indexes, without deleting existing index data.

This only works for new installations. But the install script already does this then.

We need to change it a bit, since there is a lot of existing data in the search_api and you want to replace this, because it's incorrect. 

Since you only made changes to the group search, I recommend you do a search-api-clear (drush_search_api_clear) and a search-api-index (drush_search_api_index). But only for the social_group index.

jochemvn:
Also changed the HTT a bit:
- [ ] Check the code
- [ ] Perform a fresh installation on 8.x-1.x
- [ ] In the group search for the word: local
- [ ] Notice only 1 group shows up
- [ ] Checkout this branch `git checkout feature/2896186-group-search`
- [ ] Update with the `/var/www/scripts/social/install/update.sh` script
- [ ] See the search is being re-indexed
- [ ] Log in as any user and create a closed group (don't use the word 'local' in it)
- [ ] In the group search for the word: local
- [ ] Notice that now 2 groups show up
- [ ] Search for the group
- [ ] See it's shown in the "All" and "Groups" tab

bramtenhove:
After releasing 1.4 we received an issue in the issuequeue regarding fatal errors in an update hook, see https://www.drupal.org/node/2906090. This is caused by using drush functions in `social_search_update_8103()`, but these are **not** available when you update via the **interface**, e.g. update.php.

See changes in https://github.com/goalgorilla/open_social/pull/520.

PR #503, opened 2017-08-16
Hotfix - Update Drupal Core from 8.3.5 to 8.3.7

jaapjan:
Don't merge yet.

Let's see what Travis will do.


jaapjan:
If this is green I think it can be merged. 

Notice that I already did a Drupal.org release for Version 1.3, see: https://www.drupal.org/project/social/releases/8.x-1.3

PR #510, opened 2017-08-21
Always use absolute URLs in message templates

Kingdutch:
Messages may be sent out by e-mails or be displayed in contexts
where we're not currently at the domain name where the piece of
content is hosted. In these scenarios, relative URLs break so we
should always use absolute URLs.

## Messages affected
create_comment_community_node
create_comment_community_post
create_comment_group_node
create_comment_group_post
create_event_community
create_event_group
create_post_community
create_post_group
create_topic_community
create_topic_group
create_comment_reply_mention
create_mention_comment
create_mention_post

## How to test
- Create a piece of content that triggers one of the affected messages (for example mention someone in a comment). 
- Observe that the first link in an e-mail for this message is not absolute and thus doesn't work.
- Apply the patch
- Create a piece of content that triggers one of the affected messages.
- Observe that the first link in the e-mail for this message has an absolute link and correctly leads to the website.


robertragas:
Is working but you forgot the social_admin notifications

PR #512, opened 2017-08-22
ECI-2 by ribel: Fix article in Open group behat test

ribel:
**Changes:**
Replaced **a** with **an** in behat test, to be consistent with code.

**HTT:**
- [x] Run this behat test
- [x] See it pass


frankgraave:
I've updated the PR and fixed all other issues that were occurring locally. I've also created a PR (https://github.com/goalgorilla/open_social_scripts/pull/11) to enable this test once again.

PR #519, opened 2017-08-31
DS-4122 - Remove libraries and refactor search script

MaikelGG:
- Updated waves script, so it can be easily updated
- Removed modified version of contrib search script
- Add new interaction for search block in navbar
- The block is in the conten top region, and hidden. There is a trigger to show the form
- On mobile the form remains the same
- Remove contrib scripts that are not used
- Added a gulp watch task for scripts


MaikelGG:
@bramtenhove Travis fails because of missing composer dependencies of drupal_social feature branch

themes/socialbase/src/Plugin/Preprocess/SocialBaseBlock.php
```diff
@@ -128,11 +128,15 @@ public function preprocess(array &$variables, $hook, array $info) {
...
-      $variables['content']['search_form']['actions']['submit']['#addsearchicon'] = TRUE;
-
       if ($region == 'hero') {
         $variables['content']['search_form']['#attributes']['class'][] = 'hero-form';
         $variables['content']['search_form']['#region'] = 'hero';
+        $variables['content']['search_form']['actions']['submit']['#addsearchicon'] = TRUE;
+      }
+      elseif ($region == 'content_top') {
+        $variables['content']['search_form']['#region'] = 'content-top';
+        $variables['content']['search_form']['search_input_content']['#attributes']['placeholder'] = t('What are looking for ?');
```
bramtenhove: @MaikelGG, the text is incorrect. Could you change it to something like `What are you looking for?`

bramtenhove:
@MaikelGG, as the interaction changes for the search form I'd like another pair of eyes. Who do you suggest can do this best?

MaikelGG:
@bramtenhove I can demo it to Xinyu next week. Maybe Moritz can join too.

MaikelGG:
I have updated the text

bramtenhove:
Alright, sounds like a plan. I'll complete testing the functionality and will then run all the stability tests locally.

PR #524, opened 2017-09-06
Increase specificity for selectors in LU homepage my events test

Kingdutch:
The LU homepage my events test tests for the existance of some
events but is not specific enough to target only the "My Events"
block. This causes the test to fail when the overview block of
"all community events" is also in the sidebar.

This commit adds a region map for the my events block to test the
existance of the enrolled and non-enrolled event there causing the
test to function correctly.


Kingdutch:
Apparently this only fails locally.

PR #527, opened 2017-09-08
Issue #2907539 by alex.ksis: Add default value callback for content v…

ksis:
Related issue: https://www.drupal.org/node/2907539

## HTT
- [ ] Sign in as a LU.
- [ ] Go to the **stream** page of your profile.
- [ ] Make sure that there are no any errors.

### DIFF
--- a/modules/custom/entity_access_by_field/entity_access_by_field.module
+++ b/modules/custom/entity_access_by_field/entity_access_by_field.module
@@ -12,7 +12,7 @@ use Drupal\Core\Cache\Cache;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Entity\EntityFormInterface;
 use Drupal\Core\Entity\EntityInterface;
-use Drupal\social_post\Entity\PostInterface;
+use Drupal\Core\Field\FieldDefinitionInterface;
 
 // @todo: Prevent creating multiple visibility fields on one node.
 // @todo: Add support for multiple entity types.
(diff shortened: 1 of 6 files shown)


Kingdutch:
Your changes seem sensible but for me they don't touch on the error message that's reported on https://www.drupal.org/node/2907539

That still existed after adding this PR as a patch to the Pachamama website.

It appears that the logic in the `isEmpty()` method of the `Dropdown` `FieldType` is incorrect. I've added a commit to correct this logic. `isEmpty` now returns true if `$this->values['value']` is empty or if its value in string form equals 0.

modules/custom/dropdown/src/Plugin/Field/FieldType/Dropdown.php
```diff
@@ -132,7 +132,7 @@ public function setValue($values, $notify = TRUE) {
    * {@inheritdoc}
    */
   public function isEmpty() {
-    return empty($this->values['value']) || (string) $this->values['value'] === '0';
+    return (!isset($this->values['value'])) || (empty($this->values['value']) && (string) $this->values['value'] !== '0');
```
Kingdutch: I'm unsure when exactly we say that the Dropdown is empty due to the string cast and the many brackets. It feels like this is a lot more complex and error prone than it should be.

Lets evaluate this new if statement for four possible values of `$this->values['value']`:
`$this->values['value']` is:
- undefined
- null, false
- '0'
- 0
- any other value

For undefined:
`return (!isset(undefined)) || (empty(undefined) && (string) undefined !== '0'`
`return (!FALSE) || (TRUE && error!)`
luckily results in:
`return TRUE || (doesn't matter, not executed)`

For null, false:
`return (!isset(NULL)) || (empty(NULL) && '' !== '0'`
`return (!TRUE) || (TRUE && TRUE)`
`return FALSE || TRUE`
`return TRUE`

for '0':
`return (!isset('0')) || (empty('0') && (string) 0 !== '0')`
`return (!TRUE) || (FALSE && TRUE)`
`return FALSE || FALSE`
`return FALSE`

for 0 (integer):
`return (!isset(0)) || (empty(0) && (string) 0 !== '0'`
`return (!TRUE) || (empty(0) && '0' !== '0')`
`return FALSE || (TRUE && FALSE)`
`return FALSE`

any other value:
`return (!isset(x)) || (empty(x) && (string) x !== '0'`
`return (!TRUE) || (FALSE && TRUE)`
`return FALSE || FALSE`
`return FALSE`

Why do we test for a `(string) value !== '0'`? `empty()` already counts the values `'0'` (string) and `0` (integer) as 'empty' because they are `falsy`. It thus appears that the intention of the `ifEmpty` function could be captured in the statement `return empty($this->values['value']);` but that is not what the code does.

PR #529, opened 2017-09-08
DS-4076: add timezone settings in user profile if enabled in settings

jaapjan:
When a user h as the permission to change his/her timezone it does not show up on the user edit form. Let's add this field to the form display of account settings.

HTT:

- [x] Login as LU
- [x] Go to your account settings and verify there is no timezone field.
- [x] Login as admin, enable timezone configuration for authenticated users on /admin/config/regional/settings
- [x] Login as LU and go to your account settings and change timezone.
- [x] Create some content and verify with another user that the displayed date is different depending on the timezone settings.

### DIFF
--- a/modules/social_features/social_user/config/install/core.entity_form_display.user.user.default.yml
+++ b/modules/social_features/social_user/config/install/core.entity_form_display.user.user.default.yml
@@ -18,6 +18,18 @@ third_party_settings:
         required_fields: true
         id: ''
         classes: ''
+    group_locale_settings:
+      children:
+        - timezone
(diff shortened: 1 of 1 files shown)


bramtenhove:
@jaapjan Just after following the steps in this PR and merging it I was looking at the signup process. Upon signup you have the email address and username field, but we now also have a timezone dropdown.

I assume that is desired behaviour, but just wanted to mention it to you. When checking the box to allow users to set their own timezone it already explicitly says it asks upon registration as well.

PR #534, opened 2017-09-12
#2900982 - Search 'all' tab won't show pages and books

jochemvn:
## Description
When searching for content on the 'all' tab in OS, book pages and basic pages don't show up. This is now fixed.

## Drupal
https://www.drupal.org/node/2900982

## HTT
- [x] Do a re-install from 8.x-1.x
- [x] Enable the book module
- [x] Create some book pages
- [x] Create some basic pages
- [x] Search for them and notice they don't show up on the 'all' tab
- [x] Notice that they DO show up on the content tab
- [x] Switch to this branch
- [x] do an updatedb (either interface or drush)
- [ ] Notice they now DO show up in the 'all' tab

### DIFF
--- a/modules/social_features/social_page/config/install/core.entity_view_display.node.page.search_index.yml
+++ b/modules/social_features/social_page/config/install/core.entity_view_display.node.page.search_index.yml
@@ -1,0 +1,47 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - core.entity_view_display.comment.comment.default
+    - core.entity_view_mode.node.search_index
+    - field.field.node.page.body
+    - field.field.node.page.field_content_visibility
+    - field.field.node.page.field_files
+    - field.field.node.page.field_page_comments
+    - field.field.node.page.field_page_image
+    - node.type.page
+  module:
+    - comment
+    - text
+    - user
+id: node.page.search_index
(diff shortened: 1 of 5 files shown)


jochemvn:
@bramtenhove good point. I think we should merge them both then... less work for me 👍

jaapjan:
@bramtenhove @jochemvn merging them both will not solve the issue. Remember that social_search_update_8103 is already in 1.4 and thus already executed by developers who actually have drush installed.

bramtenhove:
Good point @jaapjan. Sorry @jochemvn, you will still need to do change this.

jochemvn:
@jaapjan another good point.. I will add an update hook. I think I'll merge your PR first @bramtenhove.... otherwise we'll get a merge conflict

jochemvn:
@bramtenhove  update hook added

bramtenhove:
@jochemvn I checked all the steps, but unfortunately am still encountering the issue. Book pages do not show up in the Search All tab.

I created a book page and noticed at the social_all index that the index count went up with 1. But searching for either the book content or the book title did not give me any results.

bramtenhove:
Sorry @jochemvn, I didn't revert the features. Fine now!

PR #536, opened 2017-09-14
DS-3820 - Overview of content i follow

jochemvn:
## Note
This is purely the development part. This needs some theming by @MaikelGG. Please don't merge until that's done!

## Description
As a LU I want to have a general overview of all content I follow on the platform. I should be able to see and manage it in a single overview

## HTT
- [x]  Try the fresh install from this branch, but also using the update script
- [x]  Login as a SM+
- [x]  Create some pages
- [x]  Login as another user
- [x]  Follow some topics, events and some of the created pages
- [x]  Open the user menu and notice there a new item called 'Follow'
- [x]  Click it
- [x]  Notice you're now on an overview with all content you follow
- [x]  See that it's sorted by last followed on top
- [x]  See there are filter options in the sidebar (for each content type)
- [x]  See you can also sort on title and on type
- [x]  Click unfollow some content and see that the page is reloaded and the unfollowed item no longer shows up in the list
- [x]  Enable the social_book module
- [x]  As a SM+ create some books
- [x]  Login as the original LU again and follow some of the books
- [x]  Notice on the follow page, you see the books
- [x]  Notice you can now also filter on books.


MaikelGG:
@jochemvn I noticed the alias for this page is /following?user=7 where the number 7 should represent the user with that ID, but in fact it is the showing the items for the account you are using. So if I where to login with user ID 123 and go to /following?user=7 I get the results for user 123. I think it would be better that I see the correct results of user 7 if I have permissions for this (CM+) or I get an access denied (LU). This allows for CM to get some more insights or help members if that is needed on such a page.

jochemvn:
Hi @MaikelGG, 
The parameter in the URL was an error I made when I created the menu link. The parameter itself did not do anything. We've agreed that people can only see their own content. There's no permission for a CM/SM to view content that other people follow.

Ready for review again: @bramtenhove

modules/social_features/social_follow_content/social_follow_content.info.yml
```diff
@@ -3,17 +3,32 @@ description: 'Provides "Follow Content" flag type and related functionality.'
...
   - drupal:node
   - drupal:options
-  - social_event:social_event
-  - social_topic:social_topic
   - drupal:system
   - drupal:user
   - drupal:views
+  - dynamic_entity_reference
+  - dynamic_entity_reference:dynamic_entity_reference
+  - field
```
bramtenhove: @jochemvn according to the coding standards modules need to be prefixed. E.g. Drupal core modules should be: `drupal:field`. Other modules are something like `flag:flag`.

Please fix the dependencies in this file.

See https://www.drupal.org/docs/8/creating-custom-modules/let-drupal-8-know-about-your-module-with-an-infoyml-file.

bramtenhove:
@jochemvn next to my previous comment, I also seem to have basic issues installing this branch from scratch. I'm using the same reinstall alias as you are: `dreset`.

jochemvn:
@bramtenhove Can you be a bit more specific? What problems do you run into

bramtenhove:
We need to make sure the install_update job works as this will otherwise mess up all these build jobs in the other branches.

bramtenhove:
Functionality is tested and works. We just need to solve the build now.

PR #537, opened 2017-09-14
Issue #2908922 by ribel: Fix removing author and published date info from basic page

ribel:
**Related issue:** https://www.drupal.org/node/2908922

**HTT:**
- [ ] Create basic page
- [ ] See that Author and Date is not displayed


bramtenhove:
@ribel can you take a look at @MaikelGG's comment? This might be a possible duplicate.

ribel:
@bramtenhove yes, it looks like @MaikelGG provided a better solution.
We already use patch (diff) from this PR on the project, but I think you can close this PR, as it still should be usable by composer. And later we will switch to a patch from #533 or when it will be in new OS version.

PR #538, opened 2017-09-18
DS-4078 - Alternative e-mail logo to fix e-mail clients not supporting .svg images

jochemvn:


modules/social_features/social_swiftmail/social_swiftmail.module
```diff
@@ -67,5 +66,29 @@ function social_swiftmail_mail_alter(&$message) {
...
+  $request = \Drupal::request();
+  $host = $request->getSchemeAndHttpHost();
+  // Default we use the logo image.
+  $logo = $host . theme_get_setting('logo.url', $theme_id);
+
+  // It could be overridden here.
+  $email_logo = theme_get_setting('email_logo', $theme_id);
+
+  // Must be a non-empty array.
+  if (is_array($email_logo) && !empty($email_logo)) {
```
Kingdutch: Is it possible to support a string URI here as well? I'm not sure how difficult it is to generate this array. For example for most Enterprise sites we probably don't need a form for the site manager to change it (like implemented in SocialBlue) but we'd just want to set it to `/themes/custom/<enterprise>/logo.png`

bramtenhove:
@MaikelGG could you please check this pull request?

PR #539, opened 2017-09-18
Issue #2839327 - personalised homepage

jaapjan:
Changes I've made:
- Created a custom filter which is designed especially for the homepage stream:
-- Shows nodes outside groups user has access to
-- Shows nodes inside groups user is a member of and has access to
-- Shows posts inside groups user is a member of
-- Shows posts outside groups and user has access to
-- Shows comments on nodes inside groups user is a member of
- Disabled comment on posts in group in the homepage stream since this clutters the homepage too much imho.
- Behat tests.

HTT:

- [x] Re-install without demo content
- [x] Create two user accounts: user x, user y
- [x] Login as User X and create a group, one topic inside that group, one post inside that group and comment on both the post and the node.
- [x] Login as User Y and create a new group, one topic inside that group, one post inside that group and comment on both the post and the node.
- [x] Create some content outside of groups: one node, one comment on that node, one post and a comment on that post.
- [x] Now check the homepage stream as both these users and make sure the result is as expected.
- [x] Logout and check the AN homepage stream.
- [x] Create a Third user account and login as that user. Check the homepage stream and note that you only see items which is not posted in any group.

### DIFF
--- a/modules/custom/activity_viewer/activity_viewer.module
+++ b/modules/custom/activity_viewer/activity_viewer.module
@@ -1,0 +1,21 @@
+<?php
+/**
+ * ${NAME}
+ */
+
+
+/**
+ * Implements hook_views_data().
+ */
+function activity_viewer_views_data() {
+  $data['activity']['activity_filter_personalised_homepage_filter'] = array(
+    'title' => t('Filter on activities for the personalised homepage.'),
+    'filter' => array(
+      'title' => t('Filter activities for personalised homepage'),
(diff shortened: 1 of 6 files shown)


Kingdutch:
@jaapjan What would be the impact here if an enterprise (or distro) site has things like custom groups? Do they no longer show up in the stream?

jaapjan:
@Kingdutch they show up in the stream, but only if the user is a member of the group.

PR #541, opened 2017-09-19
Issue #2909958 by chmez: add checking if a user account has profile.

OH17L:
Issue: https://www.drupal.org/node/2909958

How to test:

```
drush pm-enable social_demo -y
drush sda create_test_user
```


jaapjan:
Thanks, that plugin (create_test_user) does not exist in the distro but the changes make sense and the demo content creation still works as expected. Thanks!

PR #546, opened 2017-09-21
Issue #2896168 by maikelkoopman: replace autogrow with autosize library

MaikelGG:
### How to test:
- create a new post with a few paragraphs
- edit the post, the textarea should be as tall as the text.
- See that autogrow behaviour still works as expected on comment fields


MaikelGG:
@bramtenhove After your comment I have have tested again:
- Post edit forms seem to work fine. This was also one of the acceptance criteria for this story. Maybe we can have a look together tomorrow?
- Node edit forms do not have this library attached, so I was wondering what you are testing. Can you explain the steps you took?

bramtenhove:
Hi @MaikelGG,

- I checked the Post edit form again, works fine now.
- I was simply monkey testing at the Node edit form. But it makes sense that it doesn't work there, 
I was just wondering.

PR #550, opened 2017-09-21
Issue #2875334 by maikelkoopman

MaikelGG:
Display image field default for new content types

Solution:
When a SM/developer creates a new content type, we do not know what the field name of the image will be, if they decide to name it image and this defaults to the machine name field_image or they reuse the image field of an existing type this solution will work and the image gets displayed. If they create a unique name, we cannot provide a template solution and manual theming is required.

We intend to create a follow up to look if we can use Display Suite module to allow a site builder role to use a new content type without having to add theming.

##how to test
- create a new content type
- add a image field and just call it image or reuse one of the existing image fields
- Configure the form and display
- Create new content
- Add the CT to an existing view (for example all-topics)
- Check the view: the fields should display like the other existing content type (image on the left)

https://www.drupal.org/node/2875334


themes/socialbase/templates/node/node--teaser.html.twig
```diff
@@ -80,15 +80,30 @@
...
     {% endif %}
 
-    {% block card_image %} {% endblock %}
+    {# allows for new content types to just use the plain field_image name #}
+    {% block card_image %}
+      {{ content.field_event_image }}
+      {{ content.field_topic_image }}
+      {{ content.field_page_image }}
+      {{ content.field_book_image }}
+      {{ content.field_image }}
```
bramtenhove: It won't work if I have named my image differently correct? E.g. `field_different_image`.

This will solve part of the problem for sure, but I can imagine we might need to make it more robust still. What do you think @MaikelGG?

MaikelGG:
@bramtenhove I have updated the template of teasers. Can you review again?

bramtenhove:
Thanks @MaikelGG, works fine now. I'll merge it.

Would be nice if you could take a look at the comment I placed earlier with regard to the naming of the image fields and how to proceed with that in the future.

PR #552, opened 2017-09-22
DS-4194 - Module updates

jochemvn:
## HTT

- [x] Check modules that are updated
- [x] Pay some extra attention to profile


bramtenhove:
@jochemvn as discussed, could you add the permissions in a way that we do not reset other permissions as is happening now? Cheers!

PR #554, opened 2017-09-27
Simplify `isEmpty` check for Dropdown widget

Kingdutch:
We currently have a very convoluted if statement in the `isEmpty` function which boils down to doing what `empty` does. We should thus just use `empty`.

A longer more complicated way of saying the above was first discussed in https://github.com/goalgorilla/open_social/pull/527#pullrequestreview-64358763


jochemvn:
hey @Kingdutch, could you check why this is failing? Maybe merge 8.x-1.x back into this branch could help.

Kingdutch:
@jochemvn I'm not sure why this is failing. As far as I can deduce this shouldn't break anything :(

bramtenhove:
@Kingdutch I see @jochemvn merged back `8.x-1.x` into this branch a few days ago. Unfortunately the tests are still failing. It doesn't look like a random fail, could you take a look? Maybe see if these failing tests works locally?

Kingdutch:
The latest commit slightly alters the requirements for the Dropdown::isEmpty check

Nowhere in the Dropdown do we require that the value that is 
supplied is actually an integer. This causes some implementations
where we pass in a string '0' instead. To account for this we
change our checks to see where we either have no value at all or
where we have a value but it's neither an integer nor a string.

The previous check also already did this but only for the value '0' so 
it would break for the value '1' even though 1 would be a valid value.

PR #531, opened 2017-09-11
DS-4122 - Move front-end libraries to composer.json

bramtenhove:
## Description

This PR has a lot of changes, sorry for that. However we already split it up a little bit, see https://github.com/goalgorilla/open_social/pull/519.

The idea is to remove all of the front-end libraries from the codebase and require them via a package manager. This means we removed a lot of code! And @MaikelGG rewrote the search script to make it work.

See also PR https://github.com/goalgorilla/open_social/pull/530 and https://github.com/goalgorilla/drupal_social/pull/320.

## Important

Make sure you followed instructions at https://github.com/goalgorilla/open_social/pull/530. 

## How to test

If https://github.com/goalgorilla/drupal_social/pull/320 is merged, it makes it easier for you.
- [x] Install this new version, see that the assets are now installed in html/libraries
- [x] Also install it via the makefile, it should also be in html/libraries
- [x] Make sure **all** tests pass, be aware that the travis builds will fail unless https://github.com/goalgorilla/drupal_social/pull/320 is merged

Please note that we did not really update any library versions, except for Waves. Morris and Raphael didn't change in version either and are only used by the analytics module, so hard to test. Now test the different libraries by:
- [x] Waves library by clicking on buttons (should cause a wave in the button)
- [x] Timepicker library by creating an event, you can pick the time via this tool
- [x] Click through the site for the Bootstrap library, create a group, join an event, check tabs on user profile page, etc.

PR #556, opened 2017-09-28
DS-4228 - Remaining coding issues

frankgraave:
# Background
So this is another one of those coding standards improvements pull requests for the benefit of [our security advisory application](https://www.drupal.org/node/2898988). Since we now moved all the libraries to be included in a proper and better maintainable way (https://github.com/goalgorilla/open_social/pull/531) we have had a better [overview on the remaining true issues and warnings](https://pareview.sh/node/2789) that we should fix. 

### Why are there so many commits?
Reason for this is to be safe and maintain a better overview about making all the changes with a lot of dependency injections, which can cause some trouble sometimes.

# HTT
Kinda hard to test a PR like this, but in general all the tests should pass.


frankgraave:
Closing this PR and opening a new one for this. Somehow the PR decided the feature state is not default anymore.

PR #560, opened 2017-10-02
Issue #2910851 - Move cropper library

frankgraave:
Related issue: https://www.drupal.org/node/2910851

# HTT
- [ ] Set this branch in your composer and do an composer update.
- [ ] See that `"fengyuanchen/cropper": "dev-master"` is being included in the libraries folder.


jochemvn:
Removing the library is one thing. Updating to v2.0 is a bit tricky. We cannot just do this and assume all goes well for enterprise and distro users. Need some extra investigation. Perhaps also moving to focal point is an option

PR #561, opened 2017-10-02
Issue #2913269 by ribel: Fix group_link variable when node is referenced from other node

ribel:
**Related issue:** https://www.drupal.org/node/2913269

**HTT**
- [x] Create Content type A which has reference field to Content type B
- [x] Create instances of Content type B can in the group
- [x] Display Teasers of Content type B in full view of Content type A
- [x] Use variable `group_link` from `socialbase` in teaser templates of Content type B
- [x] See that `group_link` is not displayed
- [x] Apply this patch and check that `group_link` is displayed


jochemvn:
Hey @ribel, Could you provide some more information both here and on Drupal.org?

ribel:
Hi @jochemvn,
I have updated the description on parent node added HTT here. It looks complicated, but by the fact, it's simple fix which should is backwards compatible and shouldn't break anything.
We use this patch on one of the project.

PR #563, opened 2017-10-03
Maintenance - Bugfixes that were caused by coding standards

frankgraave:
# Background
The constants that were moved during previous coding standards fixes, were moved to the wrong class. This is now fixed. Besides that the search was not working since it had the wrong dependeny, also fixed.

# HTT
- [ ] Check that the search is working (confirm the correct dependency is used)
- [ ] See you won't get a notice on your local environment about social mention constants any longer


jochemvn:
@frankgraave 
No need to put these in the release notes, since they were caused in this release in the first place

PR #564, opened 2017-10-03
DS-4228 - Fix more coding standards

frankgraave:
# Background
There are still 36 warnings open, but with this PR all coding standard **errors** should now be fixed! Now shooting in this PR to keep the coding standard changes smaller per PR.

### DIFF
--- a/modules/custom/social_demo/src/DemoSystem.php
+++ b/modules/custom/social_demo/src/DemoSystem.php
@@ -21,14 +21,14 @@ abstract class DemoSystem extends DemoContent {
   /**
    * The user storage.
    *
-   * @var
+   * @var \Drupal\Core\Entity\EntityStorageInterface
    */
   protected $blockStorage;
 
   /**
-   * the Config factory.
+   * The Config factory.
    *
-   * @var
+   * @var \Drupal\Core\Config\ConfigFactory
    */
   protected $configFactory;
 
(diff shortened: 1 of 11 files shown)


modules/custom/social_demo/src/Plugin/DemoContent/Book.php
```diff
@@ -143,49 +143,4 @@ protected function prepareImage($uuid) {
...
 
-  /**
-   * The function for creating a book from the demo content.
-   *
-   * @param \Drupal\node\Entity\Node $entity
-   *    The related entity.
-   * @param $book
-   *    The book.
-   */
-  public function createBookLink(Node $entity, $book) {
```
bramtenhove: @frankgraave, I assume this function was no longer necessary after refactoring took place? Is that why it is removed?

modules/custom/social_demo/src/DemoSystem.php
```diff
@@ -89,23 +89,24 @@ public function createContent() {
...
-      // This uuid can be used like this since it's defined in the code as well (@see social_core.install)
+      // This uuid can be used like this since it's defined
+      // in the code as well (@see social_core.install).
       $block = $this->blockStorage->loadByProperties(['uuid' => '8bb9d4bb-f182-4afc-b138-8a4b802824e4']);
       $block = current($block);
 
       if ($block instanceof BlockContent) {
 
-        $this->replaceANblock($block, $data['homepage']);
+        $this->replaceAnBlock($block, $data['homepage']);
```
bramtenhove: OK! I found no other calls to this method.

modules/custom/social_demo/src/Plugin/DemoContent/Book.php
frankgraave: You are correct! It's no longer necessary since @jochemvn worked his magic on refactoring it in a cleaner way!

PR #566, opened 2017-10-05
2913784 No admin profile in overviews (search / members) AND 2913513 forward slash

jochemvn:
## Description
The admin user account has a profile entity since 1.4. A negative side effect is that the admin profile is now shown in the following overviews:
- newest members overview
- newest members block on the homepage
- search all
- search members

## Solution
2 solutions. For the newest member views user 1 is filtered out. For the search indexes all and users, the admin user will no longer be indexed in there. 
Also a behat scenario has been added in search_users.feature to check you cannot find admin user

## Drupal
https://www.drupal.org/node/2913784

## HTT
- [x] Check outside this branch first, and notice that 'admin' is in the views mentioned above
- [x] Check from update
- [x] Check from fresh install
- [x] Check all the views mentioned and also check the search indexes not containing the admin profile anymore.
- [x] Also check that using a forward slash will not break the search


modules/social_features/social_search/config/install/views.view.search_all.yml
```diff
@@ -42,15 +42,17 @@ display:
           sort_asc_label: Asc
           sort_desc_label: Desc
       pager:
-        type: mini
+        type: full
```
bramtenhove: @jochemvn was this intentional?

jochemvn: yes. the all search pager was inconsistent with the other searches

PR #568, opened 2017-10-06
#2914343 Edit icon is missing from post & comment

MaikelGG:
Log in as a normal user and...

- [x] Create a post with an image. When viewing the post (not the form) the image is not elevated
- [x] When looking at the post on the right top corner I see a arrow which I can click to edit the post 

https://www.drupal.org/node/2914343


frankgraave:
Don't know why, but all the tests fail on the PR because of the feature state @MaikelGG

PR #569, opened 2017-10-06
Issue #2913198 by bramtenhove: Update to automatically set cropping for new imagestyle

bramtenhove:
## Description
We introduced a new image style and cropping configuration in version 1.4. Upon updating the cropping configuration is not added to the new image style. This update will fix it for anybody that hasn't done it manually yet.
See request at https://www.drupal.org/node/2913198.

## How to test
- [ ] Install version 1.3
- [ ] Look at the AN hero image (remember the cropping)
- [ ] Switch to this branch
- [ ] Run `drush updb -y` and `drush fra -y`.. just like normal
- [ ] Observe that the image is nicely cropped in the same manner
- [ ] Now remove the image and upload one more (*do not crop it*)
- [ ] Re-run the update hook
- [ ] Observe that nothing happened as we have nothing to do

### DIFF
--- a/modules/social_features/social_core/social_core.install
+++ b/modules/social_features/social_core/social_core.install
@@ -6,6 +6,8 @@
  */
 
 use Drupal\block\Entity\Block;
+use Drupal\block_content\Entity\BlockContent;
+use Drupal\crop\Entity\Crop;
 use Drupal\Core\Database\Database;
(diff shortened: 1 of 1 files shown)


jochemvn:
@bramtenhove better use: /var/www/scripts/social/install/update.sh instead of 'run drush updb -y and drush fra -y.. just like normal'

PR #575, opened 2017-10-11
Issue #2910851 by chmez: Update the Image Widget Crop module to version 2.0

OH17L:
## Problem
If we have two widgets with cropping on a page, the second field doesn't work properly.
The second field gets settings from the first field. Also if we have that widget inside a paragraph, the selected area does not persist.

## Solution
Update the Image Widget Crop module to 2.0 version.

## Issue tracker
https://www.drupal.org/node/2910851

## HTT
- [ ] Open page to edit some initiative
- [ ] Open section for crop image in **Image** field
- [ ] You do not see borders of an area for crop image


bramtenhove:
Thanks! I noticed there is a merge conflict that needs to be resolved. Could you take a look at it?

OH17L:
Check again, please.

PR #578, opened 2017-10-12
#2912755 - Enable inline form errors module and add styling

jochemvn:
## Description
Inline form errors will be stable in release 8.4. Let's add support for it in the social themes. This will make the overal experience a lot better.

## Drupal
https://www.drupal.org/node/2912755

## HTT

- [ ] login as admin and Go to extend and enable inline form errors module
- [ ] login as a normal user
- [ ] Go to Add Event form and submit an empty form. You should see well themed inline form errors which are dismissible alerts. Also on top you see a message with links. The links should be bring you to the correct field. 
- [ ] Go to edit your profile and leave all fields blank and submit the form. Look at the form feedback and inline form error placement. 

Note: The module is not turned on by default. The order of errors in the message is not in the correct order. This is not part of this story.


jochemvn:
@MaikelGG, please add some test steps.

If the tests are green you can merge

PR #579, opened 2017-10-13
DS-1050 Notification emails in digest form

bramtenhove:
## Settings functionality
Please be aware that we split the changes for the settings form from this digest functionality. This code is in this branch, but has already been reviewed and merged. Please see https://github.com/goalgorilla/open_social/pull/577.

## Changes

- Refactored send email queue worker to just pass the queue item along the EmailFrequency class
- Immediately class now takes care of sending immediate emails
- _Digest_ classes (daily, weekly) now add the emails to a new database table
- New cronjob that checks if daily or weekly emails need to be sent (added to new queue)
- New queue worker that sends digest emails
- Emails are now added in a template

## How to test

*Pro tip*
You can change the intervals manually for the _Daily_ and _Weekly_ emails so you can test it better. Simply change to the `interval` setting in the EmailFrequency class to a shorter period than a day or a week.

- [x] Make sure you have at least ran the update script (there are new tables to be installed)
- [x] Go to user *X* and change the notification settings to cover all options

- [x] Stay logged in as user *X* and log in as user *Y*
- [x] Trigger an action that would send a notification to user *X* `immediately` (ps. run `drush cron` if needed)
- [x] Observe that user *X* does not receive the notification via email
- [x] Log out as user *X* and repeat above
- [x] Observe that user *X* now received an email with the notification

- [x] As user *Y* trigger an action that should not send a notification to user *X* (e.g. he has the trigger set to _None_)
- [x] Observe that user *X* does not receive an email

- [x] As user *Y* trigger multiple actions that would send a daily mail to user *X*
- [x] Observe that user *X* receives a digest mail after some time

- [x] As user *Y* trigger multiple actions that would send a weekly mail to user *X*
- [x] Observe that user *X* receives a digest mail after some time

- [x] Make sure the interval of the daily and/or weekly mail are set to their original settings and that the day or week hasn't passed yet
- [x] As user *Y* trigger multiple actions that would send a daily or weekly mail to user *X*
- [x] Observe that no email is sent to user *X* yet


bramtenhove:
@MaikelGG could you fix the merge conflict?

Also please note that we still have to fix one link in the email.

modules/custom/activity_send/modules/activity_send_email/templates/digestmail.html.twig
```diff
@@ -0,0 +1,11 @@
+<h3>{{ notification_count }}</h3>
+
+<ul>
+    {% for notification in notifications %}
+        <li>{{ notification|striptags('<a><strong>')|raw }}</li>
+    {% endfor %}
+</ul>
+
+<p>&nbsp;</p>
```
jochemvn: 🙈

modules/custom/activity_send/modules/activity_send_email/templates/directmail.html.twig
```diff
@@ -0,0 +1,5 @@
+{{ notification|striptags('<p><a>')|raw }}
+
+<p>&nbsp;</p>
```
jochemvn: 🙈

jochemvn:
One more thing about the email. I don't get the link in the bottom setion. Also the list items wrap a bit weird, or is this intended? https://screencast.com/t/m8FaqfPY3CNU

modules/custom/activity_send/modules/activity_send_email/activity_send_email.module
```diff
@@ -115,6 +116,10 @@ function activity_send_email_form_user_form_alter(&$form, FormStateInterface $fo
       ],
     ];
 
+    $form['email_notifications']['description'] = [
+      '#markup' => t('This requires a nice description.'),
```
bramtenhove: This still requires a nice description!

bramtenhove:
Thanks for your comments @jochemvn.

I fixed several issues:

- Link in footer email for digest now works
- Text in email now reacts to notification counts, either singular or plural
- I made it explicit the notifications in the digest are sorted by created date, although apparently they already were sorted correctly

I added a todo to add an explanatory text in the fieldset.

The specific time on which the emails are sent is indeed not been determined. It will now be send on the moment the cron runs the first time. We'll have to create a followup to make it configurable.

PR #580, opened 2017-10-13
2915997 Fix redirect to edit profile after edit account on register

OH17L:
## Problem
Do not redirect to edit profile after edit account on register

## Issue tracker
https://www.drupal.org/node/2915997

## HTT
- [x] Go to on-site how anonymous user
- [x] Click **Sign up** link in the top bar
- [x] Fill required fields on **Create new account** page
- [x] Press **Create new account** button
- [x] Go to on-site from a link of E-mail letter which you received
- [x] Press **Log in** button on **Set your password** page. Before doing this you must sure that user creation time and user update time are different
- [x] Fill password and press **Save** button

After that, you redirecting to stream page but you should be on profile edit page.


ribel:
It's a weird bug, but we didn't find better solution for now. Workaround provided by @OH17L is better than current state.

jochemvn:
hi @OH17L,
I've written a comment on d.org. Could you please elaborate a bit more on this issue?
Jochem

bramtenhove:
I'm reluctant to merge this in the distro. We should rather see if we can fix this properly. Using timings is a bit dodgy.

Is your problem that you need to check if the user has never logged in before? Isn't it an option to check if the profile was ever edited? Or maybe check the access count of the user account?

OH17L:
Hi @bramtenhove 
Look my updates again, please.

OH17L:
Check again, please.

ribel:
@OH17L Need to add check if user Profile exists. In case if user don't have profile it gives error after password resed, and it's not possible to reset password:

`Error: Call to a member function getCreatedTime() on boolean in social_user_form_user_form_alter() (line 74 of /app/html/profiles/contrib/social/modules/social_features/social_user/social_user.module) #0 `

OH17L:
@basvanos It does not work when you have some custom code which changes users. For example, you can have hook_cron which will do something with user entities and when this hook executed between pressing **Create new account** and using one-time login link from an e-mail then changing the date of new user will be updated. So, a user creating date and user changing date is different and redirect to profile edit form will not be work.

bramtenhove:
Thanks @OH17L for the explanation.

PR #581, opened 2017-10-13
DS-4215 url embed in posts and comments

jochemvn:
## Description
URLs of known providers should not be converted to links, but to iframed embeds. 

## HTT
- [x] Install the social_embed module
- [x] Test by posting and commenting
- [x] Test links from the following providers: facebook, flickr, instagram, spotify, twitter, vimeo, youtube, ted and linkedin


modules/social_features/social_embed/social_embed.info.yml
```diff
@@ -0,0 +1,8 @@
+name: Social Embed
+type: module
+description: Social Embed for embedding social media links
+core: '8.x'
+package: Social (experimental)
+dependencies:
+  - drupal:embed
```
bramtenhove: This should be `embed:embed` and `url_embed:url_embed`. I will update it.

modules/social_features/social_embed/src/Plugin/Filter/SocialEmbedUrlFilter.php
```diff
@@ -0,0 +1,79 @@
...
+    return [
+      'facebook.com\/(.*)\/videos\/(.*)',
+      'facebook.com\/(.*)\/photos\/(.*)',
+      'facebook.com\/(.*)\/posts\/(.*)',
+      'flickr.com\/photos\/(.*)',
+      'flic.kr\/p\/(.*)',
+      'instagram.com\/p\/(.*)',
+      'open.spotify.com\/track\/(.*)',
+      'twitter.com\/(.*)\/status\/(.*)',
+      'vimeo.com\/\d{9}',
```
bramtenhove: Can also be 7 or 8 digits:
https://vimeo.com/3113655
https://vimeo.com/31136871

modules/social_features/social_embed/src/Plugin/Filter/SocialEmbedUrlFilter.php
```diff
@@ -0,0 +1,79 @@
...
+      'youtube.com\/watch[?]v=(.*)',
+      'youtu.be\/(.*)',
+      'ted.com\/talks\/(.*)',
+      'linkedin.com\/in\/(.*)',
```
bramtenhove: Doesn't really work, maybe we should remove it?

MaikelGG: Linkedin? Yes maybe we shoudl remove it for now

bramtenhove:
I tested this feature, code works fine.

1. Do we inform the end-user which links are currently supported (LGOS, description)? Should we do this later perhaps as this is still very experimental?
2. The embed from Spotify is quite large, it fills up the whole screen, it isn't looking very nice
3. The site feels quite slow, I think it has to do with the functionality we introduced here

PR #549, opened 2017-09-21
Issue #2910585 by ribel: Disable Restrict images for Basic HTML text format

ribel:
**Related issue:** https://www.drupal.org/node/2910585

**HTT**: 
- [ ] Try to upload image using Basic HTML text format at any NOT local site
- [ ] See that image was removed and replaced with warning placeholder
- [ ] Apply patch, and see that you can upload image now

PR #582, opened 2017-10-13
Feature/2890462 support inline figures

MaikelGG:
How to test:

- [x] create a node with some text and insert images with captions and different alignments
- [x] after inserting the image, do not forget to insert the caption text
- [x] also test with long caption texts
- [x] use both small and large images in your testing
- [x] save the node and test how it looks on mobile and desktop sizes

Related issue: https://www.drupal.org/node/2890462


MaikelGG:
@jochemvn The problem is caused by the filter_filter_secure_image_alter function. In /admin/config/content/formats/manage/basic_html the following filter blocks the image from debing displayed. You can also see in the inspector that it has been filtered.

screen shot 2017-10-17 at 11 20 09

I assume the other PR is a fix for this bug? https://github.com/goalgorilla/open_social/pull/549/ @ribel ?

jochemvn:
@MaikelGG, I don't know if that's the case. The images are displayed. But they're just very tiny.

MaikelGG:
Yes that is what the filter does, check out filter_filter_secure_image_alter in filter.module. That is what you see (16x16 images). So I think we should continue this story with either first merging the other bug and then test again or uncheck the filter in the basic html first. At least don't let this bug block this feature, as it is not related to this feature. I will update the PR solving the conflict.

jochemvn:
👍  and please test the other story too...

MaikelGG:
@jochemvn you can test again

PR #583, opened 2017-10-13
DS-3906 - Themed mails when send mails from admin theme

jochemvn:
## Description
When sending emails as an admin, the theme used for the emails would be the admin theme.. This is now changed to the default theme.

## HTT
- [x] Login as admin
- [x] Create a new user (and notify the new user)
- [x] See that the new mail is now well formatted
- [x] Also unblock a bunch of users in bulk
- [x] See that those mails are also well formatted
- [x] Should work from update or from fresh install

### DIFF
--- a/modules/social_features/social_swiftmail/social_swiftmail.install
+++ b/modules/social_features/social_swiftmail/social_swiftmail.install
@@ -18,9 +18,18 @@ function social_swiftmail_install() {
   $swift_settings->set('format', 'text/html')->save();
   $swift_settings->set('respect_format', FALSE)->save();
   // Alter mailsystem settings.
-  $mailsystem_settings->set('theme', 'current')->save();
+  $mailsystem_settings->set('theme', 'default')->save();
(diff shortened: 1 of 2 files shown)


modules/social_features/social_swiftmail/social_swiftmail.install
```diff
@@ -18,9 +18,18 @@ function social_swiftmail_install() {
...
   $mailsystem_settings->set('defaults.sender', 'social_swiftmailer')->save();
   $mailsystem_settings->set('defaults.formatter', 'social_swiftmailer')->save();
   $mailsystem_settings->set('modules.swiftmailer.none.sender', 'social_swiftmailer')->save();
   $mailsystem_settings->set('modules.swiftmailer.none.formatter', 'social_swiftmailer')->save();
 }
+
+/**
+ * Set which theme to use when sending emails.
+ */
+function social_swiftmail_update_8002(&$sandbox) {
```
bramtenhove: @jochemvn shouldn't this be `8001`? 🤔

PR #585, opened 2017-10-16
DS-4253 - Fix search issue with forward slash

frankgraave:
# Information
When pasting only an forward slash, the preg_replace strips the / and then wants to build an url->toString on NULL which caused an WSOD. I restructured the if/else statements to check for this and act accordingly. I also removed double statements in the breadcrumbs regarding search.

# HTT
- [x] Checkout the code changes
- [x] Search for `"/"` and see an WSOD.
- [x] Search for anything else and see the breadcrumbs have Duplicate text Duplicate text.
- [x] Checkout this branch: `git checkout feature/DS-4253-support-fix-search`
- [x] Verify that the above issues are no longer there.


bramtenhove:
@frankgraave @jochemvn What was the reason for replacing these slashes with a space?

Have you considered how this affects searching for an url like `https://github.com/goalgorilla/open_social/pull/585`? I can be find the content, but the input is changed into `https: github.com goalgorilla open_social pull 585`.

PR #586, opened 2017-10-17
Issue #2900559 by Kingdutch: clarify content reach of open groups

Kingdutch:
https://www.drupal.org/node/2900559

When creating a group, the Open Group option says "This is a open group. Users may join without approval and all content added in this group will be visible for non members as well."

Only LUs can see the Topics and Events, so when you say "non members" you're referring to the Group rather than to the community, which isn't immediately obvious. Perhaps to "This is an open group. Users may join without approval and all content added in this group will be visible to all community members."


jochemvn:
@Kingdutch, better also fix the behat (group-create-open) test then ;)

PR #587, opened 2017-10-18
Issue #2916984 by chmez: Fix deleting post comment.

OH17L:
## Problem
After deleting the comment to post on some group I see following error:
> **Warning**: Invalid argument supplied for foreach() in **social_comment_upload_comment_delete()** (line **73** of **profiles/contrib/open_social/modules/social_features/social_comment_upload/social_comment_upload.module**).

## Issue tracker
https://www.drupal.org/node/2916984


jochemvn:
Hey @Kingdutch and @OH17L,

And here's developer nr 3 with his own opinion on this: 

I'm not really happy with the idea of leaving a Drupal hook based on the a multi field being filled out or not. Potential new code that goes under that check might not be executed if the comment field is not filled out.

Since the foreach loop will not execute anyway if no value is in the field, we might just as well only check for the availability of the field, so:

`
if ($entity->hasField('field_comment_files')) {
    foreach ($entity->field_comment_files as $filefield) {
      // Try to load the actual file object.
      $file = File::load($filefield->target_id);
      // Check if it's an actual file object.
      if ($file instanceof File) {
        // Delete the file object.
        $file->delete();
      }
    }
  }
`

Kingdutch:
@jochemvn To avoid illegibility due to if statement nesting I usually prefer a guard clause at the top of a function. I understand your point about not wanting to risk someone not noticing this when extending the hook.

Would a middle road of taking the original code (with guard if nonexist -> return) into a `social_comment_cleanup_comment_files($entity)` function and just calling that from the hook implementation be a good middle road?

OH17L:
Check again, please.

jochemvn:
Hmm too bad this is still open. I kind of assumed someone was going to fix this, but apparently not. Marking it as needs work still

Kingdutch:
Missed this, assigning myself to pick it up coming Monday

Kingdutch:
@jochemvn Had some time to implement the discussed solution. Hopefully we can finally resolve this longstanding PR :)

PR #592, opened 2017-10-20
Issue #2917616 by RobertRagas, bramtenhove: Search index Social_users not indexing properly if updating from rc8.

bramtenhove:
## Description
When upgrading from Open Social **rc8** to 1.0 or above it could happen that user search index was not indexing users anymore. Installs of 1.0 or above do not have this problem.

Just fixing the profile entities does not solve it. Afterwards the index has to be disabled and enabled to actually force the profile entities to be properly indexed.

## HTT

- [ ] In the `profile` database table change the values of the column `is_default` to NULL for a few cases
- [ ] Run the update hook, check the database table to see the value is set to 1
- [ ] See that the user search index is enabled and reindexed (or still indexing)

### DIFF
--- a/modules/social_features/social_search/social_search.install
+++ b/modules/social_features/social_search/social_search.install
@@ -169,3 +169,76 @@ function social_search_update_8105() {
     }
   }
 }
+
+/**
+ * Set all existing profiles as default to fix search indexing.
+ */
+function social_search_update_8106(&$sandbox) {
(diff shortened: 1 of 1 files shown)


bramtenhove:
I also included an update hook that ensures the `Skip user 1` Search index processor is reverted before the index is cleared and reindexed.

PR #594, opened 2017-10-23
DS-4304 by frankgraave: update the search_api again since someone mer…

frankgraave:
During a merge conflict, the search_api was set back to 1.4. re-update it again in composer.json.


bramtenhove:
See original PR at https://github.com/goalgorilla/open_social/pull/584.

PR #595, opened 2017-10-24
Feature/2853910 questions about activity stream

miechiel:
The problem was that when select the the activity "ContentInMyGroupActivityContext" we got an wsod. We renamed the file to make sure the dependency injection is working properly. 

I also changed the plugin so that it actually is giving back the user of a group. 

See https://www.drupal.org/node/2853910


jochemvn:
Merged this into: https://github.com/goalgorilla/open_social/pull/624

PR #602, opened 2017-10-25
2918046 Better description when user removed from group

valgibson:
Try testing this solution:

- Make sure you have the right permission to remove a member from a group
- Go to a group membership overview page (e.g. group/4/membership)
- Verify that when the operations dropdown of a member was clicked, the word 'delete' has been replaced with 'remove'
- Click on 'remove' 
- In the following screen (e.g. group/4/content/20/delete): notice the clear page title, and the better handler description.


jochemvn:
Merged this into: https://github.com/goalgorilla/open_social/pull/623

PR #604, opened 2017-10-27
Move content attribute classes to preprocess function.

kedramon:
As was described in https://www.drupal.org/node/2907221 there is no way to change content_attributes classes array as they are pre-populated in template.

All logic was moved to preprocess, so now user can alter in own theme this classes like this:
```
use Drupal\Core\Template\Attribute;

function THEME_preprocess_page(&$variables){

  if ($variables['content_attributes'] instanceof Attribute) {
    if (SOME_CONDITION) {
      $variables['content_attributes']->removeClass('layout--with-complementary')
    }
  }
}
```
Also `<section>` is now in a twig block.


bramtenhove:
I just restarted the build, let's see what happens now.

PR #605, opened 2017-10-30
DS-3885 by frankgraave: make icons in AccountHeaderBlock configurable

frankgraave:
# Information
With the private message feature coming up, there's an additional icon in the main navigation bar. When viewing the site on mobile, this extra icon can make the navbar twice as big. So this should now be configurable.

# HTT
- [x] Check out the code
- [x] Checkout this branch `git checkout feature/DS-3885_icon-configuration`
- [x] Go to `/admin/config/opensocial/navigation-settings`
- [x] See that you can determine to display the **"My Groups"** icon and the new **"Social Private Message"** icon. _(**Note:** the setting for the private message icon is only shown when the module is enabled)_
- [x] Switch one or both off
- [x] See that the icon is no longer shown in the main navigation


modules/social_features/social_user/src/Plugin/Block/AccountHeaderBlock.php
```diff
@@ -155,29 +163,33 @@ public function build() {
...
-          'link_attributes' => '',
-          'icon_classes' => 'icon-group',
-          'title' => $this->t('My Groups'),
-          'label' => $this->t('My Groups'),
-          'title_classes' => 'sr-only',
-          'url' => Url::fromRoute('view.groups.page_user_groups', [
-            'user' => $account->id(),
-          ]),
-        ];
+        if ($navigation_settings_config->get('display_my_groups_icon') === 1) {
```
jochemvn: module exists wel laten staan

PR #621, opened 2017-11-07
Fix logic after move content attribute classes to preprocess function.

kedramon:
Fixed logic from suggestion: https://www.drupal.org/node/2907221#comment-12331052


kedramon:
@MaikelGG please check

jochemvn:
Merged into: https://github.com/goalgorilla/open_social/pull/622

PR #625, opened 2017-11-08
#2853910 - Fix group activity context

jochemvn:
## Problem
The problem was that when select the the activity "ContentInMyGroupActivityContext" we got an wsod. 

## Solution
Fixed the WSOD, renamed the file to make sure the dependency injection is working properly. Also changed the plugin so that it actually is giving back the user of a group.

## Issue tracker
https://www.drupal.org/node/2853910

## HTT
- [x] Checkout the code
- [x] Checkout this branch and clear caches
- [x] Create a new message using this activity context
- [x] Enable for entities: group_content and posts
- [x] Use notifications as destination
- [x] Test with 2 group members
- [x] Create some content as user X
- [x] See notification as user Y


jochemvn:
Merge when green

jochemvn:
@miechiel 💪🏼

PR #626, opened 2017-11-08
DS-4292 - embed performance improvement

jaapjan:
## Problem
When using the URL embed module and sharing embeds in Posts there is a significant performance issue. This is due to the fact that:

1. There is no render cache for the entity
2. URL Embed and Embed filters both make a call to the Embed service which does a CURL.

## Solution
@bramtenhove and I looked into the cache tags and how we can make sure that Posts are 'cacheable'. We removed the unnecessary 'timezone' cache context from the Post entity and in addition we removed the `$variables['#cache']['max-age']` from the `template_preprocess_post`. The first one is not necessary because the time ago is always a few seconds or minutes or days ago, etc. The timezone does not matter for the time ago. We noticed that the cache max age did not any value since the time ago still works when we remove the cache max age for posts.

**Note:**

- Caching only works in Streams. Not on Post full display.
- Caching is user-specific. You can login as user X and build cache for user X. But when you login as user Y you can not use the cache of user X. (so: first load is potentially still slow). We find this acceptable for now and fixing the caching completely seems out-of-scope. The next couple of months we will work on this when we will work on BigPipe support.

## Issue tracker
- https://www.drupal.org/node/2867668 we can move this issue forward in URL_Embed issue queue by changing the status to 'reviewed and tested by the community'.
- https://jira.goalgorilla.com/browse/DS-4292

## HTT
- [x] Enable social_embed module
- [x] Enable caching on your local machine (settings.php)
- [x] Create a Post with a youtube video
- [x] Open the stream a few times. Notice the performance difference on first load and second load.
- [x] Check if the post is in cache_render table (query by cid CONTAINS "post:X" OR data CONTAINS "youtube.com")


jochemvn:
Merge when green

PR #627, opened 2017-11-08
Issue #2921915 by chmez: As a LU I want to enroll to an event without having to go through group sign up

OH17L:
## Problem
As a LU I want to enroll to an event without having to go through group sign up

## Solution
1. Enroll button should be visible for users that are not in the group.
2. Automatic enroll people to groups when they enroll to events that are part of the group.

## Issue tracker
https://www.drupal.org/node/2921915

## HTT
- [x] Login as a SM
- [x] Go to **Event settings** page (/admin/config/opensocial/event) 
- [x] Select **Open group** and press **Save configuration**
- [x] Create open group and add event to this group
- [x] Log in how default user
- [x] When you go to created event page then you should see **Enroll** button
- [x] When you click **Enroll** then after that you should be enrolled to event and you should be joined to group of this event


jochemvn:
Please check the coding standards. 
docker exec -it social_web bash /var/www/scripts/social/check-coding-standards.sh

jochemvn:
Also please use the PR template that was originally in the PR field when you created it, so with 
Problem / Solution / Issue tracker / HTT

OH17L:
@jochemvn Coding standards were fixed. Check again, please.

jochemvn:
@OH17L, thanks for your changes. Really appreciate the effort! Will look into this PR shortly! 👍

ribel:
I have tested this for Pachamama project and it works as expected!
But I think it would be good to add some behat tests for this new functionality.
@jochemvn can Shipsters team do it? Since we don't have budget left for this feature.

jochemvn:
*Changes made:*
1. On event settingsform. I've built an extra check to see if the group types even use events. If not, don't show them on the settings form.
2. EventSettingsform now has its own permission. This is only given to sitemanager and NOT to contentmananger, which do have the permission that was previously used. Don't think we want contentmanagers to have this permission anyway.
3. The default settings for the distro is set, that it's tuned on by default for open_group. I've disabled this and it's now not turned on for any group type.

*Impact:*
1. No impact, just a 'fix'
2. If you want CM to have the permission in pachamama, you need to re-set it there.
3. Since this feature is already enabled on pachamama, the update hook responsible for the default setting will no be run again, so you should be good for this one.

*Conclusion:*
Only thing to look at is CM permission for this item.

ribel:
All changes looks good to me. And I think we don't need to do any additional actions on pachamama.

jochemvn:
Allright. Merge when green!

PR #634, opened 2017-11-10
DS-3666 - Embed in WYSIWYG (merge after DS-4292)

jaapjan:
## Problem
We can not add embeds in WYSIWYG enabled fields for now.

## Solution
Added a button to the WYSIWYG editor for basic html and full html. This will allow users and content managers to add embed just by pasting in a link via the Embed button.

Created a Behat test as well.

For basic_html it will convert normal links (which are not added via the "LINK" button in the Editor) to Embed objects. For full HTML i disabled this. I think it is desired/acceptable behaviour. Follow-up for that in: DS-4418

## Issue tracker
- DS-3666

## HTT
- [x] Enable social_embed module
- [x] Login as user X and create a topic
- [x] Click on the "Embed"/"URL" button in the Body description field
- [x] Paste in a youtube link: https://www.youtube.com/watch?v=kgE9QNX8f3c or a Twitter link
- [x] Save the node and notice that it is working as expected.
- [x] Test more edge-cases by pasting in other links, normal twitter and youtube links in the description etc.
- [x] Note that the functionality is not yet perfect because users do not know which URLs they will be able to add and when you enter an unsupported link in the Embed dialog it will not give the user any feedback that the URL is not embeddable.

**Do not merge it before DS-4292 embed performance is merged.**


jochemvn:
Merge when green!

PR #640, opened 2017-11-17
DS-2906 Make navbar colors configurable

MaikelGG:
## Problem
As SM I want to be able to giev a different color to the default navbar and secondary navbar.

## Solution
Implemented color configuration with color module.
The colors are not printed inline via the improved_theme_settings module anymore, but are now in brand.css. How it works is that all color variables need to be unique. They are listed in color.inc and brand.css. When you change the color in the theme settings form, the module will create a new brand.css file (which can then be cached) where it overrides the hex colors of the original version. So two original vars cannot have the same hex color, as then the module does not know what to replace. 

I have also optimised some of the other colors, which conflicted or were used in the navigation and could lead to contrast problems. 

The preview is too much work for now, as when you tell the module to include some socialblue stylesheets, the whole interface will make use of these stylesheets and that is not what we want (the seven theme should not be affected)

## HTT
- [ ] Fresh install , cache clear
- [ ] Login as SM+ and visit socialblue theme settings
- [ ] Change all colors and think of good contrast between bg and text
- [ ] View your changes in the front end, test as much pages as possible
- [ ] Also test mobile sizes


frankgraave:
@MaikelGG Only the hover of a CTA's needs to be adjusted (e.g. Post button or Join group/event). Everything else looks great!

MaikelGG:
@frankgraave I have updated the PR with fixes for the bugs you found

PR #646, opened 2017-11-25
fix typo in drupal_set_message

rosenstrauch:
## Problem
"Argument 2 passed to t() must be of the type array, string given"

## Solution
fix typo


jaapjan:
Thanks @rosenstrauch good find!

PR #649, opened 2017-11-28
DS-4407 - Group notification about activity in my group

frankgraave:
## Problem
As a member of a group, you did not receive notifications about new content in this group.

## Solution
Use the recently fixed activity context and create new message template to make this work.

## Issue tracker
DS-4407

## HTT
- [ ] Do a `drush fra -y --bundle=social`
- [ ] Log in as susanwilliams and post something in `/group/1/stream`
- [ ] In an incognito tab, log in as chrishall and notice you have a new notification
- [ ] Try this for topics and events
- [ ] Also check out the emails in mailcatcher
- [ ] Check out /user/x/edit
- [ ] See that you have a new email notification setting about notifications from a group you've joined
- [ ] Test the emails for these users in mailcatcher (post something in a group, run cron and see the mails)


jochemvn:
In the email notifications I noticed the following:
- post link doesn't work
- node links go to node/<group content id>, which leads to a page that does not exists

frankgraave:
@jochemvn I've addressed these issues and it should be working properly now.

modules/social_features/social_group/social_group.tokens.inc
```diff
@@ -105,34 +86,43 @@ function social_group_tokens($type, $tokens, array $data, array $options, Bubble
                     $replacements[$original] = $group_content_type->label();
                   }
                 }
+//                if ($name === 'content_url') {
+//                  $content_url = Url::fromRoute('entity.node.canonical',
+//                    ['node' => $entity->getEntity()->id()],
+//                    ['absolute' => TRUE]
+//                  );
+//                  $replacements[$original] = $content_url->toString();
+//                }
```
jochemvn: Remove this please

jochemvn: slashed out code

jochemvn:
Hey @frankgraave , This is still not working. IMHO that also not possible since you didn't change the notification center code in activity.page.inc, or am I missing something?

PR #652, opened 2017-11-29
Issue #2927180: As a LU I want to get a notification when someone comments on a topic commented on

ghost:
## Problem
As a LU I want to get a notification when someone comments on a topic commented on

## Solution

1. User should get a notification when another user comments on a topic you commented.
2. User automatically follows a topic if he comments on it.
3. User should be able to toggle those notifications on and of per item (via unfollow).
4. This auto follow should be mentioned in the success message when you post a comment.

## Issue tracker
https://www.drupal.org/project/social/issues/2927180

## HTT
- [x] Open some topic where exist link **Follow content**
- [x] Write the text of new comment and press **Comment**
- [x] You should see following success messages: **You were followed to a content of this topic.** and **Your comment has been posted.**
- [x] **Follow content** link should be changed to **Unfollow content**
- [x] Add a new comment on this topic by another user
- [x] Go back to the previous user
- [x] Run cron
- [x] Open **Notification Center** drop-down menu and you should see **[Another User] commented on [Topic Author User]'s topic you are following**


ribel:
Add needs work label to match with status in drupal.org issue: https://www.drupal.org/project/social/issues/2927180#comment-12373096

jaapjan:
Just updated the PR with the improved text. 

Found this issue; which can probably be annoying for some users.. Not sure if we already should merge this PR before improving that..

[follow-topic-when-commenting.mp4.zip](https://github.com/goalgorilla/open_social/files/2159816/follow-topic-when-commenting.mp4.zip)

jaapjan:
After reviewing marking this one as "needs work" again. 

When you comment you are automatically following the content, which is nice! But if you then unfollow the content manually and comment on the node again you are automatically following the topic again. My fear is that this will be super annoying for some users. And this feature is only there to accommodate for people who can't find the Follow button themselves. 

The only solution for the issue above is if we can somehow detect if someone unfollowed the content before and then we can prevent to let the user follow the content automatically again. It won't be easy since the Flag module does not currently keep records of previous flags in the database.

I suggest not to merge this before the issue above is solved.

Kingdutch:
Testing

Kingdutch:
Testing

Kingdutch:
@jaapjan While I agree that it might be annoying for some I think we could also argue that it's the correct behaviour (GitHub does the same):

If someone actively unfollows something then they probably don't want to interact with it anymore. If they then place another comment (interact again) it makes sense that they are notified of new updates as a result of their interaction. I think people can learn (like on GitHub) that if you don't want this, after your comment you simply "Unfollow" again.

The inverse situation that you're suggesting would be more annoying. A user comments and then unfollows because they're not longer interested. A while later they see something in a discussion and comment again, now they're not following any longer. Now they suddenly miss new activity as a result of their comment because they didn't press the "Follow" button (something they didn't need to do the first time round).

I think the behaviour that is a result of this pull request is actually the more consistent experience.

A quick test on Facebook shows that it implements the same behaviour as GitHub (always follow when you comment, even if you've unfollowed manually).

xinyumaGG:
@Kingdutch and @jaapjan 
I see the points from both of your remarks. 

My suggestion would be: we keep the behaviour consistent by make people follow a content every time they make a comment, but in the blue notification, we give people a button to unfollow (or even another button:don't follow automatically). This way, we give people more control over this.

jaapjan:
OK, thanks for taking the time to give your feedback @Kingdutch 

Given the fact that it might indeed give some value to some users as well I'm OK with getting this functionality in. But I do agree with @xinyumaGG that we need to improve on the notification then by adding an "unsubscribe" link. Could one of you work on that and let us know when it can be reviewed again?

Kingdutch:
I have not fixed the issue that was raised in the discussion with Jaap Jan but I've rebased this on 8.x-4.x and updated the target branch in an effort to help move this project forward as well as make it possible to update platforms where this was already put to use.

bramtenhove (approved):

bramtenhove:
@jaapjan, @ronaldtebrake, I think we can merge this one when tests are green.

We've had discussions about the annoyance of having to _Unfollow_ the content each time you content, but I believe we agreed that this does not outweigh the benefits this is giving with regard to engagement.

bramtenhove:
Just noticed a remark from Xinyu. I'll try to get the unfollow link in as well.

PR #655, opened 2017-11-30
DS-3946: Remove hero area when editing and deleting user, group or content

MaikelGG:
## Problem
Remove hero area when editing and deleting user, group or content

## Solution
Changed block config and conditions for showing the hero

## HTT
- [x] Login as chrishall and create an event
- [x] View event shows normal hero with image
- [x] Edit the event, the hero is gone and title says Edit event _event title_
- [x] Click delete button at bottom of the form, Delete confirmation screen show no hero, good title
- [x] Open user dropdown menu and go to edit profile. Hero is gone, it says "edit profile" 
- [x] Click Delete button, delete confirmation page shows no hero area, just plain page title


modules/social_features/social_core/src/Plugin/Block/SocialPageTitleBlock.php
```diff
@@ -23,6 +23,7 @@ public function build() {
     // Take the raw parameter. We'll load it ourselves.
     $nid = \Drupal::routeMatch()->getRawParameter('node');
     $node = FALSE;
+    $current_path = \Drupal::service('path.current')->getPath();
```
jochemvn: The current path service needs to be injected with Dependency injection

MaikelGG:
@jochemvn can you check the new commit, is this correct now?

PR #657, opened 2017-12-01
3 extra color schemes

basvanos:
## Problem
help site manger with styling the community 

## Solution
Added 3 extra color schemes, so the site manger has more option to choose from 

## HTT
- Go to Social Blue theme setting 
- Choose a different color set
- Check if the colors are changed


jochemvn:
Would be great to have this in saas too... what you you think @MaikelGG @jaapjan

MaikelGG:
Tested and approved

PR #664, opened 2017-12-05
2928647 As a LU I want to be notified if someone comments on a post that I have commented on

ghost:
## Problem
As a LU I want to be notified if someone comments on a post that I have commented on

## Solution
1. If someone comments on a post that I have commented on a notification is send.
2. If it is your post you should not get a extra notification.
3. Should be aggregated (email and internal).
4. Email notifications should be possible to turn off.

## Issue tracker
https://www.drupal.org/project/social/issues/2928647

## HTT
- [ ] Login as **User 1**
- [ ] When you open user menu and click **Following** then you should be on **/following** page with **Content and posts I follow** title
- [ ] When you open the drop-down list of **Type** field in the **Filter** block then you should see the **Post** item
- [ ] Create a post on the mainstream with the following text - **My first post!**
- [ ] Login as **User 2** and create comment to post which was created by **User 1**
- [ ] When you open the **/following** page then you should see one record in the table with the following data: **My first post!** link in the **Title** column, **Post** text in the **Type** column, **Unfollow post** button in the **Operations** column
- [ ] When you click **My first post!** then you should be on the post overview page
- [ ] Login as **User 3** and create comment to post which was created by **User 1**
- [ ] Login as **User 2** and when you open **Notification Center** drop-down menu then you should see **User 3 commented on a post that you have commented on**
- [ ] Login as **User 2** and open **/following** page
- [ ] Press **Unfollow post**
- [ ] Login as **User 4** and create comment to post which was created by **User 1**
- [ ] Login as **User 2** and when you open **Notification Center** drop-down menu then you should not see **User 4 commented on a post that you have commented on**


jochemvn:
@OH17L , The tests are failing. Feature state is overridden. Please check

Kingdutch:
I've added a comment to https://www.drupal.org/project/social/issues/2928647#comment-12383740

modules/social_features/social_follow_content/src/Plugin/ActivityContext/FollowContentActivityContext.php
```diff
@@ -39,20 +39,21 @@ public function getRecipients(array $data, $last_uid, $limit) {
...
-    $storage = \Drupal::entityTypeManager()->getStorage('flagging');
-    $flaggings = $storage->loadByProperties([
-      'flag_id' => 'follow_content',
-      'entity_type' => $related_entity['target_type'],
-      'entity_id' => $related_entity['target_id'],
-    ]);
+    $storage = $this->entityTypeManager->getStorage('flagging');
+
+    $flaggings = $storage->getQuery()
+      ->condition('flag_id', ['follow_content', 'follow_post'], 'IN')
```
Kingdutch: This entire change to using the query conditions isn't necessary. Looking at the `loadByProperties` function you can just pass `'flag_id' => ['follow_content', 'follow_post']` there and its underlying query mechanism will automatically use `IN` (it already does this for single values by default: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityStorageBase.php/function/EntityStorageBase%3A%3AbuildPropertyQuery/8.2.x )

Doing it simply using the load method allows us to better use the framework, taking advantage of caching, query altering and possible future improvements.

modules/social_features/social_follow_content/social_follow_content.module
```diff
@@ -9,44 +9,65 @@ use Drupal\Core\Entity\EntityInterface;
...
 
 /**
- * Automatically follow content when user enrolls the event.
+ * Automatically follow content or post.
  */
-function social_follow_content_event_enrollment_follow(EntityInterface $entity) {
-  if ($entity->field_enrollment_status->value) {
-    // Check if user already follow this content.
-    $account = \Drupal::currentUser();
+function _social_follow_content_action(EntityInterface $entity) {
```
Kingdutch: This function has so many checks for `if ($event)` in it that it is probably better to just create a separate function to handle when a user comments on a post. Any code that is then the same between this function and the new function could be extracted for re-use to a separate function to keep things DRY.

modules/social_features/social_follow_content/social_follow_content.install
```diff
@@ -18,6 +18,29 @@ function social_follow_content_install() {
...
+  $block = \Drupal::entityTypeManager()->getStorage('block')
+    ->load('exposedformfollowingfollowing');
+
+  if ($block) {
+    $block->delete();
+  }
+
+  $config_names = [
+    'views.view.following',
+    'block.block.exposedformfollowingfollowing',
```
Kingdutch: Is this not a duplicate of the `$block->delete()` above?

modules/social_features/social_follow_content/social_follow_content.module
```diff
@@ -9,44 +9,65 @@ use Drupal\Core\Entity\EntityInterface;
...
+  $properties['uid'] = \Drupal::currentUser()->id();
+
+  $flaggings = \Drupal::entityTypeManager()->getStorage('flagging')
+    ->loadByProperties($properties);
+
+  if (empty($flaggings) && $flagging = Flagging::create($properties)) {
+    $flagging->save();
+
+    if ($event) {
+      $message = t('You have successfully enrolled to this event. Also you are now following the event, which means you will receive notifications when a new comment is placed.');
```
Kingdutch: Should we also show a message to the user when we make them follow a post or some other content so they are aware of this?

modules/social_features/social_follow_content/social_follow_content.routing.yml
```diff
@@ -0,0 +1,7 @@
+social_follow_content.overview:
+  path: '/following'
+  defaults:
+    _controller: '\Drupal\social_follow_content\Controller\Overview::handle'
+    _title: 'Content and posts I follow'
```
Kingdutch: I think just keeping this "Content I follow" as it is now works fine because `posts` is included in the term "content". This also saves us translating a new string.

modules/social_features/social_follow_content/src/Controller/Overview.php
```diff
@@ -0,0 +1,206 @@
...
+use Drupal\Core\Url;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Class Overview.
+ *
+ * @package Drupal\social_follow_content
+ */
+class Overview extends ControllerBase {
```
Kingdutch: What is the reason that we're creating a new controller here? Why does the view not fit our purposes?

The view already loads the flags, so the view base table is not the problem. We could just add a post relationship and load the posts when a flag references one. Then the fields we display would show both the post and the content title (as only one is ever available).

The controller is very inflexible. If a site has decided to display 30 or 10 lines on a page then that is now no longer changeable by a site manager. The controller does no more then what views was designed to do.

Any changes that a site builder has made to the content overview (add an author field for example) will also be thrown out and require coding to get back.

modules/social_features/social_user/src/Plugin/Block/AccountHeaderBlock.php
```diff
@@ -385,7 +385,7 @@ public function build() {
             'title' => $this->t("View content I'm following"),
             'label' => $this->t('Following'),
             'title_classes' => '',
-            'url' => Url::fromRoute('view.following.following'),
+            'url' => Url::fromRoute('social_follow_content.overview'),
```
Kingdutch: Changing a route that's used for something can severely break sites. Even though we do this in a major release, we should try to stay backwards compatible.

themes/socialbase/templates/block/block--social-follow-content-filter.html.twig
```diff
@@ -0,0 +1 @@
+{% extends "block--views-exposed-filter-block.html.twig" %}
```
Kingdutch: The fact that this works shows that we've tried to rebuild a view without views.

modules/social_features/social_follow_content/config/install/flag.flag.follow_post.yml
```diff
@@ -0,0 +1,26 @@
+langcode: en
```
Kingdutch: Do we need a new flag here? I realise that `follow_content` may not be entirely accurate anymore in name. However, they are functionally equivalent so I don't think that separating them makes much sense? Is there any functional requirement to distinguish the two?

Kingdutch (changes requested):

modules/social_features/social_follow_content/social_follow_content.install
OH17L: Yes, it does the same

modules/social_features/social_follow_content/social_follow_content.module
OH17L: What message should be for posts?

themes/socialbase/templates/block/block--social-follow-content-filter.html.twig
OH17L: Do you propose paste code in this template from template **block--views-exposed-filter-block.html.twig** without any changes?

modules/social_features/social_follow_content/src/Controller/Overview.php
OH17L: @Kingdutch I used the controller because Views table sorting does not work correctly when using few entity types.

modules/social_features/social_follow_content/config/install/flag.flag.follow_post.yml
OH17L: @Kingdutch Yes, we should add a new flag because one flag can be related only to one entity type.

modules/social_features/social_user/src/Plugin/Block/AccountHeaderBlock.php
OH17L: @Kingdutch do you propose to use the **view.following.following** route for new page base on the controller?

bramtenhove:
@ronaldtebrake, @Kingdutch,
Was just looking at this and we're doing *a lot* of things to get posts in the following view. This feels like overkill for what we're trying to achieve.

At this moment I think we need to determine if the issues @OH17L faced with Views integration should be solved by replacing it with custom code or trying to figure out what causes the issues. I can already think of the Post entity not supporting all Views integration, but it could be something else as well.

In the worst case scenario, with Views, we could consider leaving out the sorting.

Curious to hear what you think.

PR #669, opened 2017-12-07
#2924676 Proper check if values exist before adding css

MaikelGG:
## Problem
After updating from 1.x -> 1.7 users are getting errors on non-existing border-radius properties.

## Solution
Check if these values even exist, also check if they are integers, otherwise convert them. If there is nothing filled in, don't insert the css.

## Issue tracker
https://www.drupal.org/project/social/issues/2924676

## HTT
- [ ] Before re-install, remove lines 2 - 4 from socialblue.settings.yml (so you don't import the border radius values)
- [ ] re install the site
- [ ] You do not get any errors
- [ ] Inspect the source code, no inline CSS is shown in the <head>
- [ ] Log in as SM+ and go to socialblue theme settings
- [ ] Fill in values for radius (card, form, button) and save the form
- [ ] Check the site, your radius values should be applied and visible

This scenario represents updated sites with custom subthemes based on socialbase, which do not have these radius properties in place yet. If you would have the following settings in the subtheme the bug will never show:

```
card_radius: '10'
form_control_radius: '5'
button_radius: '10'
```


modules/custom/improved_theme_settings/improved_theme_settings.module
```diff
@@ -184,3 +173,16 @@ function improved_theme_settings_page_attachments(array &$page) {
...
 }
+
+/**
+ * Function that returns a valid integer or FALSE.
+ */
+function improved_theme_settings_get_setting($setting, $themename) {
+  // Fetch setting from the theme.
+  $themevar = Xss::filter(theme_get_setting($setting, $themename));
+  // In this case '' is something different than 0.
+  if ($themevar !== trim(intval($themevar))) {
```
Kingdutch: @jochemvn 
I think this change causes a regression from a previous version. E.g. `'5 '` would previously be counted as a valid value because it'd be parsed to `5` but `trim(intval('5 '))` evaluates to `'5'` which does not equal `'5 '`. 

If the setting ends up being an object due to a misconfigured subtheme this code will also produce a notice instead of disabling the feature without errors.

jochemvn: No it doesn't. The field is a number field, you cannot add values like '5 ' there.

Kingdutch: There is no validation (like a config schema) for that anywhere else so currently it's allowed for a subtheme to have the value '5'. This is why a simple `is_null || !is_integer` did not suffice.

PR #676, opened 2017-12-12
#2929953 Waves: JS errors in IE11 for menus

MaikelGG:
## Problem
Clicking on profile menu or plus button not working in IE11.

## Solution
in waves.js function getWavesEffectElement iterates over target.parentElement recursively. This leads to problems in IE11 SVG-Elements where parentElement is undefined. Changing this to target.parentNode works as expected in all browsers.

## Issue tracker
https://www.drupal.org/project/social/issues/2929953

## HTT
- [ ] Composer update
- [ ] Test in IE11
- [ ] Login as chrishall
- [ ] Click on plus icon or user profile (both in user menu). No errors should show in Inspector


kedramon:
Hi guys,
```
- Applying patches for bower-asset/waves

    https://www.drupal.org/files/issues/2929953-waves-error-ie11-3.patch (getWavesEffectElement iterates over parentElement - should use parentNode instead)

   Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/2929953-waves-error-ie11-3.patch
```
patch is faling, looks like this bug is fixed in 0.8-dev branch.

T2L:
Tests are passing, but IE11 still doesn't like the navbar, due to:

1. classList property is not supported on SVG Element
1. IE (and Edge) doesn't propagate click from inside of SVG to A parent element, see https://github.com/facebook/react/issues/4963 and https://github.com/facebook/react/issues/4963 and https://stackoverflow.com/questions/24078524/svg-click-events-not-firing-bubbling-when-using-use-element

T2L:
See SVG click issue in #681

PR #677, opened 2017-12-13
Create update hook to reset the existing theme setting colors to the new layout

jochemvn:
## Problem
When updating form OS 1.7 to 1.8 with custom colors in your DEFAULT theme. The colors are not migrated. This causes the new settingsform not to load any of the color fields.

## Solution
Migrate the settings in an update hook

## HTT
- [x] Install SaaS 1.12 (with open_social "1.7" and socialsaas "dev-master#bc7cc90")
- [x] Change colors in theme settings to some custom values
- [x] git checkout 1.13 (first checkout composer.lock)
- [x] set this PR as patch in composer.
- [x] run composer update and run the update.sh script.
- [x] See that the colors are now properly migrated and that there's a nice new preview.

### DIFF
--- a/modules/custom/improved_theme_settings/improved_theme_settings.install
+++ b/modules/custom/improved_theme_settings/improved_theme_settings.install
@@ -14,3 +14,51 @@ function improved_theme_settings_install() {
   // Set weight.
   module_set_weight('improved_theme_settings', 50);
 }
+
+/**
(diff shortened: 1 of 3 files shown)


frankgraave:
@jochemvn I've tested it out. Turned out that I needed to change the socialsaas in composer as well of course.. So I've adjusted the HTT. But this solution works, and it works nice!

PR #681, opened 2017-12-15
#2931073 - Click events are swallowed by SVG <use> elements in IE/Edge

T2L:
## Problem
Icons in navbar implemented as inline SVG with <use> element inside of <a> element. When clicking on the <use> element in IE/Edge onClick event is not propagated to the parent anchor element, so dropdown does not work.

## Solution
Apply `pointer-events: none` CSS rule to the navbar icon SVG element.

## Issue tracker
https://www.drupal.org/project/social/issues/2931073

### Before
- [x] Open OS in IE/Edge
- [x] Click on + icon in the navbar. Aim to the icon body (not around it, not inside the plus sign itself - icon is transparent). Nothing will happen. 
- [x] Hash (#) is being added to the URL

### After
- [x] Open OS in IE/Edge
- [x] Click on + icon in the navbar. Aim to the icon body (not around it, not inside the plus sign itself - icon is transparent). Dropdown should be showed
- [x] No hash (#) is being added to the URL


left23:
@jaapjan This is good to go. Please merge

PR #684, opened 2017-12-18
2931510 AS LU I would like to create public groups

OH17L:
## Problem
AS LU I would like to create public groups

## Solution
1. AN+ can see the group in search and overviews
2. AN+ can see all content in the group
3. AN+ can see all members, but they can't click on members.
4. AN+ can see all activities of the group
5. Public group activities will also be seen in home and explore streams
6. It should be clear to everyone that this is a public group
7. Only LU can join this type of group, they can join directly.
8. AN will see join button, but after clicking they will go to sign up page, users will be redirect to this group after sign up. In this sign up process, profile creation is skipped due to redirect.
9. Other behavior is the same as Open Group, such as member management, how to join and leave, what happens when deleting content and group.
10. Enable for new installations

## Issue tracker
https://www.drupal.org/project/social/issues/2931510

## HTT
- [ ] Login as authorized user.
- [ ] Click "Plus" button on the navigation bar and choose "New group" in a drop-down menu.
- [ ] Choose "Public group" radio button and press "Continue".
- [ ] Fill fields and press "Save".
- [ ] Then you should be on "Groups" user tab where you should see new public group which you created.


frankgraave:
Hi @OH17L 

Thank you for this feature! Really looks great, and works well.

While testing I came across a few things:
1. When creating an public group -> the save button is on the left, but should be on the right.
2. Local nav(secondary nav) button "Groups" does not show correct mouse hover? (check on /user/xx/groups)
3. When AN clicks "Join" in a public group, he is being redirected to the login page, not the sign up page.

jochemvn:
Hey @OH17L ,
I've also taken a look at the feature. it's a great contribution. Thanks for that. I do have a few remarks. Please take a look:

# When creating a group, the group type selection page text for public groups should match a bit better with the other group types.
> So: for non members = to all community members
# Upload requirements shoud be fixed (see open group)
# AN can now go to the group overview and see ALL (!) groups. Same for the search. All private groups (the titles) are now out in the open. This is not acceptable.
# Why are content items from a public group being deleted when a public group is being deleted. The content was already public to begin with.
# In social_group_update_8010, please update only the permission you want to set for this feature. People using the distro might have changed the permission on their platform

OH17L:
@frankgraave, exists problem with a private file which not related to this issue. Check pull request again, please.

PR #685, opened 2017-12-18
Issue #2931507 by kedramon, ribel: Add Social Landing Page module

ribel:
## Problem
Most of the customers, using Open Social, requesting an alternative homepage, where they can promote the organisation and content within the community.

## Solution
- Add landing_page content type with Paragraphs
- Implement hero, about, featured topics and events, and activity stream blocks
- Consider featuring other content types
- Blocks can be rearranged, turned on and off, and edited
- It should be a separate page which can be configured as home page for existing users.

## Issue tracker
- https://www.drupal.org/project/social/issues/2931507
- https://jira.goalgorilla.com/browse/DS-4130
- https://jira.goalgorilla.com/browse/SDG16-1

## HTT
- [x] Checkout to this branch
- [x] Run composer update
- [x] Enable module **social_landing_page**
- [x] Login as SM
- [x] Create Landing Page content using different sections
- [x] Check that Landing Page is visible for AN
- [x] Check that Landing Page doesn't appear in overview pages, stream and search
- [x] Check Behat test


jochemvn:
**General:**
- Would be great to have the option to say you want this landing page to be the homepage
- I've created an alias '/stream'. This works, which means the stream is no longer accessible,
- Requires language. We don't want this in the distribution
- Requires a LOT of modules. Does it really?
- It's not added to the [+] button

**Hero:**
- Get the remark: The image is too small; the minimum dimensions are 1500x370 pixels. (that's quite big isn't it?)
- Hero is not attached to the menu, is that by design?
- The button link is not validated
- Button color can only be set for all buttons. On a per button basis would be better.
- After saving a page with a hero section, go back to edit and the buttons section is empty (although I did add a button)

**Block:**
- Does not play nice with social_renamer (I've added a topics block). Could be a social_renamer issue. Not sure.
(Error: Cannot create references to/from string offsets in social_renamer_preprocess_block() (line 145 of /app/html/modules/contrib/social_renamer/social_renamer.module) #0 /app/html/core/lib/Drupal/Core/Theme/ThemeManager.php(287): social_renamer_preprocess_block(Array, 'block', Array) #1 /app/html/core/lib/Drupal/Core/Render/Renderer.php(437): Drupal\Core\Theme\ThemeManager->render('block', Array) #2 /app/html/core/lib/Drupal/Core/Render/Renderer.php(195): Drupal\Core\Render\Renderer->doRender(Array, false) #3 /app/html/core/lib/Drupal/Core/Template/TwigExtension.php(490): Drupal\Core\Render\Renderer->render(Array) #4 /app/html/sites/default/files/php/twig/5a38719c449ca_field.html.twig_pvGwifcEDbAuvq6ZYE4oF0OBp/ePJgjhfBlIpr3co09A_qxTr3MeG_M-J4ZOk8BJdw6_8.php(132):)

ribel:
@jochemvn Thanks for good review!
I think first two options are _nice to have_ and we can add them later. They also valid for all other content types (like basic Page), so not only to Landing Page.
Regarding other items all they seems valid. I will take a look to them tomorrow and will try to fix.

jochemvn:
I also just noticed the 2nd item isn't entirely true, since the stream is posted below the homepage, so it actually works as it should (I think)

jochemvn:
I can also verify that the bug with the social_renamer is a bug in THAT module and not in this one!

jochemvn:
Jaap has the following additions / questions:
- Can we add node visibility to landing page
- Can we add possibility to link any content a user has access to

jochemvn:
One more thing:  

> Check that Landing Page doesn't appear in overview pages, stream and search

It does show up in the search and also in the content search there's a filter `landing page`

kedramon:
> Button color can only be set for all buttons. On a per button basis would be better.
@jochemvn one button is a paragraph that contain two links, one for AN one for LU user, so it act like one button and the idea is that style is same for both.
You can create one button with logged in link with style primary and leave empty AN link (not will be printed for AN user) and create one button with only AN link with default style.

jochemvn:
merge when green 🍏

PR #686, opened 2017-12-19
DS-4352 - As LU/SM I want to be able to tag content

jochemvn:
## Problem
Feature request, not solving a bug here. See issue tracker for more information.

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-4352

## HTT
- [x] Login as an admin an turn on the module
- [x] Login in as a SM and add some terms to the taxonomy social_tagging
- [x] Create a topic / event / page
- [x] Notice you can tag the content
- [x] Notice the tag widget is always the second last card
- [x] As a SM go to "Configuration > Open Social Settings > Tag settings" and notice you can turn tagging off
- [x] Turn off and add a node
- [x] Notice the tagging field(s) are not there anymore
- [x] Turn it back on
- [x] Save a node with some tags
- [x] Delete a tag that is selected on a node
- [x] See that the tag is removed from the node
- [x] Enable the social_book module
- [x] Notice that when adding a book, you can now tag books too

### DIFF
--- a/modules/social_features/social_tagging/social_tagging.module
+++ b/modules/social_features/social_tagging/social_tagging.module
@@ -28,13 +28,15 @@ function social_tagging_help($route_name, RouteMatchInterface $route_match) {
       return $output;
 
     case 'entity.taxonomy_vocabulary.overview_form':
-      /* @var \Drupal\taxonomy\Entity\Vocabulary $vocabulary */
-      $vocabulary = $route_match->getParameter('taxonomy_vocabulary');
-      if ($vocabulary->id() === 'social_tagging') {
-        return '<p><strong>' . t('Notice: Drag and drop has intentionally been disabled for this vocabulary.') . '</strong></p>';
-      }
+      $tag_service = Drupal::getContainer()->get('social_tagging.tag_service');
 
-    default:
+      if ($tag_service->allowSplit()) {
+        /* @var \Drupal\taxonomy\Entity\Vocabulary $vocabulary */
+        $vocabulary = $route_match->getParameter('taxonomy_vocabulary');
+        if ($vocabulary->id() === 'social_tagging' && $tag_service->allowSplit()) {
+          return '<p><strong>' . t('Notice: Drag and drop has intentionally been disabled for this vocabulary.') . '</strong></p>';
+        }
+      }
   }
 }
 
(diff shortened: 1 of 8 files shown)


jochemvn:
See also this [SaaS](https://x-bitbucket-pr-100-wf5foyq-rmio5d4m7mmka.eu.platform.sh) branch

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -0,0 +1,230 @@
...
+      }
+
+    default:
+  }
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function social_tagging_form_taxonomy_term_social_tagging_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
jaapjan: I think we should consider making this configurable with a setting (so it can be overridden in the distro) even if it is only a simple on/off toggle.

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -0,0 +1,230 @@
...
+  // Check if tagging is turned on/off.
+  if (!Drupal::getContainer()->get('social_tagging.tag_service')->active()) {
+    // Remove the field from the form.
+    $form['social_tagging']['#access'] = FALSE;
+    return;
+  }
+
+  if (isset($form['social_tagging'])) {
+    // Get the taggingservice.
+    $tag_service = Drupal::getContainer()->get('social_tagging.tag_service');
```
jaapjan: Why do you get the $tag service here again? You already do this on line 101.

jochemvn:
Jaap has the following additions / questions:
- Can we add node visibility to landing page
- Can we add possibility to link any content a user has access to

jochemvn:
1. I have made it more 'configurable'. You can now set wether you want to use the splitting of the main items into categories on the node forms. If not, you'll just get a single input field with all items.

2. Figured this to be more part of stories that are still upcoming (the actual showing of the labels and filtering on them)

3. I've added a link from the settings form to the taxonomy overview.

4. Hmm not sure. It IS part of the acceptance criteria atm.

5. I've made sure that when you delete a term in this vocabulary, the items are removed from the node tables (node__social_tagging and node_revision__social_tagging)

PR #687, opened 2017-12-21
2932350 by robertragas : Change profile creation update hook to make …

robertragas:
…it less heavy

## Problem
Mysql server gone away because of trying to run this update for 50k + users.

## Solution
Make the begin query more efficient to it's less heavy.

## Issue tracker
https://www.drupal.org/project/social/issues/2932350#comment-12393715

## HTT
- [ ] Check out the code changes on a site with 50k+ users 
- [ ] Run drush updb
- [ ] Notice mysql server goes away ( at least on platform.sh )
- [ ] Checkout to this branch
- [ ] Retry and see it works.


bramtenhove:
Thanks for the patch! Unfortunately this is for quite a few releases ago and I'm reluctant to include it now.

If people are getting this issue they can always refer back to the PR you wrote. As far as I understood this has run successfully for a client with a large number of users.

PR #694, opened 2018-01-10
Issue #2930486 Posts are orphaned when a user is deleted

Kingdutch:
The proper hooks are not implemented for the Post entity type to ensure posts are correctly reassigned an author or deleted if a user is removed. 

This pull request implements those methods by looking at how it was done for Nodes.

How to test:
- Ensure /admin/content/post does not contain any errors.
- Create a post in the stream as User1
- Delete User1 from the platform
- Observe that the /admin/content/post page now throws an error
- Apply this PR and run updates
- Observe that /admin/content/post no longer throws an error. The post for User1 has now been deleted.
- Create a post in the stream as User2
- Delete User2
- Depending on the site settings the post created as User2 has been deleted or assigned to the anonymous user. Confirm this by looking at /admin/content/post


robertragas:
I made 1 change to the post update hook.

It was placed inside the .install file while the documentation on drupal.org states 

"These implementations have to be placed in a MODULE.post_update.php file."

In case of using this patch for a customer the last implementation didn't work and kept going in a loop while doing a drush updb.

jaapjan:
I see in this pullrequest that you are unpublishing posts as well. I guess they still show up in activity streams then? We might want to finish this one as well:
#715

Kingdutch:
@jaapjan This mostly mimics what Nodes already do and I think it's fairly self contained. The issue that this solves also keeps breaking some stuff for at least 1 Enterprise client. I think we can merge this before we fix #715 which seems like a pretty big issue.

Kingdutch:
@jaapjan @bramtenhove #715 has since been merged but this is still open. Can we merge it as the problem isn't fixed yet.

Kingdutch:
Reopened, clicked wrong button :(

jaapjan:
I think we should move this to 2.x and put it in 2.2 milestone. Could you update the PR @Kingdutch ?

jaapjan:
No action taken. Marking for 3.0 instead..

Kingdutch:
This fails on PAC because this PR is used as a patch there (obvious conflict but tested to work).

This fails on SHNL with an AJAX error for a topic preview which appears unrelated.

CBI is having issues in general due to a misnamed branch.

Other than that it looks good to merge.

Kingdutch:
Good point! Still getting used to the flow of MOSS :)

jaapjan:
Tests on Moss are green as well :-)

PR #701, opened 2018-01-19
DS-4637 - Use Landingpage as replacement of homepage

jaapjan:
## Problem
Anonymous and logged-in users frontpages are currently not very flexible. It is virtually impossible to feature content. It would be an option to be able to use the landingpage feature for this since it will give the SM the possibility to create a visually appealing site and feature content, groups.

## Solution
The landingpage can not be set as the homepage on "/stream" currently since it will add new blocks automatically. Actually when using a landingpage it should not add any blocks to the hero, content area or sidebars. 

## Issue tracker
DS-4638
https://www.drupal.org/project/social/issues/2931507

## HTT
- [x] Check out the code changes and especially the refactoring of the Post blocks and the Views plugin code
- [x] Login as user with SM rights and create a landingpage and use an alias which is NOT "/stream" but something else like "/frontpage"
- [x] Try to recreate the homepage by making a hero, adding activity stream and adding the correct blocks on the right
- [x] Notice it looks like the homepage and you can even add new blocks (unlimited) to any region
- [x] Set the page alias as the Homepage on Site Configuration administration page
- [x] Verify that the page looks good as LU
- [x] Try to recreate the Anonymous Homepage as well and set the visibility of the content to AN

### DIFF
--- a/modules/custom/alternative_frontpage/src/EventSubscriber/RedirectHomepageSubscriber.php
+++ b/modules/custom/alternative_frontpage/src/EventSubscriber/RedirectHomepageSubscriber.php
@@ -1,0 +1,99 @@
+<?php
+
+namespace Drupal\alternative_frontpage\EventSubscriber;
+
+use Drupal\Core\Path\PathMatcher;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Drupal\user\UserData;
+use Drupal\Core\Session\AccountProxy;
+use Drupal\Core\Config\ConfigFactory;
+
+/**
+ * Class RedirectHomepageSubscriber.
+ */
+class RedirectHomepageSubscriber implements EventSubscriberInterface {
+
+  /**
+   * Protected var UserData.
+   *
+   * @var \Drupal\user\UserData
+   */
+  protected $userData;
+
+  /**
+   * Protected var ConfigFactory.
+   *
+   * @var \Drupal\Core\Config\ConfigFactory
+   */
+  protected $configFactory;
+
+  /**
+   * Protected var for the current user.
+   *
+   * @var \Drupal\Core\Session\AccountProxy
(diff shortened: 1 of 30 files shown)


modules/custom/alternative_frontpage/src/EventSubscriber/RedirectHomepageSubscriber.php
```diff
@@ -0,0 +1,99 @@
...
+    $this->configFactory = $config_factory->get('alternative_frontpage.settings');
+    $this->currentUser = $current_user;
+    $this->pathMatcher = $path_matcher;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    // 280 priority is higher than the dynamic and static page cache.
```
bramtenhove: I don't really understand where this 280 comes from or is set too. Are we sure it is now in the correct order?

modules/custom/alternative_frontpage/src/EventSubscriber/RedirectHomepageSubscriber.php
```diff
@@ -0,0 +1,99 @@
...
+      return;
+    }
+
+    // Ignore non index.php requests (like cron).
+    if (!empty($_SERVER['SCRIPT_FILENAME']) && realpath(DRUPAL_ROOT . '/index.php') != realpath($_SERVER['SCRIPT_FILENAME'])) {
+      return;
+    }
+
+    $isFrontPage = $this->pathMatcher->isFrontPage();
+    if ($isFrontPage) {
```
bramtenhove: I encountered some odd behaviour. I think it comes from this section.

1. Set AN homepage to an event (e.g. /node/19 of demo content)
2. Set LU homepage to /stream
3. As an LU navigate to the AN homepage (e.g. /node/19)
4. Observe that you are redirected to the stream

I guess this means that LU will never be able to visit the AN homepage right? SM will have to edit the AN homepage via the content overview?

jaapjan: Yes, true. The redirect will always happen for LU. Just like in the current situation where the SM can not see the AN page.

PR #703, opened 2018-01-23
Don't use an abbreviation for the PmAddBlock

Kingdutch:
When looking at the Private Message functionality everything is written out full, so I search for "Private Message". I have no easy way of knowing that for the "Private Message Add Block" we suddenly use the abbreviation "PM". We should just write it out in full.

We should probably also do this for the class names and ids but if we've already marked this module as stable then that would break backwards compatibility. If we haven't marked the module as stable yet it's a change still work making.

This single line change has no Drupal.org issue associated with it.

### DIFF
--- a/modules/social_features/social_private_message/src/Plugin/Block/PmAddBlock.php
+++ b/modules/social_features/social_private_message/src/Plugin/Block/PmAddBlock.php
@@ -13,7 +13,7 @@ use Drupal\Core\Access\AccessResult;
  *
  * @Block(
  *  id = "pm_add_block",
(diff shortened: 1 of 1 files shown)


jochemvn:
@Kingdutch this is fine of course, but next time please use the PR template, so with a problem, solution issue tracker (can also mention there, that there's no d.org issue) and htt. We have that in place, since more and more people are working on this and not everyone know where to look for with each PR.

Kingdutch:
@jochemvn Sorry, will do, missed the template.

PR #705, opened 2018-01-24
Issue 2934118 After editing a comment on a post, upon save you are direct to the edit screen.

BSaxion:
## Problem
After editing a comment on a post, upon save you are direct to the edit screen. This is not correct.
When editing a comment on a topic, it does work correct: you return to the topic detail page. Same for reply on a comment: works like a charm!

## Solution
Expected result:
- I think best is to return to the post detail page, scrolled down to your comment.

## Issue tracker
- https://www.drupal.org/project/social/issues/2934118

## HTT
Steps to reproduce the issue:
- [ ] create a comment on a post
- [ ] edit the comment, press save
- [ ] observe that you end up on the edit comment screen.
- [ ] You do get a message that your comment has been posted. 

Steps to reproduce the fix:
- [ ] create a comment on a post
- [ ] edit the comment, press save
- [ ] observe that you end up on the post screen and you scroll down to edited comment.
- [ ] You do get a message that your comment has been posted.

### DIFF
--- a/modules/social_features/social_post/social_post.module
+++ b/modules/social_features/social_post/social_post.module
@@ -67,19 +67,30 @@ function social_post_form_post_form_alter(&$form, FormStateInterface $form_state
  */
 function social_post_form_comment_post_comment_form_alter(&$form, FormStateInterface $form_state, $form_id) {
   // Reset title display.
-  $form['field_comment_body']['widget'][0]['#title_display'] = 'invisible';
-
-  // Set the action of the form to the current uri.
-  // This needs to be done, because we cannot override $form['#action']
-  // without breaking functionality.
-  $uri = Url::fromRoute('<current>')->toString();
-  // Store in a hidden field.
-  $form['redirect_after_save'] = [
-    '#type' => 'hidden',
-    '#title' => t('Redir'),
-    '#default_value' => $uri,
-  ];
-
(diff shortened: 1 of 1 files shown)


jochemvn:
Hi @BSaxion, the automated tests are failing. This is because you don't have access to the to the repo. Let's take a look at your setup tomorrow.

https://screencast.com/t/gvyAAj3iwt4

jaapjan:
Hey Brhane, I think there is another way to fix this which might be cleaner. I suggest to do it like this:

1. Create logic in social_post_form_comment_post_comment_form_alter() to determine if you are editing a comment and if yes: change the hidden value of redirect_after_save to the post of the comment you are editing at that moment. 
2. Then you don't need the logic you added to the _social_post_comment_post_comment_form_submit() so that's a lot cleaner.

It could also be that by default the comment form always goes back to the comment/[id] you are editing. If that's the case then instead of the solution I suggest in option 1 you could also consider adding the hidden form value on for new comment forms. 

In addition I think now you have the rights to make pullrequests in this repository directly. Perhaps you could create a new pullrequest to fix this issue?

Thanks in advance!

bramtenhove:
@BSaxion do you think I can remove the "needs work" label?

jaapjan:
Closing PR to keep the overview for now. Can be created later on the latest dev-x branch.

PR #706, opened 2018-01-24
2937317, 2937255 and 2936748 - Fix tagging bugs

jochemvn:
## Problem
The tagging feature came with some minor bugs

## Solution
- When no tags available, consider tagging turned off
- Remove tagging field from search when turned off
- Turned off tagging on landing_page node type

## Issue tracker
- https://www.drupal.org/project/social/issues/2936748
- https://www.drupal.org/project/social/issues/2937255
- https://www.drupal.org/project/social/issues/2937317

## HTT
- [x] Enable the social_tagging module in your OS installation or updatedb
- [x] Check the settings and notice tagging is turned on (/admin/config/opensocial/tagging-settings)
- [x] Also notice there's now a content type section
- [x] Notice that tagging is off for landing_page, but on for other types
- [x] Check the taxonomy and notice there are no tags yet
- [x] Check all node/add pages and notice the card with tags doesn't show up
- [x] Check out search/content and notice the filter doesn't show up
- [x] Add 1 main item and recheck the pages above (still nothing visible)
- [x] Add 1 sub items and recheck the pages above (tagging field visible)
- [x] Go to tagging settings and turn tagging off
- [x] Recheck the pages above (nothing visible again)
- [x] Go to tagging settings and turn tagging on again
- [x] Enable social_landing_page and add a page (No tag card visible on node/add)


modules/social_features/social_tagging/social_tagging.module
```diff
@@ -105,10 +105,24 @@ function social_tagging_form_taxonomy_overview_terms_alter(&$form, FormStateInte
  * Implements hook_form_FORM_ID_alter().
  */
 function social_tagging_form_node_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+
+  $ignored_node_types = [
```
jaapjan: @jochemvn just wondering: why not make this a configuration option? it will increase the flexibility of the feature as well for custom sites.

jochemvn: did have a lot of time.... but you are  right... will do it somewhere next week

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -105,10 +106,32 @@ function social_tagging_form_taxonomy_overview_terms_alter(&$form, FormStateInte
...
+    if ($config_value === FALSE || $config_value === 0) {
+      $node_type_tagging = FALSE;
+    }
+  }
+
   // Get the taggingservice.
   $tag_service = Drupal::getContainer()->get('social_tagging.tag_service');
   // Check if tagging is turned on/off.
-  if (!$tag_service->active()) {
+  if (!$tag_service->active() || !$tag_service->hasContent() || $node_type_tagging === FALSE) {
```
jaapjan: Why did you add `!$tag_service->hasContent()` that makes no sense to me? That means that the field will never be shown when there is no tagged content? So how do you create your first content item with tags?

jochemvn:
2 more items needed here:
1. when splitting is off, you should not be able to select main category items on node add/edit
2. Better explanation of this particular vocabulary on LGOS (maybe even a link)

jochemvn:
@jaapjan, I've solved the issue and I have let Xinyu know that the documentation on LGOS needs some polishing.

jaapjan:
I found two more things which should be easy to solve. You can consider yourselves if you want to do the 2nd in a follow-up or not. I think the first one needs to be solved before merging.

1. When you disallow the category split there are notices in the form (I did test it on a Dutch platform so that might be why, not sure?)

Notice: Undefined index: #title in social_tagging_form_node_form_alter() (line 188 of profiles/contrib/social/modules/social_features/social_tagging/social_tagging.module).

Notice: Undefined index: #default_value in social_tagging_form_node_form_alter() (line 190 of profiles/contrib/social/modules/social_features/social_tagging/social_tagging.module).


2. When you have created content with tags and disable the tagging feature afterwards then the block "Tags" is still shown on the node view page without any tags in them. I think this blocks should be hidden then?

jochemvn:
@jaapjan , took some time 🙈 but I've decided to use a simple select widget and prepopulate the options array with only the level 1 items so only those can be selected in the case when 'split' is turned off!! 🎉

PR #711, opened 2018-01-25
2939495 Site name instead of word "Copyright" in email

ghost:
## Problem
Show word "Copyright" in email row in Gmail.

## Solution
Use site name instead of this word.

## Issue tracker
- https://www.drupal.org/project/social/issues/2939495
- https://jira.goalgorilla.com/browse/ECI-645

## HTT
- [ ] Create a post by a user with Google email address.
- [ ] Log in as some other user and add a comment to created post.
- [ ] Run cron.
- [ ] Open Gmail of an author of the post and you should see word "Copyright" in email row.

### DIFF
--- a/themes/socialbase/src/Plugin/Preprocess/Swiftmailer.php
+++ b/themes/socialbase/src/Plugin/Preprocess/Swiftmailer.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Drupal\socialbase\Plugin\Preprocess;
+
+use Drupal\Core\Config\ConfigFactory;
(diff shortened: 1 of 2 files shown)


Kingdutch:
I added a comment in the d.o issue

Kingdutch:
@jaapjan Given the changes in the e-mail system, I think this particular PR can be changed. It now says the site name twice.

However, I do think we need to leave the original issue open to see if the summary shown in mail clients is correct.

jaapjan:
Agree; let's close it!

PR #712, opened 2018-01-25
DS-4491 activity stream multilingual

bramtenhove:
## Problem
In the activity stream it is currently not possible to have the message displayed in the language of the user. If an user has selected the Dutch language or is visiting the Dutch website the notification is still shown in English (e.g. "User x created an event in group Y").

In emails there is a similar problem, but other texts are also not translated in the language of the recipient user.

## Solution
Translating the message template text upon **display** of the item. Basically this means we retrieve the message template and generate the output text (with tokens replaced) on an entity view. And the same happens when sending notifications in emails.

_As an added bonus I fixed the URL in email notifications for groups. This was, as far as I could tell, the only item which was using a relative URL (which doesn't work well in emails). I changed it to an absolute URL like was done in other places._

The output text is saved in the database when creating the activity object. We don't really need it anymore, and thus we should remove it at some point.

## HTT
- [x] Check out the code changes.
- [x] Install the site, enable Interface Translation, set-up the Dutch language and import the Open Social translations.
- [x] Configure Language detection to use (at least) the users language.
- [x] Set the user language for user X to Dutch.
- [x] _You may choose to generate demo content again now._
- [x] Post some stuff which user X will receive, mention user X and create a topic in the group of user X.
- [x] As user X login to the platform, observe that activities are now shown in Dutch in the homepage stream.
- [x] Observe that activities are now shown in Dutch in the group stream.
- [x] Observe that activities are now shown in Dutch in the profile stream.
- [x] Observe that activities are now shown in Dutch in the notification centre.
- [x] Check the email of user X and see that the notification(s) he received are in Dutch.
- [x] Configure some notifications to be shown in a digest email (e.g. daily) for user X.
- [x] Logout user X and create some activities that will generate a digest email for user X.
- [x] Wait or force the digest to be send.
- [x] Check the digest mail to see that the notifications are translated there as well.
- [x] Verify that other languages (e.g. Spanish) work just as described for Dutch.

When adding another language, next to Dutch, be sure to translate the message template notification messages via the UI translation page!

### DIFF
--- a/modules/custom/activity_creator/activity_creator.module
+++ b/modules/custom/activity_creator/activity_creator.module
@@ -209,6 +209,7 @@ function activity_creator_entity_view(array &$build, EntityInterface $entity, En
   }
 
   // Generate the output text for the user in his selected language.
+  // Get language van recipient user.
   if ($entity instanceof \Drupal\activity_creator\Entity\Activity) {
     $message = Message::load($entity->get('field_activity_message')->target_id);
     if ($message instanceof Message) {
@@ -216,6 +217,8 @@ function activity_creator_entity_view(array &$build, EntityInterface $entity, En
       $output = \Drupal::service('activity_creator.activity_factory')->getMessageText($message);
       // Replace the old text with the correct one.
       $build['field_activity_output_text'][0]['#text'] = $output[0];
+
+      // Caching context met current language toevoegen als nodig (build array).
     }
   }
 }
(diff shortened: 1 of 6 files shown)


bramtenhove:
@jaapjan @jochemvn this is ready for review. I suppose either one of you (or both) can review this?

I'm a little bit in doubt about 6fde9161bf8ef7b1d330c4a9145375e9e498845a, but wasn't able to find a working solution within 30 minutes. If you have suggestions please let me know ;)

jochemvn:
@bramtenhove  Coding standards :trollface:

bramtenhove:
Thanks for your comments @jochemvn! I will improve the code asap.

PR #715, opened 2018-01-30
Feature 2940643 post publish status

slowflyer:
## Problem
At the moment posts can be created and are published in the same moment and they can be deleted by Content Managers and Site Managers.

But there is no possibility in the UI to simply unpublish posts.

fe. a post with dubious content is created and you as a content- or site manager want to hide it until the issue is clarified and then can keep it published with good feeling.

Therefore it would be great to have the status available on edit form

## Solution
Added the publish status to the post edit form for users with permission 'edit any post entities' fe. Content Managers and Site Managers

## Issue tracker
- https://www.drupal.org/project/social/issues/2940643

## HTT
- [x] Login as normal user  and you should see no changes, even when editing a post you created
- [x] Login as content- or site manager and edit any post. You should see and be able to use the status publish / unpublish checkbox

This is a first step in a row of more because:
Even if you unpublish the post, the post is shown in streams, the views don't take care about it correctly. But that will be a separate issue.


jochemvn:
@slowflyer , I've merged back the master branch due to some composer errors

slowflyer:
Travis reports: "The feature state is not default!"

Is there a way how I can fix that?

Thanks,
Ullrich

jochemvn:
I've merge back 8.x-1x once more. There are some module updates that could be responsible

jaapjan:
@slowflyer Actually I think this is happening because the changes made in src/Entity/Post.php. Here you add the status field but this should also be reflected in the configuration files delivered in the social_post features. I suspect you need to update the configuration files for the form displays and entity displays. 

You can do this by doing a re-install in this branch. Then checking the feature status (`drush feature-diff social_post`) and updating the features afterwards (`drush feature-update social_post -y`). Then you can commit the necessary changes. Although you have to be careful not to commit the uuid values and hashes in the configuration files and unnecessary module dependencies.

slowflyer:
@jaapjan that helped to solve the issue. But now one test still fails with:

The command "if [ "$TEST_SUITE" = "install_update" ]; then bash scripts/social/ci/build-install-update.sh $PR $BRANCH $TRAVIS_COMMIT; fi" failed and exited with 1 during .

jaapjan:
@slowflyer That test fail is actually a bug which we are still trying to solve. It affects all pull requests at the moment and we haven't pinpointed it yet. 

I manually reviewed your pull request and it works. It is a good first step but I do think we should not merge it yet because it is important that the item is not shown in the stream anymore. Also I think we should make the label less "robotic". It now says "A boolean indicating whether the Post is published." I think we can remove that description and just name the label of the checkbox "Published". This is the way it is done for nodes so users should be familiar with that naming convention.

slowflyer:
After modifying the views to show only content with status = 1, I got lost with the travis errors.
Any idea how I can move on?

Thanks

jaapjan:
Hi @slowflyer how is it going with this issue? Do you need any help? Perhaps we should chat on Drupal Slack regarding this one?

jaapjan:
Hi @slowflyer Looks like the feature status is not default. When updating a feature it is important that you do this when there are no optional modules enabled. So for example addtoany is an optional module and it could be one of the reasons why Travis is now saying that the feature state is not default. It is because the module is not enabled after the installation of Open Social but it is installed in your project I guess?

slowflyer:
Yes it is installed. But that's not the major issue. Just found the bug in Views and try to fix it. After that I will do a new install with this setup locally - this time without AddtoAny - and commit changes again.

Hope to get this done before lunch.

slowflyer:
@jaapjan still the same amount of errors ...
The first test ( #6555.1 failed) fails with:

> Social Follow Content    social_follow_content   Installed      Changed 
> ...
> The feature state is not default! 

I tried to solve it with a drush feature-diff and drush fateure-update, but no luck. I honestly have no idea where my changes affect this module.

But the most frustrating part is the fail of test 2,3,(4) (#6555.2 failed) with this:

>     When I am on "/user"                                                       # SocialMinkContext::visit()
    Then I should see "My Behat Topic created"                                 # SocialMinkContext::assertPageContainsText()
      The text "My Behat Topic created" was not found anywhere in the text of the current page. (Behat\Mink\Exception\ResponseTextException)

I tried to reproduce it manually and everything seems to be ok.
There was an issue with the handling of comments I solved this morning. That had caused that activities for comments have not been shown up in the stream. But that did'nt solved the issue. And Travis complains about the content missing not comments or replies.

jaapjan:
@slowflyer 

Ok, when I look into your changes I can not think of anything which might have impact on any of those failing tests so I understand your frustration. I'm looking into the changes you made and I do not understand why these files are here:

- field.field.message.create_post_profile_stream.field_message_related_object.yml
- field.field.message.create_post_profile_stream.field_message_destination.yml
- field.field.message.create_post_profile_stream.field_message_context.yml
- field.field.message.create_content_in_joined_group.field_message_related_object.yml
- field.field.message.create_content_in_joined_group.field_message_destination.yml
- field.field.message.create_content_in_joined_group.field_message_context.yml

Perhaps it's a good idea to check those again and see if those are needed and why? Maybe you did a `drush feature-update` for social_activity and those files where added as well? 

Another thing I noticed is that when I'm logged in as a normal user and I try to create a post in a group it does not show up in the stream. So this could explain the failing Behat tests as well. Perhaps you only test it with user 1?

Anyway, I suggest you to look over all the changes again and try and eliminate certain parts until the problem is located? Hope that helps a bit, let me know if you need further help!

slowflyer:
@jaapjan Have you done a drush entup before creating the post? Because otherwise it would not have a status field and therefore not be shown anymore in stream. Because the updated views will look for post status =1.

I'll dive into the other remarks.

jaapjan:
I have done the following steps in this order:

```
drush -y cr
drush -y updatedb
drush -y entup
drush -y fra --bundle=social
```

But I'll look again after you've checked the other remarks.

slowflyer:
@jaapjan I removed unnecessary files. Still same amount of failed tests, but now they more related to posts.
Does Behat really creates the posts with Drupal functionallity, or directly using SQL? Is there a chance to get the database of a test?

jaapjan:
@slowflyer Allright, it at least points us in a direction now! 

The Posts are created in the browser directly. We don't have a database dump from the tests; but you can run the Behat test locally and use VNC Viewer to look at the tests while it is running, maybe that helps pinpoint the issue:
https://www.drupal.org/docs/8/distributions/open-social/how-to-run-the-behat-tests-locally

I suspect that the posts are created but maybe have the wrong status or the filter is not correct. It could also be that there is an error on creation of the post, but we don't have screenshots or .sql dump of the failing test unfortunately so it's best to test this locally on a fresh installation in this branch.

slowflyer:
@jaapjan all lights green ...

jaapjan:
Hi @slowflyer 

It mostly works and the approach is great, but we noticed 3 issues still:

1. On form save we need to clear the cache tags for activities (so the activity streams are updated for everyone)
2. We need to make clear that a post is unpublished, we can check how this is done for nodes?
3. When the post is unpublished (and cache cleared) then it is not visible anywhere, not even for SM (could be a permission thing, so it might be fine).

I can look into it tomorrow some more and come-up with a solution for the above which will also work for other entities in the activity creator. Will keep you "posted".

jaapjan:
@slowflyer  I've pushed some changes for number 1. It will update activities cache tags whenever an entity is updated. I think this totally makes sense to be solved generic. For number 2 and 3 I think we should create a follow-up. Given that in distro we have the `/admin/content/post` overview there is always a way to reach these posts for SM.

jaapjan:
Created a follow-up here:
https://www.drupal.org/project/social/issues/2968935

jaapjan:
I think we should leave it like it is now. No reason to have unpublished posts in activity stream for now. In future we can add it after the new entity access control layer has landed in core.

bramtenhove:
@slowflyer, thanks for your patience and hard work! Much appreciated!

PR #716, opened 2018-02-01
DS-4680 - Add ability for forked repositories to run Open Social tests on Travis CI

ronaldtebrake:
## Problem
Currently forks don't run automated tests due to the nature of the .travis.yml file where we hardcore `goalgorilla/open_social` as repo name. 

## Solution
By setting the forked repository as a Composer git repo we can overwrite the goalgorilla/open_social package which is pulled in and build.

Also we adjusted the commit has and branch as these are different for PR's.


bramtenhove:
This PR will be replaced by https://github.com/goalgorilla/open_social/pull/811 since we need more checks than just this forked repo PR build.

PR #720, opened 2018-02-07
Issue #2942747 by ribel: Improve messages on failed login, and password reset

ribel:
## Problem
See drupal.org issue

## Solution
See drupal.org issue

## Issue tracker
- https://www.drupal.org/project/social/issues/2942747

## HTT
- [ ] Try to reset password or login with not existing users
- [ ] Check how current messages look
- [ ] Checkout to this branch
- [ ] Try to reset password or login with not existing users and check new messages

## Documentation

### DIFF
--- a/modules/social_features/social_user/social_user.module
+++ b/modules/social_features/social_user/social_user.module
@@ -11,6 +11,8 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\user\Plugin\Action\BlockUser;
 use Drupal\social_user\Plugin\Action\SocialBlockUser;
+use Drupal\Core\Url;
+use Drupal\Core\Link;
 
 /**
  * Implements hook_action_info_alter().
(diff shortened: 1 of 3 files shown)


jaapjan:
Hey @ribel 

Marking this as needs work. Although it seems like a small change I'm really hesitant about merging this. The reason is that we are exposing the e-mail address of the site in the message. Drupal core never does this. Only for account e-mails etc. (which you only receive when you fill in site contact form, you create account, or password reset which can be turned of on certain sites). So I think this should either be made configurable or we don't show that part.

The new messages are a big improvement already!

jaapjan:
@ribel could you finish the work on this? Maybe together with IT-Crowd :-)

ribel:
@jaapjan, @bramtenhove, @jochemvn I will add a new checkbox under site email address, see how it will look: https://prnt.sc/li82zw
By default, it will be disabled, and we will not display email address in the status messages.
But for some projects, like Pachamama were they already use this as a patch they will need just enable this setting after the update.
What do you think about this approach and texts of a new setting?

Issue #726, opened 2018-02-08
"Add the hero styled image" fails when using custom content types and not following convention: "field_{$node->getType()}_image", or if you want to reuse an image field from another content type


g3r4:
If you extend and use a custom content type (not provided by Open Social) and happens to have a long name, the validation will fail because machine names can only be 26 characters long.
In order to at least have the possibility of having an image set, we need code that helps to load an image field as hero, as long as it has the word "image" on its field name and it is an image field.

I will add a pull request with this extra handling for your consideration

jaapjan:
Hey @g3r4 Thanks for your contribution! I think this is a good idea. I've created an issue on drupal.org for this since we use that for issue tracking:

https://www.drupal.org/project/social/issues/2944419

I will continue there and in the pullrequest created and close this issue for now.

PR #727, opened 2018-02-08
Added extra handling for hero images on custom content types fixes is…

g3r4:
Issue #726

## Problem
If you use a custom content type (not provided by Open Social) and happens to have a long name, the validation above will fail because machine names can only be 26 characters long.
In order to at least have the possibility of having an image set, new code needs to load an image field as hero, as long as it has the word "image" on its field name and it is an image field.

## Solution
I've written code to handle this special case

## Issue tracker
Created the issue directly on GitHub

## HTT
- [ ] Login as user admin and create a content type with a "long" name (21 chars or more), then assign an image field following convention "<content type name> image" so machine name would be field_<content_type_name>_image, create a new node based on that content type including uploading an image.
- [ ] Notice it doesn't work, hero image won't load
- [ ] Checkout to this branch
- [ ] On the content type you just created, if you add an image field that has the word "image" in its machine name (you can even re-use one of the image fields established for the default content types) and is an image field, the hero image will load.

## Documentation

### DIFF
--- a/modules/social_features/social_core/social_core.module
+++ b/modules/social_features/social_core/social_core.module
@@ -101,8 +101,10 @@ function template_preprocess_page_hero_data(array &$variables) {
         $field = reset($image_fields);
         if ($field != NULL){
           if ($field->getFieldDefinition()->get("field_type") == "image") {
-            $variables['hero_styled_image_url'] = ImageStyle::load('social_xx_large')
-              ->buildUrl($node->get($field->getName())->entity->getFileUri());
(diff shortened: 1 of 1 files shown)


jaapjan:
I've created a new PR for this https://github.com/goalgorilla/open_social/pull/728 in order to run the necessary Travis checks. Will let you know once it is reviewed.

jaapjan:
I've merged the mentioned PR and I've given you credit in the commit message @g3r4 Thanks again for your contribution!

PR #734, opened 2018-02-16
DS-4632 landing page follow up

left23:
## Problem
Landing pages needs some small style tweaks, and the feature section items that do not have images instead show a grey space which can give the appearance of something being broken.

## Solution
When there is no image, we keep the space where the image should be and add a hr to separate it from the text below it. Removed grey background.

## Issue tracker
DS-4632 & DS-4791

## HTT
- [x] Check out this branch
- [x] Login as admin and enable landing page module
- [x] Create landing page with demo content for all sections
- [x] Add landing page node to one of the Featured items

## Documentation

TO DO:
The block sections have many permutations based on the display title config options. This is difficult to cater for. TBD


jaapjan:
As discussed: the alignment of the titles are not correct yet. After fixing that we can merge!

PR #735, opened 2018-02-16
Issues #2945254 by serhiy-chornobay: Fix flexbox issues in IE11

Crazylord:
## Problem
- Members table does not display correctly for Managing members tab in groups
- Filter button on search pages does not display correctly

## Solution
- Set min-width for members name.
- Set justify-content: space-between for button's wrapper. Remove margin-right: auto for second button.

## Issue tracker
https://www.drupal.org/project/social/issues/2945254

## HTT
- [x] Check out this branch
- [x] Login as admin and clear cache
- [x] Go to members search page and checn filter buttons.
- [x] Go to page group/1/membership and check the member's name.

## Documentation


themes/socialbase/assets/css/form-elements.css
```diff
@@ -2,7 +2,6 @@ label {
   display: inline-block;
   max-width: 100%;
   margin-bottom: 5px;
-  font-weight: bold;
```
left23: @Crazylord I'm curious why the font-weight here was removed.

Crazylord: @left23 
Looks like somebody added this only in css file and after generating styles - this was removed.
I don't know if this is need or not. If yes - need add to scss file this rule.

left23: Oh yes, I see. I don't know if it's needed either. Thanks for the response

PR #736, opened 2018-02-16
Number of likes to be added to the CSV export

BSaxion:
## Problem
Number of Likes to be added as a column when exporting user to CSV

## HTT
- [ ] Try to export a user or users to CSV(using the user_export module)
- [ ] Then you will not have a column heading "Likes with a number of likes a user did"
- [ ] Adding such a feature may be useful to track the activity of a user

### DIFF
--- a/modules/social_features/social_user_export/social_user_export.module
+++ b/modules/social_features/social_user_export/social_user_export.module
@@ -6,6 +6,7 @@
  */
 
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Session\AccountInterface;
 use Drupal\user\UserInterface;
 use Drupal\views\Views;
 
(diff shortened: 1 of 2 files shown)


modules/social_features/social_user_export/social_user_export.module
```diff
@@ -189,6 +189,26 @@ function social_user_export_groups_count(UserInterface $account) {
...
+ *
+ * @param \Drupal\user\UserInterface $account
+ *   The account to get the data for.
+ *
+ * @return int
+ *   number of likes
+ */
+function social_user_export_likes_count(UserInterface $account) {
+  $query = \Drupal::database()->select('votingapi_vote', 'v');
+  $query->addField('v', 'type');
```
robertragas: I think you don't need this addField since in the end you will count the results. You can see the other export queries where they also don't use the addField.

robertragas:
Looks good, approved it! The failing test is because of a bug in the multilanguage it seems , since it fails on switching language which is not related to this code.

Now we must wait till @bramtenhove finds time to check/merge this PR.

PR #742, opened 2018-02-20
Don't use deprecated RawDrupalContext property

Kingdutch:
## Problem
Interacting directly with the `RawDrupalContext::$user` property has been deprecated. Use `RawDrupalContext::getUserManager->getCurrentUser()` instead.

This causes tests to fail for some projects.

## Solution
Follow the suggestion and replace `$this->user` with `$this->getUserManager->getCurrentUser()` in SocialDrupalContext

## Issue tracker
No drupal.org issue

## HTT
- [ ] Tests don't fail in a new way

## Documentation
No documentation change needed

### DIFF
--- a/tests/behat/features/bootstrap/SocialDrupalContext.php
+++ b/tests/behat/features/bootstrap/SocialDrupalContext.php
@@ -33,7 +33,8 @@ class SocialDrupalContext extends DrupalContext {
    * @Given I am viewing my :type( content):
    */
   public function assertViewingMyNode($type, TableNode $fields) {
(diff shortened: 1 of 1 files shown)


Kingdutch:
This PR will need to be applied when we upgrade our docker set-up to drupal-extension 3.4

This issue was triggered in a project because no specific version of the drupal-extension was required. The project had a "*" version specification for the behat drupal-extension and Open Social doesn't specify its dependency on a specific version anywhere for composer to find out.

PR #745, opened 2018-02-20
Redirect to post comment after editing

BSaxion:
## Problem
After editing a comment on a post, upon save you are direct to the edit screen. This is not correct.
When editing a comment on a topic, it does work correct: you return to the topic detail page. Same for reply on a comment: works like a charm!
## Solution
I think best is to return to the post detail page, scrolled down to your comment.
## Issue tracker
- https://www.drupal.org/project/social/issues/2934118

## HTT
 - [ ] create a comment on a post
 - [ ] edit the comment, press save
 - [ ] observe that you end up on the edit comment screen.
 - [ ] You do get a message that your comment has been posted.

Steps to reproduce the fix:

 - [ ] create a comment on a post
 - [ ] edit the comment, press save
 - [ ] observe that you end up on the post screen and you scroll down to edited comment.
 - [ ] You do get a message that your comment has been posted.

### DIFF
--- a/modules/social_features/social_post/social_post.module
+++ b/modules/social_features/social_post/social_post.module
@@ -67,25 +67,36 @@ function social_post_form_post_form_alter(&$form, FormStateInterface $form_state
  *
  * Alter the comment_post_comment_form form.
  */
-function social_post_form_comment_post_comment_form_alter(&$form, FormStateInterface $form_state, $form_id) {
-  // Reset title display.
-  $form['field_comment_body']['widget'][0]['#title_display'] = 'invisible';
-
-  // Set the action of the form to the current uri.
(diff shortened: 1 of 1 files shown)


modules/social_features/social_post/social_post.module
```diff
@@ -104,6 +107,7 @@ function _social_post_comment_post_comment_form_submit(array $form, FormStateInt
   if (UrlHelper::isValid($uri)) {
     // Get the Url from the Form value and redirect to this url.
     $url = Url::fromUserInput($uri);
+    var_dump($url);
```
ronaldtebrake: Hi Brhane, make sure to remove these. 

A good idea is to always check in github what code you have actually pushed to make sure no debug messages land in the finald push.

PR #746, opened 2018-02-20
DS-4870 - Fix SM permissions and configuration errors

bramtenhove:
## Problem
We encountered several problems related to each other.

1.  We added the permission for SM to add and remove languages, but did not give them the permission to translate texts. This is inconsistent.
2. Also the SM was not able to translate account emails and basic site information although that was the intention.
3. And in the case that there was just a single language the translate tab didn't make much sense for account emails and basic site information.
4. In the case the language negotation was _not_ set to User we were still showing the field in the users profile. This wouldn't do anything for the user and would thus create confusion.

## Solution
1. The permission to add and remove languages has been removed.
2. Custom permissions for translating account emails and basic site information were added to the SM and access check was based on that.
3. In the case of the single available language the translate tab was removed for account details and basic site information.
4. If the language negotation was not set for User the language field was removed from the users profile.

## HTT
- [x] Log in as SM and go to /admin/config/people/accounts and see that there is _no_ Translate tab
- [x] Go to /admin/config/system/site-information/translate and see that there is _no_ Translate tab
- [x] As an administrator add another language, e.g. Dutch and set it as default
- [x] Login as SM and verify that you _cannot_ add or remove languages
- [x] Go to /admin/config/people/accounts and see that there is a Translate tab
- [x] Translate a few emails for the added language, e.g. the register email
- [x] Go to /admin/config/system/site-information/translate and see that there is a Translate tab
- [x] Translate the Site name
- [x] As an anonymous user register yourself, observe that you have an email in the other language
- [x] Observe that the site name is as you just translated as the SM
- [x] As an administrator set the language negotation at /admin/config/regional/language/detection to _not_ include User
- [x] Go to your settings and observe that the language field is _not_ there
- [x] Set the language negotation at /admin/config/regional/language/detection to include User
- [x] Go to your settings and observe that the language field is there
- [x] Change your preferred language and observe that the site takes your preferred language into account

_Note: I have observed that removing English as a language is a bad idea and it removes the ability to translate emails and site information. Keeping English, but setting for example Dutch as the default language still gives you the ability to customise emails. I will write documentation for this feature on Drupal.org and include this._

## Documentation
No changes needed since it was already added to the release notes and this bug is solved before the release.


OH17L:
@bramtenhove Fix coding standards.

PR #757, opened 2018-02-27
Issue #2948273 by ribel: Add alter hook for social groups

ribel:
…refactoring in socialbase and social_group form alters

## Problem
There are various customizations for interaction with groups in function social_group_form_alter() of social_group module. But this function contain hardcoded arrays of group types. And if developer wants to reuse the same customization for custom group types he need to copy/paste this code, which is not a good way due to fact that social_group code might be changed in new versions

```
$action_forms = [
    'group_content_open_group-group_membership_group-join_form',
    'group_content_open_group-group_membership_group-leave_form',
    'group_content_open_group-group_membership_add_form',
    'group_content_open_group-group_membership_delete_form',
    'group_content_closed_group-group_membership_group-join_form',
    'group_content_closed_group-group_membership_group-leave_form',
    'group_content_closed_group-group_membership_add_form',
    'group_content_closed_group-group_membership_delete_form',
  ];
```

## Solution
Refactor code of social_group_form_alter() and add an ability to alter the hardcoded array of social group types and add new.

```
  $social_group_types = [
    'open_group',
    'closed_group',
    'public_group',
  ];
  \Drupal::moduleHandler()->alter('social_group_types', $social_group_types);

  $join_forms = [];
  foreach ($social_group_types as $social_group_type) {
    $join_forms[] = "group_content_$social_group_type-group_membership_group-join_form";
  }
```

## Issue tracker
- https://www.drupal.org/project/social/issues/2948273

## HTT
- [x] Create new group type and compare join/leave/edit forms and redirects with open, closed or public groups
- [x] Notice the difference
- [x] Checkout to this branch
- [x] Try the same interactions and see that they are consistent with default group types

## Documentation


Kingdutch:
I really love this addition. I was thinking about this exact problem just today. It can make the life for Enterprise updates a lot easier. I think this is a great first step. We should try to reuse this pattern when creating new functionalities (for Groups). Instead of applying them to a specific set of objects we can use an alter hook to make it opt-in. 

This can lead to a system where you can create new functionality simply by opting in to a set of building blocks.

Kingdutch:
This might be premature optimization but if we don't do it know I'm not sure it's ever possible. Could we perhaps limit the alter hook to be `group_type` and not `$group`?  Is there a scenario where we really need to group to figure out what the overview page is? 

The reason I ask this is is that calling an alter hook can be pretty expensive. We should try to do two things:

1. Base the alter hook on the group type rather than the group
2. See if we can use the routing's system of describing what parameters are needed for a route instead of providing these in alter hook.

With those changes we could implement caching for this alter hook (later), even if it's just a drupal_static call. One of the pages where I think this will have an impact is in the group overview page.

The reason we don't want something like a `\Drupal::currentUser()->id()` in the alter hook implementation is because then if we cache this for multiple users things might break.

ribel:
@Kingdutch thanks for your questions. To clarify for others I think you was talking about `hook_social_group_overview_route_alter`. I will try to answer:

1. I think group type should be enough to determine what overview page to use. I was thinking about possible more flexibility for developers, that's why included full `$group` entity. As you can see in the example I use `$group->bundle() == 'challenge'` to determine what overview page is needed. But if it impacts on the performance we can provide only `$group_type` as a parameter.

2. Regarding `\Drupal::currentUser()->id()` i just moved it from direct call in `_social_group_action_form_submit()` and `_social_group_edit_submit_redirect()` to `_social_group_get_overview_route()`, where alter hook is called. So now it shouldn't change anything regarding performance or cache. I am not sure how we can get current user route if it's called in submit callback of group creation form.

ribel:
@bramtenhove Thank you for the review! Conflicts were resolved.

PR #758, opened 2018-02-28
Issue #2948653 by Heine. Fix create access check on group types selec…

HeineDeelstra:
## Problem
The group/add page shows group types the user cannot create, because the role doesn't have create access.

## Solution

Adds a check on create access. Also added the appropriate caching metadata to the array element.

## Issue tracker

https://www.drupal.org/project/social/issues/2948653

## HTT
- [ ] On an existing installation
- [ ] Remove create [group type] access permision for one group type from authenticated users.
- [ ] Login as user with that role and go to /group/add
- [ ] Notice the option selection still includes the group type from above
- [ ] Checkout to this branch
- [ ] Go to /group/add and see that the group is no longer in the option list.

## Documentation

### DIFF
--- a/modules/social_features/social_group/src/Form/SocialGroupAddForm.php
+++ b/modules/social_features/social_group/src/Form/SocialGroupAddForm.php
@@ -5,6 +5,7 @@ namespace Drupal\social_group\Form;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormBase;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Render\RendererInterface;
 use Drupal\Core\Url;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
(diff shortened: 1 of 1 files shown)


HeineDeelstra:
I see that the CI check failed, but the messages do not give me any clue whether the CI fail is due to the content of the pull request. What can I do to cause the CI check to pass / figure out where the problem is?

bramtenhove:
@HeineDeelstra, we have some trouble with running tests for the CI with pull requests from forked repositories.

I'd be happy to add you to the contributor group of this repository. That would make it easier for you to open Pull Requests and run them on the CI. Let me know if you'd like that.

In the meantime I'll close this PR and re-open it based on the Open Social repository.

jochemvn:
@HeineDeelstra 
cc: @bramtenhove 
There are two instantiations of SocialGroupAddForm in the social_group.module file. That is the reason this PR fails. I think that that way the code in social_group.module has been written should be changed. Otherwise every new DI that is made here will result in crashes in those two places. The idea is that these functions try to get the group type element. This could maybe be done statically with an input array of groupstypes the current user is allowed to use?

Then the element can be taken out of the form with SocialGroupAddForm::getGroupTypeElement();

jochemvn:
Move from `limoengroup` to `opensocial` in https://github.com/goalgorilla/open_social/pull/797

PR #760, opened 2018-03-01
Issue #2948787 by ribel: Add separate route for the group stream page

ribel:
## Problem
Currently Group Stream page use `entity.group.canonical` route. Which makes a problem if you want to make Group About page as default when you click on the group. We often have such requirements for custom group types, and we always need to override it in custom modules.

## Solution
I propose to add separate route `social_group.stream` to the `social_group.routing.yml`, which will not break or change current behaviour but will make groups more flexible for custom modules. Then it will be possible to have the custom controller with logic like this:

```
class SocialChallengeController extends ControllerBase {

  /**
   * Redirect to default group home page.
   */
  public function redirectToGroupHomePage(GroupInterface $group) {
    // Redirect non-members of challenge group to the about page.
    // Redirect to the group stream by default.
    $route = 'social_group.stream';
    $account = \Drupal::currentUser();

    if ($group->bundle() === 'challenge' && !$group->getMember($account)) {
      $route = 'view.group_information.page_group_about';
    }

    return $this->redirect($route, [
      'group' => $group->id(),
    ]);
  }

}
```

## Issue tracker
- https://www.drupal.org/project/social/issues/2948787
- https://jira.goalgorilla.com/browse/FOR-1

## HTT
- [x] Checkout to this branch
- [x] Login as normal user and go to group pages as usual
- [x] Notice that all works like before
- [x] Try to add code like in solution, for example for `public_group` (you also need to add route to custom module `.routing.yml` and `RouteSubscriber.php`
- [x] Cleare the cache
- [x] Check that non-member of public groups is redirecting to the group about page
- [x] Check that in all other cases user is redirecting to the group stream page

## Documentation


modules/social_features/social_group/social_group.install
```diff
@@ -339,7 +338,6 @@ function social_group_update_8010() {
     else {
       user_role_grant_permissions($role->id(), [
         'create public_group group',
-        'view group stream page',
```
Kingdutch: We shouldn't be changing previous update hooks. Adding a non-existant permission in Drupal doesn't give any errors. If we want to explicitly remove the permission from a role then we should write an update hook anyway.

modules/social_features/social_group/social_group.group.permissions.yml
```diff
@@ -0,0 +1,3 @@
+view group stream page:
```
Kingdutch: Does this need an update hook to ensure that users that now have the 'view group stream page' permission get the 'view group stream page' group permission?

modules/social_features/social_group/social_group.install
ribel: You are right - adding non-existant permission in Drupal doesn't give any errors, but adding these permissions was not correct. I think it makes sense to remove it, to avoid confusion for a developer who will read this. Anyway, with removing this line or not - it doesn't change anything functionally.

modules/social_features/social_group/social_group.group.permissions.yml
ribel: We set group permission in config files. And this one _view group stream page_ was already there, so with this file it will just appear in Group Permissions UI. 
But thanks for a notice!

ribel:
@bramtenhove Why you marked this PR with `needs work` label? 
I think it's ready to be reviewed and merged, also since it's already used in few ECI projects it's important to get it merged.
I assume you add this label because of one failing behat test, but if you take a look it's not related to changes introduced in this PR, and was also in the parent branch.
I see you add it to 3.0 milestone, so we need to merge 8.x-3.x branch here, but I see the same and one more behat test is failing in 8.x-3.x branch are failing as well.
Could you fix those behat tests there, then I will merge it here and it will be clear if issues is in PR or in destination branch. Also to have destination branches green is important for other PR's

//cc @jaapjan

jaapjan:
Landing-page-create is a false negative. It's on our radar and we hope to make this test a bit more stable soon (same as create-new-private-message).

Tbh if the status is good enough I'm also fine with merging it for 2.2. I think this one is a bit on the line of new feature/minor improvement..

jaapjan:
@ribel I think it is also on `needs work` because this branch ( `test-distro/dev-feature/2948787-group-stream-route` on eci_demo project is failing. See Project Moss test result. I think the patch need to be removed from the composer.json of that project.

bramtenhove:
@ribel, thanks for your comment.

I marked it for 'needs work' because both the distro and Moss tests are failing. The distro one seems to be a random fail, I restarted it. We're looking into these tests, it's a bit annoying sorry.

But the Moss tests are failing because the patch is not being applied anymore in several projects. Take a look at https://primate.getopensocial.com/distro/pr/36.

bramtenhove:
@jaapjan we need to be careful about these patch fixes. If the patch is removed from the project and the tests fail we need to put it back so other PRs don't trip over the removed patch.

bramtenhove:
Good point @jaapjan.

modules/social_features/social_group/social_group.group.permissions.yml
jaapjan: I don't think we can write an update for that. You can't make assumptions about this permission and what it means per group type :(

PR #762, opened 2018-03-01
Issue #2949044 by Kingdutch: Use Drupal Core's phpunit.xml.dist file …

Kingdutch:
…instead of our own

By adding `@group social` to Open Social's test we allow the
Open Social tests to be run with `phpunit --testsuite unit --group
social` against Drupal Core's phpunit.xml.dist file.

## Issue tracker
- https://www.drupal.org/project/social/issues/2949044

## HTT
- [ ] First merge https://github.com/goalgorilla/open_social_scripts/pull/18
- [ ] Verify that after that's merged the unit tests run like normal

The following before and after output can be observed:

Before:
```
root@aaba00209d50:/var/www/html# php -d xdebug.extended_info=0 -d xdebug.remote_autostart=0 -d xdebug.coverage_enable=0 -d xdebug.profiler_enable=0 -d xdebug.remote_enable=0 /var/www/vendor/bin/phpunit -c /var/www/html/profiles/contrib/social/phpunit.xml.dist --testsuite social
PHPUnit 4.8.36 by Sebastian Bergmann and contributors.

..................................

Time: 12.29 seconds, Memory: 10.00MB

OK (34 tests, 134 assertions)
```

After:
```
root@aaba00209d50:/var/www/html# php -d xdebug.extended_info=0 -d xdebug.remote_autostart=0 -d xdebug.coverage_enable=0 -d xdebug.profiler_enable=0 -d xdebug.remote_enable=0 /var/www/vendor/bin/phpunit -c /var/www/html/core/phpunit.xml.dist --testsuite unit --group social
PHPUnit 4.8.36 by Sebastian Bergmann and contributors.

..................................

Time: 36.88 seconds, Memory: 90.00MB

OK (34 tests, 134 assertions)
```

## Documentation


Kingdutch:
Tests failed as expected until goalgorilla/open_social_scripts#18 is merged

jaapjan:
@bramtenhove If you agree with this solution perhaps you could look into finishing this one with @Kingdutch ?

bramtenhove:
Currently on _Don't Merge_ until answers are clear.

jaapjan:
@Kingdutch could you answer the question above?

Kingdutch:
So our own PHPUnit configuration actually does a lot of path mapping to distinguish Drupal core's tests from our own. 

Drupal's PHPUnit configuration splits the tests up by test type and uses a few discovery classes to find all the tests that are of a certain type. By assigning the `@group social` to the Open Social tests we can already split the Open Social tests from other tests. This allows us to achieve the same results without a custom PHPUnit configuration.

I hope that answers your question, if not, please send me a Slack message :)

jaapjan:
Thanks! Marking it as `needs review` for now. Although the unit tests are still failing (probably because we have to update all the unit-test scripts everywhere in saas, open_social scripts, enterprise projects etc. tbh I'm not sure yet if it is worth the hassle; but we can check it when we review it.

I remember we also created this in the beginning because we had to exclude specific address unit tests because those were failing. Just running the social test suite could solve that indeed.

ronaldtebrake:
Unfortunately couldnt put this in 3.0 due to: 
Performance impact of +20 seconds per stability test run vs what we gain is not something i'm comfortable with merging in right now. I'd like to see this change before merging it in the distro.

PR #764, opened 2018-03-01
bugfix for intro section in landing page by left: remove blockquote e…

left23:
…lement, align title with paragraph text


bramtenhove:
As a follow-up I suggest we do create a story to add a "quote" block as was our first intention. Is this something you can do @left23?

PR #766, opened 2018-03-02
2919743 - Improve accessibility of Open Social, so it complies with WCAG2.0 AA

Crazylord:
## Problem
- There is no top level heading (H1) on the AU homepage (related issue: https://www.drupal.org/project/social/issues/2946448)
- Add post form is missing a description(related issue: https://www.drupal.org/project/social/issues/2946451)
- Helper text is unreadable, to small and to less contrast(related issue: https://www.drupal.org/project/social/issues/2946485)
- For Keyboard users it is hard to use buttons(related issue: https://www.drupal.org/project/social/issues/2946492)
- Set the title attribute of event enrolled user picture links to their user name(related issue: https://www.drupal.org/project/social/issues/2938820)
- Landing page teaser should have a read more, so there are more consistent(related issue: https://www.drupal.org/project/social/issues/2946518)
- For keyboard user is not possible to change settings in the settings table of content(related issue: https://www.drupal.org/project/social/issues/2946488)

## Solution
- Add visually hidden H1 title for page, what doesn't have title.
- Add visually hidden label for post form
- Change contrast ratio for Helper text
- Add hover, focus efect for all buttons
- Add title attribute for user picture links
- Add read more button for Landing page teasers
- Add patch for bootstrap vertical tabs for fix problem navigation with keyboard

## Issue tracker
- https://www.drupal.org/project/social/issues/2919743

## HTT
- [ ] Checkout to this branch
- [ ] For reproduce steps - please check relevant issue

## Documentation


themes/socialbase/components/00-config/variables/_forms.scss
```diff
@@ -5,7 +5,7 @@ $input-height-base:              ($line-height-computed + ($padding-base-vertica
 $form-control-max-width:         23rem;
 
 // Color help texts that go with form-controls
-$help-block:                     $gray-light;
+$help-block:                     #464545;
```
Kingdutch: Can we move this color into a variable to describe what it's used for? Perhaps it needs to be used elsewhere too? That might also help indicate how this color was chosen over `$gray-light`.

themes/socialbase/templates/profile/profile--compact.html.twig
```diff
@@ -20,5 +20,7 @@
  */
 #}
 <div class="avatar">
-  {{ content }}
+  <a href="{{ profile_stream_url }}" title="{{ profile_name|render }}">
```
Kingdutch: This now breaks the widget settings "link to content" for site builders (perhaps some site had it disabled). Is it not possible to add the "#title" of the URL in a preprocess hook if the URL is set?

Crazylord: There was code in .theme file, where all settings "link to content" was overrid like  link. Also this method used in some another place for profile template. If for some site need disable it - need only orevvide template and remove link.

themes/socialbase/components/00-config/variables/_forms.scss
Crazylord: I don't really sure if we need this. This is very hard to explain when you can use this variable, because this is related to background color. I don't think this will be used somewhere else. If you really want - we can create variable with the color name (http://chir.ag/projects/name-that-color/).

themes/socialblue/assets/css/brand.css
```diff
@@ -1,268 +0,0 @@
-[type="radio"]:checked + label:before {
```
Kingdutch: Why are we removing brand.css? Does this not break the theme customisation (for SaaS)?

Crazylord: I checked this and look like all styles, what was in brand css are present in theme styles. This only override styles from theme and broke styles for button hover, focus state. I think there was made before and now it not needed. If you know some specific page, where it can be broken - let me know.

Kingdutch: @Crazylord It indeed duplicated some styles from the theme because the color module would parse this file and replace those colors with the configuration of the theme.

- Check out the current 8.x-1.x branch of Open Social
- Enable socialblue and set as default
- In the theme settings change the theme colors
- Observe that this properly works

- Check out this feature branch of Open Social
- Enable socialblue and set as default
- In the theme settings change the theme colors
- I believe that those changes are now broken (if not then maybe @left23 knows what changed?)

Crazylord: Maybe then we need only add this file and doesn't include for theme?
So, we will be have some separate file with colors should be use only for theme colors module.
I will check this.

themes/socialbase/templates/profile/profile--compact.html.twig
Kingdutch: I had hoped the following snippet would work:

```php
    // Only change the URL if it's already linked to content. This ensures that
    // the widget setting "Link to content" remains functional.
    if (!empty($build['field_profile_image'][0]['#url'])) {
      $uid = $entity->getOwnerId();
      $name = $entity->getOwner()->getDisplayName();

      // Get the url from the user profile.
      $url = Url::fromRoute('entity.user.canonical', [
        'user' => $uid,
      ], [
        'attributes' => [
          'title' => $name,
        ]
      ]);

      // Add it to the build.
      $build['field_profile_image'][0]['#url'] = $url;
    }
  }
```

It would be a more elegant solution, however it runs into https://www.drupal.org/project/drupal/issues/2829990

Crazylord: This doens't work so easy. If you set url for image - then for image drupal use "image-formatter.html.twig" template and there set link with image.

Kingdutch: We could change the image formatter template to copy the alt/title from the image onto the URL?

Crazylord: Yes, we can set alt from image like title attribute for link, but for now "Alternative text" aren't mandatory for profile image and can be not so clear as user name. If this is okay - I can change this like you say.

Kingdutch: I suppose for now we can just leave it as you made it. Maybe once the Drupal Core issue is fixed we can come back to a different solution

left23: Please see comment here: https://www.drupal.org/project/social/issues/2938820#comment-12523445
If you have agreed to leave the current commit as is (with the anchor and its title attribute) then can I suggest we go further and add the alt tag this way too?
I think the alt tag is more important than the title (especially since this is an accessibility-focused PR)

eg:   

social/themes/socialbase/templates/profile/profile--compact.html.twig
<code>
<a href="{{ profile_stream_url }}">
    {{ profile_name }}{{ ' profile image'|t
    }}
  </a>
</code>

/social/modules/social_features/social_profile/social_profile.module
<code>
  // Add the profile image.
  if (!empty($profile->field_profile_image->entity)) {
    $variables['profile_image_url'] = ImageStyle::load('social_medium')
      ->buildUrl($profile->field_profile_image->entity->getFileUri());
  }
</code>

Can you see issues with this approach?

left23: I don't like having to set the image dimensions

Crazylord: @left23 For now image has alternative text as well, only need set this for user, because we doesn't have alternative text for images as mandatory and this is empty. So we will have the both title for link and alternative text for image.

left23: Ok, thanks for explaining further. 
It seems like a blank alt field is almost as ineffective as no alt tag. Same result.
How do you propose we set this for the user?

Crazylord:
@left23 Yes, I checked it and all looks good.

Kingdutch:
Adding "Needs work" because there's still active discussion in: https://www.drupal.org/project/social/issues/2938820

We should probably avoid grouping this many issues in a single PR, even if they're all closely related so that one does not stall another and reviewing/discussing remains easy.

themes/socialbase/templates/profile/profile--compact.html.twig
Crazylord: @left23 I think we need set alternative text for all image as mandatory for Open Social. But maybe for this need separate task.

Crazylord:
@left23 I think this is okay. Only one remark - maybe better for `<a href="{{ profile_stream_url }}" title="{{ profile_name }}">` add title like:  ` title="{{ profile_name ~ ' profile'| trans }}"`

Kingdutch:
@left23 This works for me. 

I think in the future we should think about adding the user's name to the default image as well. That ensures that the preview list is also available to screenreaders, it's now dependant on the user having a custom image set.

I don't think adding the "profile" suffix is needed for the title. If we do decide that that's needed then we should do it properly because @Crazylord's  suggested implementation would create a translatable string for each user on the platform.

See https://getlevelten.com/blog/mark-carver/drupal-8-twig-templates-and-translations for background and explanations.

themes/socialbase/templates/profile/profile--compact.html.twig
```diff
@@ -20,7 +20,14 @@
...
 <div class="avatar">
-  <a href="{{ profile_stream_url }}" title="{{ profile_name|render }}">
-    {{ content.field_profile_image }}
+  <a href="{{ profile_stream_url }}" title="{{ profile_name }}">
+
+    {% if profile_image_url %}
+      <img src="{{ profile_image_url }}" class="img-responsive" alt="{{ profile_name }}">
+    {% else %}
+      {# if default_image #}
+      {{ node.field_profile_image }}
```
Kingdutch: @left23 you moved from `{{ content.field_profile_image }}` to `{{ node.field_profile_image }}` here. Is that intentional?

left23: No, I was experimenting and didn't roll back. Thanks for spotting this. I've removed the trans for profile name in title attribute too.

PR #768, opened 2018-03-02
2925592 Disable e-mail notifications of people liking my comments

OH17L:
## Problem
User settings page does not have a field for controlling e-mail notification which you received when someone liking your post, comment, etc.

## Solution
Implement possibility to control e-mail notifications about liking my stuff.

## Issue tracker
- https://www.drupal.org/project/social/issues/2925592
- https://jira.goalgorilla.com/browse/DS-4477

## HTT
- [x] Login and go to user settings page for user X
- [x] You should see an option for control liking in **Email notifications** section
- [x] Select `None` and as user Y like a post of user X
- [x] User X should not have received any notifications regarding the like
- [x] Select `Immediately` and as user Y like a post of user X
- [x] User X should receive a notifications regarding the like
- [x] Select `Daily` or `Weekly` and as user Y like a post of user X
- [x] Wait a day or a week, or force the digest cron to run
- [x] User X should receive a digest notification regarding the like

## Documentation

### DIFF
--- a/modules/custom/activity_send/modules/activity_send_email/activity_send_email.install
+++ b/modules/custom/activity_send/modules/activity_send_email/activity_send_email.install
@@ -8,24 +8,6 @@
 use Drupal\Core\Database\Database;
 
 /**
- * Implements hook_install().
- *
- * Perform actions related to the installation of activity_send_email.
- */
-function activity_send_email_install() {
-  // Set email destination to given message templates.
-  $message_templates = ['create_post_profile'];
-  foreach ($message_templates as $message_template) {
-    $config_name = "message.template.{$message_template}";
-    $config = \Drupal::service('config.factory')->getEditable($config_name);
-    $activity_destinations = $config->get('third_party_settings.activity_logger.activity_destinations');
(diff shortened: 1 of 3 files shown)


modules/custom/activity_send/modules/activity_send_email/activity_send_email.install
```diff
@@ -7,24 +7,6 @@
 
 use Drupal\Core\Database\Database;
 
-/**
```
bramtenhove: @OH17L, could you tell me why you removed this part?

OH17L: @bramtenhove, I checked message template files and did not see that not all files have E-mail destination. Should I restore this code or will be better if I add this destination to all templates without it?

bramtenhove: Thanks @OH17L, I think we can leave it out. The `message.template.create_post_profile.yml` file already contains the destination so we should ok.

Issue #24, opened 2016-08-25
Unresolved Dependencies


kreativmind:
Posted this yesterday.
Trying to upgrade from alpha 6 to 7 and get the following message when running update.php

Unresolved dependency group_core_comments (Missing)

Any help is appreciated please. Thanks!

https://www.drupal.org/node/2789721

ronaldtebrake:
See linked drupal.org story for further communication.

PR #770, opened 2018-03-05
2528214 by robertragas: add new patch number for restrict image text-…

robertragas:
## Problem
When restrict images is checked in a text format in some cases it gives a wsod like the following:

Error: Call to undefined method Drupal\Core\Image\Image::setAttribute() in filter_filter_secure_image_alter() (line 835 of /app/html/core/modules/filter/filter.module) #0 /app/html/core/lib/Drupal/Core/Extension/ModuleHandler.php(501): filter_filter_secure_image_alter(Object(Drupal\Core\Image\Image), NULL, NULL) #1 /app/html/core/modules/filter/filter.module(817): Drupal\Core\Extension\ModuleHandler->alter(‘filter_secure_i...‘, Object(Drupal\Core\Image\Image)) #2 /app/html/core/modules/filter/src/Plugin/Filter/FilterHtmlImageSecure.php(25): _filter_html_image_secure_process(‘<p>In dit inter...‘) #3

## Solution
Update the patch number. See comment #24 for the solution.
https://www.drupal.org/node/2528214

## Issue tracker
- https://www.drupal.org/node/2528214

## HTT
Not sure how to reproduce it. I always test it on a client where this goes wrong.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "patches": {
             "drupal/core": {
                 "Color module html preview optional": "https://www.drupal.org/files/issues/color-optional-html-preview-2844190-2.patch",
-                "Restrict images to this site blocks image style derivatives": "https://www.drupal.org/files/issues/image_restrict_image_styles-2528214-19.patch",
+                "Restrict images to this site blocks image style derivatives": "https://www.drupal.org/files/issues/image_restrict_image_styles-2528214-24.patch",
                 "Fix SQL error when importing translations": "https://www.drupal.org/files/issues/2599228-31.patch",
                 "Optimize getCommentedEntity()": "https://www.drupal.org/files/issues/get-commented-entity-2580551.52.patch"
             },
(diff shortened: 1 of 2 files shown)


jaapjan:
@robertragas don't forget to update the .make files :-)

Kingdutch:
@bramtenhove @chmez I don't understand why all these unrelated commits are suddenly in this PR. I just ran into this issue again with the SDGC update and would really like to get this updated patch committed.

bramtenhove:
@Kingdutch quite interesting. I have no idea where this is coming from.

@OH17L any clue?

jaapjan:
Maybe I did something wrong here? Looks like I merged the remote-tracking branch there... :)

Kingdutch:
@jaapjan @bramtenhove Rebased this to clean it up, now only contains the actual changes.

jaapjan:
I see a comment related to this:

`Patch #24 does not work for 8.5.3 for embedded images with image styles.`

Is there any (enterprise) site where we could encounter this issue?

bramtenhove:
@robertragas @Kingdutch is the comment from @jaapjan relevant? Or could this be relevant in the near future?

Kingdutch:
For your comment: I don't think so. This is not something that we support yet. Adding image styles to embedded images is a different patch at the moment (I believe).

PR #771, opened 2018-03-06
Feature/export event details

BSaxion:
## Problem
Export event details

## Solution
Ability to export a list of events with name, location, event host, event managers, and enrolled participants

## HTT
- [ ] install the module and enable it
- [ ]  Go to admin/content and select some events 
- [ ] Select "export content to csv"
- [ ] Then click on Apply for selected items
- [ ] Then you will have a csv file exported for those events

### DIFF
--- a/modules/custom/social_content_export/social_content_export.module
+++ b/modules/custom/social_content_export/social_content_export.module
@@ -1,0 +1,139 @@
+<?php
+
+/**
+ * @file
+ * The social_content_export module.
+ */
+
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\views\Views;
+
+/**
+ * Returns event location.
+ *
+ * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+ *   The entity to look for data to export.
+ *
+ * @return mixed
+ *   Return location of event to take place.
+ */
+function social_content_export_event_location(ContentEntityInterface $entity) {
+  // Get location of event.
+  if ($location = $entity->get('field_event_location')->value) {
+    return $location;
+  }
+  else {
+    $location = '';
+    return $location;
+  }
+
+}
+
+/**
+ * Returns event host.
+ *
+ * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+ *   The entity to look for data to export.
+ *
+ * @return mixed
+ *   Returns address of event.
+ */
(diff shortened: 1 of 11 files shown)


modules/custom/social_content_export/social_content_export.module
```diff
@@ -0,0 +1,143 @@
...
+ * Returns event organizer.
+ *
+ * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+ *   The entity to look for data to export.
+ *
+ * @return mixed
+ *   Return event organizer.
+ */
+function social_content_export_event_organizer(ContentEntityInterface $entity) {
+  $query = \Drupal::database()->select('event_enrollment_field_data', 'eo');
```
Kingdutch: A database query is not needed for this. This is just a value on the node. Please look up the proper way on how to access a value of a field on an entity. Ask me for help if you get stuck.

Make sure you check if the entity has this field and provide a fallback value if it doesn't. Not all node types that we export have this field.

Hint: You're dealing with an entity reference relationship which sometimes is a bit more difficult. You'll have to fetch a user object and return its username.

modules/custom/social_content_export/social_content_export.module
```diff
@@ -0,0 +1,143 @@
...
+    ->fetchField();
+}
+
+/**
+ * Implements hook_file_download().
+ */
+function social_content_export_file_download($uri) {
+  $scheme = \Drupal::service('file_system')->uriScheme($uri);
+  $target = file_uri_target($uri);
+  $access = \Drupal::currentUser()->hasPermission('administer node types');
```
Kingdutch: @bramtenhove or @jaapjan what would be a more appropriate permission to use here? `administer content`?

modules/social_features/social_post/social_post.module
```diff
@@ -69,19 +69,30 @@ function social_post_form_post_form_alter(&$form, FormStateInterface $form_state
...
-  // without breaking functionality.
-  $uri = Url::fromRoute('<current>')->toString();
-  // Store in a hidden field.
-  $form['redirect_after_save'] = [
-    '#type' => 'hidden',
-    '#title' => t('Redir'),
-    '#default_value' => $uri,
-  ];
-
+  $from['field_comment_body']['widget'][0]['#title_display'] = 'invisible';
```
Kingdutch: I don't think this should be in this PR.

PR #772, opened 2018-03-06
Add username and real name to export

BSaxion:
## Problem
Add user name and actual name when exporting users".

## Solution
Apply the changes and you will get a CSV file whenever you export with FirstName, LastName and UserName

## Origin
This pull request fixes OSSUPPORT-428


modules/social_features/social_user_export/src/Plugin/Action/ExportUser.php
```diff
@@ -87,13 +87,13 @@ public static function create(ContainerInterface $container, array $configuratio
    */
   public function executeMultiple(array $entities) {
     if ($this->applyAll) {
-      $this->tempStoreFactory->get('user_operations_export')->set($this->currentUser->id(), [
+      $this->tempStoreFactory->get('user_operations_export')->set($this->current_user->id(), [
```
robertragas: Hey Brhane , just a reminder that this one is not the correct name. You said you corrected it, but I guess you need to push those changes then.

modules/social_features/social_user_export/social_user_export.module
```diff
@@ -85,6 +85,65 @@ function _social_user_export_get_view(array $query = [], $execute = TRUE) {
...
+ * Returns first name of user.
+ *
+ * @param \Drupal\user\UserInterface $account
+ *   The account to get the data for.
+ *
+ * @return string
+ *   First name of the use.
+ */
+function social_user_export_first_name(UserInterface $account) {
+  $query = \Drupal::database()->select('profile__field_profile_first_name', 'pfn');
```
Kingdutch: We should use the Profile entity to retrieve the first name for a user instead of accessing the database directly. See `social_user_user_format_name_alter` for an example.

Also make sure that no errors occur of the user has no profile yet or if the first name field does not exist or has no value (for example due to the privacy module).

modules/social_features/social_user_export/social_user_export.module
```diff
@@ -85,6 +85,65 @@ function _social_user_export_get_view(array $query = [], $execute = TRUE) {
...
+ * Returns LastName of user.
+ *
+ * @param \Drupal\user\UserInterface $account
+ *   The account to get the data for.
+ *
+ * @return string
+ *   LastName of the user.
+ */
+function social_user_export_last_name(UserInterface $account) {
+  $query = \Drupal::database()->select('profile__field_profile_last_name', 'pln');
```
Kingdutch: We should use the Profile entity to retrieve the last name for a user instead of accessing the database directly. See `social_user_user_format_name_alter` for an example.

Also make sure that no errors occur of the user has no profile yet or if the first name field does not exist or has no value (for example due to the privacy module).

modules/social_features/social_user_export/social_user_export.module
```diff
@@ -85,6 +85,65 @@ function _social_user_export_get_view(array $query = [], $execute = TRUE) {
...
+ * Returns username of user.
+ *
+ * @param \Drupal\user\UserInterface $account
+ *   The account to get the data for.
+ *
+ * @return mixed
+ *   The username of the user.
+ */
+function social_user_export_user_name(UserInterface $account) {
+  $query = \Drupal::database()->select('users_field_data', 'ufd');
```
Kingdutch: Please use `$account->getAccountName()` instead (defined on the AccountInterface interface and implemented in the User entity).

No database query is needed for this.

modules/social_features/social_user_export/src/ExportUser.php
```diff
@@ -34,6 +34,9 @@ public static function exportUserOperation(UserInterface $entity, array &$contex
       $headers = [
         t('ID'),
         t('UUID'),
+        t('FirstName'),
```
Kingdutch: Lets use the same proper formatting of the other fields and make this `First name`

modules/social_features/social_user_export/src/ExportUser.php
```diff
@@ -34,6 +34,9 @@ public static function exportUserOperation(UserInterface $entity, array &$contex
...
+        t('LastName'),
```
Kingdutch: Lets use the same proper formatting of the other fields and make this `Last name`

modules/social_features/social_user_export/src/ExportUser.php
```diff
@@ -34,6 +34,9 @@ public static function exportUserOperation(UserInterface $entity, array &$contex
...
+        t("username"),
```
Kingdutch: Lets use the same proper formatting of the other fields and make this `Username`

modules/social_features/social_user_export/src/Plugin/Action/ExportUser.php
```diff
@@ -59,11 +59,11 @@ class ExportUser extends ActionBase implements ContainerFactoryPluginInterface {
    *   The plugin implementation definition.
    * @param \Drupal\user\PrivateTempStoreFactory $temp_store_factory
    *   The tempstore factory.
-   * @param \Drupal\Core\Session\AccountInterface $current_user
-   *   Current user.
+   * @param \Drupal\Core\Session\AccountInterface $currentUser
```
Kingdutch: Please keep the snake_case here which is usual for function variable names. camelCase is usually reserved for class properties.

modules/social_features/social_user_export/src/Plugin/Action/ExportUser.php
```diff
@@ -59,11 +59,11 @@ class ExportUser extends ActionBase implements ContainerFactoryPluginInterface {
...
+   *   The current user.
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, PrivateTempStoreFactory $temp_store_factory, AccountInterface $current_user) {
-    $this->currentUser = $current_user;
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, PrivateTempStoreFactory $temp_store_factory, AccountInterface $currentUser) {
```
Kingdutch: Please keep the snake_case here which is usual for function variable names. camelCase is usually reserved for class properties.

modules/social_features/social_user_export/src/Plugin/Action/ExportUser.php
```diff
@@ -59,11 +59,11 @@ class ExportUser extends ActionBase implements ContainerFactoryPluginInterface {
...
+    $this->currentUser = $currentUser;
```
Kingdutch: Please keep the snake_case here which is usual for function variable names. camelCase is usually reserved for class properties.

Kingdutch:
Hey Brhane,

Why did you change from `UserInterface` to `AccountInterface`? I think it would be good to study the difference between the two.

The former is for storing Drupal user data. The latter is for representing the current session (which could come from a different place than Drupal user data).

Kingdutch:
Hi @BSaxion,

I think this still needs some work. I think it would be good if you looked at the concept of `Entities` in Drupal (8) and how they work.

Once you've done that I think it'd be good to look at my comment about the UserInterface vs the AccountInterface. After that I think you can see that in the `exportUserOperation` we already have a User entity so I don't think you need the `social_user_export_first_name` and `social_user_export_last_name` functions (or at least don't need to do another database query within them).

Please let me know if you have any questions.

~ Alexander

modules/social_features/social_user_export/src/ExportUser.php
```diff
@@ -21,6 +21,8 @@ class ExportUser {
    *   UserInterface entity.
    * @param array $context
    *   Context of the operation.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
```
Kingdutch: Hey Brhane,

Can you explain which line can throw an InvalidPluginDefinitionException?

~ Alexander

BSaxion: Hey Alexander,
line 102, at the snippet: \Drupal::entityTypeManager()->getStorage('profile');

Kingdutch: So this would mean that if the profile module is not installed on the platform then the user export can not be used? Can you find a solution to that? ;-)

BSaxion: Hey, Alexander,
I have tried to fix what you mentioned by specifying that profile module is required by user export module.
Brhane

Kingdutch: Hey Brhane,

I actually liked the fact that we didn't have a dependency on the Social Profile module. Perhaps we can do something like a `try catch` statement ( http://php.net/manual/en/language.exceptions.php ). This way we can use the profile field if it exists and if we don't have it enabled we just return the empty string value :)

Make sure you're specific enough in which error you catch so we don't mask other errors to other developers.

modules/social_features/social_user_export/social_user_export.module
```diff
@@ -85,6 +86,65 @@ function _social_user_export_get_view(array $query = [], $execute = TRUE) {
...
+/**
+ * Returns first name of user.
+ *
+ * @param \Drupal\user\UserInterface $account
+ *   The account to get the data for.
+ *
+ * @return string
+ *   First name of the use.
+ *
+ * @throws InvalidPluginDefinitionException
```
Kingdutch: You implemented error handling so this function no longer throws anything! :)

modules/social_features/social_user_export/social_user_export.module
```diff
@@ -85,6 +86,65 @@ function _social_user_export_get_view(array $query = [], $execute = TRUE) {
...
+    $storage = \Drupal::entityTypeManager()->getStorage('profile');
+    // Checking if Storage is not empty.
+    if (!empty($storage)) {
+      if ($user_profile = $storage->loadByUser($account, 'profile', TRUE)) {
+        $first_name = $user_profile->get('field_profile_first_name')->value;
+      }
+    }
+  }
+  catch (InvalidPluginDefinitionException $e) {
+    $first_name = "No first Name is provided";
```
Kingdutch: Returning the empty `$first_name` here is probably better than a "magic constant".

modules/social_features/social_user_export/social_user_export.module
```diff
@@ -85,6 +86,65 @@ function _social_user_export_get_view(array $query = [], $execute = TRUE) {
...
+/**
+ * Returns LastName of user.
+ *
+ * @param \Drupal\user\UserInterface $account
+ *   The account to get the data for.
+ *
+ * @return string
+ *   LastName of the user.
+ *
+ * @throws InvalidPluginDefinitionException
```
Kingdutch: You implemented error handling so this function no longer throws anything! :)

modules/social_features/social_user_export/social_user_export.module
```diff
@@ -85,6 +86,65 @@ function _social_user_export_get_view(array $query = [], $execute = TRUE) {
...
+    $storage = \Drupal::entityTypeManager()->getStorage('profile');
+    // Checking if Storage is not empty.
+    if (!empty($storage)) {
+      if ($user_profile = $storage->loadByUser($account, 'profile', TRUE)) {
+        $last_name = $user_profile->get('field_profile_last_name')->value;
+      }
+    }
+  }
+  catch (InvalidPluginDefinitionException $e) {
+    $last_name = 'No last name is provided';
```
Kingdutch: Returning the empty $first_name here is probably better than a "magic constant".

modules/social_features/social_user_export/src/ExportUser.php
Kingdutch: You implemented error handling so this no longer throws an exception :)

bramtenhove:
Possibly also include Display Name as this could be used instead of first and last name (e.g. nickname)

bramtenhove:
Thanks @BSaxion!

PR #776, opened 2018-03-07
ECI-702 by serhiy-chornobay: Make the user menu like dropdown on mobile

Crazylord:
## Problem
Currently, on small screens such as iphone 5, the header menu becomes 2 rows which takes a lot of space for mobile user and the menu looks 'broken'. 

Also for enterprise team, when the add google translate icon to the header menu, on mobile the space is not enough, even for large phones. 

## Solution
When the screen is small, such as iphone 5, the user menu should become a dropdown menu. 
When there is new notification for private message or notificaiton center, the dropdown icon will reflect there is new things unread. 
There should not be a configuration interface to decide which icon will be in dropdown on mobile. (We only take care of the icons provided in distro, not other customised icons.)

## Issue tracker
https://jira.goalgorilla.com/browse/ECI-702

## HTT
- [ ] Checkout to this branch
- [ ] Login and check User menu on mobile
- [ ] Try expand user dropdown menu and check if there are notification icon over user icon, if user have some nothification.

## Documentation

### DIFF
--- a/modules/custom/activity_creator/js/notifications.js
+++ b/modules/custom/activity_creator/js/notifications.js
@@ -23,7 +23,7 @@
                           var remaining_notifications = result['remaining_notifications'];
 
                           notification_count.html(remaining_notifications);
-                        $('.notification-bell-mobile .badge').html(remaining_notifications);
+                        $('.notification-bell.mobile .badge').html(remaining_notifications);
                       }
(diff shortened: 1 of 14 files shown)


Crazylord:
@left23 If all works good - I think this is okay!

PR #789, opened 2018-03-14
Improve social_comment_upload module quality

Kingdutch:
This pull request makes no functional changes to the social_comment_upload module but attempts to improve the contained code and avoid bugs in the future.

It makes the following changes (see the related commits for proper explanations):

- Improve social_comment_upload module file legibility
- Use a config file for settings and make update functions static
- Don't nuke third_party display settings
- Don't use getEditable for runtime configuration checking
- Provide a schema file for social_comment_upload settings

## Issue tracker
https://www.drupal.org/project/social/issues/2989044

## HTT
- [x] Login as user site manager and verify the comment upload setting functions as normal 
- [x] Login as a normal user and see that the comment upload function works as expected

## Documentation


Kingdutch:
```
It is worth reiterating that if you need to read the original configuration values in an override, for example to compare or merge, you would need to load from the \Drupal::configFactory and not \Drupal::config to avoid a nested loop into your override: -

$original = \Drupal::configFactory()->getEditable('system.site')->getOriginal('name', FALSE);
```

jaapjan:
I've updated the PR with the latest 8.x-2.x. @Kingdutch could you create a drupal.org issue for this? and add it to the PR?

Kingdutch:
@jaapjan I've created a d.o issue and updated the PR here: https://www.drupal.org/project/social/issues/2989044

PR #792, opened 2018-03-15
Issue #2953281 by ribel: Add Social Event AN Enroll module

ribel:
## Problem
As AN I want to be able to enroll to an event without having to create an account

## Solution
Design: https://marvelapp.com/390g4e6/screen/38982268

## Issue tracker
- https://www.drupal.org/project/social/issues/2953281
- https://jira.goalgorilla.com/browse/PAC-423

## HTT
- [ ] Enable module Social Event AN Enroll
- [ ] As AN user try to enrol in a public event
- [ ] Check that you receive an email
- [ ] Check that you able to cancel enrollment
- [ ] Try to enrol two times with the same account and see a message about it and 2 emails
- [ ] Login as normal user and go to this public event
- [ ] Check default avatar with a counter of anonymous enrollments
- [ ] Login as CM+ and go to this public event
- [ ] Check new tab "Anonymous enrollment" with names and emails of AN enrollees
- [ ] Check new settings page `/admin/config/opensocial/event-an-enroll ` and try to change email text

## Documentation


jochemvn:
And on a more general note: Very cool feature!! ❤️

ribel:
@jochemvn Thank you for good review!

I have fixed almost all items that you described. I am not only sure what should we do with both items 1 because it was like this by design.

jochemvn:
@ribel  Maybe @bramtenhove can ask Xinyu to take a look at those two questions? Let him come up with a solution

bramtenhove:
FYI: I pinged Xinyu

jaapjan:
@ribel 

Could you add how to test to the pull request?

jaapjan:
@ribel 

I think we need this before we can merge it to distro:

- Be able to turn it on/off per site
- Support for GDPR consent
- Nice to have: let Event Manager decide if they want to enable it per event

jochemvn:
@jaapjan , what about just making a setting PER event which is not turned on by default. In that case we might not need a sitewide setting, since it can be controlled on a per event (per role) basis?

jaapjan:
@jochemvn even then I think it would make sense to make it configurable (can be a combination with permission as well, e.g. there can be cases where sitemanager does allow it but only content manager can actually set the per-event setting)

xinyumaGG:
Hi guys, 

I like the designs in general. My questions are about the enrolments and manage enrolments pages:

- How do anonymous users look like on the enrolments page
- What happens on the manage enrolments page, when there are a lot of users signed up, will these two blocks have a pagination? 

Cheers,
Xinyu

basvanos:
Hi @xinyumaGG, 

- We still nee to work on this. We need to create a follow up for this. Already did a proposal in this story > https://jira.goalgorilla.com/browse/PAC-423
- yes i prefer to work with pagination per block, but also for this we need to think a bit more on this when we add management of LU enrollments. 

Cheers, Bas

modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
```diff
@@ -0,0 +1,223 @@
...
+          $output['#rows'][0]['#theme'] = $output['#theme'];
+          $output['#rows'][0]['#view'] = $output['#view'];
+          $output['#rows'][0]['#grouping_level'] = 0;
+          $output['#rows'][0]['#title'] = '';
+        }
+        $output['#rows'][0]['#rows'][] = [
+          '#prefix' => '<div class="avatar">',
+          '#theme' => 'image_style',
+          '#style_name' => 'social_medium',
+          '#uri' => 'public://default-profile-picture.png',
```
Kingdutch: Hey Taras,

This used to be correct but it will break once the private file system is enabled (such as is the case on the Pachamama platform). We should probably find the default image for the user profile's (because a platform may change this) and use that here. That should fix the issue that PAC is currently having.

Perhaps this can be done by loading the anonymous user and getting the image field value?

~ Alexander

Kingdutch: @ribel ^

ribel: @Kingdutch You solution sounds good. I can check it when I will be back to work. Thanks!

Kingdutch: @ribel Hi Taras, did you get a chance to look at this?

ribel: @Kingdutch I have fixed it by using existing function `social_profile_get_default_image()`

ribel:
@jochemvn I have implemented the following improvements:

- Event manager should be able to access AN Event enrollee list of his event
- Event managers should be able to determine whether events can accept anonymous registrants per event (only for public events)
- When Event manager+ change visibility of event with anonymous enroll they should get a warning
- If visibility moved or AN Enrollment disabled AN enrollees should not be shown to anyone
- SM should be able to turn feature on/off per site

So now all requested changes are made, except only GDPR. But we created follow up, and will work on GDPR next sprint.

modules/social_features/social_follow_content/social_follow_content.module
```diff
@@ -27,25 +27,27 @@ function social_follow_content_event_enrollment_update(EntityInterface $entity)
...
-      'uid' => $account->id(),
-      'entity_type' => 'node',
-      'entity_id' => $entity->field_event->target_id,
-      'flag_id' => 'follow_content',
-    ];
-    $flaggings = \Drupal::entityTypeManager()
-      ->getStorage('flagging')
-      ->loadByProperties($properties);
-    $flagging = reset($flaggings);
+    if (!$account->isAnonymous()) {
```
Kingdutch: Can we add this as a guard clause at the top. i.e.

```
// Only logged in users can follow content.
if (!$account->isLoggedIn()) {
  return;
}
```

This avoids rewriting this entire function in this PR and communicates intent.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+   */
+  public function getFormId() {
+    return 'event_an_enroll_cancel_form';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $nid = $this->routeMatch->getRawParameter('node');
```
Kingdutch: We should add this as a form argument:
```
public function buildForm(array $form, FormStateInterface $form_state, NodeInterface $node) {
```

Then when calling the form we should do:
```
getForm('Drupal\social_event_an_enroll\Form\EventAnEnrollActionForm', $node);
```
as per https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Form%21FormBuilder.php/function/FormBuilder%3A%3AgetForm/8.5.x

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Url;
+use Drupal\node\Entity\Node;
+
+/**
+ * Class EventAnEnrollActionForm.
+ *
+ * @package Drupal\social_event_an_enroll\Form
+ */
+class EventAnEnrollActionForm extends EnrollActionForm {
```
Kingdutch: Why are we extending the EnrollActionForm class? I understand that this is the enrollment but for anonymous users. However, we don't reuse any logic from the EnrollActionForm class which suggests that these are two entirely separate forms and shouldn't inherit from eachother.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
Kingdutch: This should also be fixed for the `EnrollActionForm` class or a follow-up should be created.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+        '#value' => $nid,
+      ];
+
+      $form['enroll_for_this_event'] = [
+        '#type' => 'submit',
+        '#value' => $this->t('Enrolled'),
+        '#attributes' => [
+          'class' => [
+            'btn',
+            'btn-accent brand-bg-accent',
```
Kingdutch: We should keep one class per line so that it's more easily manipulated in other places (e.g. using `attributes.removeClass(classname)`).

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+            'btn-lg btn-raised',
```
Kingdutch: We should keep one class per line so that it's more easily manipulated in other places (e.g. using attributes.removeClass(classname)).

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+    else {
+      if ($this->eventHasBeenFinished($node)) {
+        $form['event_enrollment'] = [
+          '#type' => 'submit',
+          '#value' => $this->t('Event has passed'),
+          '#disabled' => TRUE,
+          '#attributes' => [
+            'class' => [
+              'btn',
+              'btn-accent brand-bg-accent',
```
Kingdutch: We should keep one class per line so that it's more easily manipulated in other places (e.g. using attributes.removeClass(classname)).

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+              'btn-lg btn-raised',
```
Kingdutch: We should keep one class per line so that it's more easily manipulated in other places (e.g. using attributes.removeClass(classname)).

tests/behat/features/capabilities/event-an-enroll/event-an-enroll.feature
```diff
@@ -0,0 +1,67 @@
+@api @javascript @event @eventenrollment @stability @perfect @PAC-423 @profile @stability-3 @event-an-enroll
+Feature: Enroll for an event without an account
+  Benefit: In order to attend an Event
+  Role: AN
+  Goal/desire: I want to be able to enroll to an event without having to create an account
+
+  @AN
+  Scenario: Successfully enroll to an event as AN user
```
Kingdutch: This test currently tests a lot more than the scenario suggests. Is it possible to split the code up into multiple scenario's so that it's clearer what is broken if the test fails?

bramtenhove:
@ribel, @jochemvn, I noticed there is some valuable feedback from @Kingdutch in this PR. Do you think you are able to process this as well?

Thanks in advance!

ribel:
@Kingdutch @jochemvn  Thank you for the review! I think we will be able to fix those items, during GDPR improvements this sprint.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollForm.php
```diff
@@ -57,6 +59,32 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      ]);
+
+      $enforce_consent = !empty($this->configFactory->get('data_policy.data_policy')->get('enforce_consent'));
+
+      $form['data_policy'] = [
+        '#type' => 'checkbox',
+        '#title' => $this->t('I agree with the @url', [
+          '@url' => $link->toString(),
+        ]),
+        '#required' => $enforce_consent && $this->currentUser->isAnonymous(),
```
Kingdutch: @ribel Isn't the user always anonymous here? : )

ribel: You are right, I will remove it.

jochemvn:
Hey @ribel, looks and works good. Could you fix the issue mentioned by @Kingdutch? 

Also we need to look into the failing tests.

tests/behat/features/capabilities/event-an-enroll/event-an-enroll.feature
ribel: @Kingdutch I have fixed all items that you noticed, but would like to not split this behat test because of lack of time before upcoming release. Could you check again all changes and approve if you think it's ok?

ribel:
@Kingdutch Sorry, I miss those other comments, because they didn't appear here on a stream. I will check and try to fix those on Monday.

modules/social_features/social_event/modules/social_event_an_enroll/css/event_an_enroll.scss
```diff
@@ -0,0 +1,34 @@
+.event_an_enroll.row {
+  margin-bottom: 5px;
+  @media (min-width: 600px) {
+    width: 600px;
```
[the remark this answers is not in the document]
ribel: This was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/css/event_an_enroll.scss
```diff
@@ -0,0 +1,34 @@
...
+
+  .col-md {
+    width: 100%;
+    @media (min-width: 600px) {
+      width: 50%;
+    }
+  }
+
+  h5 {
+    font-weight: 500;
```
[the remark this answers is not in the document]
ribel: This was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/css/event_an_enroll.scss
```diff
@@ -0,0 +1,34 @@
...
+    margin-bottom: 20px;
```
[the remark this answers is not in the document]
ribel: This was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/css/event_an_enroll.scss
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+.avatar {
```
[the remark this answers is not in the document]
ribel: This was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
```diff
@@ -0,0 +1,275 @@
...
+}
+
+/**
+ * Override variables for the social page hero data.
+ */
+function social_event_an_enroll_preprocess_page_hero_data(array &$variables) {
+  $account = \Drupal::currentUser();
+  if ($account->isAnonymous() && !empty($variables['event_enrollment'])) {
+
+    $node = \Drupal::routeMatch()->getParameter('node');
```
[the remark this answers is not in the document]
ribel: This was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
```diff
@@ -0,0 +1,275 @@
...
+
+/**
+ * Implements hook_views_post_render().
+ *
+ * Alter "Event enrollments" views. Add number of anonymous enrollments.
+ */
+function social_event_an_enroll_views_post_render(ViewExecutable $view, &$output, CachePluginBase $cache) {
+  if ($view->id() == 'event_enrollments' && $view->current_display == 'event_enrollments') {
+    if (isset($output['#rows'])) {
+      $nid = \Drupal::routeMatch()->getRawParameter('node');
```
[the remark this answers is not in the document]
ribel: This was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
```diff
@@ -0,0 +1,275 @@
...
+    ['node' => $nid],
+    [
+      'query' => ['token' => $token],
+      'absolute' => TRUE,
+    ])->toString();
+  return $url;
+}
+
+/**
+ * Implements hook_menu_local_tasks_alter().
```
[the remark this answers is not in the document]
ribel: This was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
```diff
@@ -0,0 +1,275 @@
...
+  $is_event = $node->getType() === 'event';
+  $is_public = $node->get('field_content_visibility')->getString() === 'public';
+  $is_event_an_enroll = !empty($node->get('field_event_an_enroll')->value);
+  return $is_global_enabled && $is_event && $is_public && $is_event_an_enroll;
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function social_event_an_enroll_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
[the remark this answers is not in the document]
ribel: This hook is needed for adding #states conditions to field_event_an_enroll.
I have added the description.

modules/social_features/social_event/modules/social_event_an_enroll/src/Controller/EventAnEnrollController.php
```diff
@@ -0,0 +1,151 @@
...
+  /**
+   * Determines if user has access to enroll form.
+   */
+  public function enrollAccess(NodeInterface $node) {
+    $config = $this->config('social_event_an_enroll.settings');
+    $is_global_enabled = $config->get('event_an_enroll');
+    $is_event = $node->getType() === 'event';
+    $is_public = $node->get('field_content_visibility')->getString() === 'public';
+    $is_event_an_enroll = !empty($node->get('field_event_an_enroll')->value);
+    if ($is_global_enabled && $is_event && $is_public && $is_event_an_enroll) {
```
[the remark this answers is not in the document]
ribel: You are right, but the difference between those two are that in `social_event_an_enroll_is_enabled()` we use global `\Drupal::config` and in `enrollAccess` we use dependency injection.
And it's hard to use only `enrollAccess` since `EventAnEnrollController::__construct` needs `$routeMatch` and `$entityTypeManager`. 
If yo think it's ok to call `social_event_an_enroll_is_enabled()` in `enrollAccess` I can fix that.

modules/social_features/social_event/modules/social_event_an_enroll/src/EventAnEnrollOverride.php
```diff
@@ -0,0 +1,72 @@
...
+        'weight' => 100,
+        'settings' => [
+          'display_label' => TRUE,
+        ],
+        'third_party_settings' => [],
+        'type' => 'boolean_checkbox',
+        'region' => 'content',
+      ];
+
+      $overrides[$config_name]['third_party_settings'] = $third_party;
```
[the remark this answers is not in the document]
ribel: This was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/src/EventAnEnrollOverride.php
```diff
@@ -0,0 +1,72 @@
...
+
+    // Add field_event_an_enroll to event form.
+    $config_name = 'core.entity_form_display.node.event.default';
+    if (in_array($config_name, $names)) {
+      $config = $config_factory->getEditable($config_name);
+
+      $third_party = $config->get('third_party_settings');
+      $third_party['field_group']['group_event_visibility']['children'][] = 'field_event_an_enroll';
+
+      $content = $config->get('content');
```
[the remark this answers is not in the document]
ribel: I think it's not possible to avoid overriding of all `content` property since we need to add new element right to under it. See also `SocialEventTypeConfigOverride`  and `SocialEventManagersConfigOverride`

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
ribel: This extending is actually needed and useful because we reuse method `eventHasBeenFinished()` and also dependency injections from parent class like `$this->entityStorage`, `$this->currentUser` etc

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
ribel: It was fixed here, but I see that `EnrollActionForm` doesn't use `routeMatch`, so I haven't change anything there.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
ribel: It was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
ribel: It was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+              'waves-effect',
+            ],
+          ],
+        ];
+      }
+      else {
+        $attributes = [
+          'class' => [
+            'use-ajax',
+            'button',
```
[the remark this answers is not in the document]
ribel: You are right, it's not needed, I have removed it.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+            'button--accent',
```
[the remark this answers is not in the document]
ribel: You are right, it's not needed, I have removed it.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+            'js-form-submit',
+            'form-submit',
+            'btn-lg',
+            'btn',
+            'js-form-submit',
```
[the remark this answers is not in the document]
ribel: It was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -0,0 +1,146 @@
...
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $current_user = $this->currentUser;
+    $uid = $current_user->id();
```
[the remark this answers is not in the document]
ribel: It was fixed.

modules/social_features/social_follow_content/social_follow_content.module
ribel: It was fixed.

modules/social_features/social_event/modules/social_event_an_enroll/src/Controller/EventAnEnrollController.php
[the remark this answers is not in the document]
Kingdutch: I think the benefits of simply calling the function `social_event_an_enroll_is_enabled` outweigh the negatives of calling a static function from a class context.

We don't yet do any testing where not having this dependency injection is a problem and it ensures that if we ever change the requirements for `social_event_an_enroll_is_enabled` then this controller will also work as expected. 

If we really need  dependency injection at some point then we can always think about moving things like `social_event_an_enroll_is_enabled` into a service so that it can be used in both places.

modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
```diff
@@ -0,0 +1,275 @@
...
+  $event_forms = [
+    'node_event_form',
+    'node_event_edit_form',
+  ];
+
+  if (in_array($form_id, $event_forms)) {
+    // @TODO: Fix states when group is selected.
+    $form['field_event_an_enroll']['#states'] = [
+      'visible' => [
+        ':input[name="field_content_visibility"]' => ['value' => 'public'],
```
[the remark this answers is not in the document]
Kingdutch: @ribel I placed this as a comment to potentially fix a bug that I think @jochemvn mentioned. Is that still something worth looking at?

Kingdutch:
Loving the improvements @ribel ! Looks great!

Added one last question about a javascript field state. I think that was related to a bug that Jochem mentioned that we can maybe prevent :) 

Other than that I think this is ready to go 💯

modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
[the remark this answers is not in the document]
ribel: @Kingdutch I have tried to solve it before, and also with your suggestion, but without any luck :(
The difficulty that it's not just enough if groups were filled or not, but what exactly group was selected, becuase for public groups we need to display AN Enroll checkbox and for other - not. 
And also because Group is not a normal select field, but our custom widget, which is already altered in `social_group_field_widget_form_alter()`.

modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
```diff
@@ -0,0 +1,275 @@
...
+  $message['subject'] .= PlainTextOutput::renderFromHtml($subject);
+  $message['body'][] = $body;
+  $message['headers']['From'] = $message['from'];
+}
+
+/**
+ * Replace token in email.
+ */
+function social_event_an_enroll_token_replace($replace, $subject) {
+  return str_replace(['[event name]', '[event url]'], $replace, $subject);
```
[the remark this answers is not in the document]
ribel: You are right. I have replaced it with Drupal core way.

ronaldtebrake:
Tests for Pachamama failed due to PR https://github.com/goalgorilla/open_social/pull/981 which breaks a patch. Will be fixed in that PR after this is merged.

bramtenhove:
@ribel nice work on all the improvements and fixes you manage to do still! 👏
I'm happy it's in release 3.0 🎉

Kingdutch:
@ribel @bramtenhove It looks like we forgot an instance of the token in the help text: https://github.com/goalgorilla/open_social/blob/8.x-3.x/modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollSettingsForm.php#L83

bramtenhove:
Correct @Kingdutch, noticed this as well after the release.
Check https://github.com/goalgorilla/open_social/pull/1023 where @ribel already changed it.

PR #794, opened 2018-03-16
check if entity is instance of FieldItemListInterface

slowflyer:
## Problem
When trying to edit the visibility field of a content type it fails with an error

## Solution
Add check to verify that the entity is an instance of FieldItemListInterface, otherwise $entity->hasField('xxx') will fail, while hasField does not exist

## Issue tracker
- https://www.drupal.org/project/social/issues/2953546

## HTT
- [x] Login as admin and try to edit visibility field of a content type
- [x] Notice it doesn't work - You will get a message on screen like "The website encountered an unexpected error. Please try again later."
- [x] Checkout to this branch
- [x] Try this and this and that and see that it works now - try to edit and save the visibility field again and you should succeed

## Documentation


slowflyer:
Thanks! That works for me!

PR #795, opened 2018-03-16
ECI-709: Update styles for landing hero button

Crazylord:
## Problem
When adding more then 2-3 buttons in the landing page hero, the styling breaks.

## Solution
Update styles for buttons, so they have padding with left and right side.
Add negative margin for wrap.

## Issue tracker
- https://www.drupal.org/project/social/issues/2953554

## HTT
- [ ] Checkout to this branch and clear cache
- [ ] Add 2-3 buttons with long text for landing hero section and check how it looks

## Documentation


modules/social_features/social_landing_page/css/social_landing_page.section.hero.css
```diff
@@ -34,10 +34,12 @@
...
     font-weight: normal;
     font-size: 20px; }
+  .paragraph--hero .cover-small .hero-footer .field--name-field-hero-buttons {
+    margin: 0 -10px; }
   .paragraph--hero .cover-small .hero-footer .field--item {
     display: inline-block; }
-    .paragraph--hero .cover-small .hero-footer .field--item:not(:first-child) .paragraph--button {
-      padding-left: 20px; }
+    .paragraph--hero .cover-small .hero-footer .field--item .paragraph--button {
+      padding: 0 10px; }
```
Kingdutch: Why don't we just change this to `padding-right: 10px` then your negative margin is not needed.

Crazylord: Then content(buttons) will stretch correctly in left and right corner.

left23: I don't understand. In my testing, padding-right: 10px is enough. I don't see any negative side effects to this.
I prefer this over needing to add a negative margin.
@Crazylord can you explain please, otherwise perhaps you can review this please

Crazylord: If you set padding-right - then buttons will be break to another line earlier. In my case, button will be stick to right side

left23: Understood. Thanks for explaining

PR #797, opened 2018-03-19
#2948653 by Heine. Fix create access check on group types select

jochemvn:
## Problem
The group/add page shows group types the user cannot create, because the role doesn't have create access.

## Solution

Adds a check on create access. Also added the appropriate caching metadata to the array element.

## Issue tracker

https://www.drupal.org/project/social/issues/2948653

## HTT
- [x] On an existing installation
- [x] Remove create [group type] access permision for one group type from authenticated users.
- [x] Login as user with that role and go to /group/add
- [x] Notice the option selection still includes the group type from above
- [x] Checkout to this branch
- [x] Go to /group/add and see that the group is no longer in the option list.

## Documentation


jochemvn:
@HeineDeelstra 
cc: @bramtenhove 
There are two instantiations of SocialGroupAddForm in the social_group.module file. That is the reason this PR fails. I think that that way the code in social_group.module has been written should be changed. Otherwise every new DI that is made here will result in crashes in those two places. The idea is that these functions try to get the group type element. This could maybe be done statically with an input array of groupstypes the current user is allowed to use? It depends a bit on the usecase, but looking at it, it might work...

Then the element can be taken out of the form with SocialGroupAddForm::getGroupTypeElement();

I've made a fix, but I don't think we should allow this. I'm already using your code so I needed a quickfix.

PR #798, opened 2018-03-20
DS-4972 Make landing page block names less technical

OH17L:
## Problem
Landing page block names are technical.

## Solution
Make landing page block names less technical.

## Issue tracker
https://jira.goalgorilla.com/browse/DS-4972

## HTT
- [x] Enable **Social Landing Page** module
- [x] Go to **/node/add/landing_page**
- [ ] Press **ADD SECTION** and press **ADD BLOCK**
- [ ] You should see default name of blocks (for example **Activity overview**) and block groups (for example **Lists (Views)**)
- [ ] Checkout to this branch
- [ ] If you update page and repeat third step then you should see changed name of blocks (for example **Activity details**) and only one blocks group **Basic**


modules/social_features/social_landing_page/src/Plugin/Field/FieldWidget/BlockFieldWidget.php
```diff
@@ -0,0 +1,81 @@
...
+    $groups = [
+      $this->t('Lists (Views)')->render() => $this->t('Basic'),
+      $this->t('Social Landing Page')->render() => $this->t('Basic'),
+    ];
+
+    $blocks = [
+      'activity_overview_block' => $this->t('Full activity stream'),
+      'views_block:activity_stream-block_stream_homepage' => $this->t('Full activity stream with post field'),
+      'views_block:community_activities-block_stream_landing' => $this->t('Latest activities'),
+      'views_block:latest_topics-block_latest_topics' => $this->t('Latest topic'),
```
jaapjan: @OH17L Shouldn't this be "Latest topics"

PR #799, opened 2018-03-20
Add number of private messages to export if module exists

BSaxion:
## Problem
It is needed that the number of private messages to be added to user's CSV export

## Solution
I add some extra functionality to export number of private messages to the module social_user_export.


Kingdutch:
Hi @BSaxion,

It looks like this Pull Request contains a bit more code than you'd need for what you're trying to do. Please clean up the PR to ensure it only contains the changes to add the private messages to the export.

~ Alexander

Kingdutch:
Hey Brhane, it looks like you now have 2 commits that say "Cleaning commits and removing some not needed commits." instead of having actual cleaned up commits 😅 

You also still have changes in `modules/social_features/social_post/social_post.module` of which I'm not sure they're needed? 

You should also check out the merge conflicts that GitHub is reporting and see if you can use an entityQuery here for the private message count instead of a database query.

bramtenhove:
Thanks @BSaxion!

PR #803, opened 2018-03-21
DS-4886 As a LU I want to see a clear activity stream without duplicate info regarding posts + comments

OH17L:
## Problem
The problem is that there is a lot of duplicate information when people comment on a post in a group stream.

## Solution
Hide all activities about the commented post on group stream page.

## Issue tracker
https://jira.goalgorilla.com/browse/DS-4886

## HTT
- [ ] Go to stream page of some group
- [ ] Create comment on some post
- [ ] You should see your comment in the first position of a stream and you can see second time your comment after activity about creating a commented post
- [ ] Checkout to this branch
- [ ] You should not see your comment in the first position of a stream


jaapjan:
Behat tests are failing. Please check them and make adjustments accordingly

PR #806, opened 2018-03-22
Issue #2931336 replace string for page title and config description

left23:
## Problem
The title of the "Notifications" page (/notifications) is now "Activity Stream". 

I think this is due to the configuration yml having being copied and these strings were overlooked

## Solution
Title should be "Notifications" 

## Issue tracker
https://www.drupal.org/project/social/issues/2931336

## HTT
- [ ] Login as LU + and go to /notifications
- [ ] Notice the page title reads "activity stream"
- [ ] Checkout to this branch
- [ ] *I am not sure how you load this new configuration - I uninstalled social_activity module and reinstalled it
- [ ] Notice the page title reads "Notifications"

## Documentation

### DIFF
--- a/modules/social_features/social_activity/config/install/views.view.activity_stream_notifications.yml
+++ b/modules/social_features/social_activity/config/install/views.view.activity_stream_notifications.yml
@@ -13,7 +13,7 @@ dependencies:
 id: activity_stream_notifications
 label: 'Activity stream notifications'
 module: views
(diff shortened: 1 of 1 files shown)


left23:
@OH17L This is only a configuration change but because it is in the install configuration, I am not sure if other steps are needed. Please let me know what you think.

left23:
Thanks @OH17L 
I added this to the release notes too

PR #810, opened 2018-03-27
Event enrollment stays open past time

BSaxion:
## Problem
Events are being marked as "passed"  prior to their start time.
## Solution
This is caused due to offset from UTC and DST, so that i made a change in timezone handling. Here we have to get offset against the UTC as event dates are stored in UTC at the database.
## Issue tracker
- https://www.drupal.org/project/social/issues/2956238#comment-12545543
## HTT
- [ ] Create a user with a different timezone than the event creator
- [ ] Checkout the event status on both the creator and user
- [ ] Then you will notice a difference


Kingdutch:
Closing this because there's a solution that use's Drupal's functions in #817

PR #812, opened 2018-03-27
Feature/ossupport 519 adding all profile fields to user export

BSaxion:
## Problem
Adding all profile fields to user export

## Solution
A code is provided or added to the user export module.

## HTT
- [ ] Export a user
- [ ] Then you will notice that it is export with all its details


bramtenhove:
There's a merge conflict. Could you take a look at it @BSaxion?

PR #813, opened 2018-03-29
Update patch for Drupal core issue #2528214

Kingdutch:
Due to names of variables in the previous patch sometimes non-existant methods would be called on objects. This causes errors on platforms using the patch.

An updated patch (applied in this commit) as well as a full explanation of the problem and solution have been posted in https://www.drupal.org/project/drupal/issues/2528214#comment-12501483

## Issue tracker
No Open Social issue queue issue because a Drupal core issue queue issue exists.
- https://www.drupal.org/project/drupal/issues/2528214#comment-12501483

## HTT
- [ ] Check out the Drupal core issue with updated patch
- [ ] Check out the code changes for that patch
- [ ] See that Drupal core tests still pass
- [ ] Check that the correct patch URL has been used

## Documentation
Don't think this requires mention in the release notes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "patches": {
             "drupal/core": {
                 "Color module html preview optional": "https://www.drupal.org/files/issues/color-optional-html-preview-2844190-2.patch",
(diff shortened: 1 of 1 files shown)


Kingdutch:
I failed to find it but this is a duplicate of #770

PR #815, opened 2018-04-02
DS-4551 As a LU, I want to see the corresponding activity stream item moved and updated when a content is moved

OH17L:
## Problem
In the first version of moving a content story, when a content is moved to a group, the activity stream item is just removed from the previous location. 

The concern for this is people might not notice the creation of this content because it is not visible in any activity stream. So it could be bad for content engagement.

## Solution
When user changing parent group of event or topic in **Group** field form for editing content then made following operations:

1. Delete activity about creating content in a stream of a group which was in **Group** field before changes
2. Create activity about creating content in a group which will be in **Group** field after changes
3. After saving content it will have new creating date equal date of saving changes

## Issue tracker
- DS-4551

## HTT
- [x] Log in as content manager
- [x] Create the public group **A** and **B**
- [x] Create the topic **T** in **A** and create the event **E** in **A**
- [x] Run cron and open steam page of **A** where you should see two activity records about creating **T** and **E**
- [x] Open edit form of **T** and choose **B** in field **Group** and save your change and run cron
- [x] When you open again stream page of **A** then you should see only one activity record about creating **E** and when you open the same page of **B** you should not see any activity records
- [x] Checkout to this branch and revert **Social Activity** feature and flush all caches
- [x] Open edit form of **E** and choose **B** in field **Group** and save your change and run cron
- [x] When you open stream page of **A** then you should not see any activity records but when you open the same page of **B** then you should see one activity record about creating **E**
- [x] Also you should see that creating date of **E** was changed and now is a date when you edit **E** but not when you created **E** in **A**

## Documentation


modules/social_features/social_group/social_group.tokens.inc
```diff
@@ -61,26 +61,38 @@ function social_group_tokens($type, $tokens, array $data, array $options, Bubble
...
                   }
-                }
+                  break;
+
+                // If it's node.
+                case 'node':
+                  $content_type = $entity->bundle();
+                  $article = 'a';
+
+                  if (preg_match('/^[aeioquy]/i', $content_type)) {
```
bramtenhove: @OH17L, I don't really understand what this is for. Can you explain it to me or perhaps add some inline comments?

OH17L: @bramtenhove I will add inline comments.

modules/social_features/social_activity/config/install/message.template.moved_content_between_groups.yml
```diff
@@ -0,0 +1,31 @@
...
+      stream_group: stream_group
+    activity_create_direct: 1
+    activity_aggregate: 0
+    activity_entity_condition: ''
+template: moved_content_between_groups
+label: 'Moved event or topic between groups'
+description: 'A person moved event or topic to other group'
+text:
+  -
+    value: '<p><a href="[message:author:url:absolute]">[message:author:display-name]</a> created&nbsp;[social_group:content_type] in <a href="[message:gurl]">[message:gtitle]</a></p>'
```
bramtenhove: @OH17L I think you should remove `&nbsp;` here.

tests/behat/features/capabilities/language/change-language.feature
```diff
@@ -49,7 +49,8 @@ Feature: Multilingual Open Social
     # Switch to Dutch.
     When I select "Dutch" from "Interface language"
     And I press "Save"
-    Then I should see the text "Lokalisatie-instellingen"
+#    Then I should see the text "Lokalisatie-instellingen"
```
bramtenhove: @OH17L I guess these commented test steps can be deleted right?

OH17L: @bramtenhove, no, it is not. I did it because current Drupal core does not have a translation for this phrase. When a translation of this phrase will be restored then this line is uncommented and next line is deleted.

modules/social_features/social_group/social_group.tokens.inc
```diff
@@ -75,13 +75,20 @@ function social_group_tokens($type, $tokens, array $data, array $options, Bubble
...
                 case 'node':
                   $content_type = $entity->bundle();
-                  $article = 'a';
 
-                  if (preg_match('/^[aeioquy]/i', $content_type)) {
-                    $article .= 'n';
+                  // When a name of content name starts from a vowel letter then
+                  // will be added "an" before this name. For example "an
+                  // event".
+                  if (preg_match('/^[aeiou]/', $content_type)) {
```
Kingdutch: What about `an hour`? And `a helpful comment`?

bramtenhove: In order to handle this well we need to add proper support for indefinite articles. We can take a look at how this is done elsewhere, an example being https://raw.githubusercontent.com/Kaivosukeltaja/php-indefinite-article/master/IndefiniteArticle.class.php.

Maybe we should take this out of this PR and add a different one for solving that specific issue, what do you think?

Kingdutch: We should find an external library to do this for us and also take care of other languages (maybe there's a module for that?)

bramtenhove: I'll merge this in for now, as it is already a bit of an improvement. I will open a new PR for fixing the indefinite articles properly.

PR #819, opened 2018-04-09
#2957223 - Update to Drupal Core 8.5

jaapjan:
## Problem
Open Social is still on 8.4.

## Solution
Update to 8.5.1

## Issue tracker
- https://www.drupal.org/project/social/issues/2957223

## HTT
- [x] Check out the Change Records: https://www.drupal.org/list-changes/drupal/published?keywords_description=&to_branch=8.5.x&version=8.5.0
- [x] Update your local environment to this branch
- [x] Test the site and see everything works as expected

## Documentation

### DIFF
--- a/modules/custom/social_demo/src/DemoContentParser.php
+++ b/modules/custom/social_demo/src/DemoContentParser.php
@@ -22,9 +22,19 @@ class DemoContentParser extends Yaml implements DemoContentParserInterface {
   }
 
   /**
-   * {@inheritdoc}
+   * Parses YAML file into a PHP value.
+   *
+   * @param string $file
+   *   The filename.
+   * @param string $module
+   *   The module where the Yaml file is placed.
+   * @param string $profile
+   *   The profile used.
+   *
+   * @return mixed
+   *   The YAML converted to a PHP value.
    */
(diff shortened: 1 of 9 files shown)


modules/custom/social_demo/src/DemoContentParser.php
```diff
@@ -22,9 +22,19 @@ public function getPath($file, $module, $profile) {
...
+   *   The filename.
+   * @param string $module
+   *   The module where the Yaml file is placed.
+   * @param string $profile
+   *   The profile used.
+   *
+   * @return mixed
+   *   The YAML converted to a PHP value.
    */
-  public function parseFile($file, $module, $profile) {
```
bramtenhove: @jaapjan I was wondering why this was changed. Is it because of https://www.drupal.org/node/2937545?

jaapjan: Correct; that method was added in 3.4 I guess: https://github.com/symfony/yaml/commit/ea18e8d2f6be875713b4bd3db729d600d7f3727b#diff-6beeaadccc3b2cf24ef915866a3b088f 

And since we already used the same method (with different arguments) I thought renaming was the better option.

PR #824, opened 2018-04-11
SHN-44: Add possibility to turn of the users' topic/event/group tabs on profile

jochemvn:
## Problem
Some platforms don't want to expose the topics / events / groups people interact with on their profile. Those views and blocks had a permissions 'access user profiles' which made it difficult to remove them on a permission basis.

## Solution
For each tab on a profile (topics/events/groups) there are now TWO extra permissions:
- view <entity-type> on my profile
- view <entity-type> on other profiles

By default in an update hook the value are set based on what the 'access user profiles' permission is set to. On install LU+ get the permission, so no change in behaviour is done on existing or new platforms.

With these permissions it's now very easy to turn off these tabs on other people's profile and allow access on your own, or to turn it off altogether. 

## Issue tracker
https://jira.goalgorilla.com/browse/SHN-44

## HTT
- [x] Login as an admin and check the permissions 
- [x] Run the update script and notice the permissions are now set
- [x] Experiment with the settings and check (as LU/CM/SM) if everything works

Note that the effect should be visible like so (topic is example here):
1. No 'RECENTLY CREATED TOPICS' block visible on `/user/<userid>stream`
2. No link to 'TOPICS' on `/user/<userid>stream`
3. Access denied when trying to access `/user/<userid>topics`
4. No filter block visible on `/user/<userid>topics` (this does not apply to groups)

## Documentation
- [ ] This item is should be added to the release notes
- [ ] This item should be added to the feature overview


jaapjan:
Marking as needs work since the coding standards are not correct at the moment:
https://travis-ci.org/goalgorilla/open_social/jobs/372922536

jochemvn:
Fixed coding standards 🙈

modules/social_features/social_event/social_event.module
```diff
@@ -307,3 +312,28 @@ function _social_event_format_date($event, $view_mode) {
...
+/**
+ * Custom permission check, to see if people have access to users' events.
+ *
+ * Implements hook_block_access().
+ */
+function social_event_block_access(Block $block, $operation, AccountInterface $account) {
+
+  if ($operation === 'view' && ($block->getPluginId() === 'views_exposed_filter_block:events-events_overview' || $block->getPluginId() === 'views_block:events-block_events_on_profile')) {
+    // Here we're going to assume by default access is not granted.
+    $topicController = SocialEventController::create(\Drupal::getContainer());
```
ribel: @jochemvn Minor thing, but i think this variable should be named `$eventController` ?

modules/social_features/social_event/src/Controller/SocialEventController.php
```diff
@@ -11,6 +16,32 @@
...
 
+  /**
+   * The request.
+   *
+   * @var \Symfony\Component\HttpFoundation\RequestStack
+   */
+  protected $requestStack;
+
+  /**
+   * SocialTopicController constructor.
```
ribel: @jochemvn The comment should be renamed.

modules/social_features/social_event/src/Controller/SocialEventController.php
```diff
@@ -22,4 +53,38 @@ public function redirectMyEvents() {
     ]);
   }
 
+  /**
+   * Function that checks access on the my topic pages.
```
ribel: @jochemvn topic again :P

modules/social_features/social_group/social_group.module
```diff
@@ -1004,6 +1006,20 @@ function social_group_block_access(Block $block, $operation, AccountInterface $a
       }
     }
   }
+  // Check if we're alloweed to view the joined groups block.
+  if ($operation === 'view' && $block->getPluginId() === 'views_block:groups-block_user_groups') {
+    // Here we're going to assume by default access is not granted.
+    $topicController = SocialGroupController::create(\Drupal::getContainer());
+    $access = $topicController->myGroupAccess($account);
```
ribel: @jochemvn topics and events here should be groups.

modules/social_features/social_group/src/Controller/SocialGroupController.php
```diff
@@ -2,13 +2,45 @@
...
 
+  /**
+   * The request.
+   *
+   * @var \Symfony\Component\HttpFoundation\RequestStack
+   */
+  protected $requestStack;
+
+  /**
+   * SocialTopicController constructor.
```
ribel: @jochemvn Please fix the comment.

ribel:
I have made a review and add few comments to fix variable names and comments in the code.
But on the functionality level, it works like a charm! I think `needs work` label can be removed and it can be merged after these small code fixes.

jochemvn:
Thanks for the review @ribel! I've fixed all issues you reported and I removed the 'needs work' label

jochemvn:
@bramtenhove done!

bramtenhove:
Please review,  but don't merge it yet. We need to add a new base branch first.

jaapjan:
Project Moss is not liking this; lot's of conflicts with applied patches in projects.

jochemvn:
@jaapjan I see 1 patch failing, which is the patch of this PR right?

`    patches/824-my-profile-content-tabs.diff (Remove MY tabs)
   Could not apply patch! Skipping. The error was: Cannot apply patch patches/824-my-profile-content-tabs.diff`

jaapjan:
Yeah, that one is easy to solve indeed. But there is also . the following patch in some projects:

They appear to conflict with each other.

```
  Cannot apply patch 760 - Add separate route for group stream pages (https://patch-diff.githubusercontent.com/raw/goalgorilla/open_social/pull/760.diff)!  
```

jochemvn:
hmm 🤔 Don't think we can solve that right... we need to merge either one of the PRs and make the unmerged one up-to-date again

ribel:
I think #760 can be revived and merged. I see one failing behat tests, but if you check it's not related to introduced changes. So need to check it locally.

modules/social_features/social_event/social_event.install
```diff
@@ -139,3 +141,22 @@ function social_event_update_8002() {
   // Give SM the administer social_event settings permission.
   user_role_grant_permissions('sitemanager', ['administer social_event settings']);
 }
+
+/**
+ * Set my events permissions.
+ */
+function social_event_update_8003() {
```
Kingdutch: Super! 🎉

bramtenhove:
@jochemvn, tests are still running however one project is failing on the `Add a group` step:
```And I click "Groups"
Then I should see "Add a group"```

Could you look into that?

bramtenhove:
@jochemvn and I updated the custom tests in that project to make sure they click on the "Groups" tab instead of the first "Groups" link (which could very well be the group overview now that it is available.

Thanks @jochemvn!

jochemvn:
@bramtenhove We might need to consider to do this more often.. As in: Be explicit in which region you want to click on a certain element...

bramtenhove:
We should actually be doing this in all places. Apparently we cannot be sure in some cases if we're testing what we want to test...

PR #829, opened 2018-04-13
DS-5263 big pipe menu shift

Crazylord:
## Problem
When enable big pipe the menu shifts to the left

## Solution
Apply path for big pipe module

## Issue tracker
- https://www.drupal.org/project/social/issues/2928073

## HTT
- [x] Checkout to this branch
- [ ] Enable big pipe module
- [ ] Check if user menu position is correct

## Documentation


jaapjan:
Setting to "Needs work" because we also need to apply the patch for the .make file.

PR #832, opened 2018-04-16
Multilingual content translation

nielsvandermolen:
## Problem
Provide content translation capabilities for Open Social

## Solution
Provided configuration and a custom language switcher to work with our custom language implementation that changes all URLs to the current langauge.

## Issue tracker
http://jira.goalgorilla.com/browse/OSSUPPORT-530

## HTT
- [ ] Checkout this branch, enable the module and test the scenario below
- [ ] Make sure you are able to edit translations (OSSUPPORT-584)

Scenarios Creating new content and translate:

- [ ] As a user with the English language in the profile go to the website. See that the website is in English and that the language switcher has English enabled.
- [ ] Create a new book page and type some text
- [ ] Translate the page in Spanish
- [ ] Go back to the page (should be in english)
- [ ] Try to change the English translation
- [ ] Try to change the Spanish translation
- [ ] Play around with URL alias translations

## Documentation


modules/custom/social_language/social_language.install
```diff
@@ -43,18 +44,85 @@ function _social_language_set_permissions() {
...
+    'translate profile_tag taxonomy_term',
+    'translate topic_types taxonomy_term',
+  ]);
+}
+
+/**
+ * Give translate permissions to content managers.
+ */
+function social_language_update_8002() {
+  _social_language_set_permissions();
```
Kingdutch: Update hooks should explicitly give the permissions that they introduced based on the current permission set of a platform. e.g. if you introduce permission B and C as a replacement for A then B and C should be explicitly granted to roles that currently have permission A.

Calling `_social_language_set_permissions` here could reassign Open Social default permissions that may have been intentionally revoked on a platform by a maintainer or developer.

modules/custom/social_language/social_language.services.yml
```diff
@@ -12,3 +12,8 @@ services:
     class: Drupal\social_language\Routing\RouteSubscriber
     tags:
       - { name: event_subscriber }
+  drupal8_url_generator:
```
Kingdutch: Rename to `drupal_core_url_generator` so that this code will not confuse is if we update to Drupal 9 in 1.5 years time?

modules/social_features/social_event/config/install/core.entity_view_display.node.event.activity.yml
```diff
@@ -90,5 +90,6 @@ hidden:
   field_files: true
   like_and_dislike: true
   groups: true
+  groups_type_public_group: true
```
Kingdutch: This seems like a change that doesn't belong in this PR?

Same for all files where this line was added.

modules/social_features/social_event/config/install/field.storage.node.field_event_image.yml
```diff
@@ -23,7 +23,7 @@ settings:
 module: image
 locked: false
 cardinality: 1
-translatable: true
+translatable: false
```
Kingdutch: Images were made translatable on purpose. An Image could contain text so it may be desirable to upload a different image in one language than in another.

modules/social_features/social_group/config/install/field.storage.group.field_group_image.yml
```diff
@@ -23,7 +23,7 @@ settings:
 module: image
 locked: false
 cardinality: 1
-translatable: true
+translatable: false
```
Kingdutch: Images were made translatable on purpose. An Image could contain text so it may be desirable to upload a different image in one language than in another.

modules/social_features/social_topic/config/install/field.storage.node.field_topic_image.yml
```diff
@@ -23,7 +23,7 @@ settings:
 module: image
 locked: false
 cardinality: 1
-translatable: true
+translatable: false
```
Kingdutch: Images were made translatable on purpose. An Image could contain text so it may be desirable to upload a different image in one language than in another.

modules/social_features/social_topic/social_topic.module
```diff
@@ -34,11 +34,16 @@ function social_topic_preprocess_node(array &$variables) {
     $topic_type = $node->get('field_topic_type');
     $topic_type_entities = $topic_type->referencedEntities();
     if (count($topic_type_entities) === 1) {
-      foreach ($topic_type_entities as $topic) {
-        $variables['metadata'] = t('in @topic', ['@topic' => $topic->link()]);
+      $lang_id = \Drupal::languageManager()->getCurrentLanguage()->getId();
+      /** @var Drupal\taxonomy\Entity\Term $topic_type_entity */
+      foreach ($topic_type_entities as $topic_type_entity) {
+        if ($topic_type_entity->hasTranslation($lang_id)) {
```
Kingdutch: Is this really needed? My understanding was that Drupal automatically uses the current language's translation if it's available and uses the default if it's not.

If we need to do this for every entity/content type then this can create a lot of custom code when creating new entities/content types.

Kingdutch:
This PR has the "Needs Work" tag. It would be a good idea to describe what is the work that still needs to be done : )

modules/custom/social_language/social_language.install
```diff
@@ -124,5 +124,12 @@ YAML;
  * Give translate permissions to content managers.
  */
 function social_language_update_8002() {
-  _social_language_set_permissions();
+  // Check if the permissions for site managers have been restricted.
+  if ($role = Role::load('sitemanager')) {
+    if ($role->hasPermission('create content translations')) {
+      $permissions = _social_language_get_permissions('contentmanager');
```
Kingdutch: We provide contentmanager permissions here based on settings for a sitemanager?

NielsGG: Yes, this permission was added in the installation for site managers. If they removed the permission they probably do not want them added for content managers.

modules/social_features/social_event/config/install/core.entity_view_display.node.event.activity.yml
NielsGG: Should be in the public group PR. I do not see a problem by fixing a few small issues here and there without making separate tickets & PRs for everything. But feel free to separate it, the change should be self evident.

modules/social_features/social_topic/social_topic.module
NielsGG: Yes, it only applies for our custom taxonomy terms that have these custom preprocess functions (topic and event types). Bug report: OSSUPPORT-588

NielsGG:
TODO:
- Finish image translations
- Test email and account translations (mainly based on 9f8f60662a833bb28380a299415816208430dd60)
- Fix the Travis build
- Write documentation for LGOS

bramtenhove:
@Kingdutch, I assigned this item to you. It still requires some work, but I can help reviewing it.

Kingdutch:
Travis is green except for some coding standards which is related to the TODO item "Fix image translations".

I have a bit of a gripe with the current way image translations are implemented (and I also don't quite understand why it was done this way):

`_social_language_set_image_configuration_third_party_settings` has a hard dependency on some of the Open Social modules. It’s also not flexible enough to help us with other image fields that may have been added by Open Social.

bramtenhove:
@Kingdutch Thanks for the summary.

This is a pretty cool feature and I think it would be nice to have it in the distro.
But I personally don't feel comfortable going down the road of always having both interface and content translation enabled.

Next to that having a broad permission like this seems too much, we should be able to create our own permissions if necessary right?

It should simply be an optional module based on the `social_language` module. It was my understanding, and this is what we discussed some time ago as well, that this was already the case.

@jaapjan, @ronaldtebrake I'm curious to also know your opinions.

@Kingdutch what would you see as the way forward? Is it possible to _easily_ split this into a separate module?

Kingdutch:
@bramtenhove I think a possible way forward is to go back to commit ed567ae512e249aedcdc59170c1793331b5580af which basically did what you said.

It introduces a social_content_translation module as a submodule of the social_language module. It then applies default settings for translation configuration for supported entities if the setting to enable translation for that entity type (this is new and separate configuration) is set. The application is done using configuration override classes so it can be turned on and off.

This provides us with the following flexibility:
- Advanced sitebuilders can ignore the social_content_translation module and just use Drupal's content translation module directly and configure that as needed.
- Novice sitebuilders can use the social_content_translation module and this will provide them with sane defaults for the fields in Open Social's entity types.
- For site managers a UI can be added (not sure if this is currently included) so that they can toggle translatability for the Open Social entity types.
- Projects that extend Open Social entity types can extend the configuration override classes to add translatability settings for their own fields (alternatively we may need to add an alter_hook implementation here to allow multiple modules to hook in at the same time). Those projects then do not need to worry about whether it's enabled or not.

When moving back to this point we will have to re-introduce some fixes that may have been created after the fact (such as the language switcher and the issue with URLs and langcodes in the BubblingUrlGenerator).

We may need to evaluate whether the permissions for the social_content_translation module are granular enough to accomodate the scenario where only site managers can translate (and whether this is configurable (or should be configurable) for sitemanagers).

ronaldtebrake:
Hey guys, great to see activity in here! Distro users will definitely love the ability of easily enabling this. But I fully agree it should simply be an optional module based on the social_language module. 

Regarding the outlined flexibility in your last comment @Kingdutch that seems like a perfect fit to me.

bramtenhove:
Travis is failing on coding standards @Kingdutch

ronaldtebrake:
#1077 Is the replacement. Thanks @Kingdutch.

PR #834, opened 2018-04-16
DS-5269 by serhiy-chornobay: Fix position for preview container for LU

Crazylord:
## Problem
When viewing a preview of topic or event, the 'go back to content' bar is behind the menu

## Solution
Fix position for preview container for LU

## Issue tracker
- https://www.drupal.org/project/social/issues/2914874

## HTT
- [x] Checkout to this branch
- [x] Login as user member and Create/edit a topic or an event and press ' preview'.
- [x] Check if preview box is visible

## Documentation


jaapjan:
@Crazylord I'm wondering why all these Javascript files are changed. Are these supposed to be in this commit?

Crazylord:
@jaapjan It's only generated new files by gulp, maybe was updated some library for node modules. There no changes for any JS. This doesn't break any scripts.

PR #841, opened 2018-04-18
DS-5171 As a SM I want to keep Revisions for Data Policy

OH17L:
## Problem

1. SM Should be able to fill in data policy and translate data policy
2. SM need to be able to make new revisions of the data policy they can create
3. SM can also make a (new) revision active
4. Revisions are set in stone and can never be changed in the Interface
5. There is a list of revisions of the data policy which the SM can look into

## Solution
Create a module for integrating [GDPR Inform and Consent](https://github.com/goalgorilla/gdpr_consent) (_gdpr_consent_) module.

## Issue tracker
- DS-5171

## HTT
- [x] Enable **Social GDPR** (_social_gdpr_) module
- [x] When you go to **Sign up** page then you should see checkbox **I agree with the data policy**
- [x] When you click **data policy** link in a label of checkbox then you should see a pop-up window with data policy text
- [x] You can not create new user while you do not mark data policy checkbox
- [x] You should see **Data policy** link in the footer and when you click this link then you should see **Data policy** page with hero banner
- [x] When you log in as administrator then you should see **Data policy** item in an administrative menu in **Configuration / People**
- [x] You should see **Add new revision** button and revisions table on **Data policy** page
- [x] Checkbox **Create new revision** should be marked automatically and disabled on  **Add new revision** form
- [x] You should see following operations for revision on **Data policy** page: **View**, **Revert** (not for current revision), **Delete** (not for current revision)
- [x] When you create new user right now and open **View** page of current revision then you should see your new user in a table in **USER CONSENTS FOR CURRENT REVISION** section
- [x] When you log in as SM and click *Data policy** link in the footer then you should see data policy text page with two tabs - **Details** and **Revisions**
- [x] **Revisions** tab should have the same behavior as **Configuration / People / Data policy** page for the administrator but should have access to **Delete** operation of revisions


bramtenhove:
@OH17L is the `dont-merge` label still applicable?

OH17L:
@bramtenhove, need to fix theming before merge.

bramtenhove:
As discussed there are a few changes still required, but as agreed these will be picked up in a different story.

Notable changes:
- Not generating a default data policy
- Making sure new revisions are not turned on by default
- Revisions that have never been active should still be editable

PR #843, opened 2018-04-23
Issue #2962743 : Make group selection optional when creating content

jaapjan:
## Problem
When a LU is creating a new content item (topic/event) it is now required to select a group.

## Solution
Made the code less complex and added appropriate checks for if the `$option_category` is an array or not. Also ensures that the `_none` option is never removed from the `$options` array.

## Issue tracker
- https://www.drupal.org/project/social/issues/2962743

## HTT
- [x] Login as LU and CM check all these scenarios on /node/add and node/edit.
- [x] Platform with multiple group types and user is member of multiple group types
- [x] Platform with multiple group types and user is NOT member of multiple group types
- [ ] Platform with only one group types and user is member of multiple groups and one group
- [x] Platform with one or multiple group types and user is not a member of any group

## Documentation


jochemvn:
@jaapjan, Probably not for today, but maybe you can check if the user has permission in the 'group type', to create the current 'content type'. That way, this also works for other (non opensocial) groups types that have alternative permission settings.

jaapjan:
@jochemvn thanks for the feedback! Makes sense to include it. Checkout commit d7d4a8e :-)

bramtenhove:
Agreed @jaapjan, I will add it.

PR #849, opened 2018-04-24
DS-5173 Give SM option to Enforce consent and send user to cancel if they dont wanna

ghost:
## Problem

1. For Open Social we need to make sure that consent is enforced and that it is required. This needs to be a setting.
2. When this setting is enabled and people want to withdrawal consent they are given notice that the only option is to delete an account or keep on agreeing and fill in fewer data.
3. Needs to be clear that you can withdraw in the description of the consent checkbox field
4. When SM moves from "optional" to "enforce consent" we have to make sure the user can not use the platform before giving consent to data policy.

## Solution

1. Added checkbox for changing requirement of user consent on data policy settings page.
2. Created separated page for agreement and implemented redirecting to this page when need consent from the current user.

## Issue tracker
- DS-5173

## HTT
- [x] Enable **Social GDPR** (_social_gdpr_) module and log in as SM
- [x] When you go to the **Data policy settings** page (_/admin/structure/data_policy/settings_) then you should see choose checkbox **Enforce consent**
- [x] When you log in as X then you should be redirected to the **Data policy agreement** page
- [x] When you try to open home page without consent then you should be redirected to the **Data policy agreement** page again
- [x] You should see the **Data policy agreement** item in the user menu
- [x] When you set the **I agree with data policy** checkbox and press **Save** then you should be redirected to the homepage
- [x] You should not see the **Data policy agreement** item in the user menu
- [x] Log in as SM and go to the **Data policy settings** page
- [x] Uncheck **Enforce consent** and press **Save configuration**
- [x] When you log out and go to the **Sign up** page then you should see that the **I agree with data policy** checkbox is not required
- [x] When you log in as Y then should not be redirected to the **Data policy agreement** page and you should not see the **Data policy agreement** item in the user menu


bramtenhove (approved):

bramtenhove (approved):
I'd like to see if we can move the GDPR Consent settings page (just with enforce consent) to /admin/config/people/data-policy. Makes more sense imho.
What do you think @OH17L?

OH17L:
@bramtenhove, do you mean to add **Enforce content** checkbox on revisions overview page?

OH17L:
@bramtenhove, I moved settings page in [PR](https://github.com/goalgorilla/gdpr_consent/pull/8). I think you can merge this PR.

bramtenhove:
@OH17L, as discussed we're merging this PR.

There are still some pending changes, but these will be added at a later moment.

PR #854, opened 2018-04-26
2949116: Filter out blocked users from the search

jochemvn:
## Problem
Blocked users show up in the search and when you click on them you get an access denied page.

## Solution
A plugin that does 2 things:
1. Make sure blocked user are no longer indexed
2. Filter out blocked users on query level. 

## Issue tracker
- https://www.drupal.org/node/2949116
- https://jira.goalgorilla.com/browse/ECI-735

## HTT
- [ ] Notice it doesn't work
- [ ] Checkout to this branch

## Documentation

### DIFF
--- a/modules/social_features/social_search/config/install/search_api.index.social_all.yml
+++ b/modules/social_features/social_search/config/install/search_api.index.social_all.yml
@@ -20,7 +20,7 @@ dependencies:
     - profile
 id: social_all
 name: 'Social all'
-description: ''
+description: 'All content index created for the Social distribution.'
 read_only: false
 field_settings:
   rendered_item:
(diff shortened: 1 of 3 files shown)


ribel:
I have tested it and it works correctly now!

jochemvn:
@ribel, could you merge it, if you think it's ok? Shipsters are ok with us merging our own bugfixes

jaapjan:
Indeed, as long as the milestone and base branch is correct.

Bugfixes:
current stable, which now is 1.17 and 8.x-1.x branch

Feature requests/small improvements:
next stable, which now is 2.0 and 8.x-2.x branch

After merging you can mark the issue on d.o as "Fixed" as well.

PR #857, opened 2018-04-27
Issue #2966841: Make time optional for events

ribel:
## Problem
Currently, time field is required for events, but there are events that have no specific starting time.

## Solution

- When adding a date/time - the time should not be required for events
- If no time is set event starts at 0:01 that day
- If no time is set the time should not be displayed

## Issue tracker
- https://www.drupal.org/project/social/issues/2966841
- https://jira.goalgorilla.com/browse/CBI-357

## HTT
- [ ] Login as normal user and try to create event without time
- [ ] Notice you receive validation error about incorrect date format
- [ ] Checkout to this branch
- [ ] Try the same and check that there are no validation error
- [ ] Check event full page and teaser and see that time is not displayed
- [ ] Edit time to other value from 00:01 and check it displayed now

## Documentation

### DIFF
--- a/modules/social_features/social_event/social_event.module
+++ b/modules/social_features/social_event/social_event.module
@@ -14,6 +14,8 @@ use Drupal\group\Entity\GroupContent;
 use Drupal\group\GroupMembershipLoaderInterface;
 use Drupal\views\Plugin\views\query\QueryPluginBase;
 use Drupal\views\ViewExecutable;
+use Drupal\Component\Utility\NestedArray;
+use Drupal\Core\Datetime\DrupalDateTime;
 
 /**
  * Implements hook_form_form_ID_alter().
@@ -232,15 +234,6 @@ function social_event_flagging_insert(EntityInterface $entity) {
 }
 
 /**
- * Implements hook_ENTITY_TYPE_delete().
- */
-function social_event_user_delete(EntityInterface $entity) {
-  $storage = \Drupal::entityTypeManager()->getStorage('event_enrollment');
-  $entities = $storage->loadByProperties(['user_id' => $entity->id()]);
-  $storage->delete($entities);
-}
-
-/**
  * Formats the event start end date.
  */
 function _social_event_format_date($event, $view_mode) {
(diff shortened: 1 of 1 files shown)


jochemvn:
Hey @ribel ,

I've been testing this a bit and the happy flow works fine. However when you start to test with users that have different timezones it doesn't work anymore. EG: Setup that robertandrews has the London timezone and chrishall has the Amsterdam timezone. Then as robertandrews make an event from 2 may to 5 may and leave the time empty. On node view it looks fine. Now login as chrishall and go to the event. The event date is now: "1 May '18 23:01 - 4 May '18 23:01"

If you do it the other way around you get the problem that robertandrews sees: "1 May '18 01:01 - 5 May '18 01:01". To really solve this I think we need to entirely take out time of the equation and work with someting like a 'all day event' checkbox.

Jochem

jaapjan:
@ribel cc @jochemvn when do you guys plan to work on this?

ribel:
@jaapjan I am going to work on this today. We defined following criteria:

- Should be a checkbox to mark event as all day
- Time should disappear when checked
- Time should not be displayed when marked as all day

I assume this now going to be in milestone 3, right?

jochemvn:
Any particular reason why you created a checkbox for both start and end date? Looking at google calendar or office 360 calendar, they use 1 checkbox on top to immediately disable both time fields. Something to consider?

ribel:
@jochemvn yes, because in this way it's more flexible for some extra cases, where for example you want to specify only start or end time, but not both. Also, this solution is easier in implementation. 
Here are examples how it will look:

- https://prntscr.com/k6xg3y
- http://prntscr.com/k6xk3f

xinyumaGG:
@ribel  and @jochemvn, 

Personally I think 'All day' conflicts with the possibility of having a starting date/end date. So I think we should have the all day field above the start and end time, and once it is checked, we can hide both the starting time and end time.

ribel:
@xinyumaGG I have changed it to single 'All day' checkbox. Not sure if it's easy to put it above time input (need to ask @Crazylord), but currently it looks like this: 
- Checked: http://prntscr.com/k9zzr5
- Unchecked: http://prntscr.com/k9zzvb
What do you think? // cc @jochemvn

ribel:
@xinyumaGG By suggestion from @Crazylord I've put checkbox above the fields.

- Checked: https://prnt.sc/ka8nfw
- Unchecked: https://prnt.sc/ka8nky

xinyumaGG:
That looks awesome! Thanks a lot @ribel and @Crazylord

jaapjan:
Thanks Taras, demo looked nice. Let's try and get this in 3.0!

modules/social_features/social_event/social_event.module
```diff
@@ -282,28 +284,108 @@ function _social_event_format_date($event, $view_mode) {
...
+      // Default time should not be displayed.
+      $end_time = $date_formatter->format($end_datetime, $time_format, '', 'UTC') === '00:01' ? '' : $date_formatter->format($end_datetime, $time_format);
+    }
+
     // Date are the same or there are no end date.
     if (empty($end_datetime) || $start_datetime == $end_datetime) {
-      $date = $date_formatter->format($start_datetime, $date_format);
-      $time = $date_formatter->format($start_datetime, $time_format);
-      $event_date = "$date - $time";
+      $event_date = empty($start_time) ? $start_date : "$start_date - $start_time";
```
Kingdutch: If we want to be consistent with the other cases where a dash (`-`) is used to denote "from - to" then we should remove the dash here.

modules/social_features/social_event/social_event.module
```diff
@@ -282,28 +284,108 @@ function _social_event_format_date($event, $view_mode) {
...
+ */
+function social_event_date_validate(&$element, FormStateInterface $form_state, &$complete_form) {
+  $input = NestedArray::getValue($form_state->getValues(), $element['#parents']);
+  // Skip default validation for required time when date is required.
+  if (!empty($input['date']) && (empty($input['time']))) {
+    $input['time'] = '00:01:00';
+    $format = 'Y-m-d H:i:s';
+    $datetime = trim($input['date'] . ' ' . $input['time']);
+    $timezone = 'UTC';
+    $input['object'] = DrupalDateTime::createFromFormat($format, $datetime, $timezone);
```
Kingdutch: I don't feel too comfortable by the fact that we're forcing UTC here. The `_social_event_format_date` function also starts talking about the fact that we store dates in UTC time format but that is not necessarily true. is it possible to just let Drupal alter the time to 00:01:00 while not changing the timezone?

We have already had hard to debug bugs because we played with timezones before and should tread carefully.

modules/social_features/social_event/social_event.module
```diff
@@ -282,28 +284,108 @@ function _social_event_format_date($event, $view_mode) {
...
+
+/**
+ * Set default time to the date field if time was not set.
+ */
+function social_event_date_validate(&$element, FormStateInterface $form_state, &$complete_form) {
+  $input = NestedArray::getValue($form_state->getValues(), $element['#parents']);
+  // Skip default validation for required time when date is required.
+  if (!empty($input['date']) && (empty($input['time']))) {
+    $input['time'] = '00:01:00';
+    $format = 'Y-m-d H:i:s';
```
robertragas: Not sure, but will this work when a different format is set for a date? I know that one client uses a different format for event dates.

bramtenhove: I'm not sure if this is used for display or calculating. Display should definitely use an existing format.

bramtenhove:
@ribel, @jochemvn, do you think you can process the feedback next week? We'd love to have this in 3.0.

jaapjan:
@ribel, @jochemvn, do you think you can process the feedback next week? We'd love to have this in 3.0.

ribel:
@Kingdutch @robertragas I have fixed all items that you noticed. Could you check it again and approve if you think it's ok?

PR #858, opened 2018-04-27
DS-5310 Send user to cancel if they dont want to agree to the consent (when it is enforced)

ghost:
## Problem

1. When this setting is enabled and people want to withdrawal consent they are given notice that the only option is to delete an account or keep on agreeing and fill in fewer data.
2. Needs to be clear that you can withdraw consent in the description of the consent checkbox field.

## Solution
Create [PR](https://github.com/goalgorilla/gdpr_consent/pull/8) for module.
Change place of data policy settings page.
Add message about cancel account on agreement form.

## Issue tracker
- [DS-5310](https://jira.goalgorilla.com/browse/DS-5310)

## HTT
- [x] Log in as SM.
- [x] You should not see **Data policy** item in the **Manage > Structure** submenu.
- [x] You should see the **Settings** item in **Manage > Configuration > People > Data policy** (__/admin/config/people/data-policy/settings__).
- [x] When you log in as X the you should be on the **Data policy agreement** page and you should see following message **Agreement to the data policy is required for continue using this platform. If you do not agree with the data policy, you will be guided to the account cancellation process.**.
- [x] When you click **the account cancellation** then you should be on the **Cancel account** page.
- [x] You should not have access to following pages: **/user/<USER_ID>/edit**, **/user/<USER_ID>/stream**.

### DIFF
--- a/modules/custom/social_gdpr/social_gdpr.module
+++ b/modules/custom/social_gdpr/social_gdpr.module
@@ -43,12 +43,17 @@ function social_gdpr_form_gdpr_consent_data_policy_agreement_alter(&$form, FormS
     ],
   ];
 
-  foreach (['data_policy', 'not_agree'] as $key) {
+  $keys = ['data_policy'];
+
+  if (isset($form['not_agree'])) {
+    $form['not_agree']['#tag'] = 'small';
+    $keys[] = 'not_agree';
+  }
+
+  foreach ($keys as $key) {
     $form['fields'][$key] = $form[$key];
     unset($form[$key]);
   }
-
-  $form['fields']['not_agree']['#tag'] = 'small';
 }
 
 /**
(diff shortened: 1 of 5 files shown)


bramtenhove (changes requested):
@OH17L Looks OK overall, I found some minor changes that I think are needed:

1. We should not allow access to /user/*/stream
2. Xinyu's idea was to allow the user to press "Save" with the checkbox unchecked. They will then be redirected to the Cancel page.
3. We should change the text from "You can cancel their account if they do not agree" to "You can cancel your account if you do not agree"
4. The data policy body field should be a CKEditor (as agreed on April 30)

And I'm wondering why it we're allowing the user to go to /user/*/edit, isn't it better to also exclude that?

bramtenhove (approved):

PR #863, opened 2018-05-02
ECI-737 by jochemvn: Fix same as in #855

jochemvn:
## Problem
See #855 

## Solution
See #855 

## Issue tracker
See #855 

## HTT
See #855 
- But now with the social_profile_module enabled

## Documentation
See #855


jochemvn:
I've added the feature request label. Basically because this is a bug fix on an existing feature request.

jaapjan:
Is this tested as well? I think in that case it can be merged?

PR #869, opened 2018-05-04
2965716 User selector for mentions

Crazylord:
## Problem
User selector for mentions no longer goes on top when it doesn't fit on screen

## Solution
Set max height and scroll for user selector mentions

## Issue tracker
- https://www.drupal.org/project/social/issues/2965716

## HTT
- [x] Checkout to this branch
- [x] Login as user and try mention user in textfield

## Documentation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,8 @@
                 "Color module html preview optional": "https://www.drupal.org/files/issues/color-optional-html-preview-2844190-2.patch",
                 "Restrict images to this site blocks image style derivatives": "https://www.drupal.org/files/issues/image_restrict_image_styles-2528214-19.patch",
                 "Fix SQL error when importing translations": "https://www.drupal.org/files/issues/2599228-31.patch",
-                "Optimize getCommentedEntity()": "https://www.drupal.org/files/issues/get-commented-entity-2580551.52.patch"
+                "Optimize getCommentedEntity()": "https://www.drupal.org/files/issues/get-commented-entity-2580551.52.patch",
+                "ajax.js insert command sometimes wraps content in a div": "https://www.drupal.org/files/issues/2018-03-14/736066-296-328.patch"
             },
             "drupal/flag": {
                 "Add relationship to flagged entities when Flagging is base table": "https://www.drupal.org/files/issues/2723703_31.patch"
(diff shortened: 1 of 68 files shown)


jaapjan:
@Crazylord  cc @Kingdutch 

I'm not 100% sure why you choose this solution. Before it always opened the available mentions to the top when there was not sufficient room below the comment/post form element. What you did now is changing the max-height to 114 px? That results in only 3 users to be displayed as available mentions. This solves the issue, but I think it is:

a) Not very clear that the element is scrollable so it might not even be the best solution.
b) Not clear why you choose this solution. Did you look into why the element is not popping to the top anymore instead?

Crazylord:
Hi @jaapjan . How I see this preview worked only forlast elements(textarea) on the page. For now on mobile, when you :focus last element on page, JS add bottom padding for page. This JS fixed problen on iOS, when keyboard are over the textarea and you doesn't see text, what you type.

For me, adding max-height for it - this is good solution, because even if we make mowing mentions popup over textarea, still this be to long for mobile. 

Maybe we need set max-height not 114px, for example 95px, then user will be see - there are more people and he need scroll popup.

Crazylord:
Hi @jaapjan. Next week I will dive into this. I think we can set max-height for mobile and rework script for desktop and make open mention on top if textarea are to close to bottom of window.

jaapjan:
@Crazylord allright, sounds good. And you are right, if we choose 95px or 125px or something as max-height then at least it will be clear to users that they will be able to scroll the mentions. But for desktop let's try and see if we can use most of the space available and show more than just 2 suggestions.

robertragas:
I don't think this will work when a client has a footer (I know because I tested for Distro and Client :D)
You are getting the height of the window but because a client has a huge footer it will dissapear behind the footer. 

So either the code has to be changed where it get's the height from, or the z-index would need to change so it won't fall behind content that has been set in the footer region.

bramtenhove:
@Crazylord can you check @robertragas comment? This may need a bit of tweaking still.

Crazylord:
Hi @bramtenhove, @robertragas .
I made mention's popup opening upward if he are too close to footer

PR #875, opened 2018-05-10
Issue #2971969 by jaapjan: update swiftmailer 1.0-beta1 to 1.0-beta2

jaapjan:
## Problem
Update beta1 to beta2 for swiftmailer.

## Solution
Changed composer.json and the drupal-org.make file. And in addition removed the obsolete code in our massageMessageBody() method. I don't think we should/can remove the entire class since the social_swiftmailer is now used as a mail formatter and sender in some platforms. (See:
https://cgit.drupalcode.org/swiftmailer/tree/src/Plugin/Mail/SwiftMailer.php?id=8.x-1.0-beta1 and 
 https://cgit.drupalcode.org/swiftmailer/tree/src/Plugin/Mail/SwiftMailer.php?id=8.x-1.0-beta2 ) Original PR here: https://github.com/goalgorilla/open_social/pull/447 

## Issue tracker
- https://www.drupal.org/project/social/issues/2971969

## HTT
- [x] Check out the code changes and stay in 8.x-1.x
- [x] Generate demo content and run the cron few times (to get some e-mails in the mailcatcher)
- [x] Do some actions with system mails, e.g. register and reset password
- [x] Checkout to this branch
- [x] Repeat the above actions and compare the mails via the mailcatcher

## Release notes

## Release notes

The swiftmailer module is updated to 1.0-beta2. Also attached a patch regarding an issue which was introduced in beta2. Most notable changes in the swiftmailer is: the key module is now supported as a credential provider and HTML support for core e-mails.


jaapjan:
@jochemvn do you remember why the SocialSwiftMailer class was created? I've now changed the logic here a bit, since there seems to be a lot of improvements in the new swiftmailer module regarding that!
https://github.com/goalgorilla/open_social/pull/875/commits/6ac8b7cde546deef4c083f8bf596986cc635efe8

jochemvn:
@jaapjan it was created for system mails, which are always plain text. They look kind of weird. This code made sure line endings were converted to <br>.

https://www.drupal.org/project/swiftmailer/issues/2831959

PR #879, opened 2018-05-15
2929368: As a LU I don't want to see the "+" icon if I don't have permissions to create content

Crazylord:
## Solution
Hide plus icon from user menu, if LU can't add any content/group. Use JS for this.

## Issue tracker
- https://www.drupal.org/project/social/issues/2929368
## HTT
- [ ] Checkout to this branch
- [ ] Login as user what doesn't have permission to create content
- [ ] Chech if there are "+" icon in user menu

## Documentation

### DIFF
--- a/themes/socialbase/assets/css/base.css
+++ b/themes/socialbase/assets/css/base.css
@@ -131,7 +131,7 @@ optgroup {
 
 html {
   font-size: 16px;
-  -webkit-tap-highlight-color: transparent;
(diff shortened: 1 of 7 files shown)


jaapjan:
@Crazylord what is the reason to use JS for this? I think it should be solved with PHP instead?

Crazylord:
@jaapjan Maybe this can be solved with PHP, but this can be very complicated to foresee all cases and create clear and good conditions for it.

PR #880, opened 2018-05-15
IE11 by serhiy-chornobay: Fix styles for hero in IE11

Crazylord:
## Problem
There is the bug with vertical alignment for text in hero banner.

## Solution
Set height for bannet text only for IE11

## HTT
- [x] Checkout to this branch
- [x] Check hero banner in IE11

## Documentation

## Release notes
..

### DIFF
--- a/themes/socialbase/assets/css/base.css
+++ b/themes/socialbase/assets/css/base.css
@@ -131,7 +131,7 @@ optgroup {
 
 html {
   font-size: 16px;
-  -webkit-tap-highlight-color: transparent;
+  -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
   font-family: sans-serif;
   -ms-text-size-adjust: 100%;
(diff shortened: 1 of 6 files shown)


jaapjan:
@Crazylord could you elaborate a bit more on the problem? Which bug? is there an issue number somewhere? Could you write release notes?

Crazylord:
Hi @jaapjan. I can't find issue for this. But there was bug with vertical text alignment in IE11(text was sticky to top of the banner). Some one created a patch for this, but this patch has caused some errors with gulp, so the changes was declined. I found method what fix vertical alignment for banner text in IE11 and made this in this pull request. 

About release notes - I doesn't know how better to write it, could you help with it?

Crazylord:
@jaapjan 
I don't know, if this realy need add to realese note. this is really to small bug.

Text area of the hero block in IE11 was sticky to the top of the hero while should be vetricall centered.

PR #881, opened 2018-05-15
DS-5048 Deleting users should not break the private message functionality

OH17L:
## Problem
Currently, when a user is deleted, if i had a private message with him/her, when I click private message, it goes to an error page..

It seems that the private message is nor properly deleted after a user is deleted.

## Solution

1. Show **Deleted user** where should be name of deleted user (thread and message titles)
2. Delete message text in database of deleted user
3. Show special text instead of message text of deleted user

## Issue tracker
- DS-5048

## HTT
- [x] Enable **Social Private Message** module
- [x] Run database updates
- [x] Log in as **X**
- [x] Send private message to **Y**
- [x] Log in as **Y**
- [x] Write answer to **X**
- [x] Delete user and use the **Delete your account, delete private messages, and anonymize all other content. Reassign your groups to the super administrator.** account cancel method for that
- [x] Log in as **X**
- [x] When you go to on **/user/inbox** then you should see **Deleted user** instead of **Y** in the thread title
- [x] When you open thread then you should see **Deleted user** instead of **Y** in the second message  title and you should see **This message was deleted together with the user’s account.** instead of second message text
- [x] Create new message to **A** and **B** and **C**
- [x] Delete the **B** user
- [x] Log in as **X**
- [x] When you go to on **/user/inbox** then you should see **A, Deleted user, C** instead of **A, C**
- [x] When you open group thread you should see **X, A, Deleted user, C** in the thread title


bram-rongen:
Just testing the OS dependant checker, sorry for the spam ;)

bram-rongen:
Another spam comment..

bramtenhove:
@OH17L, changes you just made are working, but I'm still looking at a weird group thread theming issue.

screen shot 2018-05-18 at 15 51 31

OH17L:
@bramtenhove, I can not reproduce it. All works fine for me:

ds-5048

PR #883, opened 2018-05-16
DS-5052 The visiblity of activity stream item should remain the same after the actor is deleted

ghost:
## Problem
When a user is deleted, the activity stream item of him or her creating content remains and become anonymous. However, the visibility of this activity stream item becomes public when it should be the community. This needs to be investigated and fixed. 

## Solution
Hide activity of deleted user for an anonymous user when this activity related to a node with community visibility.

## Issue tracker
- DS-5052

## HTT
- [x] Log in as **X**
- [x] Create topic (not in a group) with the **Community** visibility
- [x] You should see new activity **A** on the homepage and on the **Explore Stream** page about creating the new topic by **X**
- [x] Log out
- [x] You should not see **A**
- [x] Delete **X**
- [x] Log in as AU
- [x] You should see **A**
- [x] Switch to this branch and flush all caches
- [x] You should not see **A**


bramtenhove:
@OH17L, I'm not sure but I'm having a hard time getting this fix working.

I followed the HTT steps, I cannot view the topic but I can still see (after clearing caches) the activity stream item on the /explore stream.

screen shot 2018-05-17 at 16 15 21

bramtenhove (approved):

PR #884, opened 2018-05-21
DS-5206 As a LU, I want to lookup users by their username and full name (first + last name) when sending private messages

ghost:
## Problem
Impossible choice recipient of a private message by first name or last name.

## Solution
Use extend autocomplete tagging field with the opportunity to choice recipient user by first name or last name.

## Issue tracker
- DS-5206

## HTT
- [x] Enable the **Social Private Message** module
- [x] Create new account: username - **testuser**, first name - **Firstname**, last name - **Lastname**
- [x] Go to on **/private_message/create**
- [x] When you typing **test** in the **Member(s)** field then you should see **Firstname Lastname** in the results list
- [x] When you typing **first** or **last** in the **Member(s)** field then you should not see **Firstname Lastname** in the results list
- [x] Switch to this branch and flush all caches
- [x] When you typing **test** or **first** or **last** in the **Member(s)** field then you should see **Firstname Lastname** in the results list


bramtenhove:
@OH17L I noticed the Private Message tests are now failing. Could you take a look?

OH17L:
@bramtenhove, sure, I will fix it.

bramtenhove:
@OH17L did you verify if mentions still work? On a fresh install on this branch I'm having trouble getting it to work.

Also, since I merged the other PM story this one now has a merge conflict that needs to be resolved.

OH17L:
@bramtenhove, I will check it all.

bramtenhove (approved):

PR #889, opened 2018-05-24
#2975145 by bramtenhove: Enable Page Cache and Dynamic Page Cache by default

bramtenhove:
## Problem
Page Cache and Dynamic Page Cache modules are not enabled by default for Open Social. Enabling them instantly gives a better site performance.

## Solution
Enable the two modules on install.

## Issue tracker
- https://www.drupal.org/project/social/issues/2975145

## HTT
- [ ] Reinstall the site
- [ ] Notice Page Cache and Dynamic Page Cache modules are turned on

## Documentation


bramtenhove:
We can also provide an update hook to enable (one of) the two modules, but I'm not sure if we should do that.

Any ideas about that?

bramtenhove:
Alright, then we can keep it like this @jaapjan. I marked it for the 2.0 release.

PR #890, opened 2018-05-25
DS-5302 As a blocked/deleted user I should not receive e-mails anymore

ghost:
## Problem
- After blocking user he still receiving e-mails about commenting his post etc.
- Error after blocking a user from users overview page.

## Solution
- Prohibit create a message to the blocked user when some content creating.
- Improve code for user blocking the process.

## Issue tracker
- DS-5302

## HTT
- [x] Log in as **X** and create post **A**
- [x] Log in as **SM** and block user **X**
- [x] Log in as **Y** and create comment to **A**
- [x] Run cron
- [x] User **X** should receive e-mail about commented his post **A**
- [x] Checkout to this branch
- [x] Create new comment to **A** by **Y**
- [x] Run cron
- [x] User **X** should not receive e-mail about commented his post **A**

### DIFF
--- a/modules/custom/activity_basics/activity_basics.module
+++ b/modules/custom/activity_basics/activity_basics.module
@@ -12,22 +12,6 @@ use Drupal\node\NodeInterface;
  * Implements hook_entity_insert().
  */
 function activity_basics_entity_insert(EntityInterface $entity) {
-  switch ($entity->getEntityTypeId()) {
-    case 'comment':
-      $related_entity = $entity->getCommentedEntity();
-      break;
-
-    case 'vote':
-      $related_entity = \Drupal::entityTypeManager()
-        ->getStorage($entity->getVotedEntityType())
-        ->load($entity->getVotedEntityId());
-      break;
-  }
-
-  if (isset($related_entity) && $related_entity->getOwner()->isBlocked()) {
-    return;
-  }
-
   _activity_basics_entity_action($entity, 'create_entitiy_action');
 }
 
--- a/modules/custom/activity_basics/activity_basics.module
+++ b/modules/custom/activity_basics/activity_basics.module
@@ -12,6 +12,22 @@ use Drupal\node\NodeInterface;
  * Implements hook_entity_insert().
  */
 function activity_basics_entity_insert(EntityInterface $entity) {
+  switch ($entity->getEntityTypeId()) {
+    case 'comment':
+      $related_entity = $entity->getCommentedEntity();
+      break;
+
+    case 'vote':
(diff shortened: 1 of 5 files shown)


modules/custom/activity_basics/activity_basics.module
```diff
@@ -12,6 +12,22 @@ use Drupal\node\NodeInterface;
  * Implements hook_entity_insert().
  */
 function activity_basics_entity_insert(EntityInterface $entity) {
+  switch ($entity->getEntityTypeId()) {
```
bramtenhove: @OH17L, I didn't test it yet, but I do have a question.

I like the idea of doing this before we even create an activity, but I do have a few concerns.

You now have two cases (comment and vote). How about private messages, nodes or other entities (that may not yet exist)? And what about digest (daily or weekly) emails. They will collect a bunch of activities and will simply still send that (last) one out.

OH17L: @bramtenhove, you are right. I will extend my functionality.

bramtenhove: @OH17L but I'm wondering if this is maintainable in the long run. It could be easier to just check before sending the email (two places) if the user is active or not.

See https://github.com/goalgorilla/open_social/blob/8.x-2.x/modules/custom/activity_send/modules/activity_send_email/src/Plugin/EmailFrequency/Immediately.php and https://github.com/goalgorilla/open_social/blob/8.x-2.x/modules/custom/activity_send/modules/activity_send_email/src/Plugin/QueueWorker/ActivityDigestWorker.php

OH17L: @bramtenhove, when I delete my code here and add conditions in two files where you propose then a user will receive all e-mails which were sent when this user was blocked. Is it ok?

bramtenhove: @OH17L sorry, I'm not sure I understand. You still get emails or not?

OH17L: @bramtenhove, no, I do not get emails.

bramtenhove: @OH17L ok! 👍

Then I suggest we just implement the change in those two files. Agreed?

OH17L: @bramtenhove, yes, I agree and I will do that.

bramtenhove (approved):

bramtenhove:
@OH17L once the tests are green we can merge this!

PR #895, opened 2018-05-30
SHN-96: Make sure the user entityreference also works with nickname

jochemvn:
## Problem
When enabling the social_profile_fields module, searching on nickname in entity references does not work.

## Solution
The new created entityreference for users is now updated with the nickname field (if it exists)

## Issue tracker
- https://jira.goalgorilla.com/browse/SHN-96

## TODO
This entityreference is now using a selectQuery instead of a entityQuery (which is normal for an entityreference). We need to reconsider this approach. One of the issues is that you cannot change this query with a hook_query_alter for instance. (the query is overriden)

## HTT
- [x] Enable the social_profile_fields module
- [x] Login as user `alisonhendrix` and change username to `vera` and nickname to `Skiire`
- [x] Try to add her to a group, by using all three names mentioned above
- [x] See that it works

## Documentation


jaapjan:
Nice catch Jochem!

PR #896, opened 2018-05-31
DS-5579 Users receive emails of new unpublished content

ghost:
## Problem
Group members receive e-mails about creating unpublished content.

## Solution
- Add a field for remember if a node was published.
- Prohibit sending an e-mail about creating node when the node has never been published yet.

## Issue tracker
- DS-5579

## HTT
- [ ] Log in as **X** and create group **A**
- [ ] Log in as **Y** and join to **A**
- [ ] Log in as **X** and create unpublished topic **B** in **A**
- [ ] When you run cron then **Y** should receive an e-mail about creating **B**
- [ ] Checkout to this branch
- [ ] Update entities (**drush -y entup**)
- [ ] Log in as **X** and create unpublished topic **C** in **A**
- [ ] When you run cron then **Y** should not receive an e-mail about creating **C**
- [ ] Publish **C** and run cron
- [ ] **Y** should receive an e-mail about creating **C**


bramtenhove:
@OH17L, cc @jaapjan, I don't think we should fix this with adding a new field.

Perhaps we should park this solution for now until we have a bit more time to take a look at a full solution.

jaapjan:
Closing the PR for now. We can continue with this later when the story is planned.

PR #897, opened 2018-05-31
Issue #2976549 by Kingdutch, bramtenhove, robertragas: Page title block should not be shown on landing page 404 or 403 pages

Kingdutch:
## Problem
When a landing page is used as a 404 or 403 page then the page title block is shown but this conflicts with landing page's own heroes.

## Solution
Let the page title block check if it's being rendered in a 404 or 403 context that uses a landing page.

## Issue tracker
- https://www.drupal.org/project/social/issues/2976549

## HTT
- [x] Create a landing page
- [x] In basic site settings set the landing page as 404 page
- [x] Go to a non-existent page and see that the page title block is not there
- [x] See that the included test fails before the fix and is green after the fix

## Documentation
Landing pages can now be used to retain users on Page not found pages!

### DIFF
--- a/modules/social_features/social_core/src/Plugin/Block/SocialPageTitleBlock.php
+++ b/modules/social_features/social_core/src/Plugin/Block/SocialPageTitleBlock.php
@@ -2,10 +2,18 @@
 
 namespace Drupal\social_core\Plugin\Block;
 
+use Drupal\Core\Controller\TitleResolverInterface;
+use Drupal\Core\Entity\EntityRepositoryInterface;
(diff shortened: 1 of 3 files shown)


bramtenhove:
@Kingdutch looks like the test fails on "redirect to login on 403". Could you take a look?

https://travis-ci.org/goalgorilla/open_social/jobs/386160435#L1872

Kingdutch:
@bramtenhove @jaapjan I don't know why the landing page test now fails as I didn't change anything that should affect that. Locally I'm unable to run it because adding paragraphs is really finicky due to race conditions between docker and Paragraphs AJAX.

Any ideas?

PR #900, opened 2018-06-01
DS-5542 by chmez: update features 3.5 => 3.7.

ghost:
## Description
Update Features to 3.7.

## HTT
- [ ] Test the functionality, observe it is still working the same

## Documentation


bramtenhove:
@OH17L this is still marked as `dont-merge`, is it correct?

OH17L:
@bramtenhove, you can merge it.

bramtenhove:
@OH17L did you check `drush` commands as well? I'm not getting any output when running them (I cleared Drush cache).

OH17L:
@bramtenhove, yes, I checked it but did not have this bug. I will check it again

OH17L:
@bramtenhove, I could not reproduce your bug:

bramtenhove:
@OH17L I mean can you try to use like `drush fl`? For me `drush cr` was also working but `drush` commands for the Features module did not give any output.

OH17L:
@bramtenhove, we use **drush fl** only [here]( https://github.com/goalgorilla/open_social_scripts/blob/master/check-feature-state.sh#L19). This command with **bundle** argument working correct in updated version of module. I found partical solution [here](https://www.drupal.org/project/features/issues/2949343) for this command without arguments but this patch allow to see only features with **default** bundle so we can not use any patches.

PR #901, opened 2018-06-01
DS-5586 Behat tests for GDPR

ghost:
## Problem
Behat tests for checking the functionality of the **Social GDPR** module are absent.

## Solution
Create Behat tests for that.

## Issue tracker
- DS-5586


composer.json
```diff
@@ -73,7 +74,7 @@
         "composer/installers": "^1.0",
         "oomphinc/composer-installers-extender": "^1.1",
         "drupal-composer/drupal-scaffold": "^2.0.0",
-        "drupal/core": "~8.5.3",
+        "drupal/core": "~8.5.4",
```
bramtenhove: We're using it already. Just updated the number.

bramtenhove:
I ran into the issue that the modal button for actually reverting the data policy is gone with the latest Bootstrap update.

**Before**
before

**After patch is applied**
after

bramtenhove:
I verified if the PWA prompt had any issues, but with or without the patch there was no problem.

screen shot 2018-06-27 at 22 12 13

bramtenhove (approved):

PR #903, opened 2018-06-05
Issue #2974781 by Makishima: Makes the image for featured book pages clickable

VolodymyrMartsynenko:
## Problem
When I create features on a landing page, the image in the teaser for an Event or Topic is clickable, but the image in the teaser for a book page isn't - only clicking on the title takes the user to the book page. I would expect the book page teaser image to be clickable as well.

## Solution
Make the image for featured book pages clickable and leading to the book page. Change image_link value to content in book featured configuration file.

**Note:** also contains https://github.com/goalgorilla/open_social/pull/900 for `2.x`

## Issue tracker
- https://www.drupal.org/project/social/issues/2974781

## HTT
- [x] Open the landing page which linked to the book page in the featured section
- [x] Notice that the book page teaser image isn't clickable
- [x] Checkout to this branch
- [x] Notice that the book page teaser image is clickable and leading to the book page


bramtenhove:
@jaapjan I think we should move this to 2.x. Agreed?

jochemvn:
It's a bug in 1.x right?... SHN has it too

jaapjan:
@bramtenhove why is the features update included in this one? Think we should remove that?

bramtenhove:
@jaapjan it was added in https://github.com/goalgorilla/open_social/pull/900 but not merged back in to `2.x`. Because of the rebase from `1.x` to `2.x` it ended up here. We can also move it into a different PR.

PR #909, opened 2018-06-13
Issue #2979293 by Kingdutch: Change download hover text

Kingdutch:
## Problem
Wether a file opens in the browser or is downloaded depends on the file format and the browser configuration (type, addons, etc) of the user. Currently the text "Open file in a new window" is used which gives a user the expectation of seeing the file in a new window even if this is not necessarily the case.

## Solution
Change the text to "Open or download file". This properly sets the expectation of the user and provides room for the various scenarios in the browser.

## Issue tracker
- https://www.drupal.org/project/social/issues/2979293

## HTT

## Documentation

## Release notes
Currently the text "Open file in a new window" is used for downloading files, which gives a user the expectation of seeing the file in a new window even if this is not necessarily the case. To set proper expectations we changed it to “Open or download file”.


jaapjan:
@bramtenhove Can this one be merged as well?

bramtenhove:
Yep!

PR #915, opened 2018-06-15
2979817 by jochemvn: Make sure the priority is higher than the event type config override

jochemvn:
## Problem
When having the social_event_type module enabled and creating an event from within a group. The group field on the event form is no longer pre-filled with the group you are creating the event in.

## Solution
Turns out that both the group and event type module run a config override on the same form. The group module now has a higher priority in the services file and is being called after the event type one. 

## Issue tracker
- https://www.drupal.org/project/social/issues/2979817

## HTT
- [x] Make sure you have a fresh install in which you immediately enable the `social_event_type` module (**during install, not after**)
- [x] Don't test update, because there the error didn't occur
- [x] Go to a group and from within the group add an event
- [x] Notice that the group field is not pre-filled
- [x] Try the same on this branch
- [x] Notice that the group field is now again pre-filled with the current group

## Documentation

## Release notes
In some cases when you would install the `social_event_type` module during installation. Its config override would override the `social_group` override, resulting in a problem when adding events from within a group. The group field would not be pre-filled anymore with the group you are working in.


jaapjan:
@jochemvn could you update the How to test for this?

jaapjan:
@jochemvn 

Just updated the composer.json for SHNL in the test-distro branch for this PR to remove this patch.

Could you:

- [ ] Update the how to test
- [ ] Write release notes in the description above

jochemvn:
Updated the HTT and the Release notes! 💪

jaapjan:
Solves an issue on JT as well! 🚀

PR #921, opened 2018-06-21
Issue #2980936 by Kingdutch: social_core_path_validate can break an e…

Kingdutch:
…vent node

## Problem
The  social_core_path_validate function can cause an event to be inaccessible.

The function checks for form_validate errors before adding a leading slash to the value of an alias field. However it appears that the address field is validated after the path field which causes the form to be rerendered but a leading slash is now added to an alias.

## Solution
This commit ensures that the social_core_path_validate does not
introduce a forward slash for an empty alias in the form when a
validation error occurs that happens after the alias element is
validated.

## Issue tracker
- https://www.drupal.org/project/social/issues/2980936

## HTT
- [ ] See Drupal.org issue

## Documentation


jaapjan:
Updated branch with latest changes of 8.x-2.x because Embed test on SaaS was failing.

jaapjan:
@bramtenhove did you review this? Are we OK with merging?

bramtenhove:
Yes, we're OK with merging!

Last time I checked was that Moss tests were failing but these are ok now.

PR #923, opened 2018-06-22
#2932250 by bramtenhove: Remove the interval check for items in the daily digest

bramtenhove:
## Problem
The digest selects digest items based on if their timestamp has passed (for example) 24 hours. This is not correct as it should just select anything available.

## Solution
Simply remove the check on timestamp.

## Issue tracker
- https://www.drupal.org/project/social/issues/2932250

## HTT
- [x] Make sure the value of `digest.daily.last_run` is set to the timestamp of now
- [x] Set email notification settings to Daily for user X
- [x] As user Y post on the profile of user X
- [x] Run cron a few time, you did not receive an email
- [x] Change the value of `digest.daily.last_run` to something 24 hours ago (e.g. 1)
- [x] Run cron a few times, user X should've received a digest email

## Documentation


jaapjan:
Works like a charm!

PR #929, opened 2018-06-26
Issue #2840694: update activity stream pager types to mini pagers

jaapjan:
## Problem
Pagers are not consistent across the activity streams.  There are currently 2 different pagers on 5 'stream' pages:

- group stream
- profile stream
- explore stream
- homepage stream
- all notifications (stream)

Homepage and explore have the 1 button 'older items', while the other 3 still have a numbered pager. 

## Solution
I suggest we change the all notifications, group and profile to use the same pager as the 2 other pages.

## Issue tracker
- https://www.drupal.org/node/2840694

## HTT
- [x] Login as user chrishall and create lots of content in one of your groups
- [x] Check the group stream and verify it has the right pager (Older items >>, Newer items <<)
- [x] Check the profile stream and verify it has the right pager
- [x] Check the all notifications `/notifications` stream and verify it has the right pager
- [x] Check the homepage stream and verify it has the right pager
- [x] Check the explore stream and verify it has the right pager
- [x] Verify all the tests are still passing

## Documentation

### DIFF
--- a/modules/social_features/social_activity/config/install/views.view.activity_stream_notifications.yml
+++ b/modules/social_features/social_activity/config/install/views.view.activity_stream_notifications.yml
@@ -251,7 +251,7 @@ display:
           entity_type: activity
           entity_field: created
           plugin_id: date
-      title: 'Notifications'
+      title: Notifications
       header: {  }
       footer: {  }
       empty:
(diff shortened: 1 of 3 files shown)


modules/social_features/social_activity/config/install/views.view.activity_stream_notifications.yml
```diff
@@ -360,17 +360,15 @@ display:
       display_extenders: {  }
       path: notifications
       pager:
-        type: full
+        type: mini
         options:
-          items_per_page: 10
+          items_per_page: 30
```
jaapjan: Changed this to 30 as well. I think that makes more sense considering you already see 10 items in the notification bar.

jaapjan:
Tests were passing, false negative for `landing-page-create.feature` -> to short waiting time.

modules/social_features/social_activity/config/install/views.view.activity_stream_notifications.yml
bramtenhove: Nice addition!

bramtenhove:
The Moss test is successful, but it didn't report back on it. Merging it.

PR #930, opened 2018-06-26
2936738: Make sure the local actions block is not displayed on admin/reports pages

jochemvn:
## Problem
The local actions block is displayed on admin/reports. In some cases (analytics) a sidebar second is being rendered, but only if both complementary area's are empty. Since this block was always rendered, those blocks were never considered empty

## Solution
Make sure the block is not rendered on admin/reports pages

## Issue tracker
- https://www.drupal.org/project/social/issues/2936738
- https://jira.goalgorilla.com/browse/DS-4662
- https://jira.goalgorilla.com/browse/SHN-141

## HTT
- [x] Install the social_kpi_analytics module
- [x] Go to the analytics page
- [x] Notice the layout is correct now (both on update and on fresh install)
- [x] Remove this function from saas: `_social_saas_core_fix_analytics_blocks`

## Documentation

## Release notes
The Primary admin actions block is always shown in the 'Complementary Top' region. Most of the times it's empty and that's the issue. Some pages use the 'Sidebar first' and 'Sidebar second' layout. However when there are blocks shown in the 'Complementary Top', the 'Sidebar Second' region will not show up. We’ve made sure to have great default setting so it works with the Social KPI Analytics module as well.

### DIFF
--- a/modules/social_features/social_core/config/optional/block.block.socialblue_local_actions.yml
+++ b/modules/social_features/social_core/config/optional/block.block.socialblue_local_actions.yml
@@ -1,6 +1,8 @@
 langcode: en
 status: true
 dependencies:
+  module:
+    - system
   theme:
     - socialblue
 id: socialblue_local_actions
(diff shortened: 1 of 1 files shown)


jochemvn:
😢

PR #931, opened 2018-06-26
Issue #2978381: SocialPageTitleBlock::build use of the title_resolver service

jaapjan:
## Problem
The else clause in SocialPageTitleBlock::build uses the title resolver service to get the page title. Is there a reason to prefer this resolver vs using $this->title from the parent PageTitleBlock? The service won't have information on eg dynamic views titles, whereas $this->title will contain the dynamic title.

## Solution
Use the original title from `\Drupal\Core\Block\Plugin\Block\PageTitleBlock::$title`.

## Issue tracker
- https://www.drupal.org/project/social/issues/2978381

## HTT
- [ ] Login as user X and open a few pages
- [ ] Check titles of overviews, nodes, profiles, posts, forms, etc.

## Documentation

### DIFF
--- a/modules/social_features/social_core/src/Plugin/Block/SocialPageTitleBlock.php
+++ b/modules/social_features/social_core/src/Plugin/Block/SocialPageTitleBlock.php
@@ -54,7 +54,7 @@ class SocialPageTitleBlock extends PageTitleBlock implements ContainerFactoryPlu
   protected $titleResolver;
 
   /**
-   * EventAddBlock constructor.
+   * SocialPageTitleBlock constructor.
    *
    * @param array $configuration
    *   The given configuration.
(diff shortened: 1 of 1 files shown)


jaapjan:
Actually this PR does not seem to work..

The code does not work in combination with:
`\Drupal\social_topic\Routing\RouteSubscriber::alterRoutes` which sets the _title_callback for `view.latest_topics.page_latest_topics` via this controller: `\Drupal\social_topic\Controller\SocialTopicController::latestTopicsPageTitle`. Not sure what to do here and if there is an easy fix to make the View title dynamic in another way.

jaapjan:
Closing PR for now awaiting answer in the issue queue.

PR #938, opened 2018-07-05
Issue #2978324 by robertragas: fix issue where unread private message…

robertragas:
… are not shown after max-count

## Problem
Users have a inbox for private messages. There is a count of 30 messages being displayed based on the thread updated time. This has the wrong sort order and causes that when you have over 30 message that latest unread message will not be shown.

## Solution
Reverse the order in the query of the private messages

## Issue tracker
- https://www.drupal.org/project/private_message/issues/2978324

## HTT
- [ ] Login as a user that has over 30 private messages and send one from another account so you will have 1 unread message
- [ ] Notice that the notification centre says you have 1 unread message but you don't see it in your inbox
- [ ] Checkout to this branch
- [ ] Retry and see it now shows

## Documentation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,8 @@
                 "Fix preview on node": "https://www.drupal.org/files/issues/2848080-2-preview-fails-on-node.patch"
             },
             "drupal/private_message": {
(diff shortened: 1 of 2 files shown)


jaapjan:
https://www.drupal.org/project/social/issues/2978017 this is the issue number we should use..

PR #944, opened 2018-07-06
2965655 by jochemvn: Make the group widget check for installed plugins.

jochemvn:
## Problem
Content types with the group field (mostly accidental, or upgrades), didn't check if the content type was installed in the group. When trying to add such a content type in a group you'd get a WSOD 

## Solution
Make sure the group widget only shows group types that are have the current content type installed

## Issue tracker
- https://www.drupal.org/project/social/issues/2965655

## HTT
- [x] Login as an admin and add the group field to the page content type (through the field_ui)
- [x] Create a new page and notice you can place it in a couple of groups
- [x] Select a group and submit
- [x] WSOD (Pluginnotfoundexception)
- [x] Try it on this branch.
- [x] Notice that the group selector is empty
- [x] Submit the page
- [x] Notice you now have a page
- [x] Try to create a page and node
- [x] Notice that you can still place both in all group types

## Documentation

## Release notes
The group field would in some cases appear on node add/edit screens of content types that are not installed in the group type. Would you then select a group and submit, it would crash because of a `Pluginnotfoundexception`. 
This fix makes sure the widget checks if the current content type is installed in any of the group types. If not, the group type and all of its groups are removed from the widget.

### DIFF
--- a/modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
+++ b/modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
@@ -84,11 +84,15 @@ class SocialGroupSelectorWidget extends OptionsSelectWidget implements Container
   protected function getOptions(FieldableEntityInterface $entity) {
     if (!isset($this->options)) {
 
-      // Get the bundle fron the node.
-      $entity_type = '';
-      if ($entity->getEntityTypeId() == 'node') {
-        $entity_type = $entity->bundle();
+      // Must be a node.
+      if ($entity->getEntityTypeId() !== 'node') {
+        // We only handle nodes. When using this widget on other content types,
+        // we simply return the normal options.
+        return parent::getOptions($entity);
       }
+
+      // Get the bundle fron the node.
+      $entity_type = $entity->bundle();
 
       $account = $entity->getOwner();
       // Limit the settable options for the current user account.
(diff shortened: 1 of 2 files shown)


modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
```diff
@@ -79,13 +83,43 @@ public static function create(ContainerInterface $container, array $configuratio
...
+        if (is_array($optgroup)) {
+          // Loop through the groups.
+          foreach ($optgroup as $gid => $title) {
+            // If the group exists.
+            if ($group = Group::load($gid)) {
+              // Load all installed plugins for this group type.
+              $plugin_ids = $this->pluginManager->getInstalledIds($group->getGroupType());
+              // If the bundle is not installed,
+              // then unset the entire optiongroup (=group type).
+              if (!in_array('group_node:' . $entity_type, $plugin_ids)) {
```
jaapjan: @jochemvn What if $entity_type = ''?

jochemvn: @jaapjan  then it will not be in the array and nothing will happen, since there's nothing to uninstall

jaapjan: Yeahhhh.. maybe I formulated my question wrongly. Shouldn't we check if entity_type === 'node' before executing that code?

jochemvn: Yes that would be a good addition I think. Is it even possible to add this fields to other entities? I guess yes right? I mean eg Taxonomies are also fieldable... 🤔

jaapjan: Yes in theory you can put every entity in a group

jaapjan: Marking as 'needs work' for now then :-)

jochemvn: This is what social_group does:
```
  $entity_types = ['node', 'user'];
  if (in_array($entity_type->id(), $entity_types)) {
```
So I'll limit it to nodes and users

PR #945, opened 2018-07-09
DS-5802 by serhiy-chornobay: Fix links color for preview pages

Crazylord:
## Problem
On preview and theme settings pages, buttons text color is different as on site.

## Solution
Fix styles for buttons text color in preview-base.css

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-5802

## HTT
- [x] Checkout to this branch
- [x] Login as user admin or site_manage and try change theme color settings.
- [x] Notice all buttons text color are the same as on Frontend side

## Documentation

### DIFF
--- a/themes/socialblue/color/css/preview-base.css
+++ b/themes/socialblue/color/css/preview-base.css
@@ -457,7 +457,7 @@
   margin-bottom: 0.75rem;
 }
 
-.color-preview a {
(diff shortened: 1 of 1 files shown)


jaapjan:
See the comments from Xinyu in the original issue. It will need some work/extra checking.

PR #946, opened 2018-07-09
DS-5829 by serhiy-chornobay: Fixed styles/template for edit comment page

Crazylord:
## Problem
Comment edit screen (as SM) looks not good

## Solution
Fix styles/template for edit comment page

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-5829

## HTT
- [x] Checkout to this branch
- [x] Login as user SM and try edit comment
- [x] Check if all looks good


jaapjan:
I've made a change to also remove the extra fields by default there. After discussing with Xinyu we decided that these don't add much value anyways since you can already delete, edit comments as SM. And in distro users can still freely change this behaviour.

PR #949, opened 2018-07-10
DS-5910 - Behat tests improvements (for PAC)

jaapjan:
## Problem

- Hard to find which tests are excluded in enterprise config.
- Dropdown selector for hero action buttons is dependent on xth number which can differ between installations.
- Dropdown selector for comment dropdown buttons is dependent on xth number which can differ between installations.
- Testing stuff which is not relevant for the story to test
- Using a user role to check something instead of creating new user with certain permission
- Having multiple scenarios in one test - splitting it makes it a bit easier to "ignore" tests for enterprise sites if applicable.
- The e-mail check checks for e-mail subjects. This contains the site-name which might differ between enterprise installations. I've added a configuration set for the sitename to make sure we can depend on that.
- Homepage can differ between installations
- When module is already installed in site you can not depend on "And I install module X" because in that case the configuration which is there by default may be different!
- Make tests work with BigPipe

## Solution
Will follow

## Issue tracker
Will follow

## HTT
- [ ] Will follow

## Release notes
We noticed that the Behat tests could be improved. The test were not working very well on existing installations making it harder to re-use the tests on those installations. We improved it by:
- making selectors more specific (e.g. not depend on "xth" element
- creating users with certain permission instead of using a user role for it 
- splitting multiple scenarios for a test in multiple test files
- making tests work with Big Pipe
- setting the correct configuration after installing a module
- adding more descriptive tags


jaapjan:
Note: https://github.com/goalgorilla/open_social/pull/971 is merged into this branch already.

PR #950, opened 2018-07-10
DS-5711 by serhiy-chornobay: Update styles on mobile for filter buttons

Crazylord:
## Problem
On mobile, filter buttons are not aligned properly

## Solution
Fix position for buttons on mobile

## Issue tracker
- https://www.drupal.org/project/social/issues/2980456

## HTT
- [x] Checkout to this branch
- [x] Go to event page on mobile and check filter buttons


themes/socialbase/components/02-atoms/button/button.scss
```diff
@@ -559,6 +559,9 @@ input[type="button"] {
 .block-topic-add-block,
 .block-group-add-block {
   display: inline-block;
+  @include for-tablet-landscape-down {
+    margin-top: 1rem;
```
Kingdutch: Would it be possible to move this to the place where the button is placed? So the atom describes how the button looks and we then handle positioning elsewhere. 

if we can't do this now then maybe we can add a `// TODO: ` for later :)

Crazylord: Hi @Kingdutch .
I see in scss file are comment from Piter:
// on mobile we want to "add" and "filter" buttons to be on one row
// We need to add these selectors here now, because they need to be loaded regardless of other libraries being loaded

Theoretic we can create separate css file and twig template for each block and attached relative library to this, but I don't sure if this will be best approach.

Kingdutch: I see the comment. I still think we should consider that a flaw or bug. 

I think this style should preferably be added to for example the `.block-social-topic block-topic-add-block` container and move the entire thing down instead of targetting a specific button.

I think adding a TODO is good enough for now so we can tackle that later.

PR #951, opened 2018-07-11
DS-3621 by jaapjan: change Enrolled label to 'You are enrolled'

jaapjan:
## Problem
In user testing, users are not clear what the label means. And on the event overview page of anther user, users thought that is about the user they are looking at. We should make this label clear that it is about the user him/herself. 

## Solution
Change the label from "Enrolled" to "You are enrolled".

## Issue tracker
- DS-3621

## HTT
- [x] Login as user X and enroll to an event
- [x] Check the event label on the button
- [x] Go to your profile and see the block on the right with upcoming events (same as homepage)
- [x] Go to event overview on your profile and verify your label
- [x] Go to event overview and verify label
- [x] Check changes in Behat

## Documentation

## Release notes
In user testing, users are not clear what the label means. And on the event overview page of anther user, users thought that is about the user they are looking at. We have changed the label "Enrolled" to "You are enrolled" to solve that problem.


bramtenhove:
@jaapjan could you take a look at Xinyu's comment? He mentioned that the mobile view looks a bit off.

jaapjan:
I made the changes requested. I also added the don't merge label because we have to ask Xinyu first what needs to be done for the "Joined" label for a group and if this needs to be consistent with events. If so, he would like to release this together, story created for refinement already.

jaapjan:
Too old to merge, will create new PR for 3.x

PR #952, opened 2018-07-11
#2984188 - Create activity when content is moved outside of a Group.

ronaldtebrake:
## Problem
When content is moved outside of a group, the 'create_entitiy_action' needs to be created. This is due to the fact there is no Group context anymore. Same for the other way around when content is moved to a group from the community. 

## Solution
On `hook_social_group_move()` we check if the referenced node has GroupContent if that is not the case we trigger the creation of the 'create_entitiy_action' activity. Otherwise it means the content was moved to a group and this is when we trigger the creation of the 'move_entity_action' activity.

## Issue tracker
- https://www.drupal.org/project/social/issues/2984188

## HTT
- [x] Login as user X and add a topic in the Community
- [x] Notice there is an activity message displayed on the home stream.
- [x] Move the topic in to an Open group
- [x] Notice the activity message displayed on the home stream is removed
- [x] Notice the activity message is now displayed in the stream of the group you moved the content item to.
- [x] Move the topic in to a different group
- [x] Notice the activity message is now only displayed in the stream of the group you moved the content item to.
- [x] Move the topic back outside of any group (-NONE-)
- [x] Notice the activity message is displayed on the home stream again
- [x] Notice the activity message is removed from the stream of the group you moved the content to or from before. 

## Release notes
When content is moved outside of a group, the activity item is now created and displayed in the correct stream. Same for the other way around when content is moved to a group from the community.


tests/behat/features/capabilities/group/group-edit-content-in-group.feature
```diff
@@ -34,3 +35,46 @@ Feature: Move content after creation
...
+    Then I should not see "Motorboats" in the "Main content"
+    And I should not see "Kayaking" in the "Main content"
+
+    When I am logged in as "harry"
+    And I am on the stream of group "Motorboats"
+    Then I should not see "harry created a topic in Motorboats"
+    And I should not see "I love this sport"
+    And I am on the stream of group "Kayaking"
+    And I should not see "I love this sport"
+    And I run cron
```
bramtenhove: @ronaldtebrake why is this? Should we move it maybe towards the first steps? This could have influence on the activity items.

tests/behat/features/capabilities/group/group-edit-content-in-group.feature
```diff
@@ -34,3 +35,46 @@ Feature: Move content after creation
...
+    And I should not see "I love this sport"
+
+    # Edit topic as SM to move outside of a group in community.
+    When I am logged in as "smith"
+    And I am on "/all-topics"
+    And I should see "I love this sport"
+    And I click "I love this sport"
+    And I click "Edit content"
+    And I select group "- None -"
+    And I wait for AJAX to finish
```
bramtenhove: @ronaldtebrake this is currently breaking the Travis build. See https://travis-ci.org/goalgorilla/open_social/jobs/404426725#L1610

PR #957, opened 2018-07-14
DS-5502 - Group overview available for Content Manager

bramtenhove:
## Problem
The content manager is able to edit and delete groups. However, due to the Group module not having granular permissions the overview was not reachable for that role.

This made managing groups harder for the content manager.

## Solution
There is a patch from the Group module that introduces a new permission to visit the group overview page. This has been granted to the CM and SM roles.

Next to that I took the opportunity to make the Group overview a bit more useful by removing the Group ID and introducing the numbers of members as well as the creation date.
Also I made these columns sortable!

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-5502

## HTT
- [x] Check out the code changes and run the database update with the permission changes
- [x] Login as a CM
- [x] Notice the menu item for Groups is in your admin menu
- [x] Go to the Groups overview, you should see the name, type, creator, member count and creation date
- [x] Also notice that you have the Edit and Delete operations
- [x] Try sorting the table, but notice the members column is sortable as that is a different entity
- [x] Now change the advanced outsider permissions of for example the closed group.. for example remove the 'administer group' and 'edit group' permissions
- [x] Check the overview, the button for editing the group is no longer visible

## Documentation

## Release notes
The group overview page is now accessible for the Content Manager role. The overview is also improved as the number of members and creation date are now in there. You can also sort the columns, it's sorted on latest created group by default.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,8 @@
                 "Add relationship to flagged entities when Flagging is base table": "https://www.drupal.org/files/issues/2723703_31.patch"
             },
             "drupal/group": {
-                "Add computed field for Group reference": "https://www.drupal.org/files/issues/add-computed-field-without-FieldItemListComputedInterface-2718195-34.patch"
+                "Add computed field for Group reference": "https://www.drupal.org/files/issues/add-computed-field-without-FieldItemListComputedInterface-2718195-34.patch",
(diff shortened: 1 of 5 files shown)


jaapjan:
Don't forget to add it to the feature overview later. But in general this looks good. I'll merge it already :-)

PR #965, opened 2018-07-19
DS-5881 by serhiy-chornobay: Remove Tablesaw library

Crazylord:
## Problem
Tablesaw library not used anymore

## Solution
Remove library

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-5881


Kingdutch:
This is a BC breaking change because a project using Open Social could depend on the tablesaw library so this should be done in a "major" version update.

Kingdutch:
The following file applies the `tablesaw` class which suggests that it's used: https://github.com/goalgorilla/open_social/blob/8.x-2.x/themes/socialbase/src/Plugin/Preprocess/Table.php

Crazylord:
Hey @Kingdutch. 
Thanks for your comments. I think, on this, we can delete this pull request and branch.

PR #779, opened 2018-03-12
Issue #2952203: Add Social Profile Organization Tag module

ribel:
## Problem
As a CM+ I want to indicate that user is part of an organization so that other users know where certain information comes from.

## Solution
**Acceptance criteria:**
- CM can create organization Taxonomy
- CM can associate profiles with Organization taxonomy
- This taxonomy will be an addition to the profile name of the associated profiles
- It has to be part of the name so it works in the social blue theme
- Maximum 1 value
- Should be added to the profile below other taxonomies
- Has to stand out from the rest of the name and Users can not fake it
- Separate module

**Concept:** https://marvelapp.com/c26c36j/screen/39089637

## Issue tracker
- https://www.drupal.org/project/social/issues/2952203
- https://jira.goalgorilla.com/browse/ECI-691

## HTT
- [x] Checkout to this branch
- [x] Check out the code changes
- [x] Enable experimental Social Profile Organization Tag module
- [x] Login as CM or SM
- [x] Add a new term to Profile Organization Tag vocabulary
- [x] Go to some user profile and fill the added term to Profile Organization Tag field
- [x] Check new icon near username on the different view modes
- [x] Compare it with the concept (in the activity stream message it says User from Organization for now)
- [x] Login as normal user and check that you can't edit Profile Organization Tag field and terms

## Documentation
- [x] This item is added to the release notes
- [ ] This item is documented on LGOS
- [ ] This item has been added to the feature overview

PR #967, opened 2018-07-19
Issue #2986904 by Kingdutch: Refactor the AccountHeaderBlock

Kingdutch:
## Problem
The current AccountHeaderBlock uses a very specific and large twig file to render the various cases that occur in the account header block. A single extension hook has been made to allow adding some menu items but it's otherwise not very flexible.

The render array also does some access checking to avoid printing empty URLs but this is not needed if using Drupal's link render elements.

The current method also causes a lot of coupling between the social_user module and other content type modules. Fixing this is one of the many steps towards making Open Social a decoupled project, allowing more flexibility in which parts are used.

## Solution
Rewrite all the components in the block in such a way that they use standard Drupal render arrays. For the icons on desktop with their submenu's we create an AccountHeaderElement render element that provides some shortcuts to the developer and keeps code DRY.

Two info hooks and alter hooks are then implemented that gather up all the AccountHeaderElements as well as all the links for the profile AccountHeaderElement. We deprecate the previously created alter hook as its format is ill-defined for the new structure, a glue layer between the new hook and the old hook has been implemented.

We also allow both sets of items to be sorted using weights so that the order in which hooks are called doesn't matter and sites can easily re-order them if they so choose.

## Issue tracker
- https://www.drupal.org/project/social/issues/2986904

## HTT
- [ ] Verify that the anonymous user menu still works the same
- [ ] Verify that the logged in user menu works as before
- [ ] Check that the backwards compatibility layer written for hook_social_user_account_header_links (in social_user.module) works by enabling the social_gdpr module which still uses this hook in social_gdpr_social_user_account_header_links

## Documentation
- [ ] This item receives a change notice for developers

## Release notes
The AccountHeaderBlock has been rewritten to provide more flexibility. Various hooks to alter it's behaviour have been introduced and documented in `social_user.api.php`. The previous hook to extend the user menu has been deprecated and will be removed in a future version.

## Follow up stories
A follow-up should be created to convert the hook in the social_gdpr module to the new format.


Kingdutch:
@jaapjan @bramtenhove I think this is probably ready to be committed? Would be cool if we could get this in before it misses the 4.x deadline : )

Preferrably I'd like to fix the mentioned merge conflicts in the linked issues. Would also be good to merge this before new features may expand what this refactor's

If you need me to walk you through anything, send me a message next week :)

Kingdutch:
eci_demo and SHNL fail in MOSS on their patch for PR #779 which makes a change to a template file that's removed in this PR. The change in that template file is no longer needed as a result of this PR so we can simply remove those changes from the patch.

Pachamama seems to fail on a patch for #652 which is probably due to a merge conflict in one of the .module files which means that a new patch will need to be created.

Kingdutch:
I've changed the tags here because the above changes can only be made after either this PR is merged or the other PRs are (imo).

cc @bramtenhove

Kingdutch:
I've changed the tags here because the above changes can only be made after either this PR is merged or the other PRs are (imo).

cc @bramtenhove

Kingdutch:
@jaapjan There was an issue in the social_book module. This issue was found in Behat tests using MOSS. The tests that run Travis seem to have missed this which indicates that the social_book module has no test coverage there.

ronaldtebrake:
As per testing 4.0 document fixes needed for:
- Notifications in the notification centre and the private message icon don’t show up
- The accountheader block displays the empty notification centre a but weird (not really the issue). It also displays the nr “1” on the right side of the top menu bar:
- The following modules don’t properly implement the deprecated hook_social_user_account_header_links which causes the transition code in the social_user module not to work
- Links under the [+] menu no longer look at permissions of the current user. This leads to links in the [+] menu that lead to access-denied pages for eg: Page, Landing Page, Book
In the previous version these links were not rendered.

Kingdutch:
There's currently one issue remaining that is breaking the MOSS tests which is a change in capitalisation of the "Add to content" links. In the PR all content types have been capitalised (e.g. `New Event` instead of `New event`). 

This means that this change either needs to be reverted (to `New event`) in all modules, or the tests need to be updated to look for links with the new capitalisation (`New Event`, `New Group`, etc.)

I've asked @NatashaSchon to look into this.

Kingdutch:
Discussion with Natasha an Xinyu lead me to update the tests for the new capitalisation :)

ronaldtebrake:
Blood, sweat, tears and 🍌 but it's in!

PR #972, opened 2018-07-23
DS - 5479 - As GM+ being able to add multiple people to a group at once

ronaldtebrake:
## Problem
On the "add member" page of the Manage members tab of a group you can not add multiple users to a group making it a dreadfull job adding users to a group. 

## Solution
We have implemented the same widget as for the private message. Unfortunately Group didnt support multiple entity cardinality for members yet. This means I had to go for a Distro custom solution and add my own validation and submit handler. 

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-5497

## HTT
- [ ] Login as user with GM permissions (or a SM)
- [ ] Go to your Group and go to the Manage Members tab and click Add members
- [ ] Try adding one user to the group, see that it works as before
- [ ] Try adding two users to the group, see that this also works
- [ ] Try adding two users where one or more are already members of the group, validation should fail and none of the users should be added to the groupl
- [ ] Try pasting a string with multiple users all at once like: _EAA_official (9), Alison Hendrix (13)_ and see the widget also allows this
- [ ] Try adding multiple users with a role, see that all of them have the role added as well.

## Documentation
- [ ] This item is **NOT** documented on LGOS yet
- [ ] This item has **NOT** been added to the feature overview yet

## Release notes
Want to add multiple users to a group? We’ve just made it easier for you! Group Managers are now able to add multiple users to a group by referencing them just as you are able to do in the private message functions.

### DIFF
--- a/modules/social_features/social_group/src/Element/SocialGroupEntityAutocomplete.php
+++ b/modules/social_features/social_group/src/Element/SocialGroupEntityAutocomplete.php
@@ -5,6 +5,7 @@ namespace Drupal\social_group\Element;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\social_core\Entity\Element\EntityAutocomplete;
 use Drupal\Component\Utility\Tags;
+use Drupal\user\Entity\User;
 
 /**
  * Provides an Group member autocomplete form element.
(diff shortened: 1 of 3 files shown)


modules/social_features/social_group/src/Element/SocialGroupEntityAutocomplete.php
```diff
@@ -0,0 +1,92 @@
...
+    foreach ($input_values as $input) {
+      $match = static::extractEntityIdFromAutocompleteInput($input);
+      if ($match === NULL) {
+        $options = $element['#selection_settings'] + [
+          'target_type' => $element['#target_type'],
+          'handler' => $element['#selection_handler'],
+        ];
+
+        /* @var /Drupal\Core\Entity\EntityReferenceSelection\SelectionInterface $handler */
+        $handler = \Drupal::service('plugin.manager.entity_reference_selection')->getInstance($options);
```
Kingdutch: We should probably use dependency injection here since we're using a class.

modules/social_features/social_group/src/Element/SocialGroupEntityAutocomplete.php
```diff
@@ -0,0 +1,92 @@
...
+        // them all in the _social_group_action_form_submit.
+        parent::validateEntityAutocomplete($element, $form_state, $complete_form);
+      }
+    }
+
+    // If we have duplicates, provide an error message.
+    if (!empty($duplicated_values)) {
+      $usernames = implode(', ', $duplicated_values);
+
+      $message = \Drupal::translation()->formatPlural(count($duplicated_values),
```
Kingdutch: We should probably use dependency injection instead of `\Drupal::translation()` as we're inside of a class.

modules/social_features/social_group/src/Element/SocialGroupEntityAutocomplete.php
```diff
@@ -0,0 +1,92 @@
...
+        "@usernames is already member of the group, you can't add them again",
+        "@usernames are already members of the group, you can't add them again",
+        ['@usernames' => $usernames]
+      )->__toString();
```
Kingdutch: No need to call the `__toString()` magic method. `formatPlural` returns a `PluralTranslatableMarkup` interface which is part of the chain of classes that implements `MarkupInterface`.

`MarkupInterface` (or classes implementing it) can be passed in place of a string here (it has the `__toString()` magic method) which will ensure it'll get recognised as being safe upon render. 

By forcing it to a string here we are passing around raw strings which will get escaped again upon render (causing usernames with apostrophes to show things like &ap;).

bramtenhove: Excellent point, tbh I think i copy pasted this from somewhere in core 😇

PR #975, opened 2018-07-25
Feature/ds 5671 photo swipe

Crazylord:
## Problem
Post image not working correctly for mobile uploaded pictures

## Solution
Use exif_orientation module

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-5671

## HTT
- [x] Login as user Chrishall and try upload image for post and publicate (on mobile).
- [x] Notice if click on image, image have bad orientation
- [x] Checkout to this branch
- [x] Try upload image for post and check image orientation

## Documentation


Kingdutch:
The current diff on this PR shows that a previous update hook is reused which causes this not to be executed. This should be fixed.

Furthermore if this module is truly a dependency then this should probably also be added to an .info.yml file so that it can not be accidentally disabled.

ronaldtebrake:
Thanks guys, also tested this on Jungle Town where we encountered this issue and it works for newly added pictures. Fixed the merge conflicts and will make sure in 3.x the hook_update_N has the correct number as well.

PR #976, opened 2018-07-26
Issue #2988332 by robertragas: bump up restrict images patch to #40 t…

robertragas:
…o support special characters

## Problem
When restrict images text format filter is on and there are images with special characters in them, it fails the search indexing which also will crash the cronjob.

## Solution
Bump up the patch number to 40 which includes support for special characters

## Issue tracker
https://www.drupal.org/project/social/issues/2988332#comment-12706997

## HTT
- [x] Login, and add some images with special characters in it.
- [x] clear the search index from "all" and reindex
- [ ] Notice that the indexing will crash with the message about insecure images
- [ ] Checkout to this branch
- [ ] Retry and it works.

## Documentation

## Release notes
Updated the restrict images patch that will prevent crashes during reindexing of search results when images had special characters in them.


jaapjan:
@robertragas don't forget the .make files when you make changes in patches or modules; marking as `needs work` for now

bramtenhove:
@jaapjan, @robertragas, do we want to include this in 2.4 or 3.0 still?

jaapjan:
@robertragas this will not be in 2.4 because of lack of response. We can still try to get in 3.0 but it will require action from your side.

robertragas:
Added new patch to drupal.org and applied it here where str_replace is replaced by urldecode

jaapjan:
@robertragas so the correct status is now `needs review`? Will this allow all the special characters now?

jaapjan:
P.s. base branch is still 8.x-2.x, maybe you can adjust this to 3.x as well?

robertragas:
@jaapjan correct, it will now work for all the special characters. I will fix the branch tomorrow morning right away that it's based on the 3.x branch.

jaapjan:
TODO: Also remove the patch from sdgc.

PR #981, opened 2018-07-31
OSSUPPORT-579 Improve the usability of login and register screens

Kingdutch:
## Problem
If a user is sent a link to a page that he/she can not access and
he/she does not have an account yet then the default login page of
Open Social is pretty useless.

When the social login module is used then those buttons are displayed
inside the login/register card which turns out as a weirdly nested
card.

## Solution
This commit cleans up the login page by adding a link between
login and sign up screens so that users can easily navigate between
them while preserving the post-login destination parameter.

Furthermore the commit moves the social buttons into their own
cards.

Finally a configurable text is added both for normal/social login
as well as register. This allows site-managers to guide users
during these procedures as needed. By default only the sign-up
screen has a text configured.

A follow-up issue to clean up the configuration and make it reusable
is here: https://www.drupal.org/project/social/issues/2989522

## Issue tracker
- https://jira.goalgorilla.com/browse/OSSUPPORT-579

## HTT
- [x] Check out the login and registration screens
- [x] Enable the social_auth_extra module and configure at least one social_auth source
- [x] Check out the login and registration screens
- [x] Play with the default texts on /admin/config/people/accounts
- [x] Notice that if you remove the help text it is no longer displayed

## Documentation

## Release notes
Improved the user registration and login experience by providing links between the two screens. Site managers now also have the option to include custom help texts on these pages by changing these in account settings. A default help text has been included for regular sign up (social_user module) and social signup (social_auth_extra module).


Kingdutch:
Added the Don't Merge tag because `social_user` is a feature so any editable configuration there would be overwritten by a `drush fra` which is run on each deploy.

Kingdutch:
This needs testing with the social_captcha module

Kingdutch:
After testing the new settings do not get reverted so this is safe to merge. Adjustments for the social_captcha (and to the social_captcha) module have been made.

bramtenhove:
@Kingdutch could you take a look at why the patch in Moss is not applying?

Kingdutch:
@bramtenhove The patch in PAC is a known merge conflict because both the AN Enrollment feature and this PR modify the same line in socialbase' form.inc. This has been fixed for the project itself but not the branch that Moss uses.

PR #982, opened 2018-07-31
Issue #2989533 by robertragas: reduce the scope of overrides applied …

robertragas:
…to event overview

## Problem
We have an Enterprise client with some overrides on the event overview. This applies to the filter block on the right. After an update these overrides suddenly didn't work anymore.

Checking out the Distro I see there is an override being done on this overview on the filters, but it changes all the filters with the original config and not only the one that needs to be changed. 

## Solution
Reduce the scope of the override so it only overrides the changes that it needs to.

## Issue tracker
https://www.drupal.org/project/social/issues/2989533

## HTT
- [x] Login as user X and go to the personal stream page and go to the events overview.
- [x] On the right side you see a filter block, try to create an override for this.
- [x] Notice the override isn't being picked.
- [x] Checkout to this branch
- [x] Retry and see the override is being picked.

## Documentation

## Release notes
The filter block on events page of a specific user applies a config override to add a new filter. This code overruled custom overrides. We made the override more specific so custom overrides can be used again.


jaapjan:
Works, we can merge this after 2.3 is released tonight!

PR #983, opened 2018-08-01
Issue #2989731 : Added related "Create" permissions for block display

slowflyer:
## Problem
The "Add a group", "Add topic" and "Add event" Button show allways up on the related page of the user profile, even if the user has no permission to create the related group/content.

## Solution
Added related permission to check if block is displayed or not

## Issue tracker
- https://www.drupal.org/project/social/issues/2989731

## HTT
- [x] Remove permission to create topic, event or all kind of groups from user role
- [x] Login as user with the role from above and go to the user profile
- [x] Verify that the "Add" buttons do not appear
- [x] Add permission to create topic, event or all kind of groups from user role
- [x] Verify that the "Add" buttons do not appear

## Documentation

## Release notes
<describe the release notes>

### DIFF
--- a/modules/social_features/social_group/src/Plugin/Block/GroupAddBlock.php
+++ b/modules/social_features/social_group/src/Plugin/Block/GroupAddBlock.php
@@ -7,6 +7,7 @@ use Drupal\Core\Url;
 use Drupal\Core\Link;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Access\AccessResult;
+use Drupal\group\Entity\GroupType;
 
 /**
  * Provides a 'GroupAddBlock' block.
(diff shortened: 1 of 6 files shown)


modules/social_features/social_group/src/Plugin/Block/GroupAddBlock.php
```diff
@@ -28,7 +28,7 @@ public function blockAccess(AccountInterface $account) {
     $route_user_id = \Drupal::routeMatch()->getParameter('user');
 
     // Show this block only on current user Groups page.
-    if ($current_user->id() == $route_user_id) {
+    if ($current_user->id() == $route_user_id && ($account->hasPermission("create open_group group") || $account->hasPermission("creat closed group") || $account->hasPermission("create public_group group"))) {
```
jaapjan: Do you think it will also be possible to make this access check a bit more generic? E.g. what if websites have disabled all the group types and created own custom group type? Is there a way to retrieve all the group types and then check if the user has permission to create new groups of one of the group types?

slowflyer: I'll look into it

jochemvn: @slowflyer  GroupType::loadMultiple give you all group types available... then you could check it for all available groups very easily.
    foreach (GroupType::loadMultiple() as $group_type) {
      $permissions = 'create ' . $group_type->label() . 'group'; //or id();
}

slowflyer:
cant see why continuous-integration/moss — OS dependant checker  is failing ... 

Login is required for Details

jaapjan:
@slowflyer That integration is doing some extra checks on some of our internal projects. I will check it out for you, looks like a false negative though. Great to see that Travis is passing already! Is it ready for review now?

slowflyer:
It's ready for review.

PR #984, opened 2018-08-01
Issue #2989478 - only create messages once to prevent duplicates in streams

jaapjan:
## Problem
This was introduced in https://github.com/goalgorilla/open_social/pull/952. When you create a node in a group, edit the node directly thereafter and remove the group from the node before the cron run you will end up with duplicate activities in the stream. The reason is that there are multiple messages created for the same entity with the exact same values. Whenever a message is created the function `activity_creator_message_insert` gets triggered which creates activity items in the stream. 

## Solution
For each activity action there should only be one Message as per https://www.drupal.org/docs/8/distributions/open-social/activity-stream-architecture-and-background-information . Let's make sure to check for duplicates before creating a Message. 

I have added a public function checkIfMessageExist in the ActivityLoggerFactory which is responsible for the message creation. This function is called right before the message is created in `ActivityLoggerFactory::createMessages`. The function checks if a message with the following values already exists:

- message type (e.g. 'create_topic_group' when new node is created in group)
- context (e.g. 'group_activity_context' when new node is created in group)
- destination (can be an array of destinations, e.g. stream_profile, email, etc.)
- related object (contains target_type, target_id, e.g. node/24)
- uid (uid of the actor)

If the message already exist it will just skip creating it.

## Issue tracker
- https://www.drupal.org/project/social/issues/2989478

## HTT
- [x] Stay on branch 8.x-2.x
- [x] Make sure your cron container is not running (`docker-compose stop cron`)
- [x] Login as user chrishall and create a group
- [x] Create a new node in the group
- [x] Edit the node and remove the group from the node
- [x] Make a `sql-dump` 💯 
- [x] Run the cron and notice there are two activities in the stream of chrishall, the homepage, etc.
- [x] Check out to this branch
- [x] `sql-drop` and import the dump
- [x] Run cron
- [x] Notice there is only activity
- [x] Now test a bunch more actions and make sure to run the cron; check if everything works as expected

## Documentation

## Release notes
When you create a new node in a group and edit the node (remove the group) directly after in some cases this could result in duplicate activities in the stream. This is now solved by making sure that the ActivityLoggerFactory.php never creates duplicate messages.


modules/custom/activity_logger/src/Service/ActivityLoggerFactory.php
```diff
@@ -175,4 +176,43 @@ protected function createFieldInstances($message_type, array $fields) {
...
+   * @param string $uid
+   *   The uid of the message.
+   *
+   * @return int
+   *   Returns true if the message exists.
+   */
+  public function checkIfMessageExist($message_type, $context, array $destination, array $related_object, $uid) {
+    $exists = FALSE;
+
+    $query = \Drupal::entityQuery('message');
```
bramtenhove: @jaapjan we should use dependency injection here.

jaapjan: Agree, but we don't use it for the rest of the services in here as well. So I propose to work on that in a later stage (since this is considered "hotfix")

jaapjan:
See also https://github.com/goalgorilla/open_social/pull/952

PR #987, opened 2018-08-02
Issue #2990176 : User export extended 

jaapjan:
## Problem
Multiple issues with the user export system:

- Not all the fields are available in the user export currently
- Multiple value fields (e.g. taxonomy term references) show only the first value, when you have multiple interests we need to show them
- Code was very specific and we used a lot of database queries.
- When the User was passed to a profile field value it would use the wrong id, namely the account->id instead of the profile->id. This gives weird results.
- When separate installations want to exclude certain fields or include new fields from the export this is currently not possible. 
- When social_profile_fields is enabled and some fields are not enabled they should not be in the export.

Also saw this issue: 
https://www.drupal.org/project/social/issues/2954838 which is easy to solve since this is in 3.x anyway.

## Solution
We need all these fields:

+ User ID
+ UUID
+ First name
+ Last name
+ Username
+ Display name
+ Email
+ Last login
+ Last access
+ Registration date
+ Status
+ Country code
+ Administrative address
+ Address locality
+ Postal code
+ Address line 1
+ Address line 2
+ Phone number
+ Organization
+ Function
+ Skills
+ Roles
+ Interest
+ Profile tag
+ Posts created
+ Comments created
+ Topics created
+ Events enrollments
+ Events created
+ Groups created
+ Number of likes
+ Total send private messages
+ Nickname

Renamed `UID` column to `User ID`.

Rewrote the code to make sure to use re-usable functions to retrieve field values, taxonomy term references and address field value. No more database queries there since we have the entity already. I did not change anything to the functions to retrieve stats for number of likes, content created, etc. since those are a bit harder to make re-usable (sometimes owner column in DB is user_id, sometimes uid, etc.).

I've made sure to only retrieve the Profile once and re-use that to retrieve values for the given user.

I've used the plugin system to retrieve all the user export columns.

There is now support for social_profile_fields as well. When fields are disabled they are not included in the export anymore.

NOTE: I realise all of this is not perfect yet, but given the amount of time reserved for this it's a good first step!

## Issue tracker
- https://www.drupal.org/project/social/issues/2990176
- https://www.drupal.org/project/social/issues/2954838

## HTT

- [x] Check the code
- [x] Enable social_user_export
- [x] Create some profile tag terms so a CM+ can add these to a certain user
- [x] Verify that the export works with social_private_message enabled and disabled
- [x] Fill in one or multiple profiles completely, including address details
- [x] Login as SM+ and make some exports.
- [x] You can also use my export script to make some exports with `drush scr user-export.php --user=1`: [user-export.php.zip](https://github.com/goalgorilla/open_social/files/2267266/user-export.php.zip)
- [x] Check if all the fields are completely. 
- [x] Are there any obsolete fields in here?
- [x] Verify the plugin system works
- [x] Implement an alter hook to remove a field or change the weight, e.g.
```

/**
 * Modify the list of available User export plugins.
 *
 * This hook may be used to modify plugin properties after they have been
 * specified by other modules.
 *
 * @param array $plugins
 *   An array of all the existing plugin definitions, passed by reference.
 *
 * @see \Drupal\social_user_export\UserExportPluginManager
 */
function hook_social_user_export_plugin_info_alter(array &$plugins) {
  if ($plugins['account_name']) {
    unset($plugins['account_name']);
  }
}
```
- [x] Enable social_profile_fields module and disable some fields, enable nick name.
- [x] Save the form, do not clear the cache.
- [x] Fill in the nickname for a user.
- [x] Verify that the export has removed the disabled fields and added the nickname field.

## Documentation

## Release notes
We've added more fields to the user export, most notable all the profile fields. In addition we've rewritten most of the code, making better use of Drupal's Entity API, Plugin system and removing duplicate code. It is now possible to create your own User export plugins to add new fields to the export. The headers titles and the order might have changed. In addition disabled fields will not be exported anymore and there is support for the Nickname field.


jaapjan:
@bramtenhove this PR is available for review. Let's remember to not merge it yet, since we still want to look into a way to add alter hooks to make the system more flexible.

modules/social_features/social_user_export/src/ExportUser.php
```diff
@@ -89,20 +100,42 @@ public static function exportUserOperation(UserInterface $entity, array &$contex
...
+      _social_user_export_profile_get_address('field_profile_address', 'locality', $user_profile),
+      _social_user_export_profile_get_address('field_profile_address', 'postal_code', $user_profile),
+      _social_user_export_profile_get_address('field_profile_address', 'address_line1', $user_profile),
+      _social_user_export_profile_get_address('field_profile_address', 'address_line2', $user_profile),
+      _social_user_export_profile_get_field_value('field_profile_phone_number', $user_profile),
+      _social_user_export_profile_get_field_value('field_profile_organization', $user_profile),
+      _social_user_export_profile_get_field_value('field_profile_function', $user_profile),
+      _social_user_export_profile_get_taxonomy_field_value('field_profile_expertise', $user_profile),
+      _social_user_export_profile_get_taxonomy_field_value('field_profile_interests', $user_profile),
+      _social_user_export_profile_get_taxonomy_field_value('field_profile_profile_tag', $user_profile),
```
bramtenhove: @jaapjan profile tag module is optional right? Maybe put a `moduleExists` around it just like we did for private message.

bramtenhove:
@jaapjan, looks good. I think the changes you did makes it a bit more robust and easier to use.

I added a small comment about an optional module and I noticed Moss is failing.

Kingdutch:
@jaapjan Is the field "Self introduction" also included in this export?

modules/social_features/social_user_export/src/ExportUser.php
jaapjan: I also thought that, but social_tagging (for content) is optional. This field is always there, since it is delivered with `social_user` feature. If there are no profile tag taxonomy terms then the feature will just not do anything.

jaapjan:
@Kingdutch nope, that field is not included

jaapjan:
@robertragas Not part of this story, but it would be an interesting addition. Perhaps you could create an issue for that?

jaapjan:
@bramtenhove updated the HTT and release notes with the latest plugin system changes.

modules/social_features/social_private_message/src/Plugin/UserExportPlugin/UserAnalyticsPrivateMessage.php
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @UserExportPlugin(
+ *  id = "user_private_message",
+ *  label = @Translation("Number of Private messages"),
+ *  weight = -190,
+ * )
+ */
+class UserAnalyticsPrivateMessage extends UserExportPluginBase {
+
+  use StringTranslationTrait;
```
bramtenhove: @jaapjan do we need this trait in every file? Even if the `UserExportPluginBase` already uses it?

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserAddressAdministrative.php
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @param \Drupal\user\UserInterface $entity
+   *   The User entity to get the value from.
+   *
+   * @return string
+   *   The value.
+   */
+  public function getValue(UserInterface $entity) {
+    $profile = $this->getProfile($entity);
+    return $this->profileGetAddressFieldValue('field_profile_address', 'administrative_area', $profile);
```
bramtenhove: Perhaps we can skip defining the `$profile` variable and just make it:
`return $this->profileGetAddressFieldValue('field_profile_address', 'administrative_area', $this->getProfile($entity));`

Same for other plugins.

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserId.php
```diff
@@ -0,0 +1,45 @@
...
+  use StringTranslationTrait;
+
+  /**
+   * Returns the header.
+   *
+   * @return \Drupal\Core\StringTranslation\TranslatableMarkup|string
+   *   The header.
+   */
+  public function getHeader() {
+    return $this->t('User Id');
```
bramtenhove: Nitpicking, but shall we make it `User ID`?

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserLastAccess.php
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @param \Drupal\user\UserInterface $entity
+   *   The User entity to get the value from.
+   *
+   * @return string
+   *   The value.
+   */
+  public function getValue(UserInterface $entity) {
+    if ($last_access_time = $entity->getLastAccessedTime()) {
+      $last_access = $this->dateFormatter->format($last_access_time, 'custom', 'Y/m/d - H:i');
```
bramtenhove: Let's use the `short` date format from core. As these are also translatable for other languages who have a different annotation for date and time.

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserLastAccess.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+    else {
+      $last_access = t('never');
```
bramtenhove: This is a nice addition!

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserStatus.php
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @param \Drupal\user\UserInterface $entity
+   *   The User entity to get the value from.
+   *
+   * @return string
+   *   The value.
+   */
+  public function getValue(UserInterface $entity) {
+    $status = $entity->get('status')->getValue();
+    return !empty($status[0]['value']) ? $this->t('Active') : $this->t('Blocked');
```
bramtenhove: I suggest we use `return $entity->isActive() ? $this->t('Active') : $this->t('Blocked');`

modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php
```diff
@@ -0,0 +1,234 @@
...
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   * @param \Drupal\Core\Datetime\DateFormatterInterface $date_formatter
+   *   The date formatter.
+   * @param \Drupal\Core\Database\Connection $database
+   *   The database connection.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, EntityTypeManagerInterface $entity_type_manager, DateFormatterInterface $date_formatter, Connection $database) {
```
bramtenhove: I'm not sure if there's an easy way around this, but we're adding services like `database` and `dateFormatter` that we do not user for every export plugin. It might be more clean to add those in the plugins that actually use them.

modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php
```diff
@@ -0,0 +1,234 @@
...
+    );
+  }
+
+  /**
+   * Returns the header.
+   *
+   * @return \Drupal\Core\StringTranslation\TranslatableMarkup|string
+   *   The header.
+   */
+  public function getHeader() {
```
bramtenhove: Because we're exporting data we will always need a header right? I would suggest to simply remove this function and let the plugin take care of it.

modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php
```diff
@@ -0,0 +1,234 @@
...
+  /**
+   * Returns the value.
+   *
+   * @param \Drupal\user\UserInterface $entity
+   *   The User entity to get the value from.
+   *
+   * @return string
+   *   The value.
+   */
+  public function getValue(UserInterface $entity) {
```
bramtenhove: Because we're exporting data we will always need a value right? I would suggest to simply remove this function and let the plugin take care of it.

modules/social_features/social_user_export/src/Plugin/UserExportPluginInterface.php
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Get the Profile entity.
+   *
+   * @param \Drupal\user\UserInterface $entity
+   *   The user entity to get the profile from.
+   *
+   * @return \Drupal\profile\Entity\ProfileInterface|null
+   *   Returns the Profile or NULL if profile does not exist.
+   */
+  public function getProfile(UserInterface $entity);
```
bramtenhove: The only two functions that we **really need** are `getHeader()` and `getValue()`. I don't think we have to define the other functions like `getProfile()` and `profileGetFieldValue()` in the Interface. We cannot let them live as additional functions in the pluginBase for whichever plugin wants to extend those.

@jochemvn check?

modules/social_features/social_private_message/src/Plugin/UserExportPlugin/UserAnalyticsPrivateMessage.php
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Returns the header.
+   *
+   * @return \Drupal\Core\StringTranslation\TranslatableMarkup|string
+   *   The header.
+   */
+  public function getHeader() {
```
bramtenhove: For all `getHeader()` and `getValue()` functions we can simply use:
```/**
 * {@inheritdoc}
 */```

File: modules/social_features/social_private_message/src/Plugin/UserExportPlugin/UserAnalyticsPrivateMessage.php (hunk 2 above)
Comment (jaapjan) [reply modules/social_features/social_private_message/src/Plugin/UserExportPlugin/UserAnalyticsPrivateMessage.php]: Nope, good point. I'll remove it.

File: modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php (hunk 9 above)
Comment (Kingdutch) [reply modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php]: You can also make the base class abstract and make this an abstract method (in the interface) so that other plugins are force to implement it.

File: modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php (hunk 10 above)
Comment (Kingdutch) [reply modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php]: You can also make the base class abstract and make this an abstract method (in the interface) so that other plugins are force to implement it.

File: modules/social_features/social_private_message/src/Plugin/UserExportPlugin/UserAnalyticsPrivateMessage.php (hunk 12 above)
Comment (jaapjan) [reply modules/social_features/social_private_message/src/Plugin/UserExportPlugin/UserAnalyticsPrivateMessage.php]: you're right. I'll change it

File: modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserAnalyticsCommentsCreated.php [hunk 13]
```diff
@@ -0,0 +1,52 @@
...
+   *
+   * @return string
+   *   The value.
+   */
+  public function getValue(UserInterface $entity) {
+    $query = $this->database->select('comment', 'c');
+    $query->join('comment_field_data', 'cfd', 'cfd.cid = c.cid');
+    $query->condition('cfd.uid', $entity->id());
+
+    return (string) $query
```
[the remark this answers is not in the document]
jaapjan: Allright.

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserId.php
jaapjan: sure

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserLastAccess.php
jaapjan: good idea

modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php
jaapjan: I don't think we should do that. That will make things even moe messy. I also have a feeling that the impact on performance will not be that big anyways. And especially entity type manager and database are very valuable services and you will use the most likely.

modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php
jaapjan: Yes, but this way there is always a fallback. In the future I think we should check if we can use the `label` from the annotation of the plugin.

modules/social_features/social_user_export/src/Plugin/UserExportPluginBase.php
jaapjan: Yes, thats true. But I prefer to keep it in for documentation purposes.

modules/social_features/social_user_export/src/Plugin/UserExportPluginInterface.php
jaapjan: Not sure what you are trying to say here @bramtenhove . You mean we should only put functions in the Interface which are really needed?

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserAddressAdministrative.php
jaapjan: Allright, good idea.

jaapjan:
@bramtenhove thanks for the feedback, processed it all in the last commit.

In addition I've also added the nickname field and support for social_profile_fields. When fields are disabled in the platform they are now excluded in the user export as well.

bramtenhove:
@jaapjan I found a blocking issue with regard to the `social_profile_fields` module. If the export module is not enabled and you're trying to save the profile fields form it will return a fatal error.

I already found the issue, will try to fix it now.

Kingdutch:
I just realised: we should probably remove all the this->t calls in the header field since the exports are probably meant for an import program and translating the headers could wreck that process if you’re exporting while voewonh the site in a different language.

We should also always add the langcode to each row so that we can later easily add exports of translated profiles.

jaapjan:
@Kingdutch It is an interesting idea, but I propose to leave it as is now. There can also be sites which are only used in one language and then it makes a lot of sense to use the translations. Also a "import program" could use the translated headers. 

And next to that we will have an API later this year which will be a lot more developer-friendly then the ancient .csv system.

PR #993, opened 2018-08-13
DS-5832 - As a CM+ I want to recognise unpublished comments and (un)publish them

bramtenhove:
## Problem
Comments can already be unpublished via the comments overview page. However these comments do not look different from published comments.

Next to that CM+ cannot publish or unpublish comments in the comment edit screen.

## Solution
CM+ can now publish or unpublish comments in the comment edit screen. There is also a visual indication that the comment (or comment reply) is unpublished.

When the main comment is unpublished all the replies are also hidden as the thread will have no context anymore.

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-5832

## HTT
- [x] Login as LU and add a comment to a node
- [x] Add two replies to the comment
- [x] Login as CM and unpublish the second reply
- [x] As LU notice you see the comment with a single reply
- [x] As CM notice you see the comment and the replies, it's indicating clearly which is unpublished
- [x] Notice you cannot click the Reply button nor like the comment (you can see who liked the comment)
- [x] Unpublish the main comment
- [x] As CM notice you see the comment and the replies, it's indicating clearly which is unpublished
- [x] As LU notice you see no comment nor the thread

## Documentation

## Release notes
CM+ can now clearly see which comments are unpublished. Also it is possible to publish or unpublish comments by going to their edit screen.


modules/social_features/social_comment/social_comment.module
```diff
@@ -96,37 +135,45 @@ function social_comment_form_reply_validate($form, FormStateInterface $form_stat
 function social_comment_comment_links_alter(array &$links, CommentInterface $entity, array &$context) {
   $bundle = $entity->bundle();
   $commented_entity = $entity->getCommentedEntity();
-  $account = \Drupal::currentUser();
-  switch ($bundle) {
```
bramtenhove: Refactored the `switch` statement as it was really just an `if` statement..

themes/socialbase/socialbase.theme
```diff
@@ -83,10 +84,33 @@ function _socialbase_get_visibility_icon($title) {
 function _socialbase_node_get_comment_count(Node $node, $comment_field_name) {
   $count = 0;
 
-  $comment_count = $node->{$comment_field_name}->comment_count;
-
-  if ($comment_count) {
-    $count = $comment_count;
+  $cids = \Drupal::entityQuery('comment')
```
bramtenhove: As LU cannot see unpublished comments (or replies on them), but CM+ can I needed a better distinction than we already had.

modules/social_features/social_comment/config/install/core.entity_form_display.comment.comment.default.yml
```diff
@@ -10,13 +10,18 @@ bundle: comment
...
     type: string_textarea
     region: content
+  author:
+    weight: 1
+    settings: {  }
+    third_party_settings: {  }
+    region: content
 hidden:
   subject: true
   author: true
```
jaapjan: @bramtenhove shouldnt we remove `author` from `hidden` when we add it to the `content`?

jaapjan:
@bramtenhove I have one open question in the files regarding the form display config. Could you look at that? Other than that it looks good!

modules/social_features/social_comment/config/install/core.entity_form_display.comment.comment.default.yml
bramtenhove: Nice find! Looks like I put back that part manually 😇

jaapjan:
Made two small changes:

1. Added message to make clear that replies on unpublished comments are not visible.
2. Changed form display config to make sure the author is not hidden anymore.

PR #997, opened 2018-08-14
DS-5933 by jaapjan: add more context to e-mails

jaapjan:
See full overview here:
https://docs.google.com/spreadsheets/d/16NdIBG8GZXaXQV5YpCfp17yKm7RshArRs_goKZNlU5E/edit?usp=sharing

## Problem
Not clear about which topics/events/groups etc the notification is about.

## Solution
If we know the title of the node, let's show it in the (e-mail) notification.

## Issue tracker
- DS-5933

## HTT
- [x] Test all notifications from this overview: https://docs.google.com/spreadsheets/d/16NdIBG8GZXaXQV5YpCfp17yKm7RshArRs_goKZNlU5E/edit?usp=sharing (make sure to test the notification (in-site), push notification and the e-mail notification for each.
- [x] See if any of the greyed-out or yellow notfications from the overview should have been worked on.

## Documentation

## Release notes
Will follow.

### DIFF
--- a/modules/social_features/social_follow_content/config/install/message.template.create_comment_following_node.yml
+++ b/modules/social_features/social_follow_content/config/install/message.template.create_comment_following_node.yml
@@ -20,13 +20,13 @@ label: 'Create comment on following node'
 description: 'A person commented on content I am following'
 text:
   -
-    value: "<p><a href=\"[message:author:url:absolute]\">[message:author:display-name]</a> commented on [social_comment:parent_entity_author]'s <a href=\"[message:field_message_related_object:entity:url:absolute]\">[social_comment:commented_content_type]</a> you are following</p>\r\n"
+    value: "<p><a href=\"[message:author:url:absolute]\">[message:author:display-name]</a> commented on [social_comment:parent_entity_author]'s [social_comment:commented_content_type] <a href=\"[message:field_message_related_object:entity:url:absolute]\">[message:node-title]</a> you are following</p>\r\n"
     format: full_html
   -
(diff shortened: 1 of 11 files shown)


modules/social_features/social_follow_content/config/install/message.template.create_comment_following_node.yml
```diff
@@ -20,13 +20,13 @@ label: 'Create comment on following node'
 description: 'A person commented on content I am following'
 text:
   -
-    value: "<p><a href=\"[message:author:url:absolute]\">[message:author:display-name]</a> commented on [social_comment:parent_entity_author]'s <a href=\"[message:field_message_related_object:entity:url:absolute]\">[social_comment:commented_content_type]</a> you are following</p>\r\n"
+    value: "<p><a href=\"[message:author:url:absolute]\">[message:author:display-name]</a> commented on [social_comment:parent_entity_author]'s [social_comment:commented_content_type] <a href=\"[message:field_message_related_object:entity:url:absolute]\">[message:node-title]</a> you are following</p>\r\n"
```
bramtenhove: @jaapjan was it a deliberate decision not to make `[social_comment:parent_entity_author]` a link? I can imagine we didn't do this because it wasn't a link before, just checking.

bramtenhove: Just discussed with Xinyu that we should keep it as-is.

jaapjan:
Solved the issue that likes on comments were displayed as:
```
Ben Flórez likes your comment (No subject)
```

This is because the comment object sometimes had a label() - not in demo content comments though. Anyway, when comments have a label we should always add extra context anyway. So i wrote a small exception for that. Tested all the other entities as well (post, node)

PR #998, opened 2018-08-14
DS-5981 by serhiy-chornobay: Move card link in card actionbar for Landing page

Crazylord:
## Problem
The .card__link element is supposed to be used in a .card__actionbar element that is next to the .teaser__content element.

## Solution
Move the .card__link element out of the .teaser__content element and wrap it in a .card__actionbar element.

## Issue tracker
- https://www.drupal.org/project/social/issues/2991405
- https://jira.goalgorilla.com/browse/DS-5981

## HTT
- [ ] Checkout to this branch
- [ ] Check if all teasers on landing page looks good

## Documentation


modules/social_features/social_landing_page/css/social_landing_page.featured.scss
```diff
@@ -51,7 +51,7 @@
     margin-top: 30px;
     margin-bottom: 0.75rem;
   }
-  .card__actionbar {
+  footer.card__actionbar {
```
Kingdutch: Hey Serghiy,

What is the reason that we're adding the `footer` tag here as a specifier?

Can we instead be more specific by targetting a specific place where this padding should be 0 (or remove it altogether?)

Kingdutch:
@Crazylord Just the one question about using a specific tag for reduced padding. I think that's risky especially since our examples say that semantically we should use the footer tag everywhere instead of the divs we now use in a lot of places.

Other than that it looks good :)

Crazylord:
@Kingdutch Remove the footer tag and fix styles for it.

jaapjan:
Anyone knows the status here? 

@Kingdutch you started the review of this so I'll assign it to you as it seems the necessary change was made.

jaapjan:
P.s. I see base branch is still 3.x. This maybe needs to be changed as well.

PR #999, opened 2018-08-14
Issue #2990176 by bramtenhove: Add self introduction as field to the export

bramtenhove:
## Problem
The self introduction field can be useful for measuring community activity. We did not include it in the [other PR](https://github.com/goalgorilla/open_social/pull/987) that introduces a lot of new fields to the export.

## Solution
Added the _Self introduction_ to the export and changed the weights of the plugins so it fits nicely somewhere at the end.

The output of the _Self introduction_ field is automatically escaped, just like the rest of the fields. This means that links and images are stripped. I don't think this is an issue.

## Issue tracker
- https://www.drupal.org/project/social/issues/2990176

## HTT
- [ ] Login as a SM and export users
- [ ] Observe that the between _Interests_ and _Profile tag_ the _Self introduction_ of the user is filled in

## Documentation

## Release notes
n/a


bramtenhove:
@Kingdutch FYI

PR #1004, opened 2018-08-16
2992994: URL alias add on too long and making vertical tabs autogrow on input

jochemvn:
## Problem
Whenever the URL of a website is really long, there's not a lot of space for input in the URL alias field, due to the input add-on. Also when filling out the URL Alias field, the vertical tabs on the left keep growing, making the mentioned problem above even bigger.

## Solution
Make sure the vertical tabs section doesn't grow ever. Secondly, strip protocols from the URL and trim if still longer than 40 chars.

## Issue tracker
- https://www.drupal.org/project/social/issues/2992994
- https://jira.goalgorilla.com/browse/SHN-163

## HTT
- [ ] Login as a SM and add a node or topic
- [ ] Go to the vertical tabs at the bottom
- [ ] Notice that when you enter a long alias, the pane left keeps on growing
- [ ] Also notice that if the URL if the site is very long, that there's not a lot of room for input (EG: https://demo.getopensocial.com)
- [ ] Checkout to this branch and clear the cache
- [ ] Notice both problems are fixed

## Release notes
As a SM when editing the URL Alias field, the input section is now bigger and the URL part shorter. Also when entering a long URL Alias, the pane on the left stays the same size and does not grow.

### DIFF
--- a/modules/social_features/social_core/social_core.module
+++ b/modules/social_features/social_core/social_core.module
@@ -479,7 +479,28 @@ function social_core_field_widget_form_alter(&$element, FormStateInterface $form
     $url = Url::fromRoute('<front>', [], [
       'absolute' => TRUE,
     ]);
-    $element['alias']['#field_prefix'] = '<div class="input-group input-group-expanded"><span class="input-group-addon">' . $url->toString() . '</span>';
+
+    // Protocols we can strip.
+    $protocols = [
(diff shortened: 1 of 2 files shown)


jochemvn:
@ribel,
Could you review this on wednesday, so it can be added to the 2.x release?
Jochem

PR #1005, opened 2018-08-16
Issue #2992979 : Blocked users should NOT be in the search

jaapjan:
## Problem
Blocked users are still in the search results (on All and on Users tab). This is happening because the change is in the User entity while the Profile entity is being indexed.

## Solution
When the user entity is updated, let's trigger a re-index of the related Profile entity. In some cases we need to add or remove the Profile from the index.

## Issue tracker
- https://www.drupal.org/project/social/issues/2992979
-See old (related) PR: https://github.com/goalgorilla/open_social/pull/854

## HTT
- [x] Login as SM
- [x] Block an active user
- [x] Notice that the user is still in the search results (also after running cron)
- [x] Clear all the indexes and re-index
- [x] Notice the user is not in results anymore
- [x] Go to this branch
- [x] run update script and cron and notice all is good
- [x] Make the user active
- [x] Check the search and notice the user is in the search again.
- [x] Block the user and notice the user is not in the search results anymore.
- [x] Also try the bulk operations on the `/admin/people` overview

## Release notes
Blocked users were still showing up in search results. When a user is blocked they should never be in the search results.

### DIFF
--- a/modules/social_features/social_search/social_search.install
+++ b/modules/social_features/social_search/social_search.install
@@ -297,3 +297,21 @@ function social_search_update_8109() {
     }
   }
 }
+
+/**
+ * Trigger a search_api re-index.
+ */
+function social_search_update_8110() {
+  $indexes = [
+    'social_all',
+    'social_users',
+  ];
+
+  foreach ($indexes as $index_id) {
+    $index = Index::load($index_id);
+    if ($index->status()) {
+      $index->clear();
+      $index->reindex();
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


jochemvn:
@jaapjan, I think that, since this feature was broken we should loop over all blocked users and trigger a re-index on the related profiles in an update hook. 😞

jochemvn:
@jaapjan Added two test steps here.. maybe take a quick look and merge 😇

PR #1010, opened 2018-08-22
Fix race condition in group edit content test

Kingdutch:
## Problem
There is currently a race condition in the group-edit-content-in-group.feature between selecting the discussion and the AJAX request for changing a group which could undo that selection.

## Solution
To fix it the order is changed so the topic type is selected first and is included in the pre-AJAX state so the race condition is avoided.

## Issue tracker
- https://www.drupal.org/project/social/issues/2994271

## HTT
- [x] No test behaviour should have changed.


robertragas:
+1 , saw it having trouble when this fix is not applied.

Kingdutch:
Trying to trick Travis 🤔

PR #1011, opened 2018-08-22
Issue #2994272 by Kingdutch: Increase waiting time for AJAX requests …

Kingdutch:
…in Behat tests

## Problem
Paragraphs can be quite slow to load by AJAX. Especially with a lot of elements such as what happens in some landing page tests. This can cause the 5 second timeout for the iWaitForAJAX step to be insufficient.

## Solution
To solve this we extend the time to 20 seconds which should be enough to let the AJAX requests pass even in slower conditions. Subjective tests show it takes about 10-15 seconds so this should provide a safe margin.

## Issue tracker
- https://www.drupal.org/project/social/issues/2994272

## HTT
- [ ] No test behaviour should change in this commit. However, some tests (e.g. landing pages) should be more reliable.

### DIFF
--- a/tests/behat/features/bootstrap/SocialMinkContext.php
+++ b/tests/behat/features/bootstrap/SocialMinkContext.php
@@ -152,5 +152,42 @@ class SocialMinkContext extends MinkContext {
       }
     }
   }
+  
+  
+  /**
+   * Wait for AJAX to finish.
+   *
+   * Overwrites the default iWaitForAjaxToFinish step to increase the time-out to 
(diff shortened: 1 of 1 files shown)


Kingdutch:
Although I agree with Robert's comment that this in theory allows unoptimzed code get into the distro I think we should separate performance requirements from functional testing. Especially since the current limit of 5s can be broken due to CI server hickups and our current way to fix that is to simply rerun the tests a few times until it's fast enough.

Kingdutch:
Trying to trick travis into downloading the composer package for this PR 🤔

PR #1015, opened 2018-08-29
Revert "FOR-101 by serhiy-chornobay: Fix brand color for Social Vote widget"

ronaldtebrake:
Reverts goalgorilla/open_social#1007


ronaldtebrake:
Needed to do this because base branch was 8.2 should be 8.3

PR #1018, opened 2018-09-04
Fix broken default profile image

suryanto:
## Problem
Custom default profile image is not shown when we set profile image on private files

## Solution
Add conditional check to see if current user has permission to default profile image

## Issue tracker
https://www.drupal.org/project/social/issues/2978581

## HTT
- [ ] Set a path for private files in settings.php
- [ ] Add a new default profile image
- [ ] Check some public post as anonymous user

## Release notes
Fix broken custom default image on private file and wsod for anonymous user.

### DIFF
--- a/modules/social_features/social_profile/social_profile.module
+++ b/modules/social_features/social_profile/social_profile.module
@@ -489,7 +489,6 @@ function social_profile_profile_view_alter(array &$build, EntityInterface $entit
         if (!$file->access('view', $current_user)) {
           // Load default data.
           $replacement_data = social_profile_get_default_image();
-          $default_profile = FALSE;
           // Check if we have default profile image
           if (!empty($replacement_data['id'])) {
             $default = \Drupal\file\Entity\File::load($replacement_data['id']);
--- a/modules/social_features/social_profile/social_profile.module
+++ b/modules/social_features/social_profile/social_profile.module
@@ -489,14 +489,23 @@ function social_profile_profile_view_alter(array &$build, EntityInterface $entit
         if (!$file->access('view', $current_user)) {
           // Load default data.
           $replacement_data = social_profile_get_default_image();
-          /** @var \Drupal\image\Plugin\Field\FieldType\ImageItem $imgitem */
-          $imgitem = $build['field_profile_image'][0]['#item'];
-          // Time to override the data that going to be rendered.
-          $imgitem->set('target_id', $replacement_data['id']);
-          $imgitem->set('width', $replacement_data['width']);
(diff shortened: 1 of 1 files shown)


modules/social_features/social_profile/social_profile.module
```diff
@@ -489,14 +489,23 @@ function social_profile_profile_view_alter(array &$build, EntityInterface $entit
...
           $replacement_data = social_profile_get_default_image();
-          /** @var \Drupal\image\Plugin\Field\FieldType\ImageItem $imgitem */
-          $imgitem = $build['field_profile_image'][0]['#item'];
-          // Time to override the data that going to be rendered.
-          $imgitem->set('target_id', $replacement_data['id']);
-          $imgitem->set('width', $replacement_data['width']);
-          $imgitem->set('height', $replacement_data['height']);
-          // Put replacement data back in the object that's about to be built.
-          $build['field_profile_image'][0]['#item'] = $imgitem;
+          $default_profile = FALSE;
```
robertragas: What does this do?

robertragas:
It has some coding standard errors and I don't really understand what you are trying to do. You are loading the default profile image and doing another permission check on that. Anonymous users don't have this permission and it will still fail.

This is at least what is happening in my local environment.

suryanto:
I cleaned up the code a bit. I am trying to add code that check if visitor have access to the default profile image. Because if you set the profile image field to use private files then change the default image profile. Anonymous users do not have permission to see the default profile image on private files. It will show WSOD screen for anonymous user.

This code changes fixed the WSOD issue for anonymous users. Because it will only show the default image if the user has permission to it, otherwise it will not show any profile image.

jochemvn:
Instead of changing this code, we might need to take a look at what is going on here: https://www.drupal.org/node/2107455

jaapjan:
Hey there, can anyone tell me what is the status here? Do we need to implement a custom solution or rather wait/work on the Drupal Core issue?

robertragas:
I think we will need to implement a custom solution for this to avoid us been stuck with unused permissions because the core issue seems to be heading towards a solution without a new permission.

jaapjan:
OK, then if I understand correctly this PR will prevent 404 for the default profile image. But it will not actually solve the issue that the default image is a private file which AN does not have access to.

What happens when the default image is in the public files directory, will the anonymous user still have access to it?

jaapjan:
Closing the PR and marking as needs work, because it seems it does not solve the issue. Although the changes in the PR do appear to be an improvement (will prevent 404) it won't solve the entire issue. Feel free to re-open if I'm in error.

PR #1019, opened 2018-09-05
#2996728 by ronaldtebrake: Cannot use isset() on the result of a func…

ronaldtebrake:
## Problem
For some hosting providers (?) People get a Fatal error:  `Cannot use isset() on the result of a function call (you can use "null !== func()" instead)` in `/home/public/profiles/contrib/social/modules/custom/mentions/mentions.module on line 134`

## Solution
According to [http://php.net/manual/en/function.isset.php](http://php.net/manual/en/function.isset.php)
`isset()` only works with variables as passing anything else will result in a parse error. For checking if constants are set use the defined() function.. So we make sure we only check variables.

## Issue tracker
- [https://www.drupal.org/project/social/issues/2996728](https://www.drupal.org/project/social/issues/2996728)

## HTT to be updated.
- [ ] Notice it doesn't work
- [ ] Checkout to this branch

## Release notes
TBD

### DIFF
--- a/modules/custom/mentions/mentions.module
+++ b/modules/custom/mentions/mentions.module
@@ -129,10 +129,11 @@ function mentions_crud_update($type, $mentions, $id, $author) {
     ->execute();
   foreach ($mention_ids as $mention) {
     $entity = $mentions_storage->load($mention);
+    $uid = $entity->get('uid')->getValue();
 
     // Make sure the uid value is available.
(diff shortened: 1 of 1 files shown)


bramtenhove:
@ronaldtebrake I think the tests fail because your branch name `feature/#2996728-mentions-module-fatal-error` has a `#` in it.

PR #1023, opened 2018-09-12
Issue #2998991 by ribel: Add default settings for anonymous event enrollment

ribel:
## Problem
Some users would like that anonymous enrollment for any event with public visibility to be enabled by default.

## Solution
Add new settings for Anonymous event enrollment and allow SM to:

1. Set the default value of the field.
2. Allow event managers (/event authors) to change the setting for their event.

## Issue tracker
- https://www.drupal.org/project/social/issues/2998991
- https://jira.goalgorilla.com/browse/OSSUPPORT-1246

## HTT
- [ ] Checkout to this branch and run db updates
- [ ] Login as SM and go to admin/config/opensocial/event-an-enroll
- [ ] Notice new settings
- [ ] Check _Allow anonymous users to enrol in new events_ by default
- [ ] Create new public event, see that AN Enroll is enabled by default
- [ ] Uncheck _Allow author of an event to change if anonymous users can enrol to the event_ in setiings
- [ ] Create new public event, and see that AN Enroll is disabled for editing

## Release notes
In the 3.0 release we added a feature to allow users to enroll for events without creating an account. In this release we give you more flexibility with a side-wide setting to determine whether this feature is enabled or disabled for new events. You can also determine whether a user is able to overwrite this site-wide setting on a per-event basis or whether it's fixed for all events.


robertragas:
Restarted build

modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.install
```diff
@@ -89,8 +89,8 @@ function _social_event_an_enroll_fix_blocks() {
  */
 function social_event_an_enroll_update_8001() {
   $config = \Drupal::service('config.factory')->getEditable('social_event_an_enroll.settings');
-  $config->set('event_an_enroll_email_subject', "You are enrolled in the event [event name]");
```
bramtenhove: @ribel If I'm correct the `[event name]` tokens did not work right?
I guess now we will now have to supply an update hook which updates these tokens to the new version. Not just setting a new default but actually updating the active value.

Any reason why these tokens were changed to `[node:title]` and `[social_event_an_enroll:enrolled_event]`? They do not seem very understandable for non-technical site managers.

bramtenhove: Clear!

Imho I think adding a custom token will be much more useful for site managers.

Kingdutch: I think that's outside of the scope of this issue though. Could you create a follow-up? :)

bramtenhove: Agreed, but we are changing it in this feature right?

Kingdutch: No the different tokens are already committed but the documentation string wasn't updated. That's what's being fixed here. So basically this feature also includes a bugfix. Simplifying the tokens further would be a new feature :)

Kingdutch:
We may want to add a test that tests AN enrolment.

It could look like the below test but still needs to check the box toallow An enrolment.

```
  @AN
  Scenario: Successfully enroll to an event as anonymous user
    Given "event_types" terms:
      | name    |
      | Webinar |
      | Other   |
    Given I am logged in as an "authenticated user"
    When I am viewing my event:
      | title                    | Anonymous event Enrollment |
      | field_event_date         | +3 days                        |
      | field_event_date_end     | +4 days                        |
      | field_content_visibility | public                         |
      | field_event_type         | Webinar                        |
    Given I logout
    And I open the "event" node with title "Anonymous event Enrollment"

    When I click "Enroll"

    Then I should see "Enroll in Anonymous event Enrollment"
    And I should see "Sign up as guest"

    When I click "Sign up as guest"
    And I fill in "Anonymous" for "First name"
    And I fill in "User" for "Last name"
    And I fill in "anonymous+user@pachamama.localhost" for "Email address"
    And I press "Enroll in event"
    Then I should see "You have successfully enrolled to this event."
```

jaapjan:
Could you also write the release notes?

Kingdutch:
The following texts are added or modified in this pull request.

AN Enrollment E-mail 
Subject: `You are enrolled in the event [node:title]`
Body: `You have been enrolled in the event [node:title]. \nYou can cancel your enrollment anytime using this link [social_event_an_enroll:enrolled_event]`
 SM Setting Default AN Enroll Value 
Title: `Allow anonymous users to enrol in new events by default`
Description: `When this setting is enabled by default it will allow anonymous users to enrol in newly created events. The author will still be able to change this if the author has the permission.`

SM Setting AN Enroll edit by users:
Title: `Allow author of an event to change if anonymous users can enrol to the event`
Description: `This will give authors the permission to set if anonymous users can enrol to their event.`

SM Setting AN Enrollment Notification:
Title: `Notify user after anonymous enrollment by email`


SM Setting E-mail Message Token Help: 
`To add event name and link use tokens: [node:title], [social_event_an_enroll:enrolled_event]`

NatashaSchon:
@Kingdutch asked me to review the following texts: 

You are enrolled in the event. 

You have been enrolled in the event [node:title]. You can cancel your enrollment anytime using the following link: [social_event_an_enroll:enrolled_event

When this setting is enabled, anonymous users are allowed to enroll in newly created events by default. 

The author will still be able to change this if the author has the permission.

Allow the author of an event to decide whether anonymous users can or cannot enroll in the event.

This will provide authors with the permission to decide whether anonymous users can enroll in their event.

Notify user by email after anonymous enrollment

To add the event name and link, use the following tokens: [node:title], [social_event_an_enroll:enrolled_event]

Kingdutch:
@bramtenhove MOSS seems to be having a bit of trouble with this. The event-an-enroll.feature tests aren't run at all on SaaS. For other projects it seems to be running an outdated version of the test which is causing issues. :|

I've retriggered everything. The texts still need to be updated.

Kingdutch:
I had a discussion with @NatashaSchon 

We ended up on using `event organiser(s)` because this was in line
with the functionality that could be provided by the Event Organiser
module. It was most fitting for the first setting (enabling the feature)
and was then consistently applied to the other settings.

Kingdutch:
@jaapjan @bramtenhove 
The following commit may be interesting to look at because it provides a hook for Enterprise sites to re-use tests without overwriting them when they have additional fields that need filling: https://github.com/goalgorilla/open_social/pull/1023/commits/cd3ee7d686ef0e4cc944b7e46668972c7507e8d5

Check the PAC fix-distro branch to see how it's used :)

bramtenhove:
@Kingdutch are you able to change the texts with regard to "anonymous user" this afternoon still?

bramtenhove:
FYI: failing Moss test is being fixed separately.

PR #1024, opened 2018-09-13
Issue #2999042 by Omar Alahmed: Undefined index: #plugin_id

bramtenhove:
## Problem
There are many Undefined index: #plugin_id in social_event_block_view_alter() and Undefined index: #plugin_id in social_topic_block_view_alter() (line 124

Which affects the server log and performance, so I think we should check whether the plugin_id is set or not.

## Solution
Add isset() to check for the index.

## Issue tracker
- https://www.drupal.org/project/social/issues/2999042
- https://jira.goalgorilla.com/browse/DS-6081

## HTT
- [ ] Browse through the site to see that there are no more notices
- [ ] Check the events overview and topics page on a users profile to see that these still work as expected

## Release notes
In some cases an undefined index notice was thrown for event overview block and topics profile block. An extra check was added to prevent these notices.


bramtenhove:
This should also be ported back to 2.x

bramtenhove:
Cherry-picked to 2.x

PR #1022, opened 2018-09-10
DS-6042: update Drupal to 8.6.0 and remove patch for 736066

jaapjan:
## Problem
There is a new Drupal version!

## Solution
Let's update to the newest!

For this I needed to make changes in:
- Patches used in the project.
- Injecting the correct services in PostForm.php.
- Update to dynamic entity reference 1.6.
- Use PHPUnit 6.x (also made change in drupal_social composer.json and enterprise projects in Project Moss) and changing some code in the unit test.
- Behat test changes: mostly waiting for AJAX.
- Getting feature status clean.
- Caching disabled for the anonymous event enrollment.
- Changing the title of the url path settings fieldset
- Use open Social 2.0.0 in install_update job.
- Using latest version of data_policy https://www.drupal.org/project/data_policy/issues/2999425 to prevent conflict with dynamic page cache.

## Issue tracker
- DS-6042
- https://www.drupal.org/project/social/issues/2998671

## HTT
- [x] Check out the code changes
- [x] Check that all the patches apply
- [x] Make sure changes are in the .make files and the composer files.
- [x] Check all the tests.
- [x] Check change records and release notes of D8.6.0 and D8.6.1.
- [ ] Ask me about any questions you might have.

## Release notes
-

PR #1025, opened 2018-09-13
Issue #2999361 by frankgraave: improve group-edit-content-in-group.feature

frankgraave:
## Problem
The **group-edit-content-in-group.feature** has a part where there's a topic being moved from a group to " - None - " and then saves the node. But this can cause false positives.

## Solution
Add a **_And I wait for AJAX to finish_** step just before it tries to save the node.

## Issue tracker
- https://www.drupal.org/project/social/issues/2999361

## HTT
- [ ] Make sure the tests are green
- [ ] Run the test on your local machine and see it turning green


jaapjan:
This was fixed already in: https://github.com/goalgorilla/open_social/pull/1022/files#diff-7c8a67d6e87c63db348e5dc723748935

PR #1029, opened 2018-09-19
2938410: See more threads in Social Private Message Inbox

jochemvn:
## Problem
When the inbox of a person has more than 30 threads, only the first 30 are shown

## Solution
Replace the inbox block we use with a view that has pagination

## Issue tracker
- https://www.drupal.org/project/social/issues/2938410

## HTT
- [ ] Install fresh on 8.x-2.x and create a LOT of threads (you could set the 30 to a lower number for test purposes)
- [ ] Notice that not all threads are shown (only the newest)
- [ ] Checkout to this branch and run the update script
- [ ] Go to your inbox and notice it looks the same, apart from the pagination at the bottom.

## Release notes
Currently the inbox of private message threads was limited to 30 threads. Meaning that only the most recent 30 used threads are shown. This inbox block has been replaced by a view that uses pagination.


bramtenhove:
Moss is all false positives.

PR #1030, opened 2018-09-25
3002123: Remove setting that adds 'not verified' string to Anonymous accounts in comments

jochemvn:
## Problem
Whenever a user deletes his/her account and leaves his/her content in the platform. On most placed the username is set to Anonymous. Only on comments the username is set to 'Anonymous (not verified)'. This is confusing for users and should be set to 'Anonymous' only.

## Solution
Turn off the setting, so 'not verified' is no longer displayed.

## Issue tracker
- https://www.drupal.org/project/social/issues/3002123

## HTT
- [x] Create a new user, make some comments and delete the user (not its content)
- [x] Notice that in the stream you now see stuff like: "Anonymous (not verified)" in comments
- [x] Checkout to this branch
- [x] run update script and see that the 'not verified' part is now gone
- [x] Try the same on a fresh install and notice it now works.

## Release notes
Whenever a user deleted his/her account, the name would be changed to Anonymous. In some cases it would show 'Anonymous (not verified)'. We have made this more consistent so the '(not verified)' string will not be shown anymore.


jaapjan:
Code works and it solves the issue. If Project Moss (SHNL) is green it can be merged.

bramtenhove:
This is also cherry-picked in 3.x

PR #1032, opened 2018-09-27
#2992332 by ronaldtebrake: Default visibility fix for Open Groups

ronaldtebrake:
## Problem
In PR #994 a default statement was added to make sure custom group types have all the visibility settings giving distro users more flexibility. However as you can see in the switch statement open_group wasn't covered in this scenario. This meant that "public" became an option in the visibility selector for Posts in streams for open groups and "public" was the default value. This is something we don't allow. 

## Solution
By adding a switch statement we provide the available options for posts in open groups. Only community is allowed in this case. We also made a data alteration that ensures all the posts created, that have visibility "Public" become "Community". By checking the timestamp of when #994 landed in the Distro we ensure not to many posts are loaded and changed.  

## Issue tracker
- https://www.drupal.org/project/social/issues/2992332

## HTT
- [x] Login as user X and go to / join an open group
- [x] See that on the stream page you can now add a Public post and Public is the default selection
- [x] See that when you save your public post, you go to /explore as AN user you actually see this post popping up in the stream
- [x] Checkout this branch, run the update, see that now your post is put back to community
- [x] Edit your post, see that the disabled visibility selector shows "Community" as well
- [x] See that your post is still shown in the Group stream

## Release notes
t.b.d still working on behat as well but lets see what our tests are doing in the meantime :)

### DIFF
--- a/modules/social_features/social_group/social_group.install
+++ b/modules/social_features/social_group/social_group.install
@@ -399,10 +399,11 @@ function social_group_update_8014() {
 }
 
 /**
- * DATA ALTERATION:
+ * DATA ALTERATION.
+ *
  * Update all post visibility for public posts placed in a group with group type
  * open_group set public to community.
- * See: https://www.drupal.org/project/social/issues/2992332#comment-12790905
+ * See: https://www.drupal.org/project/social/issues/2992332#comment-12790905.
  */
 function social_group_update_8301(&$sandbox) {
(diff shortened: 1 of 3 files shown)


jaapjan:
I approve this PR, good job!

bramtenhove:
@ronaldtebrake @jaapjan have you also checked if this is a also a possible issue for nodes? There is a visibility selector there as well.

PR #1034, opened 2018-10-01
IE explorer svg issue

Crazylord:
## Problem
auto: height for SVG brocked styles.

## Solution
Remove height: auto for SVG 

## Issue tracker
- https://jira.goalgorilla.com/browse/ECI-832

## HTT
- [ ] Checkout to this branch
- [ ] Login as normal user and try to add a reply in IE11 browser
- [ ] Check if height for SVG works okay

### DIFF
--- a/themes/socialbase/assets/css/base.css
+++ b/themes/socialbase/assets/css/base.css
@@ -334,8 +334,7 @@ img {
   vertical-align: middle;
 }
 
-img,
-svg {
+img {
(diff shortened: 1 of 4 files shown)


jochemvn:
Hey @Crazylord , Since this is a bug and not a feature request, the base branch should be 8.x-2.x and not 8.x-3.x. Could you change that and check if all is still good?

PR #1050, opened 2018-10-07
Alternative frontpage respect destination on login

slowflyer:
## Problem
When  using alternative frontpage users are redirceted after login to the homepage specified for LU, no matter if a destination parameter is passed on the URL.
When user has no active session and receives a notification with a link to a node or group or what ever, he is always redirected to the frontpage after login.

## Solution
Respect destination parameter

## Issue tracker
- https://www.drupal.org/project/social/issues/3004846

## HTT
- [x ] Check out the code changes
- [x ] Login and copy the URL of a group or node, than logout, try to open the URL you copied. You'll be redirected to the login form. You'll see the correct destiantion parameter added to your URL
- [x ] After login, you're redirecetd to the frontpage specified for LU not the URL specified with the destination parameter
- [x ] Checkout to this branch
- [x ] Logout and try to open the URL copied again. Now you're redirect to the page you want to open.

## Release notes
Was really hard to get the URL with destination parameter. Only appeared in the Server HTTP_Referer variable


jaapjan:
@slowflyer 

Looking at your solutions I'm afraid for side-effects. Since it doesn't only execute the code on `user/login` but on all the frontpages with a destination parameter as well. What do you think about this solution? It also solves the issue. Doesn't appear to have any side-effects and it looks a bit cleaner. I don't know if that redirect should actually be in the submitForm method though and if we can safely place this in the form.

```
diff --git a/modules/social_features/social_user/src/Form/SocialUserLoginForm.php b/modules/social_features/social_user/src/Form/SocialUserLoginForm.php
index a0be543f1..d44f25eca 100644
--- a/modules/social_features/social_user/src/Form/SocialUserLoginForm.php
+++ b/modules/social_features/social_user/src/Form/SocialUserLoginForm.php
@@ -77,14 +77,6 @@ class SocialUserLoginForm extends UserLoginForm {
 
     $this->renderer->addCacheableDependency($form, $config);
 
-    return $form;
-  }
-
-  /**
-   * {@inheritdoc}
-   */
-  public function submitForm(array &$form, FormStateInterface $form_state) {
-    $account = $this->userStorage->load($form_state->get('uid'));
     // A destination was set, probably on an exception controller,.
     // @TODO: Add validation if route exists.
     if (!$this->getRequest()->request->has('destination')) {
@@ -94,6 +86,15 @@ class SocialUserLoginForm extends UserLoginForm {
       $this->getRequest()->query->set('destination', $this->getRequest()->request->get('destination'));
     }
 
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $account = $this->userStorage->load($form_state->get('uid'));
+
     user_login_finalize($account);
   }
 
```

slowflyer:
@jaapjan: I'm ok with any smarter solution solving the issue. And if it would also be able to add <body class="**path-frontpage** 
to the alternative frontpage fo LU, it would be superior.

jaapjan:
Ah, thanks. I've added the changes. Let's see what the tests will do. Actually I'm not 100% sure still if this is the way to go. But perhaps someone else can also chip in after the tests have run.

Regarding the body class. I think we should handle that in a different issue.

ronaldtebrake:
Hi guys,

Unfortunately this causes some regression. When a user now logs in without a destination the settings for Alternative Frontpage Settings aren't respected anymore and the user is redirected to the user/x/stream page.

slowflyer:
Seems that this issue and this pull request can be closed without any modifications.
Just tried to find a better solution and figured out, that everything works now as expected, without any updates.

Not sure why, may be last Core update did it.

slowflyer:
@jaapjan anything I can do to satisfy the moss — OS dependant checker?

jaapjan:
I'm not sure what to do to satisfy Moss to be honest.. especially old (reopened) pullrequests are hard to process for project moss and tend to increase some false-negatives. I'll look into it next week!

modules/social_features/social_user/src/Form/SocialUserLoginForm.php
```diff
@@ -117,13 +117,12 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
     $account = $this->userStorage->load($form_state->get('uid'));
     // A destination was set, probably on an exception controller,.
     // @TODO: Add validation if route exists.
-    if (!$this->getRequest()->request->has('destination')) {
+    if (!$this->getRequest()->server->has('REDIRECT_QUERY_STRING')) {
       $form_state->setRedirect('<front>');
     }
     else {
-      $this->getRequest()->query->set('destination', $this->getRequest()->request->get('destination'));
+      $this->getRequest()->query->set('destination', parse_url(urldecode(str_replace('destination=', '', $this->getRequest()->server->get('REDIRECT_QUERY_STRING'))), PHP_URL_PATH));
```
jaapjan: I'm wondering: are there better ways to do this?

The way we retrieve the destination right now doesn't seem to be the way to go as it's always better to use Drupal's functions for this. Why doesn't `$this->getRequest()->request->get('destination')` work? Perhaps when we realise this we can find a different solution and still use that method to retrieve the current destination.

slowflyer: When debugging the stuff I had no values in:
$this->getRequest()->**request**->get('destination') 
that's why I switched over from the request-scope to the server-scope
$this->getRequest()->**server**->get('REDIRECT_QUERY_STRING')

slowflyer: What do you think about the latest changes? Transport destination information in a hidden form field. This keeps other stuff simple and easy to understand.

Kingdutch:
I don't think we should add a form field to the login form to solve this, we've had a lot of caching problems with that approach in this form.

I've also posted some questions in the Drupal.org issue: https://www.drupal.org/project/social/issues/3004846#comment-13080548

jaapjan:
See feedback from @Kingdutch

jaapjan:
Closing PR for now. Issue is still on `needs work` until feedback from Kingdutch is addressed

PR #1057, opened 2018-10-15
Issue #3005537 by Kingdutch: Remove hero from "leave group" page

foxtrotcharlie:
This is my first issue I'm taking on with this project so please let me know if I've missed anything in terms of the workflow, and pull request information below. I also don't know if you require an install hook or some other automated way to import the updated feature configuration, other than manually importing the feature with drush, as indicated below.
## Problem
The hero can push simple action forms (such as join group/leave group) below the fold on smaller screens. This makes it more difficult for users to perform the action. Because of this we've decided to remove the hero on the join group screen. However, this causes the experience between leaving a group and joining a group to be inconsistent, as the hero still appears on the leave group screen.

## Solution
The "Leave group" page path has been removed from the group hero block page visibility paths. This was done in the social_group feature.
## Issue tracker
- https://www.drupal.org/project/social/issues/3005537

## HTT
- [x] Login as a user who has permission to join groups, join a group, and then leave a group. On the "Leave group" confirmation page, the Hero block and image will appear. 
- [x] Checkout to this branch
- [x] Import the updated social group feature configuration, e.g. `drush fim social_group -y`
- [x] Clear the cache `drush cr`
- [x] Login as a user who has permission to join groups, join a group, and then leave a group. On the "Leave group" confirmation page, the Hero block and image will no longer appear.

### DIFF
--- a/modules/social_features/social_group/config/optional/block.block.socialblue_groupheroblock.yml
+++ b/modules/social_features/social_group/config/optional/block.block.socialblue_groupheroblock.yml
@@ -20,6 +20,6 @@ settings:
 visibility:
   request_path:
     id: request_path
-    pages: "/group/*/stream\r\n/group/*/about\r\n/group/*/events\r\n/group/*/topics\r\n/group/*/members\r\n/group/*/membership\r\n/group/*/leave\r\n/group/*/content/*"
+    pages: "/group/*/stream\r\n/group/*/about\r\n/group/*/events\r\n/group/*/topics\r\n/group/*/members\r\n/group/*/membership\r\n/group/*/content/*"
     negate: false
     context_mapping: {  }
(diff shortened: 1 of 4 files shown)


jaapjan:
Hi there, it seems the tests also need to be adjusted, see the Travis details. Marking as needs work for now.

foxtrotcharlie:
Thanks @jaapjan So I had a look and the tests that fail all include the following lines:

```
    And I click "Leave group"
    And I should see "Test open group" in the "Hero block"
```
So with the hero block removed from the "Leave group" page, the 
```
And I should see "Test open group" in the "Hero block"
``` 
test will always fail. So should I just delete that check, or do I need to replace it with something else? 

For the time being I am just going to remove that test - if you want anything to replace it, let me know what please. 

Do I need to create a new pull request with the updated code? Apologies for the questions, this is my first PR for this project.

foxtrotcharlie:
@jaapjan Although it's indicating "pending" to me, the Travis testing seems to have passed (https://travis-ci.org/goalgorilla/open_social/builds/444648963) but not the moss - I can't see the details for that so don't know how to fix. Let me know if there's a way for me to check that or if there's anything else I should do in that regard.

jaapjan:
I think Github is still slow with processing the webhooks. I'll look into changes needed for project Moss to pass.

PR #1064, opened 2018-10-24
DS-3621 by jaapjan: implement new label 'You are enrolled' for LU

jaapjan:
New PR based on changes made in #951 

NOTE: I did not (yet) check for compatibility with the AN event enrolment feature released in 3.x via PR #792

## Problem
In user testing, users are not clear what the label means. And on the event overview page of anther user, users thought that is about the user they are looking at. We should make this label clear that it is about the user him/herself. 

## Solution
Change the label from "Enrolled" to "You are enrolled". ## Issue tracker
- DS-3621

## HTT
- [x] Login as user X and enroll to an event
- [x] Check the event label on the button
- [x] Go to your profile and see the block on the right with upcoming events (same as homepage)
- [x] Go to event overview on your profile and verify your label
- [x] Go to event overview and verify label
- [x] Check changes in Behat

## Documentation

## Release notes
In user testing, users are not clear what the label means. And on the event overview page of anther user, users thought that is about the user they are looking at. We have changed the label "Enrolled" to "You are enrolled" to solve that problem.

### DIFF
--- a/modules/social_features/social_event/social_event.module
+++ b/modules/social_features/social_event/social_event.module
@@ -153,7 +153,7 @@ function social_event_node_view_alter(array &$build, EntityInterface $entity, En
       if ($current_enrollment_status === '1') {
         $build['enrolled'] = [
           '#type' => '#text_field',
-          '#markup' => t('Enrolled'),
+          '#markup' => t('You are enrolled'),
         ];
       }
     }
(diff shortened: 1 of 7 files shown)


jaapjan:
Created a story in our backlog to improve on this, support for AN Enrollment might cost some more time which is not planned for now.. Plan = to release together with label change for group member as well.

PR #1078, opened 2018-10-30
Issue #3010346 by truls1502: Add installer-paths to composer.json

truls1502:
## Problem & Solution
I thought that it would be great to have the standard modules, profiles, and themes stored automatically inside in their contrib-folder. So the new developers will learn from us to have the contribution modules under their module which we can improve our and their work regarding Drupal. My exact point to here is that we are helping to make a habit for us developers such as below:
```
modules/contrib/*
profiles/contrib/*
themes/contrib/*
```

If I am using `composer install`, these modules are installed in the normal way without contrib-folder.

## Issue tracker
- https://www.drupal.org/project/social/issues/3010346

## HTT
- [ ] Notice it doesn't work
- [ ] Checkout to this branch

## Release notes
<describe the release notes>


truls1502:
What do you think @goalgorilla/open-social or whoever? Alternatively, we should drop it? I need your feedback about it since it is your product. 😜

Kingdutch:
Hi @truls1502 ,

Thank you for the pull request. Unfortunately we believe that this configuration should be part of the project itself and not of a dependency. `drupal/social` (or `goalgorilla/open_social`) is meant to be installed as a dependency of a project, much like `drupal/core`. If you want to have a starting template for a project that has the install paths correctly configured then you're encouraged to install [`goalgorilla/social_template`](https://github.com/goalgorilla/social_template) instead.

~ Kingdutch

PR #1079, opened 2018-10-31
Update build-social.make

truls1502:
## Problem & Solution
Should not this social project be in 8.x-3.x?

## Issue tracker
- https://www.drupal.org/project/social/issues/3010741

## HTT
- [ ] Notice it doesn't work
- [ ] Checkout to this branch

## Release notes
<describe the release notes>


ronaldtebrake:
Hi Truls,

Thanks. I see it's working as it should be with the current make file at the moment, if you check out the package on d.org there is nothing wrong with the current package. Anyhow for consistency reasons I can agree this makes more sense.

truls1502:
Yes, it is working, @ronaldtebrake. But as I assume you agree with me that it looks more "logic", and also it can make us, whoever, less confuse even it working with the current make file. But of course, it is up to your staff, I only wanted to help to clean the code as possible as I can 🤓 

We can leave it, if there is some good reason to do that. If that, maybe a comment on top of the file might help us?

PR #1081, opened 2018-10-31
Issue #3010526 by ribel: Prefill Nickname together with First and Last name during social signup

ribel:
## Problem
In most cases, if you use the Nickname (provided by Social Profile Fields module) you usually want to populate it as well after signup with social networks.

## Solution
If Nickname is used prefill it, together with First and Last name.

## Issue tracker
- https://www.drupal.org/project/social/issues/3010526
- https://jira.goalgorilla.com/browse/HGT-26

## HTT
- [ ] Enable Social SSO and Social Profile Fields modules and 
- [ ] Enable Nickname field
- [ ] Try to signup using a social network (LinkedIn or Facebook)
- [ ] Notice First and Last Name was prefilled, and Nickname not
- [ ] Checkout to this branch
- [ ] Try to register again and notice that Nickname was prefilled together with First and Last name 

## Release notes
When the Nickname field is available and people sign up via Social Login, we now populate that field with the First and Last Name received from the social networks.

### DIFF
--- a/modules/social_features/social_sso/social_sso.module
+++ b/modules/social_features/social_sso/social_sso.module
@@ -14,8 +14,15 @@ use Drupal\social_auth_extra\UserManagerInterface;
  * Implements hook_social_auth_extra_profile_presave().
  */
 function social_sso_social_auth_extra_profile_presave(UserInterface $account, ProfileInterface $profile, AuthManagerInterface $auth_manager, UserManagerInterface $user_manager) {
-  $profile->get('field_profile_first_name')->setValue($auth_manager->getFirstName());
-  $profile->get('field_profile_last_name')->setValue($auth_manager->getLastName());
+  // If nickname is used prefill it, instead of First and Last name.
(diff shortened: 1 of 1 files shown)


jochemvn:
@ribel  After rethinking this a bit,  I think it's not just about the nickname field being enabled. That's just an addition to the existing situation. If the firstname/lastname are enabled they should still always be filled. The way it used to work should stay the same.. the only addition should be that IF the nickname is enabled, that field should be populated too.

bramtenhove:
@ribel, @jochemvn, I think in the case when all three fields (nickname, first and last name) we should not populate the nickname field. It's going to have duplicate information and that could be confusing for the user.

And maybe also good to think bout when there's the nickname field and just the first name, or the nickname field and just the last name. I suppose prefilling the nickname would then still not be the best choice.

ribel:
@jochemvn, @bramtenhove I have changed behaviour and description as you proposed.

bramtenhove:
Thanks @ribel for the changes. Situation described by @jochemvn is fixed.

PR #1082, opened 2018-10-31
by robertragas: make the suggestions amount in mentions configurable

robertragas:
Set default to 8.

## Problem
When creating mentions the full name is not respected and checking on either first or last name resulting in some people not being found. Since this is a complex problem a solution for now would be to be able to increase the suggestion amount.

## Solution
Create a new form field where people are able to alter the suggestion amount from 8 to another value.

## Issue tracker
- none

## HTT
- [ ] Make sure you have enough users
- [ ] Go to admin/config/people/mentions and change the default setting from 8 to another number
- [ ] Mention someone and see the amount of results is increased

## Release notes
We made it possible to alter the number of suggestions you get when mentioning a person so the chance the person you are searching for is within this list. 

You can alter this setting under the people mentions setting at  admin/config/people/mentions

### DIFF
--- a/modules/custom/mentions/config/schema/mentions.schema.yml
+++ b/modules/custom/mentions/config/schema/mentions.schema.yml
@@ -8,6 +8,9 @@ mentions.settings:
       sequence:
         type: string
         label: 'The entity type'
+    suggestions_amount:
+      type: number
+      label: 'Number of suggestions'
 
 mentions.mentions_type.*:
   type: config_entity
--- a/modules/custom/mentions/src/Form/MentionsSettingsForm.php
+++ b/modules/custom/mentions/src/Form/MentionsSettingsForm.php
@@ -77,6 +77,13 @@ class MentionsSettingsForm extends ConfigFormBase {
       '#description' => $this->t('Mentions entity will be created only for selected entity types.'),
     ];
 
+    $form['general']['suggestions_amount'] = [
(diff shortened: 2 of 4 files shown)


jochemvn:
@robertragas nice feature, however I'd like to see a minimum of 1 on the number field. Now people can put it on 0 or even a negative number. How about adding:
`#min`: Minimum value.
`#max`: Maximum value. (optional)
[See also](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Number.php/class/Number/8.2.x)

jaapjan:
Thanks guys, I've made changes to the widget and created a d.o issue. I think @robertragas still wants to improve this feature by adding a scrollbar to the suggestions which could allow the possibility to show even more suggestions (up to 100 I've set the max to right now, but can be more obviously).

robertragas:
@jaapjan I added styling for the scrollbar and also in a way that it's visible on mobile because default styling of scrollbars (on OSX) doesn't really indicate it's scrollable. 

For now it's also applied to the general autocomplete list so we can use it on multiple places making it more accessible / uniform. If you want it only for mentions though, then I would need to move the styling to the mentions.scss. So would like to hear your opinion.

jaapjan:
It works really well and I think it is mergeable already. However your suggestion to make the scrollbar background transparant could still be a good addition. Also I haven't reviewed the css in much detail.

PR #1073, opened 2018-10-29
Issue #3010064 by truls1502: Update paragraphs 1.2 to 1.5

truls1502:
## Problem & Solution
Updated _paragraphs_ 1.3 to 1.4. 
_Bugs are listed below or a full list on the link of the Drupal issue._
- https://www.drupal.org/project/paragraphs/releases/8.x-1.3
- https://www.drupal.org/project/paragraphs/releases/8.x-1.4
- https://www.drupal.org/project/paragraphs/releases/8.x-1.5

## Issue tracker
- https://www.drupal.org/project/social/issues/3010064

## HTT
- [ ] Check out the code changes
- [ ] Login as user X and do this and this and that
- [ ] Notice it doesn't work
- [ ] Checkout to this branch
- [ ] Try this and this and that and see that it works now

## Release notes
<describe the release notes>

PR #1084, opened 2018-11-01
Issue #3010064 by truls1502: Update paragraphs 1.2 to 1.5

ronaldtebrake:
## Problem
See #1073 

Unfortunately the upstream push went wrong so couldnt install it correctly locally. Also hopefully this will trigger MOSS. 

Thanks @truls1502 for your work!


ronaldtebrake:
ECI DEMO + PR are green all we needed to know :)

PR #1085, opened 2018-11-01
Issue #3010784: Remove unnecessary Node checks from social_core_menu_local_tasks_alter

Kingdutch:
<h2>Problem</h2>

<code>social_core_menu_local_tasks_alter</code> removes the <code>entity.node.edit_form</code> tab to force editing through the floating edit button. Before doing this it checks whether we're currently viewing a node of any valid node type. This logic isn't needed because if we weren't viewing a node then we wouldn't have the <code>entity.node.edit_form</code> task present.

<h2>Solution</h2>
Just remove the edit form tab if it's set (less computations => more performance!)

## Issue tracker
- https://www.drupal.org/project/social/issues/3010784

## HTT

### DIFF
--- a/modules/social_features/social_core/social_core.module
+++ b/modules/social_features/social_core/social_core.module
@@ -7,7 +7,6 @@
 
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\node\Entity\Node;
-use Drupal\node\Entity\NodeType;
 use Drupal\node\NodeInterface;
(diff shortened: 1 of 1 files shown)


jochemvn:
@Kingdutch I think we can consider this a bug and try to get it in 3.x still?

jaapjan:
Agree. When Project Moss (last project) is passing we should be able to merge this one.

PR #1087, opened 2018-11-02
SHN-226: Quickjoin groups

jochemvn:
## Problem
When joining a group there's an unnecessary threshold of a 2 step verification. This is not always needed. Also CTA links to joining a group create a threshold with the confirmation step

## Solution
On a group type basis it's now possible to override the 'Join' link to join a group instantly. Also sitebuilder can now make CTA links (EG: in landing pages) like: `/group/1/quickjoin` in order to allow users to join a group with a single click.

## Extra description
In order to make this feature work I made a change to the permissions of an outsider of closed groups. The issue was that the `join group` permission was enabled on those groups and that the code made sure we weren't able to join closed groups anyway. This has changed a but. In practice this doesn't make any difference, apart from the fact that the permissions are now in fact correct and usable.  There should not be any differences in the way closed group act.

## Issue tracker
- https://jira.goalgorilla.com/browse/SHN-226

## HTT
- [ ] Revert features
- [ ] Enabled the module
- [ ] Login as a SM and go to the settings page `configuration` > `open social settings` > `group quickjoin settings`
- [ ] Enable the feature
- [ ] Notice the (joinable) groups appear
- [ ] Save without selecting any groups
- [ ] As a LU try to join a group and notice there's still a 2 step wizard
- [ ] Try the manual link `/group/3/quickjoin` on the group you just joined
- [ ] You get a message you're already a member
- [ ] Try it for a group you're not yet a member of and notice you are immediately a member and you get feedback about this
- [ ] Try it on a closed group and notice that it doesn't work
- [ ] As SM Enable it for open and public group
- [ ] As LU join a public/open group from the join button
- [ ] Notice you're immediately added to the group

## Release notes
By enabling the social_group_quickjoin module, a site manager get the possibility to create CTA link for landing pages to allow users to join (non-closed) groups with a single click. Also a site manager has the possibility to override the 'Join' button on those groups and make sure people can join with a single click.

## Release notes extra
During the development of this feature I notice that the permission 'join group' for closed group was turned on for outsiders. This is incorrect, since outsiders cannot join closed groups. This was taken care of in code, so no real issue. The permission has now been set correct. This is mainly a developers change. It means that as a developer you can actually use that permission now to check group join permissions.

### DIFF
--- a/modules/social_features/social_group/modules/social_group_quickjoin/src/EventSubscriber/RedirectSubscriber.php
+++ b/modules/social_features/social_group/modules/social_group_quickjoin/src/EventSubscriber/RedirectSubscriber.php
@@ -107,7 +107,7 @@ class RedirectSubscriber implements EventSubscriberInterface {
       'group' => $group->id(),
     ])->toString();
 
-    // Redirect ot quickjoin.
+    // Redirect to quickjoin.
     $event->setResponse(new RedirectResponse($url));
   }
 
--- a/modules/social_features/social_group/modules/social_group_quickjoin/src/EventSubscriber/RedirectSubscriber.php
+++ b/modules/social_features/social_group/modules/social_group_quickjoin/src/EventSubscriber/RedirectSubscriber.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Drupal\social_group_quickjoin_quickjoin\EventSubscriber;
+
+use Drupal\Core\Routing\CurrentRouteMatch;
+use Drupal\Core\Session\AccountProxy;
+use Drupal\group\Entity\Group;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Class RedirectSubscriber.
+ *
+ * @package Drupal\social_group_quickjoin\EventSubscriber
+ */
+class RedirectSubscriber implements EventSubscriberInterface {
+
+  /**
+   * The current route.
+   *
+   * @var \Drupal\Core\Routing\CurrentRouteMatch
+   */
+  protected $currentRoute;
+
+  /**
+   * The current user.
+   *
+   * @var \Drupal\Core\Session\AccountProxy
+   */
+  protected $currentUser;
+
+  /**
+   * Redirectsubscriber construct.
+   *
+   * @param \Drupal\Core\Routing\CurrentRouteMatch $route_match
+   *   The current route.
+   * @param \Drupal\Core\Session\AccountProxy $current_user
+   *   The current user.
+   */
+  public function __construct(CurrentRouteMatch $route_match, AccountProxy $current_user) {
+    $this->currentRoute = $route_match;
+    $this->currentUser = $current_user;
+  }
+
+  /**
+   * Get the request events.
+   *
+   * @return mixed
+   *   Returns request events.
+   */
+  public static function getSubscribedEvents() {
+    $events[KernelEvents::REQUEST][] = ['groupQuickJoin'];
+    return $events;
+  }
+
+  /**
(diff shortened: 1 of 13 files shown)


bramtenhove:
@NatashaSchon we have the following new texts in this pull request. Care to take a look?

> You're already a member of this group.

> You don't have access to join this group.

> You've been added to this group.

> Enabling this feature gives site builders the possibility to create group quickjoin links (EG: /group/1/quickjoin). Furthermore it's possible to skip the confirmation step on a group type basis.

> Allow users to join groups with a singe click.

> Allow users to skip the confirmation step when joining a group of type @grouptype

modules/social_features/social_group/modules/social_group_quickjoin/src/Form/SocialGroupQuickjoinSettings.php
```diff
@@ -0,0 +1,125 @@
...
+
+    $form['help'] = [
+      '#type' => 'item',
+      '#markup' => $this->t("Enabling this feature gives site builders the possibility to create group quickjoin links (EG: /group/1/quickjoin). Furthermore it's possible to skip the confirmation step on a group type basis."),
+    ];
+
+    $form['social_group_quickjoin_enabled'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t("Enable quickjoin"),
+      '#description' => $this->t("Allow users to join groups with a singe click."),
```
bramtenhove: @jochemvn there's a typo in this string.

```suggestion
      '#description' => $this->t("Allow users to join groups with a single click."),
```

modules/social_features/social_group/modules/social_group_quickjoin/src/EventSubscriber/RedirectSubscriber.php
```diff
@@ -0,0 +1,114 @@
...
+    if ($settings->get('social_group_quickjoin_' . $group->getGroupType()->id()) == FALSE) {
+      return;
+    }
+
+    // Create quickjoin URL from route.
+    $url = Url::fromRoute('social_group_quickjoin.quickjoin_group', [
+      'group' => $group->id(),
+    ])->toString();
+
+    // Redirect ot quickjoin.
```
bramtenhove: Small typo.
```suggestion
    // Redirect to quickjoin.
```

modules/social_features/social_group/modules/social_group_quickjoin/src/Form/SocialGroupQuickjoinSettings.php
jochemvn: Applied!

modules/social_features/social_group/modules/social_group_quickjoin/src/EventSubscriber/RedirectSubscriber.php
jochemvn: Done!

jochemvn:
@bramtenhove  Thx for the review. All suggested changes are made! 💪

jochemvn:
@bramtenhove 
1. Not really nice I'm afraid. We could but it'll still be a seperate form (on the same route)
2.Not really sure what you mean by that... If you mean that we can only quickjoin when we `enable` and `enable open group` then I'd say no, since the CTA case is very useful for some customers, without losing the 2 step confirmation

NatashaSchon:
Sure @bramtenhove! Here are some improvements below. The rest are fine.

Enabling this feature gives site builders the possibility to create group 'quick join' links (ex., /group/1/quickjoin). Furthermore, it's possible to skip the confirmation step on a group type basis.

Allow users to join groups with a single click.

Allow users to skip the confirmation step when joining a @grouptype group.

ronaldtebrake:
Unfortunately it's not reflected in MOSS due to missing webhooks for Philanthropy & regio twente, but both have passed tests on shippable so we're all green here!!

PR #1089, opened 2018-11-02
Added translation for gtitle

slowflyer:
## Problem
Using more than one language and have translated all content, the gtitle token in the Activity Logger always shows the group->label in default language, even if a translation exists.
Solution

## Solution
load correct translation for the group label

## Issue tracker
- https://www.drupal.org/project/social/issues/3011070

## HTT
- [x] Login as administrator and setup a site with more than one language.
Create some groups and translate the group to other language(s)
Create a topic, or post in the group and translate it.
Swich the language and the acitivity stream will show the group title in default language (see screenshot at https://www.drupal.org/project/social/issues/3011070)

- [x] Notice it doesn't work
- [x] Checkout to this branch
- [x] Group title will be displayed in correct language in activity stream

## Release notes
<describe the release notes>

### DIFF
--- a/modules/custom/activity_logger/activity_logger.tokens.inc
+++ b/modules/custom/activity_logger/activity_logger.tokens.inc
@@ -112,7 +112,8 @@ function activity_logger_tokens($type, $tokens, array $data, array $options, Bub
               // If it's a group.. add it in the arguments.
               if (isset($group) && $group instanceof Group) {
                 if ($name === 'gtitle') {
-                  $replacements[$original] = $group->label();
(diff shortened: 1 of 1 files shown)


slowflyer:
Has someone an idea, why the test fails at:

And I click the element with css selector "select#edit-field-topic-type-target-id" # FeatureContext::iClickTheElementWithCSSSelector()
    And I click "News"                                                                 # SocialMinkContext::assertClick()
      Link with id|title|alt|text "News" not found. (Behat\Mink\Exception\ElementNotFoundException)

The patch has nothing in common with this filter ...

Thanks

PR #1092, opened 2018-11-04
Added translation for topic_type display

slowflyer:
## Problem
Using more than one language, the topic_type is always displayed in default language, even, when translation to the current language exists.

## Solution
load correct translation for the topic_type to display

## Issue tracker
- https://www.drupal.org/project/social/issues/3011297

## HTT
- [x] Login as administrator and setup a site with more than one language.
Translate topic_type taxononmy to all languages
Create some topics in all languages
- [x] Look at e.g at /all-topics page in different languages and you will see that the topic-type is always displayed in default language
- [x] Checkout to this branch
- [x] Look at e.g at /all-topics page in different languages and you will see that the topic-type is now displayed in the current language

## Release notes
<describe the release notes>


modules/social_features/social_topic/social_topic.module
```diff
@@ -39,11 +40,12 @@ function social_topic_preprocess_node(array &$variables) {
     $topic_type = $node->get('field_topic_type');
     $topic_type_entities = $topic_type->referencedEntities();
     if (count($topic_type_entities) === 1) {
+      $curr_langcode = \Drupal::languageManager()->getCurrentLanguage(LanguageInterface::TYPE_CONTENT)->getId();
       foreach ($topic_type_entities as $topic) {
         $variables['metadata'] = t('in @topic', ['@topic' => $topic->link()]);
```
Kingdutch: @slowflyer It looks like this may also be applicable for the metatdata variable?

slowflyer: @Kingdutch  no metadata shows up correctly translated

Kingdutch: @slowflyer I don't quite understand why `$topic->link()` does show up translated (it creates a link of the label and actually calls `$this->label()`) but the `$topic->label()` call itself does not work.

Unless I misunderstand in this case and `$variables['metadata']` doesn't need to be translated.

modules/social_features/social_topic/social_topic.module
```diff
@@ -39,10 +40,12 @@ function social_topic_preprocess_node(array &$variables) {
...
         // Set topic type link.
         $topic_type_url = Url::fromRoute('view.latest_topics.page_latest_topics', ['field_topic_type_target_id' => $topic->id()]);
+        $topic->isTranslatable() ? $topic = $topic->getTranslation($curr_langcode) : NULL;
```
Kingdutch: I think ternary expressions for an assignment are fine but I think we should avoid them for performing logic because they can reduce readability. Expanding this to an `if` statement is probably clearer.

```suggestion
        if ($topic->isTranslatable()) {
          $topic = $topic->getTranslation($curr_langcode);
        }
```

modules/social_features/social_topic/social_topic.module
```diff
@@ -39,10 +40,12 @@ function social_topic_preprocess_node(array &$variables) {
...
         $topic_type_link = Link::fromTextAndUrl($topic->label(), $topic_type_url)
           ->toString();
         $variables['topic_type'] = $topic_type_link;
```
Kingdutch: I think we can (and should) use this PR to replace this line with

```suggestion
         $variables['topic_type'] = $topic->toLink();
```

I've chosen for `toLink` because it creates a render array that can be modified instead of a string. If that doesn't work then we can use `link()` instead which will output a string.

modules/social_features/social_topic/social_topic.module
slowflyer: @Kingdutch maybe you are right ... I do not translate the link ... that may be the reason why everything looks fine for me. I will add the logic for $topic-link() as well

modules/social_features/social_topic/social_topic.module
slowflyer: @Kingdutch your suggested modification of line 53 will result in a link to:
/taxonomy/term/[termID]
instead of
/all-topics?field_topic_type_target_id=[termID]
which is not correct

jaapjan:
Marking as `needs work`. Please correct if I'm wrong :-)

modules/social_features/social_topic/social_topic.module
slowflyer: changed that based on your suggestoin

slowflyer:
> 
> 
> Looks good from a code perspective, still needs a functional review and a check on why MOSS is complaining.

@Kingdutch can you please let me know the reason why MOSS was complaining? No matter what I do, MOSS still fails every time. Even when adding 2 optional buttons, MOSS can't be aware of, MOSS is failing

jaapjan:
Hey @slowflyer Moss is checking client projects as well. So that makes it not possible to make those results public. We might look into making a summary available to the public (or in a comment here) but that being said getting the PRs through is higher priority at the moment. We were having some issues with false-negatives and the fact that there were a lot of open pullrequests had some impact on performance. (Jenkins server and Shippable minions were working full-time!)

robertragas:
Moss has been restarted. Parts of this patch were also in a patch for a client. Should turn green now after the tests are done.

slowflyer:
> 
> 
> Moss has been restarted. Parts of this patch were also in a patch for a client. Should turn green now after the tests are done.

Thank's for explaining!

PR #1093, opened 2018-11-05
DS - 5504 As a CM+ I want to be able to edit a Existing Group's Group Type

ronaldtebrake:
## Problem
Sometimes it is hard to figure out what the Group's visibility settings mean. After creating a Group with content and members it would be nice to be able to change an existing Group's Group Type so the influence on the visibility of content is matching the Group better than the existing Group.

## Solution
- Created a permission `edit group types` enabled for CM+ by default. So they can change an Existing Group's group type
- Made the Group Type selector available on a Group edit page for those who can edit it. 
- Created a new form_submit for Group Edit (only fired if the group type is editted from the original)
- Created a new helper function which checks the default visibility per Group type (public/group members/community) so when we change the group type we can easily identify what the existing content's new visibility setting should be
- Provided a hook so enterprise with custom group types can alter it and provide their own default visibility settings
- Created a new batch function that based on a Group and it's new Groups Type will update all the existing group content (users/posts/nodes) 
- Created a new behat test to check this functionality

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-5504

## Acceptance Criteria
- Being able to change the group type after creation
- Content and posts visibility will be changed accordingly as well
- Make it clear for the CM+ what happens with the group and it's data
- After changing Public group to other group types, the AN event enrollment data won't be lost and if the change is reversed, the AN event enrollment is still there.

## HTT
- [ ] Login as LU, see that you can't change Group Type
- [ ] Login as SM, see that when you go and edit a Group you now get to choose a new Group Type
- [ ] Login as SM, create a new closed Group, create a new topic in this Group, create a Post in this groups stream
- [ ] Login as LU - see nothing changed, you can't access the above data
- [ ] Login as a 2nd LU - Join this group, see that the group has 2 members and you can access the data
- [ ] Create a Post on the Group stream and see that its for Group only
- [ ] Log back in as SM, change the newly created Group to be a Public group, see that it's updating content in Batches and it all works out just fine
- [ ] See that on the Group stream, when you edit the Post it is now a Public post instead of a "Group" visibility one.
- [ ] Log in as the 2nd LU - See that you are still a member of the Public group
- [ ] See that on the group stream the Post visibility is now by default changed to Public
- [ ] Login as the first LU (outsider)
- [ ] See that you can now access the Public Post, the Topic and you can become a member of the Group.

## Release notes
Sometimes it is hard to figure out what a new Group's visibility settings mean for the content inside the group. After creating a Group with content and members it would be nice to be able to change an existing Group's Group Type so the influence on the visibility of content is matching the Group better than the existing Group. This is now fixed for CM+.


modules/social_features/social_group/social_group.module
```diff
@@ -41,10 +44,51 @@ function social_group_form_group_public_group_edit_form_alter(&$form, FormStateI
...
+function social_group_group_type_permission_check() {
+  $user = \Drupal::currentUser();
+  // Get the Group object from the route.
+  $group = Drupal::routeMatch()->getParameter('group');
+
+  // Check if we are editing a group. Sure to be sure.
+  if ($group instanceof GroupInterface) {
+    // Only for user 1 and users with the edit group types permission we can
+    // edit a group type.
+    if ($user->id() != 1) {
```
bramtenhove: @ronaldtebrake shouldn't we just check for the permission? Why are we checking on user ID here?

modules/social_features/social_group/social_group.module
```diff
@@ -41,10 +44,51 @@ function social_group_form_group_public_group_edit_form_alter(&$form, FormStateI
...
+      if ($user->hasPermission('edit group types')) {
+        return TRUE;
+      }
+    }
+    else {
+      return TRUE;
+    }
+
+    return FALSE;
```
bramtenhove: Can't we just put `return FALSE;` at the end of the function and remove the rest except for the user has the permission? Then you already `return TRUE;`.

modules/social_features/social_group/social_group.module
```diff
@@ -675,19 +769,7 @@ function social_group_field_widget_form_alter(&$element, FormStateInterface $for
...
-
-          case 'open_group':
-            $element['#default_value'] = 'community';
-            break;
-
-          case 'public_group':
-            $element['#default_value'] = 'public';
-            break;
-        }
+        $element['#default_value'] = SocialGroupHelperService::getDefaultGroupVisibility($group_type_id);
```
bramtenhove: Nice work. Much better!

modules/social_features/social_group/src/GroupContentVisibilityUpdate.php
```diff
@@ -0,0 +1,205 @@
...
+      $membershipEntity->set('type', $new_group_type);
+      $membershipEntity->save();
+      $entity = $membershipEntity;
+    }
+
+    // Add referenced entity to results. Might want to add it to the result.
+    $context['results'][] = $entity;
+
+    // Optional message displayed under the progressbar.
+    $context['message'] = t('Updating Entity ID "@id"', ['@id' => $entity->id()]);
```
bramtenhove:
```suggestion
    $context['message'] = t('Updating group content (@id)', ['@id' => $entity->id()]);
```

I don't think Entity ID means much to users.

modules/social_features/social_group/src/GroupContentVisibilityUpdate.php
```diff
@@ -0,0 +1,205 @@
...
+   * @param array $results
+   *   Contains individual results per operation.
+   * @param array $operations
+   *   Contains the unprocessed operations that failed or weren't touched yet.
+   */
+  public function updateVisibilityFinishedCallback($success, array $results, array $operations) {
+    if ($success) {
+      // Here we could do something meaningful with the results.
+      // We just display the number of nodes we processed...
+      drupal_set_message(t('@count results processed.', ['@count' => count($results)]));
```
bramtenhove:
```suggestion
      drupal_set_message(t('Visibility of @count content item(s) updated.', ['@count' => count($results)]));
```

modules/social_features/social_group/src/SocialGroupHelperService.php
```diff
@@ -60,4 +53,33 @@ public static function getGroupFromEntity($entity) {
...
+  /**
+   * Returns the default visibility.
+   *
+   * @param string $type
+   *   The Group Type.
+   *
+   * @return string
+   *   The default visibility.
+   */
+  public static function getDefaultGroupVisibility($type) {
```
bramtenhove: Not entirely sure, but don't we want to use `&drupal_static(__FUNCTION__);` here as well?

bramtenhove: The alter hook only has to be executed once per request anyway.

modules/social_features/social_post/src/Entity/Post.php
```diff
@@ -157,6 +157,58 @@ public function getDisplayName() {
...
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getVisibility() {
+    if ($this->hasField('field_visibility')) {
+      switch ($this->field_visibility->value) {
+
+        case "0":
```
bramtenhove: You can remove `case "2"` at line 175-177.
Just combine it like:

```suggestion
        case "0":
        case "2":
```

bramtenhove:
@NatashaSchon if you have a moment again, we'd love your feedback on some of the new and/or changed strings.

>Please note that changing the group type will also change the visibility of group content and the way users can join the group.

>In order to change the group type, please contact the content or site managers.

NatashaSchon:
@bramtenhove sure! Just adjusted the first one a bit.

Please note that changing the group type will also change the visibility of the group content and the way users can join the group.

bramtenhove:
Thanks @NatashaSchon! 🙇

ronaldtebrake:
Hi @bramtenhove ,

Thanks for the review. Updated your remarks in separate commits.

PR #1094, opened 2018-11-06
3011451 - Enable content tagging for groups

jochemvn:
## Problem
Currently it's possible to tag content items (nodes) with tags from the social_tagging vocabulary. For group it's not yet possible.

## Solution
Enable (by setting) that groups can also be tagged with items from the same vocabulary. In the future we can better build overviews of related content or recommended content.

## Issue tracker
- https://www.drupal.org/node/3011451
- https://www.drupal.org/project/social/issues/2972564 (unrelated but fixed here anyway)

## HTT
- [x] Enable content tagging and create a hierarchy in the vocabulary
- [x] Notice you can now tag groups on the group edit page
- [x] Notice it also shows up on the group overview pages in the complementary area
- [ ] Go to the search group page and notice you can also filter on tags (and on group types)

## BONUS:
Since I needed to add the tagging field to the filter bock on the search page, I first needed to create a filter block. Therefore I added a grouptype filter on the group search page. Also now it matches the regular group overview, since that has the same filter.

## Release notes
As of now it's possible to tag groups with items from the social_tagging vocabulary. As for any node type, this needs to be enabled as a SM on the tagging settings page.


modules/social_features/social_search/social_search.module
```diff
@@ -39,6 +41,21 @@ function social_search_form_views_exposed_form_alter(&$form, FormStateInterface
         }
       }
     }
+    if ($form['#id'] == 'views-exposed-form-search-groups-page') {
+      $options = [
+        '' => '- Any -',
```
bramtenhove: Shouldn't this be translatable?

bramtenhove:
```suggestion
        '' => t('- Any -'),
```

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -102,88 +101,99 @@ function social_tagging_form_taxonomy_overview_terms_alter(&$form, FormStateInte
...
-
-    $config = \Drupal::getContainer()->get('config.factory')->getEditable('social_tagging.settings');
-    $config_value = $config->get('tag_node_type_' . $node_type);
-
-    if ($config_value === FALSE || $config_value === 0) {
-      $node_type_tagging = FALSE;
+function social_tagging_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Act if the form has the tagging field.
+  if (isset($form['social_tagging'])) {
+    // Load tag config..
```
bramtenhove:
```suggestion
    // Load tag config.
```

modules/social_features/social_search/social_search.module
jochemvn: check!

modules/social_features/social_tagging/social_tagging.module
jochemvn: Check

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -211,12 +221,12 @@ function _social_tagging_node_form_validate($form, FormStateInterface $form_stat
 /**
  * Helper function to get the current default tagging values of a node.
  */
-function _social_tagging_node_form_defaults_values(Node $node) {
+function _social_tagging_node_form_defaults_values(Entity $entity) {
```
Kingdutch: We should be using `EntityInterface` here instead

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -211,12 +221,12 @@ function _social_tagging_node_form_validate($form, FormStateInterface $form_stat
...
   // Init.
   $default_value = [];
   // If the node exists, we need to get the default value.
-  if ($node instanceof Node && $node->id() !== NULL) {
-    foreach ($node->get('social_tagging')->getValue() as $key => $value) {
+  if ($entity instanceof Entity && $entity->id() !== NULL) {
```
Kingdutch: We should be using `EntityInterface` here instead.

This allows compatibility with classes that don't inherit Entity but do implement the methods we require.

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -277,14 +287,14 @@ function social_tagging_entity_delete(EntityInterface $entity) {
 /**
  * Build output on node view.
  *
- * @param \Drupal\node\NodeInterface $node
+ * @param \Drupal\Core\Entity\Entity $entity
```
Kingdutch: We should be using `EntityInterface` here instead.

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -277,14 +287,14 @@ function social_tagging_entity_delete(EntityInterface $entity) {
...
  *   A node.
  *
- * @return array
+ * @return array|string
  *   Hierarchic presentation of the terms.
  */
-function social_tagging_process_tags(NodeInterface $node) {
-  if (!($node instanceof NodeInterface && $node->hasField('social_tagging'))) {
+function social_tagging_process_tags(Entity $entity) {
```
Kingdutch: We should be using `EntityInterface` here instead.

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -277,14 +287,14 @@ function social_tagging_entity_delete(EntityInterface $entity) {
 /**
  * Build output on node view.
  *
- * @param \Drupal\node\NodeInterface $node
+ * @param \Drupal\Core\Entity\Entity $entity
  *   A node.
  *
- * @return array
+ * @return array|string
```
Kingdutch: Why do we return an empty string in some cases instead of just an empty array? This makes using the function more difficult because something like `foreach (social_tagging_process_tags($entity) as $tag)` will break without first checking which return type we have.

modules/social_features/social_tagging/src/Plugin/Block/SocialGroupTagsBlock.php
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+    $cache_tags = parent::getCacheTags();
+    $group = $this->routeMatch->getParameter('group');
+
+    if ($group instanceof Group) {
```
Kingdutch: We should be using `GroupInterface` here instead.

modules/social_features/social_tagging/src/Plugin/Block/SocialGroupTagsBlock.php
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $build = [];
+
+    $group = $this->routeMatch->getParameter('group');
+
+    if ($group instanceof Group) {
```
Kingdutch: We should be using `GroupInterface` here instead.

modules/social_features/social_tagging/social_tagging.module
Kingdutch: Furthermore if this is done for display purposes then something like `join([], ", ")` would properly output an empty string already :)

Kingdutch:
Hi @bramtenhove,

This is actually a bigger issue that we're experiencing on more platforms during the updates 🤔 

It's caused by the following core updates:
```
8502 -   Copy hierarchy from {taxonomy_term_hierarchy} to {taxonomy_term__parent}. 
8503 -   Update views to use {taxonomy_term__parent} in relationships. 
```

This renames a table for the taxonomies. The error you see in the `drush cache-clear` at the start of the install is relatively harmless. The errors during the update are of bigger concern. These happen because an update that is being performed, in this case `social_topic_update_8002` seems to be touching taxonomies. When this update runs with the newer codebase and before the taxonomy tables have changed then errors will occur.

We should probably do two things:
- Figure out the update hook dependencies and add them to a hook_dependencies (in this case probably in the project instead of the distribution due to possibly large step in versions taken here)
- Evaluate our use of `hook_update_N`. I feel we overuse it and should probably look to `hook_post_update_NAME` in 80% of our use cases. We haven't run into many issues before but we would probably not have this issue using that solution.

The difference between the two:
- `hook_update_N` make changes to the data model
- `hook_post_update_NAME` make changes to the data (https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_post_update_NAME/8.6.x)

In this case we're trying to change the data in `hook_update_N` before the data model has been changed which causes problems.

Kingdutch:
@bramtenhove I'm unsure, it could be that the fixture database wasn't updated after the last successful update so it's trying to test that upgrade path again.

bramtenhove:
Discussed with @Kingdutch, a check will be done to see if the problem is in the fixture database.

ronaldtebrake:
As per: https://docs.google.com/document/d/1QtuOgdCjJqXo6kaqHI4rU9VXpbt1IltKeoAwb1TGTFI/edit#heading=h.uzv08wmtdf8k

Remove the group tagging block from the leave group route

jochemvn:
@ronaldtebrake: I've removed the tags block from the following routes:
`entity.group.join` >> Join group 
`entity.group.leave` >> Leave group 
`entity.group.edit_form` >> Edit group 
`entity.group.delete_form` >> Delete group 
`entity.group_content.collection` >> Manage members

bramtenhove:
Seems like the one failing Moss project is no problem and we can merge it.

PR #1095, opened 2018-11-06
Issue #3011721 by Kingdutch: Ensure login and registration forms have proper weight

Kingdutch:
## Problem
The user form items have no weights at the moment which makes it hard to add custom fields.

## Solution
Provide weights for the form elements.

## Issue tracker
- https://www.drupal.org/project/social/issues/3011721

## HTT
- [x] See that the login and registration form hasn't changed for various settings affecting these forms.

## Release notes
The elements in the login and registration forms now have proper weights to ensure that you can properly position custom fields in these projects.


jochemvn:
@Kingdutch not sure if relevant, but what about `social_sso` elements?

Kingdutch:
> @Kingdutch not sure if relevant, but what about `social_sso` elements?

I don't see any render elements for `social_sso` there's the `social_auth_extra` module for social login but I'm not sure whether render elements are needed on the social buttons (does their order matter?). The containers for the social login buttons and the login with e-mail fields already have a weight so that should work :)

Otherwise we can create a follow-up story as child of drupal.org/node/3010776

PR #1097, opened 2018-11-08
DS-6143 by ronaldtebrake: show the visibility of posts

ronaldtebrake:
## Problem
With the ability of editing groups and the already great amount of visibility options per group type it's more and more necessary for users to see what visibility the posts have they are replying on. Also since comments on these posts get their visibility inherited of the post. 

## Solution
The post already has a visibility option, it's now added to their activity stream views and post entity views. Also a hook has been added if other visibility options are available or other svg icons are available so users can easily alter it.  

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-6143

## HTT
- [ ] Login as LU
- [ ] See that when you post on different streams the post now get a correct visibility setting
- [ ] See that when you visit streams already existing posts have their visibility set

*INFO* Xinyu already approved the design (gave him a demo)

## Release notes
With the ability of editing groups and the already great amount of visibility options it's more and more necessary for users to see what visibility the posts have they are seeing and maybe acting on. Also since comments on these posts get their visibility inherited of the post we now added post visibility to the post output. 

If you have a custom theme you can add something similar as below to your twig template for posts / activities and showing their visibility.

`{% if visibility_icon and visibility_label %}
  <svg class="margin-left-s icon-visibility">
    <use xlink:href="#icon-{{ visibility_icon }}"></use>
  </svg>
  <strong>{{ visibility_label }}</strong>
{% endif %}`


ronaldtebrake:
Wrong base branch, cherry picked it in to #1104

PR #1098, opened 2018-11-08
Add "create bookpage" button

slowflyer:
## Problem
Open Social does currently not provide buttons (blocks) to create new book pages in the same way as it is done for topics or events

## Solution
Add the blocks to the distro

## Issue tracker
- https://www.drupal.org/project/social/issues/3012343

## HTT
- [x] Checkout to this branch
- [x] Login as admin and place the related button's either inside or outside a group context, following same logic as it's done for e.g. topics
- [x] Navigate to pages where the buttons are visible and create new bookj pages by clikcing on the button


jaapjan:
Doubting if this should be added to 4.0 instead of 3.x.

Also should we consider adding blocks to existing installations and adding them to new installations?

What do you think @slowflyer ?

slowflyer:
@jaapjan that's up to you if you prefer 4.0 instead of 3.x. I would like to see it as soon as possible, because I need to fix my project on every update. And adding the 2 files should not conflict with existing installations, seems that nobody missed them before. By "adding blocks" you mean place the block (created with this patch) inside a region? 
I currently show the block only on book-pages. Maybe that's a good point to start and if someone needs to adjust he is free to do this. If you agree I can the config file.

jaapjan:
Ok, understood. I think only contentmanagers have access to create books and by default books can't be created in Groups in Open Social. Perhaps we can supply the blocks as you created them now and not place them in new and existing installations. If we do that I'm ok with adding them to 3.x still. (if we get the tests to pass! 💯)

slowflyer:
Ok, than keep fingers crossed that moss test will succeed.

jaapjan:
@slowflyer 

Tests are passing, but I still have some feedback which needs to be processed before it is ready to be merged.

- [ ] The link text should be `Create book page` instead of `Create Bookpage`. I prefer it to be consistent with what we use in the header as well.

book-create-link

- [ ] I think the GroupAddBookBlock should be added to the social_book module instead of the social_group module since it is an optional module it is better to have everything related to book in the social_book module directly.

- [ ] Styling looks off. Not sure why this is happening. I've positioned the group add book block on every group page (stream in this example).

group-create-book-block

slowflyer:
@jaapjan fixed your items 1 & 3. The styling was "fine". You created the block with the option "Display title" enabled. That's how it looks for all :-), if this is done...  I set default to not display title

Your item 2 is hard, I did not get it work when GroupAddBookBlock is part of social_book. There are functions called from social_group module ...

jaapjan:
@slowflyer OK, cool

Regarding point 2. 

There is only the function `_social_group_get_current_group()` right? If we just check if the module exists before calling that function that should be enough?

https://drupal.stackexchange.com/a/188323

And in addition moving the file to social_book and change the namespace on the top to social_book. :-)

slowflyer:
> 
> 
> @slowflyer OK, cool
> 
> Regarding point 2.
> 
> There is only the function `_social_group_get_current_group()` right? If we just check if the module exists before calling that function that should be enough?
> 
> https://drupal.stackexchange.com/a/188323
> 
> And in addition moving the file to social_book and change the namespace on the top to social_book. :-)

I promise, I exactly tried it before and failed ... now it's working

jaapjan:
FYI: still 1 out of the 8 projects failing. Is false-negative. Sorry about that. Will merge now!

PR #1100, opened 2018-11-08
3012256 - Replace addtoany with shariff 👮🏼‍♂️

jochemvn:
## Problem
We don't want to use addtoany anymore

## Solution
Use shariff instead 

## Issue tracker
- https://www.drupal.org/node/3012256

## HTT
- [ ] Enable social_sharing
- [ ] Notice on public content you have the sharing icons just like before
- [ ] Notice it's not addtoany (pretty important 😛)
- [ ] Make sure that as a SM you can edit the buttons here: `/admin/config/services/shariff`
- [ ] Go back to the 8.x-4.x branch an do a reinstall... after that immediately install social_sharing
- [ ] Go back to this branch and run the update.sh script
- [ ] Notice all is still good and notice it's not addtoany (pretty important again 😛) 

## Release notes
"Facebook, Google+ and Twitter supply official sharing code snippets which quietly siphon personal data from all page visitors. Shariff enables visitors to see how popular your page is on Facebook and share your content with others without needless data leaks."


bramtenhove:
Can be merged when Composer issues are solved and builds are green.

jaapjan:
Will this also work for subthemes of socialblue by default?

bramtenhove:
@jaapjan, I believe it works in the same way as AddToAny worked before.

In install hook the AddToAny blocks are removed in, I think, a generic enough way to also remove the "custom theme" blocks.

jochemvn:
@jaapjan : What @bramtenhove says. I still have some time this week to look into another issue we have, so I'll look at that too

jochemvn:
👋 I've update .travis.yml and made sure composer (not php) has a no memory limit anymore. I've also update some package version nrs in the `composer.json`. The packages are not actually updated, but only set closer to the actual version they're using.

jochemvn:
@ronaldtebrake , I assume this is now good to go? or....

jaapjan:
@jochemvn Mr Moss still says no. False negative or?

jochemvn:
yes

PR #1102, opened 2018-11-11
#3013127 - node view statistics

bramtenhove:
## Problem
To increase and measure engagement we want to track views per node. And we want to display them on full node display.

## Solution
An optional module that tracks page views per user, displays the amount in full node display and adds a new KPI block if the `social_kpi_lite` module is enabled.

## Issue tracker
- https://www.drupal.org/project/social/issues/3013127
- https://jira.goalgorilla.com/browse/SHN-132

## HTT
- [ ] Login as SM and configure node types that should be tracked (default is event and topic) at /config/opensocial/node-statistics
- [ ] Create public topic and view it a couple of times as LU, AN, etc.
- [ ] Create community topic and view it a couple of times as LU
- [ ] View the community topic as AN, observe (e.g. in the database) that it won't count as a page view
- [ ] At the full node display see the amount of views for node types that are enabled
- [ ] See that the page is still cached and amount of views does not visually increase
- [ ] Wait 15 minutes (and 7-10 page views) or clear the cache to see the amount of views increasing
- [ ] Go to the KPI analytics report to see a block on the right with th amount of views per month

## Release notes
A new optional module is added that can be used to track and display the amount of the page views per node type. On the full view display of the node the page views are shown. And if you use the Social KPI Lite module you will also get a new KPI block.


bramtenhove:
@jochemvn I fixed several issues in 3ddc17e.

- Only track page views on node canonical url
- Cache invalidation settings are now configurable
- Page view request time is now correct
- Parent menu link is added
- Uninstall hook added to remove the KPI Analytics block
- Added hook_modules_enabled to add KPI Analytics block when `social_kpi_lite` module is enabled

ronaldtebrake:
Unfortunately it's not reflected in MOSS due to missing webhooks for Philanthropy & regio twente, but both have passed tests on shippable so we're all green here!!

PR #1105, opened 2018-11-13
DS-6092 - Ability to (don't) add post block to community or personalised stream in landing pages

jaapjan:
## Problem
One can only add a community stream without a post option. What if you want the community stream in a landing page but also want the post block?

## Solution
Let's add more flexibility by allowing community stream _with_ post block and in addition allow personalised stream _without_ the post block.

## Issue tracker
- DS-6092

## HTT
- [x] Install social_landing_page on 8.x-4.x branch.
- [x] Create landing page with personalised stream
- [x] Create landing page with community stream.
- [x] Notice personalised stream contains post block and community stream doesn't.
- [x] Switch to this branch and run necessary updates.
- [x] Notice the pages didn't change.
- [x] When you edit the landingpages note that you have now extra options, to add community stream with post block and a personalised stream without a post block. Test these out and reload the landingpages.
- [x] Notice it works as expected.
- [x] Also test out a fresh installation and notice social_landing_page installs perfectly on this branch.

## Release notes
We've made the landing page functionality even more flexible. When creating a landing page it is now possible to add personalised stream blocks without post option and a community stream block with a post option.


ronaldtebrake:
Works like a charm!

PR #1112, opened 2018-11-20
DS-5914: Post comment size

jaapjan:
## Problem
Post comment field is made too tall (potentially due to big pipe)

## Solution
Update script autosize for comment fields

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-5914

## HTT
- [x] Enable big_pipe
- [x] Check the homepage in the base branch, notice the comment post field height is not correct
- [x] Checkout to this branch
- [x] Check post comment field height and notice the height changes when you click on it.

## Documentation

## Release notes
The post comment form was shown in double the size as usual. We fixed this.


ribel:
Hi @jaapjan,
@Crazylord is no longer work on Open Social, so I think he will not have time to work on it. 
But since as you mentioned there is some issue only with BigPipe and on SaaS, I think any other developer can also take a look to it.

jaapjan:
I did a `gulp scripts` and the diff is a bit bigger, but now it seems to work.

jaapjan:
Ah, it only works after a refresh on the page or when resizing the window. I think it still needs a bit of work though, but an extra pair of eyes is welcome :-)

ronaldtebrake:
Cherry picked to 4 and 5 🥇

PR #1116, opened 2018-11-21
DS-6236, #3005538, 2965717: Mention improvements

jaapjan:
## Problem
On platforms which multiple users it is hard to find the correct user in user reference fields or in the mentions. This is already partly improved in PR #1082 but this is only part of the solution. If your platform has more than 50 John's it is still hard to find `John Doe` as there is no sorting by relevance and it is not possible to search for space-separated strings.

- [#2965717 - When searching for a user to mention the suggestions should sort by relevance](https://www.drupal.org/project/social/issues/2965717)
- [#3005538 - When searching for a user to mention the suggestions should sort by relevance](https://www.drupal.org/project/social/issues/3005538)

## Solution
The solution is two-fold. 

### 1. Better sorting
By providing better sorting it is easier to find a given user. Let's say you have the following users:

- Robert Johnson (username: robertjohnson, uid: 5)
- John Doe (username: harn8, uid: 3)
- John Marksen (username: xiou, uid: 8)
- Noah Farais (username: johnny, uid: 2)

When searching for `John` the results would have been ordered like this:

- Noah Farais
- John Doe
- Robert Johnson
- John Marksen

This was confusing to users as the sorting didn't make sense. By implementing a sort with the following rules the results are much better:

1. Users whose have first name starting by the given string;
2. Users whose have last name starting by the given string;
3. Users whose have nickname starting by the given string;
4. Users whose have username  starting by the given string;
5. Users containing the string.

### 2. Allow searching with a space
By allowing multiple strings and filtering on a concatenated value of first name and last name it allows users to narrow down a search with a combination of first name and last name.

E.g. when you have a list of those 4 users again and you know that the user you want to reference is named `John Doe` you can just search for `John Doe` and it will return the user.

## Issue tracker
- DS-6236
- https://www.drupal.org/project/social/issues/3005538
- https://www.drupal.org/project/social/issues/2965717

## HTT
- [ ] Import a bunch of demo users (Jaap has demo content folder you can use) or try it out on a big enterprise site or create users with similar names (like the examples above)
- [ ] Try it out on the 8.x-4.x branch and notice that the experience to mention a user is not so nice and might be unusable on bigger sites
- [ ] Checkout to this branch
- [ ] Notice you can find users because the search sorting is a lot better
- [ ] Notice you can find users by specifying a combination of first name and last name
- [ ] Enable the optional social_profile_fields module and test the nickname search as well

## Release notes
On platforms with a lot of users it was sometimes hard to mention a user or add a user to a group. The provided sort was not logical for users. This has now been improved. It is now also possible to narrow down your search by providing a combination of first and last name.

### DIFF
--- a/modules/social_features/social_profile/src/SocialProfileTrait.php
+++ b/modules/social_features/social_profile/src/SocialProfileTrait.php
@@ -46,6 +46,19 @@ trait SocialProfileTrait {
 
       case SOCIAL_PROFILE_SUGGESTIONS_FULL_NAME:
       case SOCIAL_PROFILE_SUGGESTIONS_ALL:
+        $strings = explode(' ', $name);
+        if (count($strings) > 1) {
+          $query->where("CONCAT(fn.field_profile_first_name_value, ' ', ln.field_profile_last_name_value) LIKE :full_name", array(':full_name' => '%' . $name . '%'));
+          $query = $this->sortQuery($query, $name, $suggestion_format);
+          $results = $this->endQuery($query, $count);
+
+          if (count($results) > 0) {
+            return $results;
+          }
+          // Fallback to creating a new query if there is no hit on full name.
+          $query = $this->startQuery();
+        }
+
         $or_query = $query->orConditionGroup();
         $or_query
           ->condition('fn.field_profile_first_name_value', '%' . $name . '%', 'LIKE')
--- a/modules/social_features/social_profile/src/SocialProfileTrait.php
+++ b/modules/social_features/social_profile/src/SocialProfileTrait.php
@@ -48,7 +48,7 @@ trait SocialProfileTrait {
       case SOCIAL_PROFILE_SUGGESTIONS_ALL:
         $strings = explode(' ', $name);
         if (count($strings) > 1) {
-          $query->where("CONCAT(fn.field_profile_first_name_value, ' ', ln.field_profile_last_name_value) LIKE :full_name", array(':full_name' => '%' . $name . '%'));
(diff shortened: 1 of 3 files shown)


jaapjan:
Last commits some small improvements, namely:

- Using leftJoin instead of join to also add users which didn't enter first or last name. Somehow this was already wrong for setting SOCIAL_PROFILE_SUGGESTIONS_FULL_NAME but since most platforms just use the SOCIAL_PROFILE_SUGGESTIONS_ALL it wasn't noticed before.
- Removed trailing and leading spaces when searching for a string with a space in it. Reason is that otherwise people who enter their first name as ` John ` with last name `McDonald` may not get find when searching for `John McDonald`.
- Removed the left space from the search query. Reason is that otherwise when you would have searched for `@ John` it would return `Robert Johnson` while the user probably expected to find someone with first name John.

Kingdutch:
> Removed the left space from the search query. Reason is that otherwise when you would have searched for @ John it would return Robert Johnson while the user probably expected to find someone with first name John.

Is it possible to write a message such as `Hey want to meet up @ McDonald's?` without searching for a user? I think it makes sense to require a character right after the @ sign (much like other platforms do). This also prevents something that Jira sometimes does which is replace 3 sentences after you performed a mention without an option for undo (ctrl + z), leaving you to retype a bunch of text.

jaapjan:
Sure, that would be a nice addition. But let's leave that out-of-scope for now. It currently does not work like that as well and was not in the scope of this story to solve that issue.

By the way: you can always click outside of the suggestion box when you don't want to tag someone so that is already possible.

ronaldtebrake:
After checking with Robert the SDGP failure is a false negative and we agreed to merge knowing this.

PR #1117, opened 2018-11-22
Use hirak/prestissimo for faster composer installs

Kingdutch:
## Problem
Installing all composer dependencies can take 3-5 minutes per Travis run. 

## Solution
Use hirak/prestissimo. The package parallelises downloads ahead of time instead of downloading
packages one by one. This can speed up package installs by about 60-80%.

## Issue tracker
No issue for this single line CI change.

## HTT
- [ ] All tests should remain green and the `composer install` run should be faster.

## Release notes
N.A.


bramtenhove:
This sounds pretty interesting. Let's see how Travis and Moss performs and also compare that to other PRs.

Kingdutch:
The installation of `hirak/prestissimo` takes about 9 seconds. The installation of Open Social with composer afterwards takes about 95 seconds: https://travis-ci.org/goalgorilla/open_social/jobs/458296034#L542

Comparing this to a different PR where it takes 180 seconds https://travis-ci.org/goalgorilla/open_social/jobs/458303145#L532

That's a 1.5 minute decrease per run.

jaapjan:
Nice, thanks. In Shippable it saves on average 2-3 minutes per site, quite significant in the long run :-)

PR #1121, opened 2018-11-22
Issue #3010778 by Kingdutch: Ensure group local tasks have proper weights

foxtrotcharlie:
## Problem
The items in this render array should have #weight properties that are properly spaced (e.g. 10, 20, 30, 40) so that their order remains consistent across releases and other modules can easily add their own items in the desired place.

## Solution
Adjusted social_group_menu_local_tasks_alter to provide #weights for all known $data['tabs'][0] children.

Set weights for group related links in social_group.links.task.yml

## Issue tracker
- https://www.drupal.org/project/social/issues/3010778#comment-12866568

## HTT

I'm not really sure how we should test this fix, as it's not going to have any visible impact.

## Release notes
Group local tasks now have default, more spaced out weights, allowing for insertion of other tabs at any position.


jaapjan:
Hey @foxtrotcharlie thank you for creating the PR. It does seem your branch is not entirely up-to-date with the latest changes in goalgorilla/open_social `8.x-3.x` branch. Could you merge the upstream branch in your branch? This will retrigger the tests and after that we can pick up the review :-)

foxtrotcharlie:
My apologies, have merged.

jaapjan:
Thanks!

jaapjan:
Hi @foxtrotcharlie I think in this case we need to do only one of the two actions. Either add it to the social_group.links.task.yml or social_group_menu_local_tasks_alter. I've removed it from the social_group_menu_local_tasks_alter. The issue is that social_group_menu_local_tasks_alter is changing it for all the tabs. Also for example on nodes, users, and whereever there are tabs. This can cause some unexpected changes.

You can see this for yourself when using xdebug and pausing it in your IDE in that function. 
social_group_menu_local_tasks_alter

You can see for example that the social_group.members tab has a weight of 90 instead of the expected 60.

After the tests are green I'll merge this for 3.6 release. Thanks for your work!

PR #1125, opened 2018-11-23
#3015692 -  #3009970 Performance PM and database prefix fix

bramtenhove:
## Problem
**Performance issue**
On every page load the private message unread count indicator in the header is checking for well, unread private messages.

Considering you have 86 private message threads, it will execute 86 queries to determine how many unread private messages you have. This is not ok and should be limited to one query for all threads.

**Query error**
Also fixed an issue where a query was not using table placeholders and thus breaking sites with a database with a prefix.

## Solution
**Performance issue**
Rewrote the logic to determine the count of unread private messages. It now uses a single query.

Tested with a user that has approximately 90 PM threads, for users with less threads impact will be lower.
screen shot 2018-11-23 at 12 28 14

**Query error**
Rewrote the query to use table placeholders.

## Issue tracker
- https://www.drupal.org/project/social/issues/3015692
- https://www.drupal.org/project/social/issues/3009970

## HTT
- [x] Login as user X and create a new private message, send it to user Y
- [x] Create a new private message and send it to user Y and Z
- [x] Observe that the unread count in the header is still 0
- [x] Login as user Y, notice the PM unread count is 2
- [x] Send some messages to each other and notice the unread count is as it is supposed to be

## Release notes
The unread private message indicator in the header would check every thread one-by-one (using SQL queries) to determine if you have seen the latest messages. This was inefficient, specially with a large number of threads. We rewrote the logic so it runs a single query for all the threads saving a lot of possible overhead.

### DIFF
--- a/modules/social_features/social_private_message/src/Service/SocialPrivateMessageService.php
+++ b/modules/social_features/social_private_message/src/Service/SocialPrivateMessageService.php
@@ -9,7 +9,6 @@ use Drupal\Core\Database\Connection;
 use Drupal\Core\Entity\Entity;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Session\AccountProxyInterface;
-use Drupal\private_message\Entity\PrivateMessageThread;
 use Drupal\private_message\Mapper\PrivateMessageMapperInterface;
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
Cherry picked in 8.x-4.x.

PR #1127, opened 2018-11-23
Issue #2948195: fix installation of social_book during profile installation

jaapjan:
## Problem
Installation error when selecting social_book during installation. Reason is that it can't find the necessary dependencies.

## Solution
Move the `core.entity_view_display.node.book.featured.yml` file to social_landing_page/config/optional instead. That will solve the issue.

## Issue tracker
- https://www.drupal.org/project/social/issues/2948195

## HTT
- [x] Check out the code changes and description in the d.o issue
- [x] Try installation on 8.x-3.x branch with social_module as optional module to install
- [x] Checkout to this branch and retry and notice it is working
- [x] Install the social_landing_page (and run update.sh afterwards) and see that the book node has a nice featured entity display and you can create a featured block in a landing page and feature a book node.

## Release notes
When running installation via the UI and selecting the Book Functionality as an optional feature to enable it would result in an installation error. This has now been solved.


Kingdutch:
@jaapjan The fact that this is green makes it look good to me : ) 

One thing we may be able to think about is whether this issue should've been caught by tests in the first place? Can we modify Travis to test for this scenario so that we catch any regressions?

PR #1134, opened 2018-11-24
Issue #2947898: Missing schema for entity_access_by_field.settings

jaapjan:
## Problem
The entity_access_by_field config schema declaration has no entry for the entity_access_by_field.settings config object.

## Solution
Add this to configuration part to the top of entity_access_by_field.schema.yml.

## Issue tracker
https://www.drupal.org/project/social/issues/2947898

## How to test
- [x] Install the site and see everything is imported and installed correctly
- [ ] Set the configuration for entity access by field and notice everything still works as expected.

## Release notes
The entity_access_by_field config schema declaration now has an entry for the entity_access_by_field.settings config object.

### DIFF
--- a/modules/custom/entity_access_by_field/config/schema/entity_access_by_field.schema.yml
+++ b/modules/custom/entity_access_by_field/config/schema/entity_access_by_field.schema.yml
@@ -1,5 +1,15 @@
 # Schema for the configuration files of the Entity Access By Field.
 
+entity_access_by_field.settings:
+  type: config_object
+  label: 'Entity Access by Field Settings'
+  mapping:
+    disable_public_visibility:
+      type: boolean
+      label: 'Disable public visibility'
+    default_visibility:
+      type: string
+      label: 'The default visibility for fields'
 field.storage_settings.entity_access_field:
   type: mapping
   label: 'Entity access field settings'


jaapjan:
I tested this but it is not working yet as it does not save the settings correctly. 

Test by:
- Login as sitemanager.
- Disable public visibility here: /admin/config/opensocial/visibility
- Login as LU.
- Go to /node/add/topic and notice that you can still select public (this should not be possible).

jochemvn:
@jaapjan 
Could it be because: https://www.drupal.org/project/social/issues/2947898#comment-12891198

jaapjan:
That is actually one of the reasons to add the missing schema 👍

ronaldtebrake:
Unfortunately this will need some more work. 

`-          if ($config->get('disable_public_visibility') === 1 && !$account->hasPermission('override disabled public visibility')) {
+          if ($config->get('disable_public_visibility') === TRUE && !$account->hasPermission('override disabled public visibility')) {`

This will need to be done together with possibly an update path of the current 0 value which is set in the entity_access_by_field.settings.yml. Or we need to go for a == solution so checking for 1 and TRUE will work.

jaapjan:
Closing. We can re-open when it's actually solved 🙈

PR #1141, opened 2018-11-30
Issue: #3012005 all day event timezones

jaapjan:
## Problem
The all day event sets the time to 0:01 and also does some trickery with timezones. When you save an event on a website with a timezone below GMT (e.g. America/Los_Angeles for GMT - 8) then this trickery causes the 0:01 to change to the previous day which causes an event created for 10 November 2018 to be displayed as taking place on 9 November 2018.

Using "all day" option with Timezone of "New York" was displaying the date 1 day behind to the front end users + when you go back to edit the event, the date shifts also. So, if you were not paying attention to the date when re-saving the node, you could keep shifting the dates with every save of the node..

## Solution
veronicaSeveryn: And I think I found the issue what goes wrong in this case. I have Drupal 8.6.3.
When you save "all day" event, it disables "time" form element, which causes Drupal's default handling of date/time form widgets fail in DateTimePlus::createFromFormat() with "'The date cannot be created from a format.'" exception because the Time is not passed in and the format is not correct.

I agree that Open Social shouldn't be manipulating the timezones.. and just to make the things work for now I created a patch that fixes the issue for me on Open Social 8.x-3.5

## Issue tracker
- https://www.drupal.org/project/social/issues/3012005

## HTT
- [ ] See the failing test as a sign of success
- [ ] See problem and solution for more test steps


jaapjan:
Will be 3.7 as well.

Marking as needs work, because:
- behat test is still failing
- Coding standards
```

FILE: .../social/modules/social_features/social_event/social_event.module
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 432 | ERROR | [x] A cast statement must be followed by a single
     |       |     space
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

ronaldtebrake:
Cherry-picked to 4.x and 5.x ofc :)

PR #1145, opened 2018-12-04
Refactor Group Edit Test

Kingdutch:
## Problem
The test breaks on some MOSS platforms because there are many groups/topics on the overview.

## Solution
This tries to avoid the group and topic overviews to visit a group
because this may not work in some scenarios on some platforms.

## Issue tracker
n.a.

## How to test
- [x] Check that Travis/Moss are green.

## Release notes
internal, n.a.


jaapjan:
Merged, didnt cherry-pick for 3.x, but will be in 5.x

PR #1146, opened 2018-12-05
Issue #2994207: Checkboxes are not shown in the Social Blue theme because checkboxes need titles

jaapjan:
## Problem
Checkboxes are not shown in the Social Blue theme because checkboxes need titles.

## Solution
Let's not test for the title. The title can be empty so the label can still be printed with a description.

## Issue tracker
https://www.drupal.org/project/social/issues/2994207

## How to test
- [ ] Test all the checkboxes and form elements.
- [ ] Install the filebrowser (see issue) and test it.

## Release notes
In some cases (contributed modules) checkboxes in forms were not shown when there was no title filled in. Now the label element is always printed.

### DIFF
--- a/themes/socialbase/templates/form/form-element-label.html.twig
+++ b/themes/socialbase/templates/form/form-element-label.html.twig
@@ -18,22 +18,25 @@
 #}
 
 {%-
-  set classes = [
+    set classes = [
     'control-label',
     title_display == 'after' ? 'option',
     title_display == 'invisible' ? 'sr-only',
     title_display == 'above' ? 'control-label--above',
     required ? 'js-form-required',
     required ? 'form-required',
-  ]
+]
 -%}
 
 {{ element }}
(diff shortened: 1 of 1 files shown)


jaapjan:
@Kingdutch Would be great if you can review this one next week?

ronaldtebrake (commented):
Wouldn't something like this be more suitable?

`
{{ element }}
{%- if title is not empty -%}
  <label{{ attributes.addClass(classes) }}>
      {{ title }}
    {%- if description -%}
      <p class="help-block">{{ description }}</p>
    {%- endif -%}
  </label>
{%- endif -%}
{%- if required and title_display != 'invisible' -%}
  <span class="form-required">*</span>
{%- endif -%}
`

That way we can still influence if we want titles or not (for example for VBO this was rendering wrong)

jaapjan:
Cherry-picked in 5.x and 4.x as well.

PR #1147, opened 2018-12-05
Issue #3018215 by robertragas: Add context for comment button to aid …

robertragas:
…in translation

## Problem
Open Social has the ability to be multilingual. The "Comment" text though can have different meanings in another language which require a better context.

Example:
- Notifications center: xxx likes your comment.
- Post comment button: comment.
These are perfectly fine but when translated to dutch

- Notifications center: xxx vindt jouw reageer leuk.
- Post comment button: reageer
Here the first line has the incorrect context

## Solution
Add context to the comment button. 

## Issue tracker
https://www.drupal.org/project/social/issues/3018215

## How to test
- [ ] Use a Dutch version of Open Social
- [ ] Create a comment and let someone else like it
- [ ] It will show a notification like above
- [ ] Check out the code and see the translations now has more context and able to translate more accurately.

## Release notes
Added a translation context to the comment button to be able to translate more accurately without conflicting with the comment entity type.

### DIFF
--- a/modules/social_features/social_comment/social_comment.module
+++ b/modules/social_features/social_comment/social_comment.module
@@ -287,7 +287,7 @@ function social_comment_comment_view(array &$build, EntityInterface $entity, Ent
           ],
         ];
         $commented_entity_url = $commented_entity->toUrl('canonical', $link_options);
-        $build['comment_link'] = Link::fromTextAndUrl(t('Comment'), $commented_entity_url)->toRenderable();
+        $build['comment_link'] = Link::fromTextAndUrl(t('Comment', [], ['context' => 'Comment Button']), $commented_entity_url)->toRenderable();
       }
 
       $comment_count = $commented_entity->{$comment_field_name}->comment_count;


Kingdutch:
Should we change the context from the place that it's used to the meaning in which it's used?

E.g. use `verb` as context? If we then have the action "comment" somewhere else then we can re-use the context and don't need a separate translation.
```
For the Dutch among us:

verb <-> Werkwoord
noun <-> Zelfstandig Naamwoord
```

jaapjan:
I don't think that works in all the situations. For example this scenario: https://www.drupal.org/project/social/issues/2948329 

I do think adding translation context like is done by Robert now is actually ok. See also what is done here.. Not sure why this is not there anymore 🤔 
https://github.com/goalgorilla/open_social/pull/650

Here also a related issue: https://www.drupal.org/project/social/issues/2891575 by the way, perhaps an idea to solve it all at once?

Kingdutch:
I understand that there will not always be a distinction between a noun and a verb but sometimes a distinction between two different nouns. My point was that I think it's better to use the context to describe the function of the word instead of where it's used. That way when a translation is used in multiple places but has the same function it only needs 1 translation instead of one for each place.

In the issue you linked it could be for example: "noun (as used in programming)" vs "noun (job/organization title)"

In this PR it would be: "noun" vs "verb"

The issue you linked to actually links on to https://www.drupal.org/project/social/issues/2901337 where there's also a discussion between "verb" and "noun" but that's unfortunately not what ends up being used.

PR #1148, opened 2018-12-05
 #3018168 - Hotfix 4.1 - Fix read count placement and updateReadCount on click.

ronaldtebrake:
## Problem
In 4.0 the AccountHeaderBlock got a massive overhaul. This resulted in the count indicator moving from the Top of the icon to the bottom.
Also the notification reading results in an JS error resulting in messages not being marked as "read" therefore the counter remains the same even after reading the notifications.

## Solution
1. We will update the count to make sure accessibility stays in place and the count shows on top again.
2. We will fix the JS issue that marks messages as read.

## Issue tracker
- https://www.drupal.org/node/3018168

## How to test
- [ ] Login as LU
- [ ] See that he indicator for Notifications is shown on the bottom of the Icon
- [ ] See that when you click the icon you get a JS error and the count doesn't update

- [ ] Checkout this branch, repeat the steps above and see that it now worsk.

## Release notes
In 4.0 the AccountHeaderBlock got an overhaul. This overhaul resulted in two bugs which we managed to fix and release in this hotfix. The count indicator should now be updated again once you click on the Notification Centre icon. And the indicator is shown on the top part of the icon again.

### DIFF
--- a/modules/social_features/social_activity/social_activity.module
+++ b/modules/social_features/social_activity/social_activity.module
@@ -23,6 +23,12 @@ function social_activity_social_user_account_header_items(array $context) {
   $account = $context['user'];
 
   $notifications_view = views_embed_view('activity_stream_notifications', 'block_1');
+  // Mark all notifications as seen.
+  $notifications_view['#attached'] = [
+    'library' => [
+      'activity_creator/activity_creator.notifications',
+    ],
+  ];
 
   $account_notifications = \Drupal::service('activity_creator.activity_notifications');
   $num_notifications = count($account_notifications->getNotifications($account, [ACTIVITY_STATUS_RECEIVED]));
--- a/modules/social_features/social_activity/social_activity.module
+++ b/modules/social_features/social_activity/social_activity.module
@@ -30,6 +30,15 @@ function social_activity_social_user_account_header_items(array $context) {
     ],
   ];
 
+  // Make sure we check if attached is part of the render array.
+  if (!isset($notifications_view['#attached'])) {
+    $notifications_view['#attached'] = [];
+  }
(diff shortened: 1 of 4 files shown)


modules/social_features/social_activity/social_activity.module
```diff
@@ -23,6 +23,12 @@ function social_activity_social_user_account_header_items(array $context) {
   $account = $context['user'];
 
   $notifications_view = views_embed_view('activity_stream_notifications', 'block_1');
+  // Mark all notifications as seen.
+  $notifications_view['#attached'] = [
+    'library' => [
```
Kingdutch: This could theoretically break if the view already has a library or other attachment.

Can do something like (untested):
```php
if (!isset($notifications_view['#attached']) {
  $notifications_view['#attached'] = [];
}
if (!isset($notifications_view['#attached']['library']) {
  $notifications_view['#attached']['library'] = [];
}
$notifications_view['#attached']['library'][] = 'activity_creator/activity_creator.notifications';
```

Kingdutch: It's more verbose but it it's also more robust.

Kingdutch:
From a code perspective this looks good to me other than the single improvement in my comment 👍 

I did not have time to do any functional testing other than what you showed me which looked as though it did fix the bugs :)

jaapjan:
Seems to work. Only thing I found is that when there are no notifications that it doesn't look very nice (bit too small).

no-notifications

jaapjan:
Double checked and on 3.x it looks like this.

no-notifications-3x

Perhaps good to check the entire header and compare it on 4.x again with 3.x?

PR #1150, opened 2018-12-06
#3018558 by bramtenhove: Add group memberships to the user export

bramtenhove:
## Problem
We're already exporting a lot of data with the user export, yet we do not have any group memberships for the user.

## Solution
Added two export plugins:
1. Exports the amount of groups a user joined
2. Exports the group names and IDs the user joined

## Issue tracker
- https://www.drupal.org/project/social/issues/3018558

## How to test
- [ ] Enable the user export module
- [ ] Join a few groups for several users
- [ ] Export all users via /admin/people
- [ ] The amount of groups a user joined and the group names and IDs should be in the exported file

## Release notes
When you now export users you will also be able to see the amount of groups they joined and which specifically.

## Change record
https://www.drupal.org/node/3026220


modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserAnalyticsGroupMemberships.php
```diff
@@ -0,0 +1,33 @@
...
+   */
+  public function getHeader() {
+    return $this->t('Group memberships');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getValue(UserInterface $entity) {
+    return count(social_group_get_all_group_members($entity->id()));
```
jochemvn: The problem is that you now call the drupal container inside this plugin, because that function does a ` \Drupal::database()->select.....`

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserGroupMemberships.php
```diff
@@ -0,0 +1,40 @@
...
+   */
+  public function getHeader() {
+    return $this->t('Group memberships (specified)');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getValue(UserInterface $entity) {
+    $group_memberships = Group::loadMultiple(social_group_get_all_group_members($entity->id()));
```
jochemvn: Same as above

jochemvn:
Works as designed! See remarks above

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserAnalyticsGroupMemberships.php
bramtenhove: This has been replaced by moving that function to the social_group.helper_service.

modules/social_features/social_user_export/src/Plugin/UserExportPlugin/UserGroupMemberships.php
bramtenhove: This has been replaced by moving that function to the social_group.helper_service.

PR #1155, opened 2018-12-10
Issue #3019458 by robertragas: Give possibility to add additional hea…

robertragas:
## Problem
Currently the notification emails can be quite plain without any additional general information a sitemanager might want to add.
We might want to add the possibility for a sitemanager to add a header or footer to the email notifications.

## Solution
Add a field under the swiftmailer configuration to add a general header or footer with a basic text format to allow 'a', 'em', 'strong', 'u', 'br'. 

## Issue tracker
https://www.drupal.org/project/social/issues/3019458

## How to test
- [ ] Go to admin/config/opensocial/swiftmail through the url bar of interface
- [ ] First save without filling the fields in.
- [ ] Trigger email notification and see fields are not included
- [ ] Fill in the fields with some basic formatting provided by the text format
- [ ] Trigger email notification again and see fields are included
- [ ] Try to add Xss to the fields and see that they are not included

## Additional note
Apart from the restricted text format that I created, it was possible to use the source editor or chrome inspector to add <script> and save without it being properly stripped. This is why I also sanitized it on a deeper level, also making sure that should people switch text format to full basic they cannot include any additional markup.

## Release notes
We have added the possibility to add an additional header or footer to the email notifications, giving you the possibility to give a personal touch to the notifications.


modules/social_features/social_swiftmail/social_swiftmail.module
```diff
@@ -54,6 +55,21 @@ function social_swiftmail_preprocess_swiftmailer(array &$variables) {
     $variables['site_slogan'] = '"' . t('Make your people bloom') . '"';
   }
 
+  // Add variables header and footer if set.
+  if ($social_swiftmail_config->get('template_header')['value']) {
+    // Sanitizing variables first to filter out Xss.
+    $template_header = Xss::filter($social_swiftmail_config
+      ->get('template_header')['value'], ['a', 'em', 'strong', 'u', 'br']);
+    // Create markup so it will be converted to html in the email.
+    $variables['header'] = Markup::create($template_header);
```
Kingdutch: We should replace most of this with:
```suggestion
    $template_header = $social_swiftmail_config->get('template_header');
    $variables['header'] = [
      '#type' => 'processed_text',
      '#text' => $template_header['value'],
      '#format' => $template_header['format'],
    ];
```

Although I realise that this would allow administrators with the `full_html` format to enter any arbitrary HTML, I'll argue that this is desired behaviour. When we trust a user on a website with this text_format then we should not limit what they can do (but merely instruct that something is a bad idea). If you agree then we may be able to remove the mail_html format altogether for any sites that do want to force a certain text format beyond what the distribution offers this can easily be done with an alter hook. If you disagree then we should force the use of html_mail here but that may cause some distribution bug reports.

modules/social_features/social_swiftmail/social_swiftmail.module
```diff
@@ -54,6 +55,21 @@ function social_swiftmail_preprocess_swiftmailer(array &$variables) {
...
+  }
+
+  if ($social_swiftmail_config->get('template_footer')['value']) {
+    $template_footer = Xss::filter($social_swiftmail_config
+      ->get('template_footer')['value'], ['a', 'em', 'strong', 'u', 'br']);
+    $variables['footer'] = Markup::create($template_footer);
```
Kingdutch: We should replace most of this with:
```suggestion
    $template_footer = $social_swiftmail_config->get('template_footer');
    $variables['footer'] = [
      '#type' => 'processed_text',
      '#text' => $template_footer['value'],
      '#format' => $template_footer['format'],
    ];
```

themes/socialbase/templates/swiftmailer.html.twig
```diff
@@ -359,7 +359,13 @@
                             <table border="0" cellpadding="0" cellspacing="0">
                                 <tr>
                                     <td>
+                                        {% if header %}
```
Kingdutch: If we set default empty values for our variables then we may not need the if statement here but can just always print it.

themes/socialbase/templates/swiftmailer.html.twig
```diff
@@ -359,7 +359,13 @@
...
+                                            {{ header }}
+                                        {% endif %}
                                         {{ body }}
+                                        {% if footer %}
```
Kingdutch: If we set default empty values for our variables then we may not need the if statement here but can just always print it.

modules/social_features/social_swiftmail/social_swiftmail.module
Kingdutch: Found in https://drupal.stackexchange.com/a/235028/2166 👍

robertragas:
@Kingdutch I have made the alterations like discussed. Currently, the footer will now be rendered within the activity workers, the header is a render array which needs to be rendered to markup there just because swiftmailer otherwise can't handle with it. And I did some additional tag stripping in the templates. Just to keep it a bit safer than just accepting everything.

jaapjan:
By the way: discussed this last year offline with Robert but it's fine to move the strip tags to a new (low prio) task. It seems out of scope of this one. So you can focus on getting this green and merged ;-)

ronaldtebrake:
Won't make it in 3.10 - Robert will have time planned coming sprint for this though! Added 3.11 so it doesn't slip our attention

robertragas:
The Travis is now green, but alot of moss projects still failing because of patches and what not :(

robertragas:
> I think we lost the momentum to put this in the 3.x branch. Could you mark it for 4.x or 5.x instead? And create a new PR? 🙈

Created a new PR for branch 5.x
https://github.com/goalgorilla/open_social/pull/1239

closing this PR

PR #1158, opened 2018-12-12
#3020038 - Group content manage members tab order weight

ronaldtebrake:
## Problem
Manage members is now the first tab in the group local tabs. Due to #3010778.

## Solution
Made sure that we set the weight only when the tab is available.

## Issue tracker
https://www.drupal.org/project/social/issues/3020038

## How to test
- [x] See that manage members is now the last tab. (Is only there for people with permissions)

## Release notes
The order of the Group tabs is now fixed so Manage Members is right after Members.


jaapjan:
Cherry-picked for 8.x-4.x and 8.x-5.x as well.

PR #1161, opened 2018-12-14
Issue #3020522: Read more link is shown twice in featured

frankgraave:
## Problem
The "Read More" link is shown twice in Featured blocks on the Social Landing Page. This is happening because certain template files in the social_landing_page module are extended from node--featured.html.twig where the this one already has the card__actionbar with the card__link. These elements were also in the extended template file, which caused it to show twice.

## Solution
Remove the card__link element from node--topic--featured.html.twig and node-landing-page--featured.html.twig since only these two had the duplicate elements.

## Issue tracker
https://www.drupal.org/project/social/issues/3020522

## How to test
- [x] Check out the changes
- [x] Create a landing page with featured items, test any or all kinds.
- [x] Verify that you only have one Read More link in the featured card (on all content types).

## Release notes
Fixed an issue where featured blocks in landing pages showed the Read More link twice.

### DIFF
--- a/modules/social_features/social_landing_page/templates/node--event--featured.html.twig
+++ b/modules/social_features/social_landing_page/templates/node--event--featured.html.twig
@@ -26,13 +26,4 @@
         </span>
       {% endif %}
     </small>
-
-
-  {% endblock %}
-  {% block card_link %}
-    <div class="card__link">
-      <a href="{{ url }}" rel="bookmark">{{ 'Read more'|t }}
-        <span class="visually-hidden">{{ 'about'|t }}{{ label }}</span>
-      </a>
-    </div>
   {% endblock %}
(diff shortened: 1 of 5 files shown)


Kingdutch:
This seems to have been introduced in #998

robertragas:
Looks logical, but not sure why this got introduced in the first place.

jaapjan:
Merged it to 8.4 and 8.5 branches. If you think it should be in 3.x as well, you can cherry-pick `9e9f35b` and push it to 8.x-3.x

jaapjan:
> We can actually get away with removing these blocks from node--event--featured.html.twig and profile--featured.html.twig. There the blocks were properly changed but they don't actually differ from what they're inheriting.

Btw, this is not completely true. We need that code in profile--featured because otherwise you get that issue again:
https://www.drupal.org/project/social/issues/3019760

Will solve it in a separate PR.

PR #1163, opened 2018-12-14
#3020575: Fix configuration schema for modules/social_features/social_page/config/install/core.entity_form_display.node.page.default.yml

ideb:
PR for https://www.drupal.org/project/social/issues/3020575

## Release notes
The configuration value core.entity_form_display.node.page.default.content.status.settings.display_label was installed as an integer but it is now converted to a boolean.


jaapjan:
Thank you for the PR, it works. Also checked all the other entity form templates and there it is already a boolean. Furthermore, will also cherry-pick for 3.x and 5.x :-)

PR #1164, opened 2018-12-17
3020943: Fix image styles that use the wrong effect

jochemvn:
## Problem
Fix image styles that use the wrong effect. Preview images look skewed. Also when mass importing images, the images itself get skewed. 
The interface was working fine btw.

## Solution
Make sure the image styles use the `image_scale_and_crop` effect rather than then `image_resize` effect.

## Issue tracker
- https://www.drupal.org/project/social/issues/3020943

## How to test
- [ ] Look at the code changes
- [ ] Create an image in code (profile load > file create > profile set new image > profile save)
- [ ] In the old way, you can see on `/all-members` that images are skewed
- [ ] Go to the profile and notice the images are good
- [ ] Checkout to this branch
- [ ] Do the same
- [ ] See all is good

## Release notes
Some images used to be resized using the image_resize, rather than the scale and crop typ. This could cause images to look skewed. This has now been fixed.

### DIFF
--- a/modules/social_features/social_core/config/install/image.style.social_an_hero.yml
+++ b/modules/social_features/social_core/config/install/image.style.social_an_hero.yml
@@ -17,9 +17,10 @@ effects:
       crop_type: hero_an
   c4885faf-b34c-446f-b95b-e0a43dd71080:
     uuid: c4885faf-b34c-446f-b95b-e0a43dd71080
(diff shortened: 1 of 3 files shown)


jaapjan:
Cherry-picked in 4.x and 5.x as well.

PR #1167, opened 2018-12-19
GPI - 6 geolocation filter

ronaldtebrake:
## Problem
WE have a separate module available at https://www.drupal.org/project/social_geolocation

## Solution
But we want to make sure no tests fail, therefore committed the work in here and moved this to 8.x-1.x in https://www.drupal.org/project/social_geolocation after. To see if we don't break tings. 

## Issue tracker
https://www.drupal.org/node/3019950

## How to test
- [x] Enable social_geolocation
- [x] See that there are now 2 options for geolocation data providers on : `/admin/config/opensocial/social-geolocation` OpenStreetMaps is the default, this is free so doesn't need an API key. for PAC we also added Google, and made sure on hook_update_n() we cover their API key as well. 
- [x] See that when you now save location details on Events, User Profiles, Groups they get stored in the database under a new hidden field_[type]_geolocation 
- [x] See that when you go to search there is now a search block with the new filter and see if it works according to your filled in data
- [x] See that on admin/people also there is a filter right now that works :) 

## Acceptance criteria it should match

- Users location is stored automatically based on address details when saving profile (in background)
- Each NRO fills in a Google Maps API key which can be used (or delivered by GPI?) => SM should be able to fill this in
- LU can filter on location on /search/users page by providing a city and a distance (in kilometers/miles) 
- Should work by default for events and groups as well:
-- Filter on content, groups and user tabs in search
- Publish the module on drupal.org as social_geolocation project
- It should be clear that the two fields in the filter belong to each other

## TODO
- [ ] Make it in to a nice Views custom filter
- [ ] Sometimes at random addresses can't be found, something to improve in UAT maybe. 
- [ ] Make a new note for demo content, the ones in distro have non existing addresses, would be lovely to make them realistic 
- [ ] Opened up a support request to make sure PAC also gets this module when finished. 

## Release notes
TBD.


composer.json
```diff
@@ -133,6 +133,7 @@
         "npm-asset/jquery.caret": "^0.3.1",
         "npm-asset/diff": "^3.5",
         "npm-asset/photoswipe": "^4.1.2",
-        "npm-asset/shariff": "^3.0.1"
+        "npm-asset/shariff": "^3.0.1",
+        "drupal/geocoder": "^2.0@beta"
```
ronaldtebrake: Only there for testing purposes, part of the composer.json and dependencies of the module :) but easier install for us :D

PR #1168, opened 2018-12-20
Issue #3021721: As an Event organiser, I want to manage event enrollment easily

ghost:
## Problem

- See events enrollments tab in a table view (columns should be sortable)
- Bulk operation on enrollees:
  - Remove
  - Change role
  - Send email
  - Export enrollments
- Optimise interaction for mobile devices
- Bulk operations should work on guest enrolments as well.
- Solve tabs issue on event (guest enrolments, enrolments):
  - Solution could be to get rid of guest enrolments tab

## Solution
Create a feature for that.

## HTT 
- [ ] Enable the social_event_manager module
- [ ] Test as a Event Organiser (Not Event author!) and a SM 
- [ ] There is now a new event management tab
- [ ] You can bulk perform the above 4 items
  - Remove
  - Change role
  - Send email
  - Export enrollments
- [ ] You can singular Edit / Delete event enrollment
- [ ] AN Event enrollment is also supported 

## Issue tracker
- https://www.drupal.org/project/social/issues/3021721
- https://jira.goalgorilla.com/browse/GPI-10

## Release notes
We have added event management features to the event managers module. This allows you to do bulk operations on event enrollment, including removing, exporting and sending e-mails. If you are a site manager you can even add enrollees to the event. Also we have simplified the amount of tabs on the events. This feature now works together with the anonymous event enrollment feature and it will replace the existing Enrollments tab with a table of event enrollments to improve the information density.


ronaldtebrake:
Finally with the fixes for 5.x the travis jobs are green, so lets merge this one :)!

PR #1173, opened 2018-12-28
Issue #3020007 : prevent error on post list when field_post does not exist

jaapjan:
## Problem
If the field `field_post` does not exist on a post it throws an error on `admin/content/post`.

## Solution
Let's check if the entity has the field before trying to get the field value from the entity.

## Issue tracker
https://www.drupal.org/project/social/issues/3020007

## How to test
- [ ] Login as user 1, remove the field `field_post` from the post or post_photo entity.
- [ ] Check `admin/content/post` and notice an error.
- [ ] Switch to this branch and notice the error is gone.

## Release notes
The `admin/content/post` overview only worked if all the post entities had the field `field_post. Now it checks if the field exists before trying to display the field value.


jaapjan:
Also cherry-picked in 4.x and 5.x.

PR #1176, opened 2018-12-28
Issue #3018268 by jaapjan: don't send notification for liking own content

jaapjan:
## Problem
You get a notification when liking your own entity.

## Solution
Let's implement the same solution as we did in #464 this will prevent an actor from being marked as recipient.

## Issue tracker
https://www.drupal.org/project/social/issues/3018268

## How to test
- [ ] Like your own post, comment and topic
- [ ] Notice you don't receive a notification = great success.
- [ ] You can debug with Xdebug as well if needed.

## Release notes
From now on you will not receive notifications when you like your own comment or post.


ronaldtebrake:
Cherry picked in 4.x and 5.x

PR #1180, opened 2019-01-14
Fix styling for social sharing buttons

Kingdutch:
<h2>Problem</h2>
Switching from AddToAny to Shariff caused a visual regression because the styling that was present in the AddToAny button was not replicated for Shariff.

Related PR: #1100 
See: https://www.drupal.org/files/issues/2019-01-14/Screenshot%202019-01-14%20at%2012.44.10.png

The Shariff buttons don't have rounded corners or centered icons by
default which causes a regression compared to the previously present
addtoany buttons.

<h2>Solution</h2>

Add the centering of icons to Socialbase and the default border-radius to Socialblue. Also ensure that this changes with the button rounding setting in the theme settings.

Result: https://www.drupal.org/files/issues/2019-01-14/Screenshot%202019-01-14%20at%2012.44.30.png

This commit adds a default border-radius for the socialblue theme and
uses the socialbase theme to center the icons. It also ensures that when
a theme sets a custom border-radius for buttons that it's honored on the
sharing buttons.

## Issue tracker
https://www.drupal.org/project/social/issues/3025741

## How to test
- [x] Using the socialblue theme
- [x] Ensure the social_sharing module is enabled
- [x] Visually inspect the "Share this page" block on a public topic
- [x] Change the border-radius under theme settings for socialblue
- [x] Check that the border-radius changes.

## Release notes
The move to Shariff for our social sharing buttons caused a visual regression. Styling has been added to the themes and theme customisation module to ensure that these buttons are properly displayed again.

### DIFF
--- a/modules/custom/improved_theme_settings/improved_theme_settings.module
+++ b/modules/custom/improved_theme_settings/improved_theme_settings.module
@@ -161,6 +161,9 @@ function improved_theme_settings_page_attachments(array &$page) {
             border-radius: 50% !important;
           }
 
+          .shariff-button {
+            border-radius: ' . $button_radius . 'px !important;
+          }
         ';
       }
 
--- a/themes/socialbase/assets/css/button.css
+++ b/themes/socialbase/assets/css/button.css
@@ -414,6 +414,10 @@ input[type="button"].btn-block {
   display: inline-block;
 }
 
+.shariff-button {
+  text-align: center;
+}
+
 @media (min-width: 900px) {
   .block-group-add-event-block,
   .block-group-add-topic-block,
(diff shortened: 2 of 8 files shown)


bramtenhove:
@Kingdutch, looks good. Tests do fail though.

Kingdutch:
@bramtenhove I believe the Topic-preview test is a false positive on more test runs so we may need to look at that and add a wait for AJAX after the preview page load? :)

I'll look at MOSS tomorrow

Kingdutch:
Travis is now working after retrying the false positives. Moss is failing on unrelated issues (e.g. missed patches etc.) so I'm merging this :)

PR #1183, opened 2019-01-16
Hotfix PSA - 2019-001 and 002

ronaldtebrake:
## Problem
Security update SA 2019-001 and 002.
Checking if patches apply on the new double security release. After we hotfix it.


ronaldtebrake:
All green, woop woop!

PR #1184, opened 2019-01-21
3027295 - Alternate landing/link options for a group

jochemvn:
## Problem
When clicking on a link to go to the group, member and non-members always land on the stream and on closed groups, non-members land on the about page. There no way to change this behaviour if you'd want that

## Solution
Added 2 fields to groups that control where a member lands and one to control where a non-member lands.

## Issue tracker
https://receptive.io/app/#/case/56262
https://www.drupal.org/node/3027295

## How to test
- [ ] Checkout to this branch
- [ ] Check the code
- [ ] Enable the module
- [ ] Go to an edit group page and notice 2 extra fields
- [ ] Select some alternative routes for a groups (also test some permissions things like: a route for a non-members that he/she cannot access)
- [ ] Notice that members and non-members land correctly on the alternative route from the group. 
- [ ] Notice it also works on landing pages
- [ ] Notice that when you've selected a route that cannot be accessed by a user, the user wil go to the original default route.

## Release notes
When enabling the optional `social_group_default_route`module, group manager can now choose where members and non-members land when they click on a group link anywhere in the platform.
The group edit screen will be extended with two extra fields for a GM+ to choose where non-members and members will land.

For developers there are 2 hooks added. By default this functionality will be available for all group types. This behaviour can be changed with the `hook_default_route_group_types_alter` hook.
This functionality will also only show localtask items from the `entity.group.canonical`. This can also be altered. To alter this user the `hook_default_route_group_tabs_alter` hook.


jaapjan:
Moved to:
https://github.com/goalgorilla/open_social/pull/1197

PR #1185, opened 2019-01-21
Issue #2977782: My own Private Messages are marked as unread

ghost:
## Problem
When I went to my private message inbox today, a number of old conversations were newly marked as unread. Viewing those messages revealed that I was the last person to communicate in those threads. The viewing process did remove the green indicator dot in the inbox.

Then I created a new message and now it displays as unread in my inbox.

## Solution
My expectation is that only messages I receive would display as unread.

## Issue tracker
https://www.drupal.org/project/social/issues/2977782

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,8 @@
                 "Fix preview on node": "https://www.drupal.org/files/issues/2848080-2-preview-fails-on-node.patch"
             },
             "drupal/private_message": {
-                "Private message unread messages fix": "https://www.drupal.org/files/issues/2018-06-08/2978324-getthreads-sort-order-2.patch"
+                "Private message unread messages fix": "https://www.drupal.org/files/issues/2018-06-08/2978324-getthreads-sort-order-2.patch",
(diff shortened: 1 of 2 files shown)


Kingdutch (approved):
Patch has been reviewed by multiple people in the private_message queue. I've used a slightly newer version of the patch than the one by Oleksandr. I've also tested this in a distro installation and it works as expected.

jaapjan:
Note: this will not solve the issue for people with existing installs who run `composer update goalgorilla/open_social --with-all-dependencies` because the version of private message is not updated the new patch won't be found and applied.

jaapjan:
Doing a `rm -rf html/modules/contrib/private_message` before running composer update works :-) Will merge it anyway because it solves the issue. But for existing installs this extra step is required. I'm not aware of any other work-around :-)

jaapjan:
Also cherry-picked `31390f2` in 8.x-5.x and 8.x-3.x

jaapjan:
@OH17L could you also write release notes next time? Thanks in advance! :-)

OH17L:
Sure.

PR #1186, opened 2019-01-21
Issue #2992737: bundle inside the like modal title is not translatable

jaapjan:
## Problem
Bundle is not translatable inside like modal window

## Solution
Add a t() function around it.

## Issue tracker
https://www.drupal.org/project/social/issues/2992737

## How to test

## Release notes
Bundle is not translatable inside like modal window, this should be translatable again now.


jaapjan:
Not working because coding check is failing. `Only string literals should be passed to t() where possible`

PR #1187, opened 2019-01-21
Issue #3027247: Call to function isAnonymous() on null in template_preprocess_activity()

jaapjan:
## Problem
Call to function isAnonymous() on null in template_preprocess_activity()

## Solution
Only run the code when the `$account` is set.

## Issue tracker
https://www.drupal.org/project/social/issues/3027247

## How to test
- [ ] Check if notifications still work as expected.

## Release notes
Generation/displaying of activity items could in some cases result in an error on page, this is now solved by adding some more defensive coding.


jaapjan:
Also cherry-picked to 3.x and 5.x

PR #1188, opened 2019-01-22
Issue #2968506 by uros.ceh, jaapjan, truls1502: When group is deleted the posts in this group should be deleted as well

truls1502:
## Issue tracker
https://www.drupal.org/project/social/issues/2968506

## Release notes
When a group of any type (closed, open, public) is deleted the posts and content in this group are also deleted.

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -526,7 +526,7 @@ function social_group_form_alter(&$form, FormStateInterface $form_state, $form_i
     if (in_array($form_id, $group_forms['delete'])) {
       // Add custom submit handler to delete all content of the group.
       $group = _social_group_get_current_group();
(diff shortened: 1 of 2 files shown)


ronaldtebrake:
Manual tests work like a charm, could be the behat tests has a wrong confirmation message now? Refired travis PR to verify it still fails on the group-edit-open.

jaapjan:
FYI:
I've made a change which should solve that issue :-)

Side effect of this entire change is that when a public, open or closed group is deleted that all the posts and nodes within that group are also removed. I think that's ok. But we should make that clear in the release notes. I think it's OK if we don't add this to 3.x what do you think @ronaldtebrake ?

ronaldtebrake:
Sounds great and i agree lets keep it in 4.x +

PR #1179, opened 2019-01-08
Issue #3024525 by Kingdutch, bramtenhove: Fix project names in fully qualified dependencies

Kingdutch:
## Problem
When using the project:module_name syntax for dependency declaration the
`project` part should reflect the name under which the project can be
found on drupal.org. For all modules contained in Open Social this is
thus the `social` project.

## Solution
This change allows automated systems (e.g. automatic updates or
translation servers) to find the correct module even if a module is
contained in multiple projects.

## Issue tracker
https://www.drupal.org/project/social/issues/3024525

## How to test
- [ ] Check that all dependencies are still recognized

## Release notes
Code format change only, release notes not needed.

PR #754, opened 2018-02-26
Issue #2936477: fix dependency declaration in social.info.yml

jaapjan:
## Problem
The info file contains dependencies without the drupal: declaration. 

## Solution
All dependencies must be prefixed with the project name.

## Issue tracker
- https://www.drupal.org/project/social/issues/2936477

## HTT
- [x] Check out the code changes
- [ ] Do a new install, check that it works and the modules are enabled.

## Documentation
- [ ] This item is added to the release notes
- [ ] This item is documented on LGOS
- [ ] This item has been added to the feature overview

PR #1190, opened 2019-01-22
Issue #2936477 : dependency declaration in social.info.yml

jaapjan:
All dependencies must be prefixed with the project name, for example "drupal:"

## Issue tracker
https://www.drupal.org/project/social/issues/2936477

## How to test
- Verify installation works still as expected, wait for Travis to be completed.

## Release notes
(developer) All dependencies in social.info.yml are now with the project name, for example "drupal:".

### DIFF
--- a/social.info.yml
+++ b/social.info.yml
@@ -6,11 +6,11 @@ version: '8.x-4.3'
 project: social
 
 dependencies:
-  - config
-  - system
-  - user
-  - breakpoint
-  - features
+  - drupal:config
+  - drupal:system
+  - drupal:user
+  - drupal:breakpoint
+  - features:features
 themes:
   - seven
   - socialbase


social.info.yml
```diff
@@ -6,11 +6,11 @@ version: '8.x-4.3'
...
-  - config
-  - system
-  - user
-  - breakpoint
-  - features
+  - drupal:config
+  - drupal:system
+  - drupal:user
+  - drupal:breakpoint
+  - features:features
```
jochemvn: not `drupal:features` ?

jaapjan: I think the drupal: namespace is reserved for Drupal core modules, so it should be modulename:modulename for contrib afaik. Fact is that it didnt work in https://github.com/goalgorilla/open_social/pull/754 however, so let's wait and see 🤞

truls1502: Travis tests have passed ✅

jochemvn: > I think the drupal: namespace is reserved for Drupal core modules, so it should be modulename:modulename for contrib afaik. Fact is that it didnt work in #754 however, so let's wait and see 🤞

Sounds legit 👌

jaapjan:
Cherry-picked for 3.x and 5.x as well.

social.info.yml
Kingdutch: As described in #1179

```
When using the project:module_name syntax for dependency declaration the
project part should reflect the name under which the project can be
found on drupal.org. For all modules contained in Open Social this is
thus the social project.
```

PR #1195, opened 2019-01-22
Issue #3026830: language switcher not switching

jaapjan:
Note: only needed for 4.x and 5.x.

## Problem
The language switcher doesn't switch the language anymore in the latest release.

## Solution
It seems like the logic is flawed in the SocialLanguageMetadataBubblingUrlGenerator class with the reset language functionality. Currently, it resets to the current language on every page except for content_translation_overview pages, but it should be reversed instead. The patch above should fix the problem.

## Issue tracker
https://www.drupal.org/project/social/issues/3026830

## How to test
- [ ] Enable social_language
- [ ] Enable a new language
- [ ] Place the social language switcher block in the Header
- [ ] Try to switch the language and notice it doesn't work
- [ ] Switch to this branch and clear cache
- [ ] Notice switching the language now works.

## Release notes
The language switcher didn't work as expected. It should switch the languages now.

### DIFF
--- a/modules/custom/social_language/src/SocialLanguageMetadataBubblingUrlGenerator.php
+++ b/modules/custom/social_language/src/SocialLanguageMetadataBubblingUrlGenerator.php
@@ -46,7 +46,7 @@ class SocialLanguageMetadataBubblingUrlGenerator extends MetadataBubblingUrlGene
       $language = $this->languageManager->getCurrentLanguage();
 
       if ($options['language']->getId() != $language->getId()) {
-        $reset_language = TRUE;
+        $reset_language = FALSE;
         $unmodified_pages = [
           'content_translation_overview',
         ];
@@ -56,7 +56,7 @@ class SocialLanguageMetadataBubblingUrlGenerator extends MetadataBubblingUrlGene
         $route_parts = explode('.', $current_route);
         foreach ($unmodified_pages as $page) {
           if (in_array($page, $route_parts)) {
-            $reset_language = FALSE;
+            $reset_language = TRUE;
             break;
           }
         }


jaapjan:
@Kingdutch @robertragas Could you guys review this as this is only used on one enterprise site and not (yet) in saas or any other sites I maintain it's hard to determine the impact.

jaapjan:
Btw: it did solve the issue for me :-) Just not sure if the solution is correct.

robertragas:
@jaapjan with my installation it still not switches correctly. I can switch a first time to Dutch and that goes correctly, but after that the english path keeps saying /nl

Kingdutch:
Sorry for not getting back to this sooner. I have no time to test this but I can share my thoughts.

The reason that this code exists is as follows. There is code in Open Social to ensure non-translated content (e.g. posts) are always displayed in the currently selected language. This works because the fallback behaviour for Drupal when displaying an entity in an untranslated language is to show the default (English language). However, because it's been attempted to display in a different language, any links will preserve the current language selection.

If we were to rely on Drupal's default behaviour then any item clicked in the stream (post or activity) would always switch the language to English (which is the language of the entity that is being displayed). This would make using the site in another language like Spanish impossible because you'd constantly revert back to English.

The code that overwrites this has a drawback however. It is also in action on the "Translate entity" page. This causes a content manager to be unable to translate an entity because both the original entity and its translation will always link to the version of the currently selected language for the viewing user. To counter-act this the code that is changed in this PR has been created. This should reset the languages from our override in the correct moment so that language switching is allowed in order to translate content.

I'm not sure what causes the issue with the language switcher but believe it is somewhere in the current implementations. **I have the feeling that the change in this PR actually does not resolve the issue** but simply disables the fix that I have described above. This means that the bug in the workaround that causes the language switcher to function is removed but in turn the translation of content is probably broken again.

**Things to test when testing this PR** 
Check on a multi-lingual site whether:
- When viewing the stream with English posts in Spanish and clicking a post you're still viewing the site in Spanish
- The language switcher works as expected
- When viewing the site in English you are able to add a Spanish translation to a topic
- When viewing the site in English you are able to edit a Spanish translation of a topic
- When viewing the site in Spanish you are able to add a Spanish translation to a topic
- When viewing the site in Spanish you are able to edit a Spanish translation of a topic
- When viewing the site in English you are able to add a English translation to a topic
- When viewing the site in English you are able to edit a English translation of a topic
- When viewing the site in Spanish you are able to add a English translation to a topic
- When viewing the site in Spanish you are able to edit a English translation of a topic


I hope this can move the resolution of this issue forward.

PR #1196, opened 2019-01-22
Issue #3027600 by Kingdutch: Set font-fallback for Montserrat font to…

Kingdutch:
… sans-serif

<h2>Problem</h2>
Montserrat is by default configured to fallback on a 'serif' style font when loading. This causes a jarring experience for users because Montserrat itself is a 'sans-serif' font.

<h2>Solution</h2>
Set the default for the Montserrat font entity to fallback to a 'sans-serif' font. Also update existing Montserrat font entities to use this new default fallback. We suspect that few users will actually know about this setting and those that do probably have other fonts configured than Montserrat that will remain unchanged.

## Issue tracker
https://www.drupal.org/project/social/issues/3027600

## How to test
- [ ] Check the fallback for the montserrat font as set by the social_font module

## Release notes
Montserrat is a beautiful sans-serif font that is used in Open Social by default. When loading this font, we used to fallback to a serif font which can look odd. We've changed the Montserrat font configuration to fallback to a sans-serif font instead.


bramtenhove:
@Kingdutch, this still requires some Moss work

Kingdutch:
@bramtenhove I hadn't planned to as the code in this PR isn't in any test coverage at the moment. 🤔

PR #1197, opened 2019-01-23
3027295 - Alternative entry point for groups

jochemvn:
## Problem
When clicking on a link to go to the group, member and non-members always land on the stream and on closed groups, non-members land on the about page. There no way to change this behaviour if you'd want that

## Solution
Added 2 fields to groups that control where a member lands and one to control where a non-member lands.

## Issue tracker
https://receptive.io/app/#/case/56262
https://www.drupal.org/node/3027295

## How to test
- [x] Checkout to this branch
- [x] Check the code
- [x] Enable the module
- [ ] Go to an edit group page and notice 2 extra fields
- [ ] Select some alternative routes for a groups (also test some permissions things like: a route for a non-members that he/she cannot access)
- [ ] Notice that members and non-members land correctly on the alternative route from the group. 
- [ ] Notice it also works on landing pages
- [ ] Notice that when you've selected a route that cannot be accessed by a user, the user wil go to the original default route.

## Release notes
When enabling the optional `social_group_default_route`module, group manager can now choose where members and non-members land when they click on a group link anywhere in the platform.
The group edit screen will be extended with two extra fields for a GM+ to choose where non-members and members will land.

For developers there are 2 hooks added. By default this functionality will be available for all group types. This behaviour can be changed with the `hook_default_route_group_types_alter` hook.
This functionality will also only show localtask items from the `entity.group.canonical`. This can also be altered. To alter this user the `hook_default_route_group_tabs_alter` hook.

## Change record
- https://www.drupal.org/node/3039071


jaapjan:
FYI There seems to be a merge conflict here

robertragas:
Noticed it now myself when working on a similar functionality for profiles.
When choosing for example about as the group landingpage you as expected land on the about page, but when you are there and then click on stream it will redirect again to the landing page.

jochemvn:
Take into account more use cases. 
- Coming from dedicated link outside group may not redirect
- Coming from link inside group you may also not redirect

jochemvn:
@ronaldtebrake / @robertragas The issue of the redirect has been fixed by adding an extra 'empty' route in between. This route is only there to determine where to send the user.

modules/social_features/social_group/modules/social_group_default_route/social_group_default_route.install
```diff
@@ -0,0 +1,18 @@
...
+ * Installation file for social_group_default_route module.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function social_group_default_route_install() {
+  $extension_config = \Drupal::configFactory()->getEditable('core.extension');
+  $group_weight = $extension_config->get('module.social_group');
+  // One more plz.
```
ribel: 😂

modules/social_features/social_group/modules/social_group_default_route/social_group_default_route.module
```diff
@@ -0,0 +1,158 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\group\Entity\GroupType;
+
+/**
+ * Implements hook_form_alter().
+ */
+function social_group_default_route_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+
+  $forms = [];
+  foreach (_social_group_default_route_group_types() as $type) {
```
ribel: @jochemvn Is it safe to use constructions like this? What If this function will return an empty array or not array?

modules/social_features/social_group/modules/social_group_default_route/social_group_default_route.install
jaapjan: 1️⃣ 🆙

jochemvn: 🙃

modules/social_features/social_group/modules/social_group_default_route/social_group_default_route.module
jochemvn: 👌  I've added an social_group_default.api.php file to document the hook

PR #1198, opened 2019-01-24
Issue #3028153: Profile Tag List not displaying correctly in filter on user search page

OH17L:
## Problem
Taxonomy terms sorting by IDs.

## Solution
Sort taxonomy terms by weight and name.

## Issue tracker
https://www.drupal.org/project/social/issues/3028153

## Release notes
It's now possible to create multiple levels of Profile Tags. Similarly to content tagging the top level taxonomy term can now be used as a category by giving it children. When only a single level of profile tags exist the current behaviour will be preserved. When a user receives a taxonomy term in a category then they will also receive the tag of the category itself to facilitate searching.

### DIFF
--- a/modules/social_features/social_profile/social_profile.module
+++ b/modules/social_features/social_profile/social_profile.module
@@ -90,19 +90,14 @@ function social_profile_form_profile_profile_add_form_alter(array &$form, FormSt
  * Implements hook_form_FORM_ID_alter().
  */
 function social_profile_form_profile_profile_edit_form_alter(array &$form, FormStateInterface $form_state, $form_id) {
-  $user = \Drupal::currentUser();
-  $account = \Drupal::routeMatch()->getParameter('user');
+  $viewing_user = \Drupal::currentUser();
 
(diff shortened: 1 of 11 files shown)


Kingdutch:
This still needs theming. Additionally the current behaviour doesn't properly handle nested taxonomy terms so more changes may be needed.

jaapjan:
Also marked the d.o issue as "Needs work". When do you guys plan to add the theming?

Kingdutch:
@jaapjan Turned out nested tags weren't actually supported yet so this will need more than just theming. It's in our current sprint.

Kingdutch:
I've added a suggestion for release notes but this may still need some tweaking. Other than that the development work for this feature is complete.

Follow ups:
- Hide category tags from the user teaser
- Ensure that search uses `AND` for filtering tags instead of `OR`. This ensures that the number of search results decrease when more tags are clicked (this is considered to be a bug in the current search behaviour after talking to @ronaldtebrake).

bramtenhove:
Cherry-picked in `8.x-3.x` and `8.x-5.x`.

PR #1202, opened 2019-01-28
Issue #3028992: Mentions and embed media in the comment

ghost:
## Problem
When adding mentions and images in one comment then will be showing broken markup below image and comment in a database contain CKEditor plugin markup.

## Solution
Use a source of comment instead of markup from CKEditor content for comment data.

## Issue tracker
https://www.drupal.org/project/social/issues/3028992

## How to test
- [ ] When creating comment with mention and image then you shouldn't see broken markup below image.

## Release notes
When a mention was added to the same comment as an image then this sometimes caused a glitch in the matrix, exposing the page's inner workings. This has now been fixed.


Kingdutch (changes requested):
I think we can slightly improve how we get the data but other than that it works as expected and fixes the issue.

Kingdutch (approved):

Kingdutch:
It does look like this PR changes an external library. I'm unable to find the source to see if they may have already fixed this issue.

My proposal would be to merge this as is. To ensure that we don't accidentally revert these changes, we could consider just throwing away the coffeescript file as we don't use this anywhere else in Open Social. I'm attempting to standardise on ES6 using babel/webpack for our JS source files.

The jQuery specific plugins can possibly be replaced by React components in the distant future anyway.

(cc @ronaldtebrake for your input 😄)

ronaldtebrake:
Sounds like a good idea @Kingdutch, also can't seem to find this external library. @OH17L Don't forget to add release notes next time.

Kingdutch:
@ronaldtebrake Removed the coffeescript files. Added release notes. Ignoring the MOSS tests because we don't test this functionality currently.

Merging :)

PR #1203, opened 2019-01-28
#2983840 - Crop API fixes + Update image_widget_crop + Move patch 

ronaldtebrake:
## Problem
We use a patch from the Image Widget Crop module which included some incorrect logic which became apparent for square crops. However it turns out that the initial logic is flawed as this fixes cropping for square profile pictures but introduces other issues for very wide images with a wide crop that results in a higher crop than the image is wide. This can be experienced by uploading a very low and wide image into a topic. The earlier issue was for narrow and high profile images.

An additional problem is that the issue in Image Widget Crop is closed as "Wont't Fix". The maintainer of the Image Widget Crop and the Crop module suggests to implement it in the Crop module itself.

## Solution
- Updated image widget crop 
- Removed the patch for image widget crop
- Applied patch for crop api
- Aligned issues with https://github.com/goalgorilla/open_social/pull/1164 which seem to fix a lot as well. 

## Issue tracker
- https://www.drupal.org/project/social/issues/2983840

## How to test
- [x] Check different images and different resolutions / aspect ratios
- [x] Upload them on a profile / topic
- [x] See that for all the image styles (hero / teaser / group member / post form / profile stream) the images get cropped well automatically. So don't apply any manual cropping.
- [x] I have checked the following myself for the above scenario's and they now seem to work well:
-- 750x3171
-- 3044x4565
-- 3693x3693
-- 5184x3456
-- 1120x373 

## Release notes
Cropping images without manually selection has resulted in many images being saved incorrectly. Sometimes it was transformed incorrectly, sometimes it was very pixelated or resized incorrectly. We have made an attempt to fix all of these issues in this release.


ronaldtebrake:
Part of that follow up could also be to checkout https://www.drupal.org/project/automated_crop
This adds some new plugins for automated crop effects we could potentially benefit from. For now we didn't use it since all of the above tests were done with out it and it doesn't seem necessary as of now.
But wanted to mention it in here for when we look back at this issue.

ronaldtebrake:
And cherry-picked to 4 and 5.x ;)

PR #1205, opened 2019-01-29
Issue #3024525 by Kingdutch: Remove cyclic dependencies and fix names…

Kingdutch:
## Problem
A few modules contained a cyclic dependency, preventing them from being uninstalled. This was discovered while trying to uninstall the social_group_gvbo module and then analysed with the [Drupal Dependency Deduplicator](https://github.com/goalgorilla/drupal-dependency-deduplicator).

## Solution
Remove the cyclic dependencies where they make sense and while at it finish the work that was started in #1179 

## Issue tracker
https://www.drupal.org/project/social/issues/3024525

## How to test
- [x] Try to install and uninstall the affected modules.

## Release notes
There was an error in the declaration of the dependencies for the social_landing_page, social_user and social_group_gvbo module that has now been fixed, allowing these modules to be uninstalled when needed.


modules/social_features/social_landing_page/social_landing_page.info.yml
```diff
@@ -3,65 +3,35 @@ type: module
...
+  - image_effects:image_effects
+  - image_widget_crop:image_widget_crop
+  - paragraphs:paragraphs
+  - social:social_core
+  - social:social_event
+  - social:social_group
+  - social:social_page
+  - social:social_post
+  - social:social_profile
+  - social:social_topic'
```
bramtenhove:
```suggestion
  - social:social_topic
```

modules/social_features/social_group/social_group.info.yml
```diff
@@ -21,7 +21,6 @@ dependencies:
   - social:social_event
   - social:social_profile
   - social:social_topic
-  - social:social_group_gvbo
```
bramtenhove: I can see why it is removed here, but I do think we want this module enabled by default right @ronaldtebrake?
Is that still taken care of @Kingdutch?

ronaldtebrake: Yes we want this enabled by default. Install profile should be fine!

Kingdutch:
The `social_group_gvbo` module should be added to the `social_install_profile_modules` hook.

Kingdutch:
@bramtenhove @ronaldtebrake Your feedback has been incorporated and I believe MOSS is not caused by this story (but simply broken in general 🙁). Please merge this after final review or let me know if work is needed.

PR #1206, opened 2019-01-31
DS - 6384 - As a LU, I want to see the event title in a map follow the platform colour configuration.

ronaldtebrake:
## Problem
Currently the event title in a map is blue, when the site configuration does not contain this colour. We should make it follow the primary colour of the platform

## Solution
Change the marker title text color to `$navbar-default-link-color;`

## How to test
- [x] https://gpi-december.getopensocial.com/community-events (See that it works there perfectly now) 

## Release notes
Nothing needs to be added.


jaapjan:
It looks good. Note: release notes shouldn't be necessary since this is a new feature which should already be in the release notes.

Feel free to merge it after Travis bug is solved (which seems unrelated anyways)

ronaldtebrake:
Merging 8.x-5.x back in tot his solved the travis conflict :) after travis = green can be merged!

PR #1210, opened 2019-02-07
Issue #3031330 by Kingdutch, bramtenhove, jaapjan: Fix html entities and user result type in search suggestions

Kingdutch:
## Problem
The user type is not shown. HTML entities are escaped twice.

## Solution
The "User" entity type is not visible for non-admin users so we rewrite the user_id field (visible for everyone and only present for user entities) when it's filled instead.

Double escaping is caused by the REST API which escapes its output
within the JSON encoding. This is currently an open Drupal core bug. To
work around this we decode the entities first when receiving the
response and then just let React do its normal escaping.

See https://www.drupal.org/project/drupal/issues/2928793

## Issue tracker
https://www.drupal.org/project/social/issues/3031330

## How to test
- [x] Create a topic for Michael O'Connor & Señor Dóge back together
- [x] Search for "Robert" and see that the "User" type isn't displayed in the search suggestions
- [x] Search for "Michael" and see that the special characters in the topic are doubly escaped
- [x] Apply the fixes in this PR
- [x] Repeat the searches and observe the issues have been resolved.

## Release notes
No release notes because this fixes issues in an as-yet unreleased feature.


jaapjan:
Thanks for the PR! But I'm removing the bug label and change it to task since this is an unreleased feature. 🙈

Kingdutch:
Yay metrics! ^^

bramtenhove:
@Kingdutch, will you fix the Moss tests?

Kingdutch:
@ronaldtebrake I think this can be merged. I'm not sure why MOSS fails but it seems to be unrelated to this PR. Especially since this in a module that's not enabled for existing installs.

If this was an issue with installation then I would've expected it to fail in Travis as well.

Kingdutch:
This PR makes changes to modules that are not in use on the platforms that are failing in MOSS. The failures also appear to be for unrelated reasons. After talking to Ronald I'm merging this in its current testing state.

PR #1212, opened 2019-02-08
Update drupal core in make file to 8.6.9. (4.x branch)

jaapjan:


jaapjan:
Btw I've checked Drupal Core on 5.x branch and there it's working as expected. I think we can merge it when Travis is green.

PR #1217, opened 2019-02-12
Issue #3032327: As a LU the tags block should be updated after I add a tag to a group/node

ghost:
## Problem
After adding a second tag to a group/node, the block on node/group view is not updated with the new tags.

## Solution
Add the group to the block context for correct caching by group entity.

## Issue tracker
https://www.drupal.org/project/social/issues/3032327

## How to test
- [ ] When you add different sets of tags to a few groups then you see first set for all groups.

## Release notes
Show own tags set of a group instead of a set of the first cached group.

### DIFF
--- a/modules/social_features/social_editor/config/install/editor.editor.full_html.yml
+++ b/modules/social_features/social_editor/config/install/editor.editor.full_html.yml
@@ -59,7 +59,7 @@ settings:
     language:
       language_list: un
     stylescombo:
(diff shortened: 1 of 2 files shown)


bramtenhove (approved):

bramtenhove:
The failing test in Travis is a false positive caused by Behat cleanup...
We can merge it imho, if afterwards this remains an issue we can dive deeper into the root cause.

ronaldtebrake:
Cherry picked and pushed to 5.x ^^

PR #1218, opened 2019-02-13
#3032584 by ronaldtebrake: Remove theme override for image widget crop

ronaldtebrake:
## Problem
Image widget crop released https://www.drupal.org/node/2871137 which uses a `Drupal.theme.VerticalTabs` function in javascript to override the theme function that gets called for all the Vertical tabs.
Unfortunately this overrides `Drupal.theme.VerticalTabs` we use from bootstrap.

## Solution
Usually what we would be able to do is use a `libraries-override` in the _info.yml_ file from our theme like:

```
libraries-override:
  core/drupal.vertical-tabs: bootstrap/drupal.vertical-tabs
```
Or you could provide the correct order as per: [Dependencies should determine asset order, not "weight"](https://www.drupal.org/project/drupal/issues/1945262) but we don't want to add these dependencies.

Therefor our solution right now is to patch out the imagewidgetcrop functionality until the issue gets fixed there.

## Issue tracker
https://www.drupal.org/project/social/issues/3032584#comment-12969389

## How to test
- [x] Edit a topic
- [x] See that the vertical tabs get rendere strangely with markup inside after doing a crop
- [x] Checkout this branch
- [x] See that it renders perfectly

## Release notes
Not needed


ronaldtebrake:
also for 4.x and 5.x :)

PR #1219, opened 2019-02-13
DS-4604 - Preview header image should be in the theme folder

jaapjan:
## Problem
In some platforms (especially older ones) is the header image preview not working in the theme settings.

## Solution
We are using a file which can be removed, e.g. `/sites/default/files/styles/social_an_hero/public/homepage-block.jpg` in some installations. Let's move the file to the theme assets directly. It's still not perfect, but it's more trustworthy for sure!

## Issue tracker
DS-4604

## How to test
- [ ] Note that the header preview is still working as expected even when the `/sites/default/files/styles/social_an_hero/public/homepage-block.jpg` does not exist.

## Release notes
The header preview image on the Appearance settings for the theme is now working on all platforms even if the homepage block image is removed.


ronaldtebrake:
Tested this, travis PR is due to something in the base branch and team banana's will take a look at that.  Will merge it but for reference thats why it's on red.

ronaldtebrake:
Cherry picked to 5.x as well :)

PR #1222, opened 2019-02-14
Issue #3032998: Use longer phrase for search in groups via Behat test

ghost:
## Problem
The text in the test is quite short. This can cause the test to fail when previously unremoved test content still exists even though the functionality works. This can occur due to issues in other tests.

## Solution
Make the texts used in the test longer. The behavior of the test will remain the same but this reduces chances of false positives. Ideally, we use something unique to the test (like the test number or name) in the phrases to avoid clashes in the future.

## Issue tracker
https://www.drupal.org/project/social/issues/3032998

## How to test
- [x] Create ten groups which have the **my description** phrase in the **Description** field.
- [x] When you run Behat tests then text for searching in groups should be failed.

## Release notes
Improve testing of search in groups through use longer search phrase.


ronaldtebrake:
hi @OH17L could you make sure sdgc turns green again, seems like it's failing on a patch of this PR which needs to be removed.
Thanks :)

ronaldtebrake:
Thanks @jaapjan that did the trick! Merging :) looks good to me.

ronaldtebrake:
And also in 5.x :)

PR #1225, opened 2019-02-18
Issue #3030750 by ribel, agami4: Add Featured Items section to Landing page

ribel:
## Problem
Currently, it's only possible to have featured content section, but sometimes content managers need to highlight information, that doesn't have separate content item.

## Solution
Create a new section that allows featuring items without having separate content pages.

## Issue tracker
- https://www.drupal.org/project/social/issues/3030750
- https://jira.goalgorilla.com/browse/GPEDC-36

## How to test
- [x] Checkout to this branch and revert features
- [x] Open Add/Edit form of the Landing Page
- [x] Add new **Featured Items** section with 5 or more featured item
- [x] You should see 5 **Featured Item** in a row with as on Design
- [x] On hover of the `(i)` icon you should see description text

## Release notes
Sometimes content managers need to highlight information, that doesn't have separate content item, for this purpose we added new section to the Landing Page that allows featuring items without having separate content pages.


ribel:
Hey, @ronaldtebrake and @jaapjan!
This PR is ready for the review.
There are some failing MOSS tests, but I see they are not related to proposed changes, and it pass OK on ECI Demo, SaaS and Foraus.

robertragas:
I checked code-wise and looks good, need to functional test it later when solution for event managers is in.

ronaldtebrake:
Screen Shot 2019-03-29 at 11 45 03

This still needs work, as you can see the `<div class="teaser__info" data-toggle="popover" data-container="body" data-html="true"
           data-placement="bottom" data-content='{{ content.field_featured_item_description }}'>` doesn't render the description correctly. 

If i change the tpl to a string itself it does work for example:

`<div class="teaser__info" data-toggle="popover" data-container="body" data-html="true"
           data-placement="bottom" data-content='<p>test</p>'>` 

Could you check it out?

ribel:
@ronaldtebrake This is known the issue, but it happens only if you have twig debug enabled in `development.services.yml`. We were not able to fix that, but on production sites it should be enabled, right?

ronaldtebrake:
Thanks for the input @ribel, would be nice to know these known issues next time. 

For me changing it to `{{ content.field_featured_item_description|render|striptags|trim }}` fixed it. 
See also [this link](https://drupal.stackexchange.com/questions/223488/how-to-stop-twig-debug-from-breaking-html-when-inside-tags)

I'm not hundred percent sure what the acceptance criteria is on how much tags we should allow but that should do the trick also with twig debug enabled. Could you ellaborate on if you want to show more than just text (because there is a full WYSIWYG editor now for the description field).

ribel:
@ronaldtebrake Thanks for the review!

There were no strict acceptance criteria if this description should be rich text or not. So we can decide it here :)

I would propose to leave field storage like it is now, so in case we will need rich text for some cases we will not need to change it. But we also can add `striptags|trim` for safety by default. And if needed it's easy to just override template.

ronaldtebrake:
Thanks @ribel.. That's perfect, since I'm in the branch I will commit it that way!

PR #1227, opened 2019-02-18
Feature/configurable profile landingpage

robertragas:
## Problem
Currently when going to a user you always end up at the stream page. Some customers would like them to end up at the user profile.

## Solution
Added the possibility to set a configuration item to land on either the information page or stream.

Right now it has the two options hardcoded but can be extended in the future by getting the user tabs dynamic based on permissions set for a client (since it might be that users are not allowed to view a page).

## Issue tracker

## How to test
- [ ] drush updb or install
- [ ] Go to configuration -> open social -> user settings
- [ ] By default it lands on stream, but try to alter it to Information
- [ ] Click on a user from a stream link like mentions or username, you will end up at the profile information.
- [ ] Try to go to your own profile through the account header item, this still goes to your stream page.

## Release notes
We have added a new setting where you can configure the landing page when visiting a user's profile. This can be set under configuration -> open social -> user settings.


jaapjan:
I fixed small coding standard issue. Btw, could you create a drupal.org issue for this as well?

ronaldtebrake:
Still needs a bit of work as discussed with Robert so will be in 4.7 hopefully :)

robertragas:
Working on it today during the collaboration day.

robertragas:
Changes made and should work now like intended.

robertragas:
Closing in favor of https://github.com/goalgorilla/open_social/pull/1284 which is for branch 5.x

PR #1228, opened 2019-02-20
Issue #3034318 by jaapjan: prevent Notifications 500 errors after upd…

jaapjan:
## Problem
500 errors on certain platforms after the 4.4 update because referenced entities can't be found.

## Solution
More defensive coding by first checking if the entity exist before calling method on it.

## Issue tracker
https://drupal.org/node/3034318

## How to test
- [ ] Open old platform as group manager.
- [ ] Notice site works as expected

### DIFF
--- a/modules/custom/activity_creator/activity.page.inc
+++ b/modules/custom/activity_creator/activity.page.inc
@@ -7,6 +7,7 @@
  * Page callback for Activity entities.
  */
 
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Render\Element;
 use Drupal\file\Entity\File;
 use Drupal\Core\Link;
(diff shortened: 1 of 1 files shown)


modules/custom/activity_creator/activity.page.inc
```diff
@@ -49,26 +49,27 @@ function template_preprocess_activity(array &$variables) {
...
-
-  if ($entity->getEntityTypeId() === 'group_content') {
-    /** @var \Drupal\group\Entity\GroupContentInterface $entity */
-    if ($entity->getGroupContentType()->getContentPluginId() === 'group_membership' && $entity->getEntity()->id() != $activity->getOwnerId()) {
-      $account = $entity->getEntity();
+  if (!empty($entities)) {
+    /** @var \Drupal\Core\Entity\EntityInterface $entity */
+    $entity = reset($entities);
+
+    if ($entity && $entity->getEntityTypeId() === 'group_content') {
```
jochemvn:
```suggestion
    if ($entity instanceof EntityInterface && $entity->getEntityTypeId() === 'group_content') {
```

jochemvn: In any other case, the nullpointerexception might still occur.

jaapjan:
Also cherry-picked for 5.x

PR #1229, opened 2019-02-21
Add smaller hero as landing page section

jochemvn:
## Problem
The hero section of the landing page takes up too much space. We want another hero that is significantly smaller than the current landing page hero section

## Solution
A second section called 'Hero small', which is around 50% less high. 

## Issue tracker
https://jira.goalgorilla.com/browse/SHN-363

## How to test
- [ ] Install OS and enable landing page module
- [ ] Create new LP with the new hero small section
- [ ] Notice that TITLE text is overflown after 2 lines
- [ ] Notice that SUBTITLE text is overflown after 1 line
- [ ] Check that images aren't skewed

## Release notes
<describe the release notes>


modules/social_features/social_landing_page/css/social_landing_page.section.hero_small.css
```diff
@@ -0,0 +1,69 @@
...
+    left: 50%;
+    transform: translateX(-50%);
+  }
+}
+.paragraph--hero-small .cover-small .page-title {
+  width: 100%;
+  text-align: left;
+  padding: 30px 0 10px;
+  font-size: 2rem;
+  max-height: 6.5rem;
```
robertragas: In my case with 6.5 just a little bit of the line under was showing. 6.4 fixes it for me.

robertragas:
Working as expected. Just wonder 2 things.
1. The text on the small hero is aligned to the bottom, and not the center (vertically). Is this according to design?
2. I don't see Travis running?

modules/social_features/social_landing_page/css/social_landing_page.section.hero_small.css
jochemvn: nope

PR #1230, opened 2019-02-22
Lets make travis green again!

ronaldtebrake:
## Problem
Group edit open failed.

## Solution
Fix group-edit-open, we used to press delete but didnt check if the batch was finished and if the success message was displayed. This could potentially cause an issue where the test was green, which means the AfterScenario ` /**
     * Remove any groups that were created.
     *
     * @AfterScenario
     */
    public function cleanupGroups(AfterScenarioScope $scope) {
      if (!empty($this->groups)) {
        foreach ($this->groups as $group) {
          $group->delete();
        }
      }
    }` 
was called but in the meantime the batch job deleting the group was deleting all of the Group Content. Due to that the cachetags are duplicating in insertion instead of an update.

The solution is to wait for the job to finish, for all the Group and Group content to be deleted correctly and after that the AfterScenario step can kick in.

## Issue tracker
None.

## How to test
Travis will be green!

## Release notes
None.

### DIFF
--- a/modules/social_features/social_event/config/optional/message.template.activity_on_events_im_organizing.yml
+++ b/modules/social_features/social_event/config/optional/message.template.activity_on_events_im_organizing.yml
@@ -12,8 +12,8 @@ third_party_settings:
     activity_destinations:
       notifications: notifications
       email: email
-    activity_create_direct: 1
-    activity_aggregate: 0
+    activity_create_direct: true
+    activity_aggregate: false
     activity_entity_condition: ''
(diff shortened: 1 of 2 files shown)


jaapjan:
Yoyo, travis is green. Let's merge? @ronaldtebrake

ronaldtebrake:
Travis is green! Moss seems to fail on batch jobs. Will fix it in the necessary branches as part of the MOSS green work :)

ronaldtebrake:
8.x-5.x cherry picked.

PR #1232, opened 2019-02-25
Issue #3024448: html in pop-up for anonymous event enrolment

jaapjan:
## Problem
HTML in pop-up for anonymous event enrolment. 

## Solution
I followed this solution in Core
https://www.drupal.org/project/drupal/issues/2568149#comment-10327469

## Issue tracker
https://www.drupal.org/project/social/issues/3024448

## How to test
- [ ] Enable social_event_an_enroll
- [ ] Create public event with title `New year's party` or similar and enable anonymous event enrolment as well for that event.
- [ ] Open the event as anonymous user.
- [ ] Click on enrol button and notice the nice title as expected.

## Release notes
If an event title included any special characters (for example an apostrophe `) it would not display correctly in the pop-up window which anonymous users see after clicking on the "Enrol" button. This is now solved.

### DIFF
--- a/modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
+++ b/modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
@@ -92,7 +92,7 @@ class EventAnEnrollActionForm extends EnrollActionForm {
           ],
           'data-dialog-type' => 'modal',
           'data-dialog-options' => json_encode([
-            'title' => t('Enroll in @event Event', ['@event' => $node->getTitle()]),
+            'title' => t('Enroll in') . ' ' . strip_tags($node->getTitle()),
             'width' => 'auto',
           ]),
         ];


jaapjan:
Event enrollment test keeps on failing on one site:

event-enrollment-feature-test

Could be a false negative but I restarted the test multiple times and it happens all the time.

modules/social_features/social_event/modules/social_event_an_enroll/src/Form/EventAnEnrollActionForm.php
```diff
@@ -92,7 +92,7 @@ public function buildForm(array $form, FormStateInterface $form_state, Node $nod
           ],
           'data-dialog-type' => 'modal',
           'data-dialog-options' => json_encode([
-            'title' => t('Enroll in @event Event', ['@event' => $node->getTitle()]),
+            'title' => t('Enroll in') . ' ' . strip_tags($node->getTitle()),
```
ronaldtebrake: the word Event has been removed from the title now, that is why the behat test is failing for PAC f.e. 
should we put that back?

jaapjan: AH, that explains. I prefer to remove the word "Event" as it doesn't add value here and only makes the title a lot bigger. I'll change the test in PAC.

jaapjan: Test is passing now :-)

ronaldtebrake:
And cherry picked :)!

PR #1234, opened 2019-02-25
Issue #2992737: fix translation of the bundle in like modal title

jaapjan:
## Problem
The bundle in the title in the like modal window is untranslatable.

## Solution
Let's make it translatable. Best is to retrieve the bundle settings and retrieve the label from there. The bundle settings label respect the current active language and the label is always the label from the configuration translation of the current bundle.

## Issue tracker
https://www.drupal.org/project/social/issues/2992737

## How to test
- [x] Enable social_language and dependencies
- [x] Add a language
- [x] Go to `/admin/structure/types/manage/topic/translate` and add translation for the new language there.
- [x] Go to a topic and open the like modal window. Verify the word "topic" is translated correctly.
- [x] Now check comments, posts, other entity types as well. Verify it all still works as expected.
- [x] Check code changes.

## Release notes
The entity type in a like modal window is now showing the translated word in multi-language sites.

### DIFF
--- a/modules/social_features/social_like/social_like.module
+++ b/modules/social_features/social_like/social_like.module
@@ -70,20 +70,23 @@ function social_like_invalidate_cache(EntityInterface $entity) {
  */
 function social_like_preprocess_like_and_dislike_icons(&$variables) {
   $bundle = $variables['entity_type'];
-  if ($variables['entity_type'] === 'node') {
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
Andddd cherry-picked to 5.x

Kingdutch:
I've created a follow-up issue because a change in this issue can cause problems in certain languages: https://www.drupal.org/project/social/issues/3042115

PR #1138, opened 2018-11-29
DS-5959 Event Organizer recieves notification for Enrollments.

ronaldtebrake:
## Problem
As an Event Organizer you are currently not notified of any people actually joining your event. It's great to get direct insights in the activity of your organized event.

## Solution
We created a new Notification for E-mail & the Notification center that shows you someone enrolled to your Event. For the normal users their display name is used. For AN we use the first + last name. It's all in a new message template, with custom tokens and with the ability to extend the people that receive this notification (for now it's the Event Creator & all the Event Organizers if the module is turned on). 

## Issue tracker
https://www.drupal.org/project/social/issues/2923782

### Normal flow
- [x] Login as LU
- [x] Create an Event
- [x] Logout & Login with someone else, Enroll to this Event
- [x] See you get a message someone enrolled to your Event and also a nice Email notification

### AN Enrollments
- [x] Enable An event Enrollments
- [x] Login as LU
- [x] Create a Public event and Accept AN Enrollments
- [x] Logout and Enroll to this Event as AN 
- [x] See that the Event Creator gets a notification & email with a message the AN user (FirstName + LastName) has enrolled to your Event

### Event Organizers
- [x] Enable the Event Organizers module if not already done
- [x] Edit your Public event and add multiple Organizers
- [x] Enroll with a new AN user
- [x] See the event creator & all the organizers get the notification now.

## Release notes
As an Event Organizer you will now get a notification that someone enrolled to your event in order to get direct feedback on the activity of your event.

PR #1235, opened 2019-02-25
Issue #3035592: make sure event exists before calling toUrl method on it

jaapjan:
## Problem
When you create an event as LU, enrol to the event yourself, let some other people enrol to your event, run the cron and then delete the event you will receive a 500 error when trying to load your notification bar resulting in a non-working website.

## Solution
Let's add an extra check before blindly calling the toUrl() method on an object which might be NULL.

## Issue tracker
https://www.drupal.org/project/social/issues/3035592

## How to test
- [ ] Check code
- [ ] Stay on the 4.x branch
- [ ] Create event
- [ ] Signup to event (yourself and/or maybe others)
- [ ] Run cron few times (at least make sure you have a notification about an event signup to that event in your notification bar - https://github.com/goalgorilla/open_social/pull/1138)
- [ ] Delete the event
- [ ] Notice you receive an unexpected error immediately
- [ ] Switch to this feature branch.
- [ ] Retry above steps and notice it works now (or just refresh the page)

## Release notes
If an event was deleted and an event organizer received a notification for it this could result in an error on the page/your stream not loading correctly. This is now solved.

### DIFF
--- a/modules/social_features/social_event/social_event.tokens.inc
+++ b/modules/social_features/social_event/social_event.tokens.inc
@@ -108,11 +108,15 @@ function get_link_to_event_from_enrollment($object_id, $as_link = FALSE) {
     /** @var \Drupal\node\Entity\Node $event */
     $event = $storage->load($event_id);
 
-    if ($as_link) {
-      return $event->toUrl('canonical')->toString();
+    // Check if the event still exists.
+    if ($event !== NULL) {
+      if ($as_link) {
+        return $event->toUrl('canonical')->toString();
+      }
+
+      return Link::fromTextAndUrl($event->getTitle(), $event->toUrl('canonical'))->toString();
     }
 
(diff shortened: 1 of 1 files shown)


Kingdutch:
I thought about this scenario when I merged that code but I actually purposefully didn't add that check because I think it hides a different bug.

There should be no event enrollments for non-existing events.

The proper fix here is probably to clean up the event enrollments when an event is deleted. This way the code would already exit earlier because the enrollment does not exist.

Ideally we'd even clean up the notifications that resulted from the enrollment but that may not be possible.

jaapjan:
Feel free to comment it next time ;-)

Anyway I propose to solve it like this for now as I don't want to write an update hook to remove old event enrolments and our code should be defensive enough to not cause a crash like that. If you added the "needs work" label because of that reason I'll remove the label again or did you test my fix as well properly?

ronaldtebrake:
Thanks will also hotfix voor SaaS & merge in to 5.x

PR #1245, opened 2019-03-11
Issue #2946146 by Kingdutch, sukh.singh: Don't run entity queries for…

Kingdutch:
… activities during configuration entity modification.

## Problem
Activities should never be created for config entities (they notify of
content). However, in certain cases (feature reverts, module uninstall,
update hooks) it can cause issues when an entity query is executed for
activities during a configuration change.

This can cause fatal errors during things such as update hooks, feature imports and module uninstalls.

## Solution
The easiest way is to just stop doing things when we're triggered for a configuration entity.

## Issue tracker
https://www.drupal.org/project/social/issues/2946146

## How to test
- [ ] Uninstall the activity_creator module

## Release notes
Configuration changes could sometimes trigger an error in the activity_creator module due to transitive field configuration. This has been resolved by no longer trying to process activities as a result of a configuration change. This solution relies on the fact that activities were only supported for content changes.

### DIFF
--- a/modules/custom/activity_creator/activity_creator.module
+++ b/modules/custom/activity_creator/activity_creator.module
@@ -7,6 +7,7 @@
 
 use Drupal\message\Entity\Message;
 use Drupal\Core\Cache\Cache;
+use Drupal\Core\Config\Entity\ConfigEntityInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\activity_creator\Entity\Activity;
@@ -255,6 +256,15 @@ function _activity_creator_activity_for_entity_updater($action, EntityInterface 
   if ($entity !== NULL) {
     $entity_type = $entity->getEntityTypeId();
     $entity_id = $entity->id();
+
+    // We never create activities for configuration entities but accessing
+    // entity queries during configuration changes may cause errors.
+    if ($entity instanceof ConfigEntityInterface) {
+      return;
+    }
+
+    // TODO: The if statement below can be refactored after the code that was
+    // added above but we didn't have time for that yet.
     if ($entity_type !== 'activity') {
 
       $entity_query = \Drupal::entityQuery('activity');


modules/custom/activity_creator/activity_creator.module
```diff
@@ -255,6 +256,15 @@ function _activity_creator_activity_for_entity_updater($action, EntityInterface
...
     $entity_type = $entity->getEntityTypeId();
     $entity_id = $entity->id();
+
+    // We never create activities for configuration entities but accessing
+    // entity queries during configuration changes may cause errors.
+    if ($entity instanceof ConfigEntityInterface) {
+      return;
+    }
+
+    // TODO: The if statement below can be refactored after the code that was
```
jochemvn: I've been looking at this particular piece of code too for a few times already... Did you already think about how to fix this? IMO I think we should remove all of those entityQuery and just use the entitytype manager to load the related activities. Thoughts?

Kingdutch: I'm not entirely sure how the entityTypeManager code would look. I don't think that's necessarily a good idea because we don't really need to entities themselves, we only need the IDs to invalidate them (unless we can get the entity IDs of the related entities without loading the actual entities, then that would probably save us a roundtrip to the database altogether because the entity we're fetching related entities for is already loaded?).

I think moving to a switch can already make this more readable, or even ensuring we don't potentially run queries multiple times for the user, group and group_content types.

Something like:

```php
// Never update activities as a result of an activity update.
if ($entity_type === 'activity') {
  return;
}

$ids = [];
switch ($entity_type) {
  case 'user':
  case 'group':
    $entity_query = \Drupal::entityQuery('activity');
    $entity_query->condition('field_activity_recipient_' . $entity_type, $entity_id, '=');
    $ids = $entity_query->execute();
    break;
  case 'group_content':
    /** @var \Drupal\group\Entity\GroupContent $group_content */
    $group_content = $entity;
    $linked_entity = $group_content->getEntity();
    $group = $group_content->getGroup();
    if ($linked_entity && $linked_entity->getEntityTypeId() === 'node' && $group->id()) {
      $entity_query = \Drupal::entityQuery('activity');
      $entity_query->condition('field_activity_entity.target_id', $linked_entity->id(), '=');
      $entity_query->condition('field_activity_entity.target_type', $linked_entity->getEntityTypeId(), '=');
      $entity_query->condition('field_activity_recipient_group', $group->id(), '=');
      $ids = $entity_query->execute();
    }
    break;
  default:
    $entity_query = \Drupal::entityQuery('activity');
    $entity_query->condition('field_activity_entity.target_id', $entity_id, '=');
    $entity_query->condition('field_activity_entity.target_type', $entity_type, '=');
    $ids = $entity_query->execute();
}
```

While we're at it, we can also replace
```php
$tags = [];
foreach ($ids as $id) {
  $tags[] = 'activity:' . $id;
}
Cache::invalidateTags($tags);
```

With
```
$tags = array_map(
  function ($id) { return "activity:${id}"; }, 
  $ids
);
Cache::invalidateTags($tags);
```

jochemvn: I like the code... however, keep in mind it's not just about fetching the IDs of the activities, since there's also a (deprecated) call to `entity_delete_multiple` at the bottom.

jochemvn:
As a side note... Maybe we should also delete the configuration this module has. The config is also in `social_activity` 🙈

modules/custom/activity_creator/activity_creator.module
Kingdutch: I think we should tackle these things in a follow-up issue (hence the TODO). The change in this PR fixes a pretty serious bug, the code below can use some refactoring (and Drupal 9 proofing) but otherwise still works.

Kingdutch:
I think we should tackle these things in a follow-up issue (hence the TODO). The change in this PR fixes a pretty serious bug, the code below can use some refactoring (and Drupal 9 proofing) but otherwise still works.

(PS. Given the fact that activity_creator was written to be relatively unaware of Open Social, I argue that we shouldn't delete the duplicate configuration from the module but should clean up `social_activity` when we have a better alternative to features. Treating activity_creator as if it was a module from drupal.org in this case.)

Kingdutch:
Merged in 8.x-4.x and cherry-picked to 8.x-5.x 😃

PR #1246, opened 2019-03-11
Issue #3039084: Always show Full Name

ghost:
## Problem
Can not see Full Name when turned on the Nickname field.

## Solution
1. Add permission to the social_profile_fields module to always show Full Name.
2. Add a setting to the social_profile_fields module to limit search and mention using "Full Names" to the permission to see real name.

## Issue tracker
- https://www.drupal.org/project/social/issues/3039084
- https://jira.goalgorilla.com/browse/GPDE-65

## How to test
- [ ] Turn on **Social Profile Fields** module.
- [ ] Log in as **SM** and navigate to **Manage > Configuration > Open Social Settings**.
- [ ] You should see unchecked **Limit search and mention** checkbox.
- [ ] Enable **Nickname** field.
- [ ] Create **First Test User** (username - **ftu**, first name - **First**, last name - **Test User**, nickname - **First User**).
- [ ] Create **Second Test User** (username - **stu**, first name - **Second**, last name - **Test User**, nickname - **Second User**).
- [ ] Create **Test Role** role and add **Always show Full Name** permission to this role.
- [ ] Assign **Test Role** role to **First Test User**.
- [ ] Log in as **Second Test User**.
- [ ] When you type **@ftu** in post field then you should see **First User** in drop-down list.
- [ ] Log in as **First Test User**.
- [ ] When you type **@stu** in post field then you should see **Second Test User** instead of **Second User** in drop-down list.
- [ ] Log in as **SM** and navigate to **Manage > Configuration > Open Social Settings**.
- [ ] Check **Limit search and mention** checkbox and press **Save configuration**.
- [ ] Log in as **Second Test User**.
- [ ] When you type **@First Test User** in post field then you should see empty drop-down list.
- [ ] When you use **First Test** as search phrase then you should not see **First Test User** user in search results.
- [ ] Log in as **First Test User**.
- [ ] When you type **@Second Test User** in post field then you should see **Second Test User** in a drop-down list.
- [ ] When you use **Second Test** as search phrase then you should see **Second Test User** user in search results.

## Release notes
A user can not use the full name of other users for mention them or search them when the **Limit search and mention** option (the **Social Profile Fields** module settings page) is activated and the user does not have the **Always show Full Name** permission.


ronaldtebrake:
Should we set the default to false on install?

OH17L:
> Should we set the default to false on install?

No, we shouldn't. It will be false automatica

OH17L:
> Should we set the default to false on install?

Maybe...

modules/social_features/social_profile/src/SocialProfileTrait.php
```diff
@@ -122,20 +159,26 @@ private function startQuery() {
    *   The select query.
    */
   private function sortQuery(SelectInterface $query, $name, $suggestion_format) {
```
bramtenhove: Please add that to the `@param string $name` description that it is already sanitized.

OH17L: Is it enought?

bramtenhove: Yes, thanks!

Kingdutch (dismissed):

modules/social_features/social_profile/modules/social_profile_fields/social_profile_fields.module
```diff
@@ -148,3 +155,54 @@ function social_profile_fields_social_user_export_plugin_info_alter(array &$plug
     }
   }
 }
+
+/**
+ * Implements hook_search_api_query_alter().
+ */
+function social_profile_fields_search_api_query_alter(QueryInterface &$query) {
```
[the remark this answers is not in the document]
bramtenhove: Good explanation. The same goes for usernames, everyone should always be able to find someone by their username. This is also how it works now afaik.

Kingdutch:
I'm closing this pull request. Given the changes that have been made this will have to be introduced in 8.x-5.x instead of 8.x-4.x

PR #1247, opened 2019-03-11
3039042: Member tab in group shows the same members on the second page

jochemvn:
## Problem
Whenever a group has more than 10 people, pagination appears. However when you go to the next page, a couple of the same users appears as on the first page. Seems to be a bit random. Maybe something with sorting?

## Solution
I turned out to be related to the create date of the group content (mass added users). Since that creation data of the group content (the membership) was the same for almost all members, views is kind of clueless. I've added an additional sorting on name (label of the group content)

## Issue tracker
- https://www.drupal.org/project/social/issues/3039042
- https://jira.goalgorilla.com/browse/YANG-757

## How to test
- [ ] Add multiple members to a group at the same time (bulk). Add like 15 or something like that
- [ ] Notice that when you go to older items, there's an issue that you see the same people you've already seen on the first page.
- [ ] Checkout this branch
- [ ] Revert features and clear caches
- [ ] Retest and notice all is good now

## Release notes
In some rare cases when a lot of users were added to a group in bulk, the pagination of the members in a group would show the same results on the second page of the group members page. This is now fixed by not just sorting on the date people were added to the group, but next to that also on the name of the person added.

### DIFF
--- a/modules/social_features/social_group/config/install/views.view.group_members.yml
+++ b/modules/social_features/social_group/config/install/views.view.group_members.yml
@@ -48,7 +48,7 @@ display:
           sort_asc_label: Asc
           sort_desc_label: Desc
       pager:
-        type: mini
+        type: full
         options:
           items_per_page: 10
           offset: 0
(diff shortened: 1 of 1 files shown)


jaapjan:
@robertragas Don't forget to close the d.o issue next time :-) I've closed it for now. Let me know if I'm in error somehow.

PR #1248, opened 2019-03-11
Issue by Kingdutch #3039150: SocialEmbedConfigOverride overrides are …

Kingdutch:
…too broad

<h2>Problem</h2>
<code>SocialEmbedConfigOverride</code> contains some overrides that depend on existing configuration (because it adds to the string value of "allowed_html"). However, the <code>$overrides</code> array contained entire configuration object instead of only the array path that was adjusted.

<h2>Solution</h2>
Overrides should be as specific as possible and will use deep overriding to traverse a path of specific keys.

## Issue tracker
https://www.drupal.org/project/social/issues/3039150

## How to test
- [x] Ensure the embed button is still added to the wysiwyg

## Release notes
The configuration overrides used by the social_embed module have been narrowed in scope to decrease the risk of conflicts with other html format or text editor configuration overrides.


modules/social_features/social_embed/src/SocialEmbedConfigOverride.php
```diff
@@ -132,36 +127,36 @@ protected function addEditorOverride($text_format, array &$overrides) {
...
+            break 3;
           }
         }
       }
     }
 
+    // If the button already exists we change nothing.
     if (!$button_exists) {
       $row_array_keys = array_keys($settings['toolbar']['rows']);
       $last_row_key = end($row_array_keys);
```
Kingdutch:
```suggestion
      $last_row_key = end($row_array_keys);
      // Ensure we add our button at the end of the row.
      // We use count to avoid issues when keys are non-numeric (even though that shouldn't happen).
      // This will break if the keys are non-consecutive (which should also never happen).
      $group_key = count($settings['toolbar']['rows'][$last_row_key]) + 1;
```

modules/social_features/social_embed/src/SocialEmbedConfigOverride.php
```diff
@@ -132,36 +127,36 @@ protected function addEditorOverride($text_format, array &$overrides) {
...
-      $settings['toolbar']['rows'][$last_row_key][] = $group;
-
-      $overrides[$config_name] = [
-        'settings' => $settings,
+      // Add the button as a new group to the bottom row.
+      $group = [
+        'name' => 'Embed',
+        'items' => ['social_embed'],
       ];
+      $overrides[$config_name]['settings']['toolbar']['rows'][$last_row_key][] = $group;
```
Kingdutch:
```suggestion
      $overrides[$config_name]['settings']['toolbar']['rows'][$last_row_key][$group_key] = $group;
```

bramtenhove:
Let's wait what the tests do, changes apply just fine now.

PR #1249, opened 2019-03-11
Issue #3039170 by Kingdutch: Add styling for tables in content

Kingdutch:
<h2>Problem</h2>
Tables in content (e.g. topic bodies) aren't styled by default which makes them difficult to read.

<h2>Solution</h2>
Adding a class to user content isn't simple so extending the selector from <code>.table</code> (elements with the class table) to also include <code>.page-node .body-text table</code> (table elements in elements with the body-text class of node pages) ensures that they're properly styled when using socialbase as a theme.

## Issue tracker
https://www.drupal.org/project/social/issues/3039170

## How to test
- [x] Add a table to a topic and see that it is styled.

## Change record
https://www.drupal.org/node/3043143

## Release notes
Tables in content (topic, events, etc.) bodies are now styled by default when using (a subtheme of) the socialbase theme. This improves readability for content that includes tables.

Screenshot 2019-03-12 at 11 13 27


bramtenhove:
I'm wondering if we should add a change record for this. We are theoretically impacting styling that others already did for their themes right?

Kingdutch:
> I'm wondering if we should add a change record for this. We are theoretically impacting styling that others already did for their themes right?

That's actually a good point! I think I agree although I don't know if it would have much more substance than what we put in the release notes. It would be a good highlight.

bramtenhove:
I don't mind if we say the same as in the release notes, but it would indeed serve as a nice highlight.

Kingdutch:
A draft change record has been created here: https://www.drupal.org/node/3043143

Kingdutch:
Given that this is a purely cosmetic change, Travis is green and the MOSS failures are on features that are not impacted by this addition I'm merging this branch in its current test state.

PR #1250, opened 2019-03-12
DS-6407 DS-6373 - Sharing and collecting usage data

jaapjan:
## Problem
The Open Social team doesn't have any data on how Open Social is being extended and used. There is no accurate version information as well as if the installation profile is being extended by other modules or new entity types.

## Solution
We would like to give distro users a way to share how they use Open Social. This will help make the distro better since we can use that information to determine priority for making the project more flexible.

## Issue tracker
n/a

## How to test
- [ ] Install the module on existing installation (is done in update hook in social_core) notice sharing is disabled by default, but module is enabled.
- [ ] Install the module on new installation (automagically)
- [ ] Test sending data to the provided endpoint by running cron and queues.
- [ ] Test different settings to verify if you can turn sharing on (and specific parts on/off)

## Release notes
tbd


jaapjan:
Still need to check the text/language descriptions.

PR #1252, opened 2019-03-13
#3039585 - Composer updates cause twig to fail.

ronaldtebrake:
## Problem
Since we didnt specify which version `twig/twig` we use `composer update` updates to the latest version which causes a fatal error. 

## Solution
Core is trying to patch using: https://www.drupal.org/project/drupal/issues/3039408
this however results in a patch which doesnt work since core needs to patch composer.lock files we don't use. Custom copying that pach results in errors in Open Social with twig caching. 
We will open up a separate issue for that. First lets go for green on the tests again.

## Issue tracker
- None.

## How to test
- [ ] See that travis doesn't fail on composer update step

## Release notes
None


truls1502:
I hope it is okay to add this on 4.6 milestones? Because I believe it will solve future problems.

ronaldtebrake:
@truls1502 definitely, this is the last thing we will commit in 4.6 :) (and backport to 3.x for a new release) so people don't run in to issues with composer updates of Open Social.

truls1502:
Sounds perfect plan 👍

ronaldtebrake:
Also confirmed it works by @truls1502 , thanks 👍

PR #1254, opened 2019-03-14
#3040183 by ronaldtebrake: update translations for pagers

ronaldtebrake:
## Problem
The default pager buttons on overviews (first, next, previous, last) aren't translatable due to an issue in the descovery of these items. 

## Solution
Make sure we work around the twig discovery so potx is aware of these strings. See: https://www.drupal.org/project/potx/issues/2816601

## Issue tracker
https://www.drupal.org/project/social/issues/3040183

## How to test
- [ ] The strings get discovered already by drupal translations, it's just weblate that doesn't see them through potx discovery. After fixing this the above strings should become available on the translation server.

## Release notes
The buttons on pagination are now translatable. These being _first, next, previous, last_


ronaldtebrake:
Cherry picked to 5.x - Go weblate!

PR #1255, opened 2019-03-14
Issue #3039084: Allow full name to be hidden

Kingdutch:
## Problem
As a sitemanager I want to protect full names on my platform when users have filled in a nickname. Currently users can still be found and mentioned by their full name even if they have filled in a nickname.
 
When protecting the full name then as a developer I want to give certain roles the permission to also view and search for the full name.

## Solution
1. Add a setting to the social_profile_fields module to disallow display, search and mention using "Full Names" when a nickname is filled in.
2. Add permission to the social_profile_fields module to display the full name with the nickname and allow searching based on the full name.

## Issue tracker
- https://www.drupal.org/project/social/issues/3039084
- https://www.drupal.org/project/social/issues/3043698
- https://www.drupal.org/project/social/issues/3043714
- https://jira.goalgorilla.com/browse/GPDE-66

## How to test
See test that was added in this PR.

## Release notes
When the nickname field was enabled on a platform and a user filled in a nickname then this would be displayed on the platform. However, the user could still be found by his/her real name in search and there was no easy way for content/site managers to view a users' real name. A new setting **Limit search and mention** has been added (on the **Social Profile Privacy** module settings page). When this setting is activated users will not be able to find the user by their real name. Additionally a permission has been added: **View full name when restricted**. When users have this permission they will be shown the full name in addition to the nickname and will be able to search using the full name.

When no nickname is filled in for a user then there are no changes in behaviour and the full name is shown (when available).

## Change record
This PR requires two change records:
- Commit aaddf03 makes changes to the way that we determine the username resolution in Open Social. This gives us a more flexible system that can be used by developers.
  https://www.drupal.org/node/3043747
- Commit f9ff90332 disables immediate indexing for content in the search indexes. This means that cron is now required for search to function properly.
  https://www.drupal.org/node/3043749

## Original 8.x-4.x pull request
Due to the changes included the PR #1246 has been cherry-picked and moved to 8.x-5.x


Kingdutch:
Commit aaddf03408906478eb4528d943655d404e676db5 makes changes to the way that we determine the username resolution in Open Social. This gives us a more flexible system that can be used by developers.

A change record should be added to announce this change. The code in the `social_user.api.php` file as well as the commit message can be used for creating this.

Kingdutch:
Marking this as needs work. I believe the changes in 5e578b65ca2496f38bd8b9d0ab5d2636a29f2ee6 break the configuration override. Config overrides are merged deeply based on key. The new arrays in the override are given key `0` which means that they override the first item of the configuration instead of appending to the array.

0fd0a10e727e3e01e854ce7f03349d7357e845aa adds a `_social_profile_fields_update_search_index('social_all');` to `social_profile_fields_install` but this is already done a line earlier by `_social_profile_fields_update_search_index();`.

I do like the changes in 465d3c461570c31343e12125d2eb913714fab50b where simply not allowing search to use the fields is better than trying to filter those thing out afterwards. It should be noted that if the full name (or display name with full name) is indexed by the rendered entity however then this new solution no longer works.

Kingdutch:
@OH17L In your latest commit I see that you added the `.settings` sub-array to the configuration which felt strange to me.

Looking at the inner working of the field processors in the search_api it appears that this `settings` array is not needed and is actually an error in the base configuration for our search indexes (all of them) that we should fix. This configuration should just directly be kept under the name of the processor.

@ronaldtebrake @jaapjan Is that something we can fix here? Or should be fix that in a separate issue? It may be the cause for the field not showing up properly in this PR.

@OH17L I'm not sure whether enabling/disabling the nickname field is what's needed (`social_profile_fields_update_8002`). I think the reason that worked last time was because the nickname field was added as an override for the first time. What I do think is required is that the indexes are cleared and rebuilt. However, this needs to happen after the feature is reverted so that the first and last name are properly added, this won't be the case yet once the update hooks are run.

Furthermore, we're adding the entire configuration to our override but we should only add what's actually overridden using a new latest key.

Kingdutch:
The Travis build for this is failing on the social profile fields test which is very much related to changes in this PR so that still needs to be addressed.

ronaldtebrake:
@Kingdutch cc @OH17L 
What would be ideal in that case is to just quickly create an issue on d.org, fix it in this branch using that d.org id so we have the administration covered for users that might already have encountered this. But fixing it in here definitely would be great!

Kingdutch:
The change records have been added to the opening post.

Kingdutch:
@ronaldtebrake We have a good example of a Travis test doing its job in this PR :) It caught the fact that this new solution was missing the handling of the social profile fields settings. 

The fact that we can disable the first name and last name separately from each other and don't immediately clean this data from the profile makes the solution here quite complex. The effect of this can be seen in f4f95de5de5e41a3291b068c4b2a67914ea9ac66. 

We can simplify this code by either clearing the data when a field is disabled immediately (instead of waiting for a data flush) or we add a special case to put the first name and last name field together instead of allowing them to be toggled separately.

Kingdutch:
The tests that are failing now are a result of f9ff903. Since we no longer index search content directly the newly created content can not be found. This can be resolved by creating a Behat helper step that indexes the content.

cc @bramtenhove

Kingdutch:
@ronaldtebrake The only failing test here is "Updating group content..." text is not shown. However, I've run this locally and there I can see the text. 

This could be related to the fact that this is a batch progress page. Have we had this issue anywhere else?

jaapjan:
@Kingdutch 

Perhaps because we have to wait for AJAX transaction? 

One thing I noticed about batch pages is that they don't look pretty, see for example this page when changing role of user(s) in a group.

change-role-in-group

Kingdutch:
@jaapjan We're getting close!

There is one case that still seems to break which is when a user should be findable by his username but is filtered out anyway because his real name is hidden and it also matched. When the username differs from the username or there is no nickname (e.g. change "Secretive User" to "Secretive Person") then everything works as expected.

Do you maybe have any idea why the conditions in `social_profile_privacy_search_api_query_alter` could filter out the Secretive User (Hide my nickname) when searching for `user` (which matches her nickname)?

jochemvn (approved):
Made a small improvement about the default behaviour of the search page. Looks good and ready to merge

PR #1256, opened 2019-03-18
Issue #3040045: update to drupal/core 8.6.12

jaapjan:
## Problem
Twig is fixed at version 1.37.1 causing issues when trying to update drupal/core to 8.6.12.

## Solution
Let's remove the twig fixed version.

## Issue tracker
https://www.drupal.org/project/social/issues/3040045

## How to test
- [ ] Tests are passing
- [ ] 8.6.12 is used.

## Release notes
To release 4.6 we were forced to set twig/twig to a fixed version. That caused issues when trying to update to Drupal Core 8.6.12. This is now solved.


jaapjan:
Merging not necessary but will be in 4.7.

PR #1263, opened 2019-03-22
Issue #2936809: remove user submit form protection

jaapjan:
## Problem
There is some weird behaviour in the user forms. This might be caused by javascript code which doesn't seem to do the desired behaviour.

## Solution
Let's remove it for now. As it probably only causes side-effects and the code doesn't fix anything. We can rely on Drupal core behaviour.

Also we probably don't want users to enforce changing their passwords as in some cases it is also an option to use the password request link as a login link. If people's only intention is to login (and not change the password) then that behaviour can be annoying for them as well.

## Issue tracker
https://www.drupal.org/project/social/issues/2936809

## How to test
- [x] Use one-time login link for LU on 5.x branch.
- [x] Notice you get a pop-up message when you try to leave the page or save the form without entering your password.
- [x] Try again on this branch.
- [x] Notice you can go away without this pop-up.

## Release notes
Previously there was some behaviour which was warning users when they leave the user edit form without changing their password (after using login link). We removed this warning as we believe it's better to leave it up to the user to decide if they want to change the password.


ronaldtebrake:
merged back 8.x-5.x lets see what Moss does :) manual tet & review all look good and seem unrelated to any test failures so will merge after

ronaldtebrake:
Just checked seem to be false negatives. Topic download count on one project and event enrollment on an other. Will deal with that separately.

PR #1264, opened 2019-03-22
3021721 by jaapjan: Create routes only after the vbo module is installed

jochemvn:
## Problem
Problem with the current implementation is the forms that are written for the social_event_managers module. Those extend some forms from views_bulk_operations. When you don't have the social_event_managers enabled there's no problem, since it now has a dependency towards views_bulk_operations. However if you had already enabled social_event_managers and you take this code, you get an error, because the autoloader is not yet aware of the extended 'Drupal\views_bulk_operations\Form\ConfigureAction', causing a fatal error.

## Solution
Turns out a solution was already made in #1250 

## Issue tracker
http://www.drupal.org/node/3021721

## How to test
- [x] Install this branch on an existing platform that has the `social_event_managers` module enabled and notice the installation / update goes according to plan

## Release notes
...

### DIFF
--- a/modules/social_features/social_event/modules/social_event_managers/social_event_managers.routing.yml
+++ b/modules/social_features/social_event/modules/social_event_managers/social_event_managers.routing.yml
@@ -6,16 +6,6 @@ social_event_managers.settings:
   requirements:
     _permission: 'set social event managers settings'
 
-social_event_managers.vbo.execute_configurable:
-  path: '/node/{node}/manage-all-enrollments/configure-action'
-  defaults:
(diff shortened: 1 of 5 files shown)


jochemvn:
Only change made from #1250 is dependency injection of the module handler

ronaldtebrake:
Lets go greeeen! (Works like a charm) 💚

ronaldtebrake:
Nice, eventhough moss is / could be red i'm still merging it since now we finally have results after merging this in our PR's. From there we will get Moss green again on individual level :) so we can at least see some results.

PR #1265, opened 2019-03-22
Issue #2961232 change private message contact url when thread is deleted

jaapjan:
## Problem
When a user deletes a private message thread the contact url on the profile of participants in the thread doesn't work anymore (access denied).

## Solution
We check for access before returning the URL. In addition the cache tag is invalidated for every participant to always ensure the latest url.

## Issue tracker
https://www.drupal.org/project/social/issues/2961232

## How to test
- [x] Test first on 4.x
- [x] User 1 messages User 2
- [x] User 1 deletes the thread, but User 2 does not.
- [x] User 1 clicks on the Private Message URL in the all-members view or on the user's page and is given message: "You are not authorised to access this page."
- [x] Now try again on this branch and notice it works as expected.
- [x] Check the cache invalidation code with Xdebug.

## Release notes
When a user deletes a private message thread the contact url on the profile of participants in the thread didn't work in some cases. This issue is now solved.

### DIFF
--- a/modules/social_features/social_private_message/social_private_message.module
+++ b/modules/social_features/social_private_message/social_private_message.module
@@ -150,6 +150,31 @@ function social_private_message_thread_delete_redirect(array $form, FormStateInt
   // Force redirect to the inbox.
   $url = Url::fromRoute('social_private_message.inbox');
   $form_state->setRedirectUrl($url);
+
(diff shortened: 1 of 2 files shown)


jaapjan:
Also cherry-picked in 5.x

PR #1266, opened 2019-03-22
Issue #3041077 by ribel: As LU I want to be able to filter on tags on the all topics, events and groups overview pages

ribel:
## Problem
Currently, filter by tags is only possible on search pages, but not on the overview pages.

## Solution
Add content tagging filters to all topics, events and groups overview pages.

## Issue tracker
- https://jira.goalgorilla.com/browse/ISHN-43
- https://www.drupal.org/node/3041077

## How to test
- [ ] Check out this branch and clear the cache
- [ ] Enable Social Tagging module and create some terms and content with those tags
- [ ] Go to All topics page and see new filters by tags there, check that they work correctly
- [ ] Check that filters are also on All events, User Topics, User Event, Group Topics, Group Events and All Groups pages.
- [ ] Check with split tags option and not.

## Release notes
Before filter by tags was only possible on search pages, now we add those filters also to all topics events and groups overview pages.


ribel:
@jochemvn I have fixed this issue, please check again.

jochemvn:
I was checking it, I get some notice/error on the search page (I've sent you a video). Also I noticed only now that the filters on the topics and events overviews are single select filters... whereas on the search they are multi select filters. I think for sake of consistency and functionality we should keep it multi on both.

ribel:
@jochemvn I will check it again. Thanks for the video.

The only 2 things that I was not able to solve are:
Regarding single filters, I left a comment in a story, and now duplicating it here:

1. It was not possible to allow multiple select for social_tagging field on normal (not search) views, so I it just can be single there. I also think If user really needs multiple filters he can use search.
2. Also, I didn't add filters to User Groups page, because we don't have any filters there and so it's not possible to alter exposed form there. But I don't think that users have so many Groups so he need to filter it.

jochemvn:
@ribel Sorry... I missed that on the story :( 
1. I think we should create a follow up
2. Agree

ribel:
@jochemvn I was not able to reproduce this issue again :)
I created a branch on ECI Demo to test it, and there all looks good as well
https://x-bitbucket-pr-35-kajchqi-vncdokyquw6mc.eu.platform.sh/search/content?category1%5B%5D=5

Also, could you create follow up, and attach it to ISHN story, since I don't have access to it?

jochemvn:
Works purrrrfect 😸 

Will merge this when 💚

PR #1268, opened 2019-03-25
Issue #3043086 by bramtenhove: Add extra check to prevent notice thrown in social_profile module

bramtenhove:
## Problem
In some cases a notice is thrown in the social_profile module.

`Notice: Undefined index: field_profile_profile_tag in social_profile_form_profile_profile_edit_form_alter() (line 108 of profiles/contrib/social/modules/social_features/social_profile/social_profile.module).`

## Solution
Add an extra check to prevent a notice being thrown.

## Issue tracker
https://www.drupal.org/project/social/issues/3043086

## How to test
- [x] Take a look at the code

## Release notes
A notice could be thrown in the `social_profile` module when trying to unset the profile tag field when it was already not available. With an extra check this will no longer happen.


jaapjan:
Also cherry-picked in 5.x.

PR #1269, opened 2019-03-25
Issue #3043105: Core Page Title Block returns empty HTML when title is missing

jaapjan:
## Problem
Page Title Block returns empty HTML when title is missing

## Solution
Check if title is empty or null before returning the array.

## Issue tracker
https://www.drupal.org/project/social/issues/3043105

## How to test
- [x] Verify the titles still work
- [x] Verify tests are passing.
- [x] Check the code.

## Release notes
The Page Title block outputs empty markup when there is no title set. Since it renders empty markup the block which renders it, gets rendered as HTML and css (which comes with it) creating unnecessary spacings and margins.


jaapjan:
It seems there are some issues, mostly related to:

- [ ] Messages (indicating you enrolled to event for example)
- [ ] Access denied page (probably the title can't be found)

Marking as needs work.

PR #1262, opened 2019-03-21
Issue #3042119 by Kingdutch: The socialbase like_and_dislike_service …

Kingdutch:
…should use formatPlural

<h2>Problem</h2>
Socialbase contains an overwrite for the like_and_dislike_service to change how we display likes and dislikes. It contains an if statement that call <code>Drupal.t</code> which makes translation more difficult and increases the number of occurrences of "like" in translation files.

<h2>Solution</h2>
Use <code>Drupal.formatPlural</code> so that we can re-use the translation of <code>@count like</code> and <code>@count likes</code>

## Issue tracker
https://www.drupal.org/project/social/issues/3042119

## How to test
- [ ] Check that liking a post still shows the correct string and is translatable

## Release notes
The like_and_dislike_service in SocialBase now translates the like count correctly.

PR #1274, opened 2019-03-26
Issue #3004793 by truls1502: Updated like_and_dislike from 1.0-alpha2 to 1.0-beta1

truls1502:
## Problem & Solution
[Full Changelog](https://cgit.drupalcode.org/like_and_dislike/diff/?h=8.x-1.x&id=8.x-1.0-beta1&id2=8.x-1.0-alpha2)
- https://www.drupal.org/project/like_and_dislike/releases/8.x-1.0-alpha3
- https://www.drupal.org/project/like_and_dislike/releases/8.x-1.0-beta1

## Issue tracker
- https://www.drupal.org/project/social/issues/3004793

## How to test

- [ ] Install on 4.x. Update to this branch and run update.sh script.
- [ ] Login as LU, notice the likes functionality is still on all the correct entity types (post, content, etc.)
- [ ] Like a post, see the number of likes is updated instantly.
- [ ] When you click on the number of likes the likes are displayed in a modal window.
- [ ] Do a re-install and repeat testing of the like functionality.
- [ ] Check the logs for any warnings or errors
- [ ] Check the code changes.

## Change record


truls1502:
I am sorry for bothering you, @goalgorilla/open-social: Does anyone of you have time to check this PR? I would like to know if there is something more I need to do to get your existing projects to work smoothly, and does the update_hook works fine? 😊🤓

truls1502:
Trying retrigger MOSS

jaapjan:
Tested it, but doesn't update the number of likes unfortunately. I think that still needs some work.

See the GIF:
number-likes-not-updating

jaapjan:
Also I think it needs a change record because we change the templates. If we don't change the templates functionality will not work anymore? So this might become a problem in subthemes.

truls1502:
@jaapjan - If you are planning to release 4.9 and 5.0 then I suggest have it in 4.9 + 5.0 (rather than 5.1-2/3 etc). However, if you are planning 5.0 + 4.10 - then it should be fine in 4.10 milestone.

Yes, the subtheme needs to change update the template to get it working. 


However, it worked fine before I merged #1262 into this PR? Do you mind to test it without #1262 and let me know if it works or not? If not, then it is strange it worked fine here on local.

Kingdutch:
Hi @truls1502 and @jaapjan 

Can you explain why we need these changes in the template and styling selectors? It also feels like we're changing the Javascript logic to do something different than we were doing instead of only adhering to the new response structure.

I think this would be a breaking change that we'd need to push to 5.x (and that I'd prefer to explore a bit more because we may be able to move even closer to the module itself).

Some explanation of why the changes were made can help in evaluate whether this is truly necessary for the update or whether we can separate the module update from improvements to our usage of the module 🙂 

~ Kingdutch

jaapjan:
Closing this PR since it will be very likely we'll need a new one for 8.x or 7.x 🙈

PR #1280, opened 2019-03-28
Issue #3044423 by jochemvn: As a SM I want to be able to control the height of the hero of the group pages

jochemvn:
## Problem
The hero of the group pages takes up a lot of space. Some platform maintainers don't like this.

## Solution
On the group settings form you can now choose to use a smaller hero image. 

## Issue tracker
- https://jira.goalgorilla.com/browse/SHN-309
- https://www.drupal.org/project/social/issues/3044423

## How to test
- [x] Switch to this branch and run update script
- [x] Go to the group settings form and notice a select for crop types
- [x] Select the alternative one
- [x] Go to a group with an image and a group without an image
- [x] Edit the group with the image
- [x] Notice an extra crop type in the cropping area
- [x] Set some cropping and save
- [x] Notice the smaller hero area on both groups with or without image

## Release notes
As a sitemanager you can now select a smaller group hero type. This leaves more space on the group pages for content.


themes/socialbase/templates/group/group--hero.html.twig
```diff
@@ -28,7 +28,19 @@
   ]
 %}
 
-<div{{ attributes.addClass(cover_classes) }} {% if group_hero_styled_image_url %} style="background-image: url('{{ group_hero_styled_image_url }}');" {% endif %}>
+{%
+  set cover_wrap_classes = group_hero_small ? 'cover-wrap small' : 'cover-wrap'
+%}
+
+{%
+  set hero_image = group_hero_styled_image_url
```
ribel: @jochemvn Why not just set correct image style to `group_hero_styled_image_url` in preprocess?

jochemvn: yeah... that sounds like a pretty good idea 😅... Fix incoming

jochemvn: @ribel I've made that change... Can you quickly check? 😇

ribel: Yes, looks much better now 👍

PR #1287, opened 2019-04-01
Issue by ribel: Revert PSR-12 standard for PHP7 to avoid fatal error in classes that extend social_group RouteSubscriber

ribel:
## Problem
On sites where Social Course module is enabled all pages, there are WSOD with a message:

`Fatal error: Declaration of Drupal\social_course\Routing\RouteSubscriber::alterRoutes(Symfony\Component\Routing\RouteCollection $collection) must be compatible with Drupal\social_group\Routing\RouteSubscriber::alterRoutes(Symfony\Component\Routing\RouteCollection $collection): void in /app/html/modules/contrib/social_course/src/Routing/RouteSubscriber.php on line 15`

## Solution
It can be fixed by removing :void from social_features/social_group/src/Routing/RouteSubscriber.php chnage introduced in PR #1197 

## How to test
- [ ] Check this branch on ECI Demo

## Release notes
Not needed. Issue found during 5.0 testing day.


jaapjan:
Hi @ribel If you check the original pull request I see the `:void` in multiple places. https://github.com/goalgorilla/open_social/pull/1197/files

Any reason why you didn't change it everywhere?

ribel:
Hi @jaapjan only this one cause error in my case. And I was not sure if I need to change it everywhere.

jaapjan:
Ok @ribel  , let me rephrase, do you think it is theoretically possible the same error can occur in those cases? If so: perhaps change those as well?

ribel:
@jaapjan Theoretically yes, so then let's fix it everywhere.

ribel:
@Kingdutch, @bramtenhove, @OH17L I want to bring your attention to this issue.
Me, @ronaldtebrake and @jaapjan agreed that we will not use PSR-12 standard until it will be approved in Drupal Core.
We want to revert all changes that were made before to avoid fatal errors like in the example above.
Please make sure that you will not use it in new and existing PR's.

Kingdutch:
@goalgorilla/open-social (mentioning everyone given the impact/consensus, maybe we should move this to a JOSTM).

Can we nuance this to using it only on new code? I think adding type hints is always a breaking change. The following for example throws an error:
<br /><b>Fatal error</b>:  Declaration of Child::foo($x) must be compatible with Base::foo(int $x): array in <b>[...][...]</b> on line <b>13</b><br />
```
<?php

class Base {
    public function foo(int $x) : array {
        return [];
    }
}

class Child extends Base {
    public function foo($x) {
        return ["array"];
    }
}

$c = new Child();

echo $c->foo("bar");
```

This means we can't add typehints to code that is already being used somewhere in a minor version because it'll break sites. This is probably also an impediment for Drupal implementing it anywhere in existing code before Drupal 9 (may be good to file an issue?).

Reversing the example and adding typehints when we extend it provides the following warning.
<br /> <b>Warning</b>: Declaration of Child::foo(string $x): string should be compatible with Base::foo($x) in <b>[...][...]</b> on line <b>13</b><br />

```
<?php

class Base {
    public function foo($x) {
        return [];
    }
}

class Child extends Base {
    public function foo(string $x) : string {
        return "great " . $x;
    }
}

$c = new Child();

echo $c->foo("bar");
```

This means that we definitely should introduce type-hints in code that is not used and does not extend anything in Drupal. Adding it in those places later will be as difficult as adding typehints to existing code now. It also means that indeed, we can't use type-hints for any code that is already in the codebase or builts on Drupal Core. 😞

ribel:
@Kingdutch What is JOSTM? :)
And, so do you agree that we need to remove type-hints from existing code that extends Drupal Core classes?

Kingdutch:
@ribel Yes I agree with that :)

I just didn't want to go as far as to not use it anywhere. Anything new that we create ourselves (like helper classes or services) that does not extend something made by Drupal can still use type hints.

jaapjan:
@ribel what still needs to be done to get this PR in?

ribel:
@jaapjan Time as always :)
But I can work on this tomorrow.

ribel:
@jaapjan I have removed PSR-12 standard from all classes that exists already in 4.x OR extends Drupal Core, as was proposed by @Kingdutch. So the only class that was left with this is `SocialGroupHero` - because it was added just in 5.x AND doesn't extend Drupal Core classes.

Let me know if it's ok, and please merge if tests will pass.

jaapjan:
Thank you @ribel ! 🚀

Kingdutch:
For anyone interested, I've opened up a Drupal core issue about this problem here: https://www.drupal.org/project/drupal/issues/3050720

PR #1292, opened 2019-04-02
3045214: Destination parameter in login/signup destination is cached

ghost:
## Problem
The destination parameter on the login and signup pages ensures that a user is redirected to the proper page after login or sign-up. However the values for these parameters are cached which means that if a user receives a 403 access denied, please login, screen and then create an account instead they will go to a different page than originally intended.

## Solution
Currently the link is rendered in markup which can not take the caching into account. Instead the link should be rendered as an inline template where the caching of the link can be properly handled by rendering the link from its own render array.

## Issue tracker
https://www.drupal.org/project/social/issues/3045214

## How to test

- [x] When you try to open **/node/add/event** page as **AU** then you should be on the **/user/login?destination=/node/add/event** page.
- [x] When you click **Sing up** then you should be on **/user/register?destination=/node/add/event** page.
- [x] When you try to open **/node/add/topic** page then you should be on the **/user/login?destination=/node/add/topic** page.
- [x] When you click **Sing up** then you should be on **/user/register?destination=/node/add/topic** page.

## Release notes
Destination argument of the **Sing up** link on the **Log in** page will be the same as a destination in page path when an anonymous user trying to open few denied pages.

### DIFF
--- a/modules/social_features/social_user/social_user.module
+++ b/modules/social_features/social_user/social_user.module
@@ -194,6 +194,9 @@ function social_user_form_user_register_form_alter(&$form, FormStateInterface $f
     $form['account']['login-link'] = [
       '#markup' => new TranslatableMarkup("Have an account already? @link", ["@link" => $login_link]),
       '#weight' => 1000,
+      '#cache' => [
+        'contexts' => ['url.query_args'],
+      ],
     ];
   }
 
--- a/modules/social_features/social_user/src/Form/SocialUserLoginForm.php
+++ b/modules/social_features/social_user/src/Form/SocialUserLoginForm.php
@@ -90,6 +90,9 @@ class SocialUserLoginForm extends UserLoginForm {
       $form['username_login']['sign-up-link'] = [
         '#markup' => $this->t("Don't have an account yet? @link", ["@link" => $sign_up_link]),
         '#weight' => 1000,
+        '#cache' => [
+          'contexts' => ['url.query_args'],
+        ],
       ];
     }
 


jaapjan:
@Kingdutch could you review this tomorrow? If it's still good we can get it into 4.9 which is scheduled for Wednesday.

Kingdutch:
Although the code looks good, this doesn't actually fix the problem:

afbeelding

Kingdutch:
Some input on why this is probably the case.

The Drupal static cache enabled for the login page (which is usually fine). However, the content that you've set to be uncacheable is not dynamically rendered so it's still included in the static page cache.

jaapjan:
I will mark this for 4.10!

OH17L:
@Kingdutch Can you provide steps for how to reproduce it?

Kingdutch:
Given I am an anonymous user in browser 1
And I am on `user/login?destination=/admin/config`
Then the link to `Sign Up` should equal `user/signup?destination=/admin/config`

Given I am an anonymous user in browser 2
And I am on `user/login?destination=/admin/people`
Then the link to `Sign Up` should equal `user/signup?destination=/admin/people`
[Exception: link `user/signup?destination=/admin/config` does not equal `user/signup?destination=/admin/people`]

The destination parameter should still be correct after you've clicked on the "Sign Up" and look at the "Log in" link in either (or both) of the browsers.

Kingdutch:
I'm also able to still reproduce the issue on a non-local environment:

If I go to `user/login?destination=/admin` and I click on `Don't have an account yet? Sign up` then I am on `user/register?destination=/admin`

Then I go to `user/login?destination=/admin/content` and I click on `Don't have an account yet? Sign up` then I am on `user/register?destination=/admin` instead of `user/register?destination=/admin/content`

That means that if I sign-up I’ll get redirected to the wrong page because the link is still cached on the page.

Kingdutch:
I've created a new branch with the changes based 6.x so that we can merge there and cherry-pick to 5.x/4.x as needed. This requires Travis to re-run :(

Kingdutch (approved):

Kingdutch:
Tested the fix locally and it resolves the problem :)

Kingdutch:
Merged and cherry-picked to 8.x-4.x and 8.x-5.x

PR #1293, opened 2019-04-03
Updated search_api 1.6 to 1.12.

slowflyer:
Updated search_api 1.6 to 1.12.

Issue tracker
https://www.drupal.org/project/social/issues/3045160


slowflyer:
@jaapjan I thouhgt I found the issue why tests have been failed on the previous PR from Truls, but now I see a lot of:

>     And I should see "Access denied."                                                                                 # Drupal\social\Behat\SocialMinkContext::assertPageContainsText()
>       The text "Access denied." was not found anywhere in the text of the current page. (Behat\Mink\Exception\ResponseTextException)

https://travis-ci.org/goalgorilla/open_social/jobs/515358978 line 1310

and have no idea, why my changes should trigger that behaviour or have influence on it.

The reason, why tests failed: Tests did not respect configuration (at least in parts)
for search all and search group the group label and node title have been defined as String, not as Fulltext.

> String | String fields are used for short, keyword-like  character strings where you only want to find complete field values,  not individual words.

This means, you can't search for parts of it:
The result of the test: https://travis-ci.org/goalgorilla/open_social/jobs/515358983 line 2889 is correct, but that is not what you would expect from search and I changed configuration to Fulltext.

jaapjan:
Hi Ullrich,

Right now there seems to be an issue with the latest symfony packages which causes some tests to fail (probably for the right reason). I don't think that is related to your code. Truls already started solving this in a separate PR:
#1295 

We are on it!

Regarding this PR I think it's important that the search tests work and the logic is correct. I don't see any issue with making the search configuration Fulltext, but I haven't looked into it in detail yet.

jaapjan:
I've merged 5.x back into this branch. Let's see what the tests will do now 🤞

slowflyer:
@jaapjan only one error left https://travis-ci.org/goalgorilla/open_social/jobs/516099219 line 2264 ...
The Tjakka headache ...
The weird thing, I can't reproduce this error on my local environment ...

jaapjan:
@slowflyer I can try to reproduce this next week. Not sure what's causing this. Did you try locally with a new installation (exact steps as in the Travis scripts?)

slowflyer:
@jaapjan No I used my current config, based on the patch. When you try to reproduce verify that the group label is set to fulltext => https://dev.crowdcreation.eu/search_api.jpg and content is indexed based on these settings

slowflyer:
@jaapjan After trying different setups over the weekend with no findings, I did a new install this morning and can still not reproduce the error. Search works as expected.

slowflyer:
@jaapjan  After starting from scratch today with version 1.6, the error already occured on the next version 1.7.
Looking at the change log for this version I found this issue, which was fixed with 1.7: [https://www.drupal.org/node/2922525](https://www.drupal.org/node/2922525)

It seems that "Index items immediately" does not work in the way content is added in the Behat tests. To solve it, I added a cron run to the test, before searching for content. 

Having this in mind, it's clear why it worked when testing locally and manually.

slowflyer:
Last Travis CI build passed -> one last clean up

jaapjan:
Good catch @slowflyer Let's hope the build is passing now! 🤞

slowflyer:
Yeah! Travis green! Moss false negative green, @jaapjan ?

jaapjan:
It does seem to work partially, I fear it still needs work.. :'( I'm trying to solve it still but there is an issue with direct indexing causing the All tab and the autocomplete in search loupe to malfunction.

Test steps:

- [ ] Create node of type topic with title "Shaka".
- [ ] Click on search loupe in top and search for "Shaka". No results. (this worked before on the old installation of 5.x).
- [ ] Go to content tab on search and search for "Shaka". You can find result.
- [ ] Go to all tab on search and search for "Shaka". No result :-(
- [ ] Run cron
- [ ] Refresh, still no result.
- [ ] Clear the search index and reindex (sapi-c; sapi-i;)
- [ ] You can find it again on both indexes.

Weird thing is that I debugged search_api_entity_insert() and there the node is passed correctly to both indexes. It just doesn't update the all index.

jaapjan:
FYI:
I thought it might have something to when experimental module `social_search_autocomplete` is enabled but it seems it doesn't as I can reproduce the error with this module enabled or disabled.
https://github.com/goalgorilla/open_social/pull/1201

jaapjan:
@slowflyer 

If we manage to solve this before Monday 17:00 I'm ok with merging this for 5.0 (scheduled release on Tuesday) but if we don't I fear I need to move this to 5.1 :-( I won't have time to work on this actively, but I can do reviewing.

ronaldtebrake:
I did some digging (haven't been able to fix it yet) and I'm fairly sure I figured out the issue.

So what happens after creating a node, you would see that the entry in the database for the content & all index contains this item. However the "rendered_item" stays empty in the index for both items. 
See below:

Screen Shot 2019-04-12 at 11 10 45

Now if you go in to the views that show are shown the social_all view doesn't search for content > title (ofcourse it shows more than content, groups & users etc as well)

Screen Shot 2019-04-12 at 11 11 54

But search_content does have the content > title
Screen Shot 2019-04-12 at 11 11 41

Which means that it will show up in there (since the shaka title is indexed). 
- After a reindex, the rendered_item is filled and it shows up again in both the searches. 

So we should take a look at how and why the rendered_item isn't added to the index on node_save. 

If you look at social_all, the rendered_item was set to use the search_index view mode. 
This view mode never had any fields in it so that is why it stayed empty. 
I have changed it to use the teaser for now (consistent with the rest) and that makes it work. 

I was thinking about doing it for the full content mode, but the rendered_item is only a VARCHAR of 30, so the full content contains not enough information density. 😕 

@bramtenhove will review and test with me!

slowflyer:
At least, Travis is satisfied again ...

Followed the configuration of "social_content" search, I made node title and group label fulltext field instead of simple string.

This has positive effects, when searching for phrases in node title and group labels, but the weird thing is, it makes searching for phrases in the body or description working consistenly.

jaapjan:
Thanks guys, for diving into this some more. Interesting stuff!

It does work now for searching on the title/label indeed. However I still see a huge regression in the description search compared to the old situation. E.g. Create topic with:
Title: title
Description: description

Then search for description and you should be able to find the node in the old search api version, but it doesn't work yet in the new version. Even though we plan to make more improvements to the search relevancy.. I think this still needs to be solved before I'll merge this since the impact is too high.

I hope I can see what's causing this after lunch and hopefully find appropriate solution.

slowflyer:
indexitem

it's working, when removing the "Index items immediately" option, which is not recommended to use ...

jaapjan:
@slowflyer I think we could implement that as a solution indeed. It is better for bigger sites anyway, but it requires to have regular cron running often.

slowflyer:
@jaapjan But you need the cron run for activities and messages as well, already for a long time ... Site without regular cron runs is not working consistently today 
fe. On a Stream page you see latest topic on the right with a new topic, but until next cron run, you wont see a new topic in the activity stream ... 

There is already a need for regulary often cron runs without focusing on search

jaapjan:
@slowflyer yes totally agree! This is a change we can afford, also asked the other developers and everybody agrees :-)

Kingdutch:
@jaapjan @slowflyer ⚠️ 

Given the amount of commits in this PR it looks like you created a branch based on 8.x-4.x and are trying to merge it into 8.x-5.x

We should create a new branch based on 8.x-5.x and cherry-pick those commits that are absolutely necessary. Merging in this branch from 8.x-4.x may also apply bugfixes for 8.x-4.x that should not be made in 8.x-5.x (like the update of the AddToAny module).

Kingdutch:
Additionally #1255 made the change to remove the immediate search indexing. It also created an unpublished Change Record that can be used. The PR also includes fixes for the Travis tests with the introduction of a new test step.

jaapjan:
Yes, I won't merge it, but will create new branch for this :-)

jaapjan:
New PR which is a bit cleaner with amount of commits.

Let's see what our friend Travis says now! 🚀 

https://github.com/goalgorilla/open_social/pull/1325

PR #1299, opened 2019-04-04
Issue #3045199: prefer instanceof over method_exists

Kingdutch:
## Problem/Solution
We should encourage the use of Interfaces (this is exactly what they're for) instead of checking whether the method exists.

Performance wise the difference is negligible (<a href="https://stackoverflow.com/a/29496479/576060">instanceof is even slightly faster</a>). However, <code>instanceof</code> more clearly communicates intent because we're looking to use the <code>getOwner</code> function in the way that it's defined in that interface. When using <code>method_exists</code>, if someone were to implement <code>getOwner</code> but not return an instance of <code>UserInterface</code> then our code would still fall apart.

## Issue tracker
https://www.drupal.org/project/social/issues/3045199

## How to test
- [ ] Check tests

## Release notes (5.x)
The check in `ActivityLoggerFactory` to detect whether an entity has an owner now uses the `instanceof` method to check against Drupal's `EntityOwnerInterface`. As a result, the `MentionsInterface` now also extends this interface and the `Mentions` entity implements the setter methods for the interface that it was missing.

## Change Record
https://www.drupal.org/node/3046773


jaapjan:
It looks good, but doesn't work:
`https://travis-ci.org/goalgorilla/open_social/jobs/516098980`

Kingdutch:
@jaapjan That sounds like Trump "Great idea, doesn't work" 😋 

The reason it didn't work was that the `MentionsInterface` doesn't extend the `EntityOwnerInterface`. However, the `Mentions` entity does implement all the methods for this interface.

Technically speaking, adding requirements to an interface is a backwards compatible breaking change. This means that adding the extension of `EntityOwnerInterface` to the `MentionsInterface` would also be backwards incompatible. However, seeing as that code dealing with the `Mentions` entity (thus also any replacing entity implementing `MentionsInterface`) already expects this method to be there, I think we're safe in releasing this in 4.x with a change record to highlight the addition.

TL;DR of the above paragraph: Not implementing the `EntityOwnerInterface` in entities implementing the `MentionsInterface` can already cause bugs. This change makes sure it just causes a clearer error sooner.

jaapjan:
Fatal error: Class Drupal\mentions\Entity\Mentions contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods (Drupal\user\EntityOwnerInterface::setOwner, Drupal\user\EntityOwnerInterface::setOwnerId) in /var/www/html/profiles/contrib/social/modules/custom/mentions/src/Entity/Mentions.php on line 35

Kingdutch:
@jaapjan My mistake! :) 

Travis is green now

jaapjan:
@Kingdutch cool, Can we release this in 4.9 or do we need to put it into 5.0 release?

Could you also create the (unpublished) change record on d.o if you didn't already?

Kingdutch:
@jaapjan I've reconsidered and think we should play it safe, moving this to 5.0. 

I don't think anyone except ourselves is using the `MentionsInterface`. However, this is a pretty fatal error during an update if someone does happen to use this interface already. This would not be expected updating to 4.x but is acceptable when updating to 5.0.

The `method_exists` bugfix can remain in the 4.9 release.

I've change record for 5.x here: https://www.drupal.org/node/3046773

jaapjan:
@Kingdutch 
I think we should indeed then move it towards 5.x branch (create new PR most likely). I'll close this for now but will keep the branch so you can cherry-pick 🚀

PR #1306, opened 2019-04-08
[WIP] Clean up installed modules after testing

Kingdutch:
## Problem
Testing optional modules can affect other modules, therefore we should clean up any modules that a test enabled. However this may cause other tests to fail if they rely on modules that were enabled elsewhere.

## Solution
This cleans up modules after testing to ensure they aren't used in the
next test.

A good initial test to see how much this would break in our suite before we can let other tests depend on this new behaviour.

## Issue tracker

## How to test

## Release notes
<describe the release notes>


Kingdutch:
So this actually exposes two interesting bugs:

- It appears that the social_private_message module leaves behind some configuration that trips up our activity related modules
- The `social_embed` module currently can't be uninstalled because it doesn't properly remove its embed filter from the text format. This could be due to the configuration override?

jaapjan:
Downside of not leaving the optional modules enabled is that you won't catch any errors the optional modules might be causing in core functionality.

jaapjan:
But... the way we enable the modules should then also be improved :D

Kingdutch:
> Downside of not leaving the optional modules enabled is that you won't catch any errors the optional modules might be causing in core functionality.

That is true, but it's easier to test those specific cases. I suppose the flip-side would be to just enable all the optional modules all the time. However, then you miss instances where things break when a certain module is not enabled, so that's not really any better.

> But... the way we enable the modules should then also be improved :D

I'm not sure I understand? Do you mean in Open Social or in the tests?

Kingdutch:
I'm closing this as it should have been a draft. It would be something that's interesting to fix in the future.

PR #1310, opened 2019-04-09
Issue #3027295: remove Translate tab from Tab management landing tab on Group

jaapjan:
## Problem
When social_language is enabled you could select the Translate tab as landing tab of a group.

## Solution
Let's remove that tab and use the alter function provided.

## Issue tracker
N/A (but check the testing 5.0 doc)

## How to test
- [x] Enable social_language and social_group_default_route and notice that you have a "Translate" tab in the options. 
- [x] Switch to this branch and notice the tab is gone in the select box.
- [x] Check code.

## Release notes
Not necessary, because it is just introduced in this release.


jaapjan:
No cherry-picks needed as it's only in 5.x

PR #1315, opened 2019-04-10
Use private file system for images in landing page

jaapjan:
## Problem
Private file system should be used for any image provided by us.

## Solution
I've added it to the social private files configuration.

## Issue tracker
n/a because 5.0 isn't released yet.

## How to test
- [ ] Check if small hero image upload still works.

## Release notes
n/a because 5.0 isn't released yet.


jaapjan:
Won't be necessary to cherry-pick to 4.x as it's a new feature in 5.x.

PR #1316, opened 2019-04-11
Issue #3045772: Alter anonymous enrollment button texts

Kingdutch:
<h2>Problem</h2>
From a customer
<blockquote>I have discovered that the wording on the buttons when an AN tries to enroll in an event can be confusing. Some people interpret "Sign up" to mean sign up for the event rather than sign up for Open Social, and members who have accounts have encountered difficulty by clicking that option.

What do you think about changing the design to have buttons that say something like "Create an Account" "Log In" "Enroll as a Guest" and removing the "Enroll with account" and "Enroll without account" headings?</blockquote>

<h2>Solution</h2>
Change "Sign up" in the modal to "Create an account" and change "Sign up as guest" to "Enroll as guest".

As discussed with Xinyu and Natasha:
<blockquote>I think the problem is the sign up button. I agree that we should change the 'sign up' button into 'Create an account'. 

I don't think we should remove the headings though, they are used to categorise these 2 options conceptually. </blockquote>
<blockquote>

<ul><li>    Create an account</li>
<li>    Log in</li>
<li>    Enroll as guest</li>
</blockquote>

## Issue tracker
https://www.drupal.org/project/social/issues/3045772

## How to test
- [ ] Anonymously try to enrol in an event

## Release notes
The texts that are shown in the modal for anonymous enrolment have been slightly tweaked to clarify the difference between the options for creating an account and enrolling without an account.

### DIFF
--- a/modules/social_features/social_event/modules/social_event_an_enroll/templates/event-an-enroll-dialog.html.twig
+++ b/modules/social_features/social_event/modules/social_event_an_enroll/templates/event-an-enroll-dialog.html.twig
@@ -3,11 +3,11 @@
 <div{{ attributes.addClass('event_an_enroll','row') }}>
   <div class="col-md">
     <h5>{% trans %}Enroll with account{% endtrans %}</h5>
-    <p><a class="btn-lg btn btn-primary" href="{{ links.register.uri }}" title="{% trans %}Sign up{% endtrans %}">{% trans %}Sign up{% endtrans %}</a>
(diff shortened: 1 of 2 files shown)


Kingdutch:
This patch was from an external contributor :D I've already reviewed it and it looks good.

Just want to see what Travis says (perhaps we also need to change some tests) before merging this.

jaapjan:
I propose only 5.x :-)

PR #1317, opened 2019-04-11
Issue #3035614 by agami4: Add the ability to move the text to a new line in the card block

agami4:
## Problem
When you have no upcoming events and use the site in another language then English that has a longer phrase to describe the absence of events, the texts doesn't wrap as it should. Unsure if it's just this block or more blocks.

## Solution
Add the ability to move the text to a new line in the card block.

## Issue tracker
https://www.drupal.org/project/social/issues/3035614

## How to test
- [x] Should add more texts to 'My upcoming events' block

## Release notes
When you have no upcoming events and use the site in another language than English that has a longer phrase to describe the absence of events, the long texts now move to a new line and displayed correctly.

### DIFF
--- a/themes/socialbase/assets/css/cards.css
+++ b/themes/socialbase/assets/css/cards.css
@@ -47,10 +47,6 @@
   padding-top: 0;
 }
 
-.card__block.dropdown-header {
-  white-space: normal;
-}
-
 .card__block--table {
   padding: 1.25rem 0;
 }
--- a/themes/socialbase/assets/css/cards.css
+++ b/themes/socialbase/assets/css/cards.css
@@ -47,6 +47,10 @@
   padding-top: 0;
 }
 
+.card__block.dropdown-header {
+  white-space: normal;
+}
+
 .card__block--table {
   padding: 1.25rem 0;
 }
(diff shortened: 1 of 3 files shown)


ribel:
@jaapjan I have reviewed this fix and it's ready to be merged, but I don't know how to manage cherry-pick thing correctly since I think it should be in both 4.x and 5.x releases?
So could you merge this PR?

ribel:
See how it looks now:

my-events-in-dutch

Kingdutch:
Hey @ribel @agami4,
It looks like this is actually caused by the fact that the empty text gets the `dropdown-header` class. 

The div that is output is also malformed: `<div class="small card__block dropdown-header" "="" role="alert">`

I think it may be better to figure out where that class comes from and why the output is malformed than the fix this by adding an extra CSS rule.

This is what happens when I remove the class:
afbeelding

agami4:
Hey @Kingdutch @ribel,
I removed class "dropdown-header" when "My upcoming events" block is empty and removed previous styles. The result the same.
Please check it.

Kingdutch:
Looks good to me! :D

jaapjan:
Also in 5.x and 6.x..

PR #1318, opened 2019-04-11
Issue #3029706 by agami4: Add svg icon for search button

agami4:
## Problem
After configuring header menu to white, then click the search icon in the menu. In the light box appears afterwards, the search icon is not visible anymore, maybe because it is also white.

## Solution
Add svg icon for search button.

## Issue tracker
https://www.drupal.org/project/social/issues/3029706

## How to test
- [ ] Should change color scheme

## Release notes
After configuring header menu to the white search icon was invisible. Now it will take color of the text - according to the color scheme.


ribel:
See how it looks now:

search-icon

ribel:
@jaapjan I have reviewed this fix and it's ready to be merged, but I don't know how to manage cherry-pick thing correctly since I think it should be in both 4.x and 5.x releases?
So could you merge this PR?

jaapjan:

PR #1319, opened 2019-04-11
Issue #3047439: Add support for unpublished groups

Kingdutch:
<h2>Problem</h2>

For the course module which uses Groups we have custom logic to be able to unpublish groups. We also have a custom extension to emulate this. 

The group module is working on incorporating this functionality into the Group module itself. Using that code would make our lives a lot easier: [#2873212]

<h2>Solution</h2>
Apply the patch from the group issue to implement the group publishing behaviour so that we can start removing our custom code.

## Issue tracker
https://www.drupal.org/project/social/issues/3047439

## How to test
- [ ] This should do nothing by default. We should add permissions for site managers, content managers to publish/unpublish groups in a follow-up.

## Release notes
The group module now supports the functionality to publish and unpublish groups. This simplifies some of the functionality that we've previously built and allows us to improve on this in the future. This is currently a developer only change.

# Release target
I think we can put this in 5.0 but if the timing of that is too short then feel free to push it to 5.1. The patch can be included in any project that needs it before that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,8 @@
                 "Add computed field for Group reference": "https://www.drupal.org/files/issues/add-computed-field-without-FieldItemListComputedInterface-2718195-34.patch",
                 "Add permission to view group overview": "https://www.drupal.org/files/issues/group-2943564-2.patch",
                 "Use VBO together with group permission": "https://www.drupal.org/files/issues/2018-12-18/vbo-and-group-permission-3020883-5.patch",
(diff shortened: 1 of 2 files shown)


ronaldtebrake:
Hi @Kingdutch, this patch isn't applying as it's a patch made for the dev branch of Group (which has a lot of rewrites already) think we should take a look at getting it in 5.1.
In the meantime i will talk to Kristiaan and see if we can get this in Group before the 1.0 release :) so we can see after 5.0 has released if we can either focus on updating group or getting this patch in depending on the time the update will take.

ronaldtebrake:
Closing at as I'm hoping / rooting / aiming for the module update :)

PR #1324, opened 2019-04-15
Issue #3019660 by agami4: Add a full width for content tagging filters

agami4:
## Problem
Content tagging filters in search, themed incorrect

## Solution
Add a full width for content tagging filters

## Issue tracker
https://www.drupal.org/project/social/issues/3019660

## How to test
- [ ] Enable 'Social Tagging' extend
- [ ] Add terms of Content tags taxonomy
- [ ] Move to 'Search page' and you will see those elements on the sidebar

## Release notes
When you use 'Content Tagging Filters', 'Select2' plugin does not wrong calculate width for those element. I added a full width select elements


jaapjan:

PR #1326, opened 2019-04-15
Issue #3045773 by agami4: Add name and title attributes for dropdown button and link wrapped the teaser icon

agami4:
## Problem
The dropdown button in the activity stream, and other places, does not have a discernable name.
Lighthouse accessibility audits also report that the links wrapped around the default content type icons in the activity stream do not have a discernable name either.

## Solution
Add name attribute to button dropdown with text "More options".
Add title attribute to link wrapped the default content type icons.

## Issue tracker
https://www.drupal.org/project/social/issues/3045773

## How to test
- [ ] You can hover to link wrapped the default content type icons and you can see title current teaser
- [ ] Should open inspect element for button dropdown and you can see name attribute with text "More options"

## Release notes
Added 'title' for link wrapped tease icons and 'name' for button dropdown attributes for improve accessibility

### DIFF
--- a/themes/socialbase/templates/comment/links--comment.html.twig
+++ b/themes/socialbase/templates/comment/links--comment.html.twig
@@ -44,7 +44,7 @@
 
   {% if links['comment-edit'] or links['comment-delete'] %}
     <div class="comment__actions pull-right">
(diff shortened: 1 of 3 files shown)


jaapjan:

PR #1327, opened 2019-04-16
Revert back to using yarn.lock file

Kingdutch:
## Problem
We've been using yarn as our package manager with the accompanying
yarn.lock file.

The yarn.lock file was accidentally deleted in
1e8ee35320f8135aabb3a3f91cbde1faff9dbf3f 

## Solution
Restore it here and
remove the package-lock.json.

## Issue tracker
Internal build-tooling, no issue in tracker

## How to test
N/a

## Release notes
Not needed


Kingdutch:
Merging, given that this is build tooling that's not even covered by tests.

PR #1328, opened 2019-04-17
Update README.md for 5.x with two more badges

truls1502:
Added Packagist version and Follow their Twitter account


jaapjan:
Good idea, adding this to 6.x and will cherry-pick to 5.x :-)

PR #1329, opened 2019-04-17
Update README.md for 4.x with two more badges

truls1502:
Added Packagist version and Follow their Twitter account


jaapjan:
Let's not make unneccesary changes to older 4.x :-) Focus on 6.x and 5.x instead

truls1502:
Ok, @jaapjan! I was thinking have it in 4.x and let the people know there is a newer version than 4.x available

jaapjan:
People won't see the 4.x readme anymore. Default branch is now 6.x

truls1502:
Aha, okay! That makes sense. 👍

PR #1330, opened 2019-04-17
Update travis.yml for 4.x

truls1502:


truls1502:
@jaapjan - is it fine with 4.0 here or should it be 3.0/3.15?

PR #1331, opened 2019-04-17
Update travis.yml for 6.x

truls1502:


jaapjan:
We can implement this for 6.x, not 5.x (install update always need to test the version before this, since it's about an install in older version and then update to the latest).

truls1502:
Thank you for letting me know. Sure. Will remember that next time 😊

PR #1334, opened 2019-04-18
Issue #3045770 by agami4: Add label to search forms

agami4:
## Problem
Text filed of search form do not have a label describing what kind of textarea it is.

## Solution
Added label for text field on search forms.

## Issue tracker
https://www.drupal.org/project/social/issues/3045770

## How to test
- [ ] You can inspected search form

## Release notes
Added label to search form to improve accessibility


jaapjan:

PR #1335, opened 2019-04-18
Issue #3029738 by agami4: Remove the duplicate buttons

agami4:
## Problem
Buttons on the Content I follow page are duplicated

## Solution
Remove the duplicate buttons

## Issue tracker
https://www.drupal.org/project/social/issues/3029738

## How to test
- [ ] Go to /following page

## Release notes
Removed the wrapper div for the unfollow link.


jaapjan:

PR #1336, opened 2019-04-18
Issue #3000026 by mdeny: Add patch for message to fix Notice: 'Undefined index: value in MessageTemplate->getText()'

denysyukm:
## Problem
On Course and Challenge group stream and about pages there we see notices:
`Notice: Undefined index: value in MessageTemplate->getText()`

## Solution
I found the related issue on drupal.org (https://www.drupal.org/project/message/issues/3000026) and applied patch

## Issue tracker
- https://www.drupal.org/project/message/issues/3000026

## How to test
This issue reproduced on enterprise projects (ECI demo).
- [x] Go to the Course and Challenge group stream and about pages you can see notice (Undefined index: value in MessageTemplate->getText())
- [x] Apply patch via composer install
- [x] Go to the Course and Challenge group stream and about pages and be sure that notice (Undefined index: value in MessageTemplate->getText()) is not appear

## Release notes
Fix notice caused by message module (Undefined index: value in MessageTemplate->getText()) on custom group types.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,6 +72,9 @@
             "drupal/bootstrap": {
                 "Dropdown toggle variable ignored when using links__dropbutton": "https://www.drupal.org/files/issues/2018-12-19/dropdown-without-default-button-3021413-2.patch",
                 "JSDeliver issues unsupported operand types": "https://www.drupal.org/files/issues/2019-01-24/3027569-8.x-3.x-8.patch"
+            },
+            "drupal/message": {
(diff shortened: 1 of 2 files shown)


ribel:
@jaapjan Can we include this in 5.2 and I think it will also work for 4.11
There is an issue in message module, which was merged to dev branch a long time ago, but don't released yet. And we see the same issue on Course and Challenge modules after update to 5.x

jaapjan:
Also in 5.x.

PR #1337, opened 2019-04-19
3049398: SocialProfileFieldsOverride throws error during update to 5.x

jochemvn:
## Problem
Error during update

## Solution
No error during update 😄 

## Issue tracker
https://www.drupal.org/project/social/issues/3049398

## How to test
- [x] Update Open Social and notice you don't get config override errors


Kingdutch:
@jochemvn I think we should check whether `all_fields` is set to `true` instead and if it's set to `false` create the `fields` array if it doesn't have any fields selected for processing yet.

This solution would still break in case the `all_fields` option was unselected but the processor was not yet given any fields to process.

Kingdutch:
This PR lists some commits that should already be in 6.x 🤔

jaapjan:
As discussed on Slack it still needs a rebase and the comment you posted is still relevant

jaapjan:
Merged to 5.x and 6.x

PR #1338, opened 2019-04-22
Add update hook to enable views_bulk_operations module.

mohit-rocks:
On 8.4.x to 8.5.x update views_bulk_operations module is disabled. So it doesn't allow to import view 'group_manage_members' and group/{group}/stream pages are broken.
Enabling module before config import will fix the issue.

## Problem
group pages wrong after upgrade to 8.5
https://www.drupal.org/project/social/issues/3048980

## Solution
Adding an update hook to enable the module will do the job and by doing config import after enabling module, will allow to import the view.

## Issue tracker
https://www.drupal.org/project/social/issues/3048980

## How to reproduce
- [ ] Switch to 8.x-4.x branch and install the site
- [ ] Notice vbo module is disabled.
- [ ] Switch code base to 8.x-5.x site
- [ ] Do partial config import
- [ ] `group_manage_members` will not be imported due to dependency of `view_bulk_operations`.

## How to test
- [ ] Switch to 8.x-4.x branch and install the site
- [ ] Switch code base to 8.x-5.x site
- [ ] Apply patch and run `drush updb`
~~- [ ] Do partial config import~~ Done automatically by update hook.
- [ ] `group_manage_members` will imported now.

### DIFF
--- a/modules/social_features/social_group/social_group.install
+++ b/modules/social_features/social_group/social_group.install
@@ -518,3 +518,14 @@ function social_group_update_8502() {
   $config->set('default_hero', 'hero');
   $config->save();
 }
+
+/**
+ * Install views_bulk_operation module.
+ */
+function social_group_update_8503() {
+  $modules = [
+    'views_bulk_operations',
+  ];
+
+  \Drupal::service('module_installer')->install($modules);
+}
--- a/modules/social_features/social_group/social_group.install
+++ b/modules/social_features/social_group/social_group.install
@@ -526,6 +526,10 @@ function social_group_update_8503() {
   $modules = [
     'views_bulk_operations',
   ];
-
   \Drupal::service('module_installer')->install($modules);
+
+  // Install default configuration from social_group module so we can reimport
+  // views from the configuration set.
+  \Drupal::service('config.installer')->installDefaultConfig('module', 'social_group');
+
 }


jaapjan:
@mohit-rocks The module is already enabled in this update hook:
`social_user_update_8007` so not sure why do we need this?

mohit-rocks:
@jaapjan When I did a fresh install, the module was disabled. I can do quick check tonight and confirm once again.
I'll try with both the branches 8.x-4.x and 8.x-5.x

jaapjan:
In 4.x the module should not be enabled since we only added this module to 5.x branch

in 5.x it should be on fresh install (it is dependency of social_user) and it should be enabled after update from 4.x (as it is enabled in the update hook social_user_update_8007).

mohit-rocks:
Alright, I'll have to investigate the sequence of update hooks then.
If update hooks of `social_group` is running before `social_user` module's update hook then we might need to enable the module, otherwise just importing configuration will also do the job.

mohit-rocks:
@jaapjan Here is the output of drush updb run after switching code base to 8.x-5.x
```
drush updb
 -------------- --------------------- --------------- -------------------------
  Module         Update ID             Type            Description
 -------------- --------------------- --------------- -------------------------
  social_core    8501                  hook_update_n   Install
                                                       social_lets_connect_usa
                                                       ge module.
  social_group   8501                  hook_update_n   Install gvbo module.
  social_group   8502                  hook_update_n   Set some extra
                                                       configuration.
  social_user    8007                  hook_update_n   Install
                                                       views_bulk_operations
                                                       module.
  social_user    8008                  hook_update_n   Create the default
                                                       configuration for the
                                                       social_user default
                                                       landing page.
  social_font    montserrat_fallback   post-update     Set font-fallback for
                                                       Montserrat font to
                                                       sans-serif.
 -------------- --------------------- --------------- -------------------------


 Do you wish to run the specified pending updates? (yes/no) [yes]:
 >

>  [notice] Update started: social_group_update_8501
>  [notice] Update completed: social_group_update_8501
>  [notice] Update started: social_user_update_8007
>  [notice] Update completed: social_user_update_8007
>  [notice] Update started: social_core_update_8501
>  [notice] Update completed: social_core_update_8501
>  [notice] Update started: social_group_update_8502
>  [notice] Update completed: social_group_update_8502
>  [notice] Update started: social_user_update_8008
>  [notice] Update completed: social_user_update_8008
>  [notice] Update started: social_font_post_update_montserrat_fallback
>  [notice] Update completed: social_font_post_update_montserrat_fallback
 [success] Finished performing updates
```

It seems `social_group` is running before `social_user` module.

ronaldtebrake:
@mohit-rocks thanks for your work! 
Still have some questions though

Usually if one dependency is run before an other we would set a 

`function social_group_update_dependencies() {
  $dependencies['social_group'][8501] = [
    'social_user' => 8007,
  ];
  return $dependencies;
}`

Although I'm not 100% convinced this will fix it in any case.
 
It seems in your case `social_group` update went through successfully, this will enable `gvbo` which in turn has a dependency on `views_bulk_operations`.. not sure how that could have been successfull.

Are you sure you went through the full update steps of Open Social? With a feature revert and all (was that succesfull?)

mohit-rocks:
@ronaldtebrake Yeah, I see we are enabling modules in `social_group_update_8501()`
Can we just perform config import programmatically for this module after modules are installed in this hook?
The reason is, many times site builders might not get the errors and takes a while to figure out things.
If we perform config programmatically, they won't see those error.

jaapjan:
Hi @mohit-rocks it seems like we missed your message here, but actually the same answer applies for this issue as we gave for this issue: https://www.drupal.org/project/social/issues/3050093#comment-13095662

Closing PR for now.

PR #1339, opened 2019-04-23
Issue #3029695 by agami4: Open navigation menu on focus and hover without js

agami4:
## Problem
The navigation menu uses Bootstrap dropdowns for a lot of its items. However, these don't work when Javascript is disabled.

## Solution
Open navigation menu on focus and hover

## Issue tracker
https://www.drupal.org/project/social/issues/3029695

## How to test
- [ ] Should disable JavaScript in your browser 
- [ ] Reload page
- [ ] Hover on dropdown element on header and submenu will open

## Release notes
The navigational menu uses Bootstrap dropdowns for a lot of its items. When JavaScript will be disabled, submenus will open.


jaapjan:

PR #1341, opened 2019-04-23
Issue #2984283 by agami4: Add loader indicator when loading image

agami4:
## Problem
When loading image from any input form (stream, topic etc.) there's no loader indicating to wait.

## Solution
Show loading indicator when loading image

## Issue tracker
https://www.drupal.org/project/social/issues/2984283

## How to test
- [ ] Should create new topic or other Content Type and load image

## Release notes
When loading image from any input form (stream, topic etc.) there's loading indicator to wait.


jaapjan:

PR #1342, opened 2019-04-23
3050192: Check if Url is instanceof Url and check if the Url is routed

jochemvn:
## Problem
Sometimes non-external Urls are still not routed. This can lead to a hard crash of the platform.

## Solution
Check if an Url is routed before fetching the routename and parameters.

## Issue tracker
https://www.drupal.org/project/social/issues/3050192

## How to test

## Release notes
Sometimes non-external Urls are still not routed. This can lead to a hard crash of the platform. This is fixed now


ribel:
@jaapjan @Kingdutch I think it would be good to include this in the upcoming release, as few people reported it and it was Reviewed & tested by the community

ronaldtebrake:
Also cherry-picked to 8.x-5.x

PR #1343, opened 2019-04-24
DS-6572 in remembrance of Allo Allo - I shall attach the gallery / click handlers only ONCE

ronaldtebrake:
## Problem
We encountered issues on SaaS (but could also be on other platforms) where the image gallery in a comment upload would not open anymore after closing.

## Solution
This was caused by the good old JS firing multiple times. Which would always reopen the modal after closing. You wouldnt be able to see this unless directly but when inspecting the code you would see it being fired several times (up to 15x). This caused a transparent overlay to appear which made sure the images weren't clickable anymore.

## Issue tracker
- https://jira.goalgorilla.com/browse/DS-6572

## How to test
- [ ] Will update SaaS with this patch so you can test it there since that is where we encountered the issue, distro should still work fine as well :) 

## Release notes
In 5.0 we added a nice image upload possibility for attachments in comments. Sometimes the image gallery would not be able to open up anymore after closing. This should now work as supposed to.

### DIFF
--- a/modules/social_features/social_comment_upload/js/comment-upload.js
+++ b/modules/social_features/social_comment_upload/js/comment-upload.js
@@ -6,14 +6,14 @@
     attach: function(context, setting) {
       $('.comment-attachments').once('socialCommentUpload').each(function () {
         var $content = $('> div', this).filter(function () {
-              return !$(this).hasClass('panel-heading');
-            }),
-            $header = $('> summary', this),
-            display = true,
-            handle = function () {
-              display = !display;
-              $content.toggle(display);
-            };
+            return !$(this).hasClass('panel-heading');
+          }),
+          $header = $('> summary', this),
+          display = true,
+          handle = function () {
+            display = !display;
+            $content.toggle(display);
+          };
 
         $header
           .on('click', handle)
@@ -28,7 +28,7 @@
       var $pswp = $('.pswp')[0];
       var image = [];
 
-      $('.photoswipe-gallery-custom').each( function() {
+      $('.photoswipe-gallery-custom').once('AttachGalleryToPhotoswipeElement').each( function() {
         var $pic     = $(this),
           getItems = function() {
             var items = [];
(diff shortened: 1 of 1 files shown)


Kingdutch:
Hey @ronaldtebrake,

Another improvement we can make here is to pass the `context` parameter that we get from Drupal as the second parameter to the jQuery `$` call. This limits the scope of finding targets to process to the items that were actually attached.

That on its own would probably also solve the problem (although `once` doesn't hurt too much here) but also ensures we don't re-process the entire page unnecessarily each time.

modules/social_features/social_comment_upload/js/comment-upload.js
```diff
@@ -6,14 +6,14 @@
...
-            display = true,
-            handle = function () {
-              display = !display;
-              $content.toggle(display);
-            };
+            return !$(this).hasClass('panel-heading');
+          }),
+          $header = $('> summary', this),
+          display = true,
+          handle = function () {
```
Kingdutch: Unrelated to this PR. I feel that this will break because it binds to a global `handle` variable (missing `var` keyword). This means that any comment attachment field would always toggle the last attachment field (due to `$content` being stored in the closure)

Kingdutch: nevermind, we're using comma's 🤦‍♂️

ronaldtebrake: @Kingdutch could you create a separate issue for this? 
We have to use this for our SaaS release today so don't want to extend the scope to much.

Kingdutch: I was mistaken, it'll work as designed. We'll tackle the formatting of our Javascript somewhere else ([goalgorilla/sweet-pea](https://github.com/goalgorilla/sweet-pea/)) :)

jaapjan:
In 5.x and 6.x 🚀

PR #1344, opened 2019-04-24
3050377: Tags block on the group manage members page

ghost:
## Problem
Tags block on the group manage members page

## Solution
Hide block.

## Issue tracker
https://www.drupal.org/project/social/issues/3050377

## How to test
- [ ] Enable the **Social Tagging** module.
- [ ] Navigate to **/admin/config/opensocial/tagging-settings**.
- [ ] Check the **Group** checkbox and press **Save configuration**.
- [ ] Add **1** taxonomy term to the **Content tags** vocabulary.
- [ ] Add **1.1** taxonomy term to the **Content tags** vocabulary as a child of the **1** term.
- [ ] Create a new group and set **1.1** in the **Tag content** field.
- [ ] When you open **Manage members** tab then you should see the **Group Tags** block.

## Release notes
Hide the tags block on the group manage members page.

### DIFF
--- a/modules/social_features/social_tagging/src/Plugin/Block/SocialGroupTagsBlock.php
+++ b/modules/social_features/social_tagging/src/Plugin/Block/SocialGroupTagsBlock.php
@@ -88,6 +88,7 @@ class SocialGroupTagsBlock extends BlockBase implements ContainerFactoryPluginIn
       'entity.group.edit_form',
       'entity.group.delete_form',
       'entity.group_content.collection',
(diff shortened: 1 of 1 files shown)


Kingdutch (approved):
We've added a new route so adding that to exclude this block seems like the right way to go. A follow-up issue will have to ensure that any blocks that purposefully are added to this page properly work in the sidebar.

See https://www.drupal.org/project/social/issues/3051380 for that follow-up

jaapjan:
Thanks guys, also cherry picked for 5.x

PR #1347, opened 2019-04-25
Issue #2992959 by agami4: Add color inherit for accent button on hover

agami4:
## Problem
if we configure the accent color of the site into a dark color, the button text is still black which makes it not readable.

## Solution
Added color of the text 'inherit' for accent button on hover

## Issue tracker

## How to test
- [ ] Should hover on accent button and you will see a change in the color of the text

## Release notes
Changed text color for accent button from black to white(inherit)


jaapjan:

PR #1348, opened 2019-04-25
Add new styles for unread thread card

agami4:
## Problem
Colors are reserved for the socialblue theme which is good. However, the unread-thread class for the private message module contains no color changes at all in the socialbase theme which can lead to UX issues.

## Solution
Let socialbase change:
- The background card color to be slightly off-white (from it's default, white).
- The read-indicator a background-color of something like rgba(0,0,0,0.2)

## Issue tracker
https://www.drupal.org/project/social/issues/2949011

## How to test
- [ ] Should enable Social Base theme
- [ ] Send private message and  check styles for unread thread card and read indicaotr

## Release notes
Added new styles for unread thread:
- The background card color to be slightly off-white 
(from it's default, white) background: rgba(0, 0, 0, .02);.
- The read-indicator a background-color of something like rgba(0,0,0,0.2)

### DIFF
--- a/themes/socialbase/assets/css/message.css
+++ b/themes/socialbase/assets/css/message.css
@@ -147,12 +147,17 @@
   background-size: 22px 22px;
 }
 
+.unread-thread .card {
+  background: rgba(0, 0, 0, 0.02);
+}
+
 .unread-thread .read-indicator {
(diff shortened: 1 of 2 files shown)


jaapjan:

PR #1350, opened 2019-04-28
Fix for #3050192 - Cannot get routename of unrouted URLs

lcorbasson:
## Problem
The social_event_managers module implement the link alter hook, which checks of an URL is not external and then tries to fetch the routename... however it's not checked that the URL is routed....

## Solution
1. Check if the URL is external
2. Check if it is routed, and then check route characteristics

## Issue tracker
https://www.drupal.org/project/social/issues/3050192

## Release notes
N/A

## Change Record
Not needed


lcorbasson:
This can and should probably be backported to 5.x too (no changes)

Kingdutch:
Closing this as a duplicate of #1342 🙂 

The result appears to be pretty much the same so that's a good sign! Thanks!

PR #1282, opened 2019-03-29
SHN-310 by robertragas: add range slider to social blue to control th…

robertragas:
## Problem
Some clients find the gradient of the hero a bit too dark and would like to have the possibility to change this.

## Solution
Create a setting in the socialblue theme where SM can set the opacity of the hero gradient with a slider.

## Issue tracker
https://www.drupal.org/project/social/issues/3044302

## How to test
- [x] Go to the social blue theme settings
- [x] Notice there is a category called "hero". In this category you will find a slider to change the opacity. By default it's set to 70 which related to the 0.7 opacity in the css. Try changing this.
- [x] See that the hero gradient changes.

## Release notes
Sitemanagers now have the possibility to control the gradient of the hero. This way you can adjust the opacity to a value that best fits your theme.

PR #1351, opened 2019-04-29
Issue #3051436 by Kingdutch: Sub-theme hero gradient broken

Kingdutch:
<h2>Problem</h2>
The usage of the gradient parameter in the theme settings does not properly check for a non existent theme setting. This causes output to be generated even when the setting is not set. Because the default output with a non-existent value results in the values used for 0, there is no hero gradient being applied.

An <code>isset</code> check is performed but the variable is set a line earlier so this check will always be true.

This change was introduced in [#3044302]. #1282 

<h2>Solution</h2>
Properly utilise the return value of the get-theme setting to check whether we have a usable value instead of using isset. The cast to integer is also not needed because it will be output as string.

## Issue tracker
https://www.drupal.org/project/social/issues/3051436

## How to test
- [x] Create a sub-theme without the setting enabled
- [x] Expect the default hero but see no hero

## Release notes
The new adjustable hero background gradient could break the existing background for sub-themes of socialblue. This has been fixed to output no overriding gradient if none is configured.

### DIFF
--- a/modules/custom/improved_theme_settings/improved_theme_settings.module
+++ b/modules/custom/improved_theme_settings/improved_theme_settings.module
@@ -172,13 +172,13 @@ function improved_theme_settings_page_attachments(array &$page) {
       $hero_gradient_opacity_end = 0.7;
 
       // Alwasys cast as an integer.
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
Also cherry picked to 8.x-5-x where it was introduced.

PR #1353, opened 2019-04-30
DS-6570 - Improvement - Can’t find someone for adding to event

ronaldtebrake:
## Problem
We noticed the Add Enrollee autocomplete input field doesn't behave the same as the others for lets say adding members to a group.

## Solution
We forgot to use the custom selection handler which ensures the correct Social EntityAutocompleteMatcher is used.
In this case:
`social_features/social_core/src/Entity/EntityAutocompleteMatcher.php`

## Issue tracker
https://jira.goalgorilla.com/browse/DS-6570

## How to test
- [x] Enable Social Event Organizers
- [x] Go to Event management tab and click Add Enrollee
- [x] Search for Tony Stark (no results found)
- [x] Try the same on Adding members to a group, you will see a result
- [x] Now checkout this branch, do a `drush cr`
- [ ] Search for Tony Stark, it has been found!

## Release notes
The behavior for adding Enrollee's to an event is now similar as the one for adding members to a group.

### DIFF
--- a/modules/social_features/social_core/src/Entity/EntityAutocompleteMatcher.php
+++ b/modules/social_features/social_core/src/Entity/EntityAutocompleteMatcher.php
@@ -34,6 +34,13 @@ class EntityAutocompleteMatcher extends EntityAutocompleteMatcherBase {
       // Loop through the entities and convert them into autocomplete output.
       foreach ($entity_labels as $values) {
         foreach ($values as $entity_id => $label) {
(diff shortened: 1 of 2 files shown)


jaapjan:
Doesn't seem to work on saas, most notable:

- Tony Stark can't be found
- Enrollees are still in the suggestions

jaapjan:
Also cherry picked for 5.x

PR #1354, opened 2019-04-30
DS-6569 Minor textual change on Adding event enrollee submit message.

ronaldtebrake:
## Problem
For non event organizers (CM/SM) we show "your event" as part of the submit message, which is a little confusing.

## Solution
We now check if it's actually an organizer or manager. If that is the case we show this message. Of it's someone else we make it this event as a better descriptive message.

## Issue tracker
https://jira.goalgorilla.com/browse/DS-6569

## How to test
- [x] Login as CM
- [x] Add someone to an event
- [x] See that the text refers to the correct plural version of 

```
$singular = '@count new member is enrolled to this event.';
$plural = '@count new members are enrolled to this event.';
```
- [x] Enable Social Event Organizers
- [x] Edit an event and add a Event organisers user
- [x] Login with that user, go to that event you just edited and add an enrollee and see the text refers to the correct plural version of 

```
$singular = '@count new member is enrolled to your event.';
$plural = '@count new members are enrolled to your event.';
```

## Release notes
Only Event Organisers see the your event context, since it's theirs.

### DIFF
--- a/modules/social_features/social_event/modules/social_event_managers/src/Form/SocialEventManagersAddEnrolleeForm.php
+++ b/modules/social_features/social_event/modules/social_event_managers/src/Form/SocialEventManagersAddEnrolleeForm.php
@@ -89,8 +89,13 @@ class SocialEventManagersAddEnrolleeForm extends FormBase {
 
       // Add nice messages.
       if (!empty($count)) {
-        $singular = '@count new member is enrolled to your event.';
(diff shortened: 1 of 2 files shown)


jaapjan:
@ronaldtebrake coding standards need some work:

https://travis-ci.org/goalgorilla/open_social/jobs/526407636

jaapjan:
Cherry-picked for 5.x

PR #1359, opened 2019-05-04
Update preprocess group function to consider default image if no imag…

mohit-rocks:
…es are given.

## Problem
I start by installing a vanilla open social with composer and then drush standard documentation.

The I enable the field_ui, because it is disabled by default.

I add a default image and create a new closed test group, and the default image does not show in "".

Then I add the same image to the newly created group, and the image shows in the header.

## Solution
Consider default image.

## Issue tracker
https://www.drupal.org/project/social/issues/3039023

## How to test
- [ ] Enable field_ui module
- [ ] Upload default image in "field_group_image" field
- [ ] Create group and don't upload group image
- [ ] Save group and default image should be visible in group hero banner area.

## Release notes
<describe the release notes>


mohit-rocks:
Closing this PR in favor of https://github.com/goalgorilla/open_social/pull/1381

PR #1362, opened 2019-05-07
Issue #3045773 by agami4: Add name and title attributes for dropdown button and link wrapped the teaser icon

agami4:
## Problem
The dropdown button in the activity stream, and other places, does not have a discernable name.
Lighthouse accessibility audits also report that the links wrapped around the default content type icons in the activity stream do not have a discernable name either.

## Solution
Add name attribute to button dropdown with text "More options".
Add title attribute to link wrapped the default content type icons.

## Issue tracker
https://www.drupal.org/project/social/issues/3045773

## How to test
- [ ] You can hover to link wrapped the default content type icons and you can see title current teaser
- [ ] Should open inspect element for button dropdown and you can see name attribute with text "More options"

## Release notes
Added 'title' for link wrapped tease icons and 'name' for button dropdown attributes for improve accessibility


themes/socialbase/templates/comment/links--comment.html.twig
```diff
@@ -44,7 +44,7 @@
 
   {% if links['comment-edit'] or links['comment-delete'] or links.report %}
     <div class="comment__actions pull-right">
-      <button type="button" data-toggle="dropdown" aria-expanded="false" aria-haspopup="true" class="btn btn-icon-toggle dropdown-toggle">
+      <button type="button" data-toggle="dropdown" aria-expanded="false" aria-haspopup="true" class="btn btn-icon-toggle dropdown-toggle" name="{{ 'More options'|t }}">
```
Kingdutch: The `name` attribute is actually used in form submissions and does nothing for accessibility. We should instead use the `aria-label`

See https://developers.google.com/web/tools/lighthouse/audits/button-name

themes/socialbase/templates/node/node--teaser.html.twig
```diff
@@ -93,7 +93,7 @@
 
     {% if view_mode == 'teaser' or no_image %}
       {% block card_teaser_type %}
-        <a href="{{ url }}">
+        <a href="{{ url }}" title="{{- label|render|striptags|trim -}}">
```
Kingdutch: Title's don't help accessibility. It does help sighted users to figure out what the icon is for. 

I suggest the following:

- Before the SVG add the node bundle type as text that is only visible to screenreaders (Drupal has a class for this)
- Change the title to have this same text so that sighted users can see what the icon means when they hover over it

See: https://silktide.com/blog/2013/i-thought-title-text-improved-accessibility-i-was-wrong

themes/socialbase/templates/system/links.html.twig
```diff
@@ -42,7 +42,7 @@
   {%- endif -%}
   {%- if links.edit or links.report -%}
     <div class="comment__actions pull-right">
-      <button type="button" data-toggle="dropdown" aria-expanded="false" aria-haspopup="true" class="btn btn-icon-toggle dropdown-toggle">
+      <button type="button" data-toggle="dropdown" aria-expanded="false" aria-haspopup="true" class="btn btn-icon-toggle dropdown-toggle" name="{{ 'More options'|t }}">
```
Kingdutch: The `name` attribute is actually used in form submissions and does nothing for accessibility. We should instead use the `aria-label`

See https://developers.google.com/web/tools/lighthouse/audits/button-name

jaapjan:
@agami4 could you implement the feedback from @Kingdutch still?

agami4:
@jaapjan, I implemented all feedback from @Kingdutch

jaapjan:
Ok good. It needs to be rebased to 7.x. Could you do that?

albertoalaejos:
@agami4 the travis test fails. Can you fix that?

jaapjan:
Cherry-picked to 6.x as well

PR #1365, opened 2019-05-07
Issue #3029695 by agami4: Open navigation menu on focus and hover without js

agami4:
## Problem
The navigation menu uses Bootstrap dropdowns for a lot of its items. However, these don't work when Javascript is disabled.

## Solution
Open navigation menu on focus and hover

## Issue tracker
https://www.drupal.org/project/social/issues/3029695

## How to test
- [x] Should disable JavaScript in your browser 
- [x] Reload page
- [x] Hover on dropdown element on header and submenu will open

## Release notes
The navigational menu uses Bootstrap dropdowns for a lot of its items. When JavaScript will be disabled, submenus will open.


themes/socialbase/templates/block/block--system-menu-block--main.html.twig
```diff
@@ -41,8 +41,8 @@
   {% endblock %}
   {% if content.links.search_block is defined %}
     <div class="navbar__open-search-control">
-      <button class="navbar__open-search-block" rel="search" title="{{ 'Click to open search box'|t }}">
-      </button>
+      <a href="{{ '/search/all' }}" class="navbar__open-search-block" rel="search" title="{{ 'Click to open search box'|t }}">
```
Kingdutch: Can we add `Search` as invisible text for accessibility here?

jaapjan:
@agami4 could you implement the feedback from @Kingdutch ?

PR #1367, opened 2019-05-07
Issue #3049363 by robertragas: Add social_path_manager missing depend…

robertragas:
…encies to the make files

## Problem
Forgot to update the .make file with the new dependencies/patch for the social_path_manager

## Solution
Add to the .make file

## Issue tracker
https://www.drupal.org/project/social/issues/3049363

## How to test
- [ ] Run drush make and see the projects get downloaded, notice also the patch gets applied to the group module

## Release notes
In release 5.0 we have added the possibility to create automated path aliases for your content. This required some additional dependencies to the pathauto, ctools and redirect module, but was not properly added to the make file resulting in the Drupal.org release to be missing these modules. They are now correctly added to the make file.


ronaldtebrake:
Cherry-picked to 5.x

PR #1369, opened 2019-05-07
Issue #2984283 by agami4: Add loader indicator when loading image

agami4:
## Problem
When loading image from any input form (stream, topic etc.) there's no loader indicating to wait.

## Solution
Show loading indicator when loading image

## Issue tracker
https://www.drupal.org/project/social/issues/2984283

## How to test
- [x] Should create new topic or other Content Type and it should be clear there is a loader indicating to wait.
- [x] Should create new comment with attachment and it should be clear there is a loader indicating to wait.
- [x] Should create new post with image and it should be clear there is a loader indicating to wait.

## Release notes
When uploading an image in any file attachment form (post, comment and topic for example) there is now a loader which indicate that the user should wait for the image to upload.

### DIFF
--- a/themes/socialbase/assets/css/ajax.css
+++ b/themes/socialbase/assets/css/ajax.css
@@ -59,18 +59,21 @@ html.js .ajax-new-content:empty {
 
 .form-managed-file {
   position: relative;
-  display: inline-block;
+}
+
+.form-managed-file .form-submit[data-drupal-selector*='upload-button'] {
+  font-size: 0;
+  color: transparent;
 }
 
 .form-managed-file .js-hide.form-submit[disabled] {
-  position: absolute;
-  display: block !important;
+  position: relative;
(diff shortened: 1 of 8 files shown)


jaapjan:
It works for topics (and I assume other file types) but does not work for post and comment yet. I think that also needs to be done as part of this story. See attached movies for example.

[loader-movies.zip](https://github.com/goalgorilla/open_social/files/3178591/loader-movies.zip)

jaapjan:
@agami4 time to process my feedback still? :-)

agami4:
Hey @jaapjan,
I have started to update this issue

jaapjan:
It works really well, thanks @agami4 this is a great improvement for 6.x!

jaapjan:
I propose to keep this only in 6.x for now.

PR #1370, opened 2019-05-07
Issue #2975701 by agami4: Add content variable without default fields

agami4:
## Problem
Currently when you change display settings of the profile to display more fields these are not displayed without making any required changes to the template.

## Solution
Add content variable without default fields

## Issue tracker
https://www.drupal.org/project/social/issues/2975701

## How to test
- [x] Should add new filed to profile page
- [x] Edit some profile page
- [x] Add some content for new field
- [x] Check input the new field on page

## Release notes
When you change display settings of the profile to display more fields these are not displayed without making changes to the profile template. The content variable is now also rendered without default fields so that when you add new fields they will display

### DIFF
--- a/themes/socialbase/templates/profile/profile--profile--default.html.twig
+++ b/themes/socialbase/templates/profile/profile--profile--default.html.twig
@@ -122,6 +122,10 @@
       {% endfor %}
     {% endif %}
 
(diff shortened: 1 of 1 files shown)


jaapjan:
Merged to 5.x and 6.x, thanks

PR #1371, opened 2019-05-07
Issue #2992959 by agami4: Add color inherit for accent button on hover

agami4:
## Problem
if we configure the accent color of the site into a dark color, the button text is still black which makes it not readable.

## Solution
Added color of the text 'inherit' for accent button on hover

## Issue tracker

## How to test
- [ ] Should hover on accent button and you will see a change in the color of the text

## Release notes
Changed text color for accent button from black to white(inherit)


Kingdutch:
I'm not sure whether inheriting the text color would fix the issue. If an accent color is used that is really light with a text color that is also light then we still have the same problem.

We also have this problem on our primary buttons where we use a light color causing a light primary color to be problematic. The only real way to tackle this is to let users change both background and foreground color of the different button types.

I'm marking this as postponed because I think it needs more discussion with the product team and may need a larger scope for our component configurability.

I'd like your thoughts @jaapjan @ronaldtebrake and Xinyu

agami4:
Hey @Kingdutch,
I have fixed it after discuss with Xinyu.

I'll show you above comment Xinyu to resolve this problem

xinyumaGG:
There might be some misunderstanding during our discussion. Sorry Serhii. 
First of all, the issue we try to solve is shown in this image https://www.drupal.org/files/issues/2018-08-16/Screen%20Shot%202018-08-16%20at%2009.51.05.png Because currently the text on accent buttons is always black, when the accent colour is a dark colour, the text is not readable anymore. 
We can try 2 solutions:
1. Make the text colour of accent buttons configurable, in the same way as how the text colour in the header menu is configured. 
2. Make the text colour of accent buttons automatically adjusted based on the accent colour. 

Maybe option 1 is more realistic. Let me know what you guys think.

agami4:
Hey @Kingdutch,
Could you answer on the comment from Xinyu, which item should be selected
If we do first point, we need BE side for it

Kingdutch:
My preference would be for more configurability (option 1). I don't think we'll be able to create something that correctly predicts a text color that the user will agree upon. I do think that if we make this configurable for accent buttons we'll have to do the same for other (primary/?) buttons. 

Making this configurable does increase scope a bit so I don't know if there's time in the team. It would however solve some issues that we've been having with a different client (where we've discussed button contrast before, Xinyu).

xinyumaGG:
Cool, for this story lets go for the first option that makes the text colour of accent buttons configurable. 

Shall we discuss the button contrast in another story Alexander? I would prefer to de-colorising all the major component to avoid these issues, such as colour buttons, text over image or colour.

agami4:
@jaapjan, this issue need BE side

PR #1374, opened 2019-05-07
Issue #2976023 by agami4: Add the ability to show default profile banner image

agami4:
## Problem
Default profile banner image is not shown

## Solution
Added the ability to show default profile banner image

## Issue tracker
https://www.drupal.org/project/social/issues/2976023

## How to test
- [ ] Should load default profile banner image (/admin/config/people/profiles/manage/profile/fields/profile.profile.field_profile_banner_image)
- [ ] Go to the profile page (if this profile has a default image -> please remove) and check it.

## Release notes
When you upload a default image banner to an image, you can see this image on the profile page if this page does not have its own default image.


modules/social_features/social_profile/social_profile.module
```diff
@@ -369,6 +369,14 @@ function social_profile_preprocess_profile(array &$variables) {
     $variables['profile_hero_styled_image_url'] = ImageStyle::load('social_xx_large')
       ->buildUrl($profile->field_profile_banner_image->entity->getFileUri());
   }
+  else {
+    if ($file_uuid = $profile->field_profile_banner_image->getSetting('default_image')['uuid']) {
```
Kingdutch: We should move this outside of the if to avoid assignment in the if statement (it doesn't add succinctness here).

We should also check we got something back before accessing `uuid` (isset/empty?) because an empty array here could throw errors.

jaapjan:
@agami4 are you processing the feedback from @Kingdutch still?

modules/social_features/social_profile/social_profile.module
agami4: I have updated it, but need fix problem with access on the imege

agami4:
@Kingdutch, 
need fix bug with access for private files on this banner.

jaapjan:
Closing PR since it has been on `needs work` for a while

PR #1345, opened 2019-04-25
by robertragas: add the social_content_block module to be able to cre…

robertragas:
## Problem
Lack of possibilities to enrich the platform with list of content in sidebar blocks apart from the ones who are defined by Open Social itself.

## Solution
Create a custom block type where people can filter on taxonomy terms and create a list of content which they then can place in landing pages.

## Issue tracker
https://jira.goalgorilla.com/browse/SHN-391

## How to test
- [x] Enable the social_content_block module
- [x] Go to the custom block library and add a new block of this type
- [x] Add some filters on taxonomy terms and more
- [x] Save the block and now create a new landing page
- [x] Add a block and in the second sidebar you are able to choose this new block
- [x]  Save and see it shows teasers now of topics.

## Release notes
We added the possibility to create custom blocks containing a list of teasers of topics. This list is defined in the block by selecting filters as Content tags, Groups and Topic types.

You can add this by creating a custom content block and when created by adding this as a secondary block in landing pages

PR #1376, opened 2019-05-07
Issue #3064560 by Makishima: Makes the activity statistics block on the landing page extensible by other modules

VolodymyrMartsynenko:
## Problem
As a LU I want to see relevant statistics for Challenges and Ideas in the Activity overview block on the landing page

## Solution
Add the alter hook in the social_landing_page module to allow other modules to extend Activity overview block items

## Issue tracker
https://www.drupal.org/project/social/issues/3064560
https://jira.goalgorilla.com/browse/FOR-163

### DIFF
--- a/modules/social_features/social_landing_page/social_landing_page.api.php
+++ b/modules/social_features/social_landing_page/social_landing_page.api.php
@@ -1,0 +1,26 @@
+<?php
+
+/**
+ * @file
+ * Hooks provided by the Social Landing Page module.
+ */
+
+/**
+ * @addtogroup hooks
+ */
+
+/**
+ * Provide a method to alter array of activity overview block items.
+ *
+ * @param array $build
+ *   Array with build data.
+ *
+ * @ingroup social_landing_page_api
+ */
+function hook_social_landing_page_activity_overview_alter(array &$build) {
+  unset($build[0]['group_info']);
+}
+
+/**
+ * @} End of "addtogroup hooks".
+ */
--- a/modules/social_features/social_landing_page/src/Plugin/Block/ActivityOverviewBlock.php
+++ b/modules/social_features/social_landing_page/src/Plugin/Block/ActivityOverviewBlock.php
@@ -4,6 +4,7 @@ namespace Drupal\social_landing_page\Plugin\Block;
 
 use Drupal\Core\Block\BlockBase;
 use Drupal\Core\Database\Connection;
+use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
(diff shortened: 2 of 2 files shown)


modules/social_features/social_landing_page/social_landing_page.api.php
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Provide a method to alter array of activity overview block items.
+ *
+ * @param array $build
+ *   Array with build data.
+ *
+ * @ingroup social_landing_page_api
+ */
+function hook_social_landing_page_activity_overview_alter(array &$build) {
```
OH17L: Add some example code.

modules/social_features/social_landing_page/src/Plugin/Block/ActivityOverviewBlock.php
```diff
@@ -274,6 +274,10 @@ public function build() {
       ],
     ];
 
+    \Drupal::moduleHandler()
```
OH17L: Use the injected service.

modules/social_features/social_landing_page/src/Plugin/Block/ActivityOverviewBlock.php
```diff
@@ -274,6 +285,10 @@ public function build() {
       ],
     ];
 
+    $this->moduleHandler()
```
OH17L:
```suggestion
    $this->moduleHandler
```

Kingdutch:
Hey I just saw this in the ECI demo and had some questions.

I think we add alter hooks a lot which opens up a large API surface that we have to support (essentially the entire block structure is now 'frozen'). Instead we should probably create a hook that simply requests additional stats in a specified format and just merges this. I'll add this as a discussion point to the Join OS Team Meeting.

Additionally we've implemented an alter hook that is at the end of the build output but this scenario would already be covered by Drupal's https://api.drupal.org/api/drupal/core%21modules%21block%21block.api.php/function/hook_block_view_alter/8.6.x which we've now duplicated.

OH17L:
I think will be better to create new plugin type instead of a hook.

OH17L:
Also, we can not use **hook_block_view_alter** because this block adding via block field widget which just calls **build** method.

Kingdutch:
@OH17L 

Creating a statistic plugin (system) may actually not be a bad idea. It should allow us to control the caching, use it in different places and easily make it configurable what is displayed.

You could also use this for the other block types  and add a context (group/user) system.

The block instance configuration form could show the user all the plugins that support the current context. The plugin could then also have a `supportsChange` method that allows you to configure it to display the change of a statistic in a certain time period.

Is this something that is possibly already available in the Social KPI module? I think it could be a really cool feature that is useful in quite some places but it will require some architecture and might be outside of the scope for this story.

@jaapjan @ronaldtebrake thoughts?

Kingdutch:
As for 
> Also, we can not use hook_block_view_alter because this block adding via block field widget which just calls build method.

This was fixed in https://www.drupal.org/project/block_field/issues/2978883 which is marked for review.

ribel:
@VolodymyrMartsynenko could you create Issue on drupal.org, reference each other and update title here?

jaapjan:
What about the feedback from @OH17L and @Kingdutch is this going to be processed? I do agree it would be nice, but I would be ok with a follow-up as well as that would require a lot more work to refactor.

jaapjan:
Marking it as `needs work` since it would be better to implement the block_field patch instead of creating our own patch which does the same.

ronaldtebrake:
The referenced PR: #1345 
Will actually already add this patch. 
I'll leave it up to you guys to either remove this issue as duplicate, or use this issue to create the statistic plugin system. Either way the patch should be able to solve the scope of this issue.
Let me know what you guys decide.

ribel:
Since we now can use `hook_block_view_alter`, we don't need to have duplicated custom hook. So I think we can close this ticket. And we will refactor existing code in `social_challenge` module.
Also, I am leaving issue https://www.drupal.org/project/social/issues/3064560 with a solution to implement a plugin system when we will have time.

PR #1378, opened 2019-05-07
#3053234: Replaced incorrect route for default profile landing page at user settings

bramtenhove:
## Problem
The default settings works fine, switching from the default (stream) to the information page also works. However when switching back to the default the profile page redirect breaks and you end up on an empty profile page.

Screenshot 2019-05-07 at 20 22 15

This feature and the issue was introduced in #1284.

## Solution
Changed the default route in the settings form from the entity canonical url to `social_user.stream`.

## Issue tracker
https://www.drupal.org/project/social/issues/3053234

## How to test
- [x] Go to a users profile, you should see the stream
- [x] Switch the default profile landing page to "Information"
- [x] Go to a users profile, you should see the information page
- [x] Switch the default profile landing page to the initial setting "Stream"
- [x] Go to a users profile, you should see the stream (where before it would show an empty page.

## Release notes
Switching back to the default profile landing page setting now delivers the expected page result to users.


bramtenhove:
Cherry-picked to 5.x

PR #1380, opened 2019-05-08
Issue #3050760: Ensure event max enroll limit takes anonymous enrollments into account

bramtenhove:
## Problem
The enroll button for anonymous users is never showing the 'no spots left' disabled button when there was at least one anonymous enrollment.

There was only a check if the current user had an event enrollment. But when there was already one anonymous enrollment, this would always be positive for anonymous visitors as the user ID would be 0.

## Solution
Refactored the check for authenticated enrollments and added a new one for anonymous enrollments if the module was enabled.

Also moved AN enroll functions to a service, where I also added a function to see if there was a "logged-in" visitor looking at the event.

## Issue tracker
https://www.drupal.org/project/social/issues/3050760

## How to test
- [x] Set max enroll limit for an event at 2
- [x] Enroll as authenticated user, you should see the "Enrolled" state button
- [x] Enroll as anonymous visitor, you should see the "Enrolled" state button
- [x] As another anonymous visitor try to enroll, you should see "No spots left" instead of the "Enroll" button
- [x] As another authenticated user try to enroll, you should see "No spots left" instead of the "Enroll" button 

## Release notes
The event enroll button is now respecting the maximum enrollment limit of events for anonymous visitors.

### DIFF
--- a/modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
+++ b/modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.module
@@ -130,7 +130,7 @@ function social_event_an_enroll_views_post_render(ViewExecutable $view, &$output
       if (!empty($view->args[0])) {
         $nid = $view->args[0];
         $node = Node::load($nid);
-        $an_count = social_event_an_enroll_count($nid);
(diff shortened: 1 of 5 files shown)


bramtenhove:
Cherry-picked to 5.x

PR #1381, opened 2019-05-09
Update preprocess group function to consider default image if no images are given

mohit-rocks:
## Problem
I start by installing a vanilla open social with composer and then drush standard documentation.

The I enable the field_ui, because it is disabled by default.

I add a default image and create a new closed test group, and the default image does not show in "".

Then I add the same image to the newly created group, and the image shows in the header.

## Solution
Consider default image.

## Issue tracker
https://www.drupal.org/project/social/issues/3039023

## How to test
- [ ] Enable field_ui module
- [ ] Upload default image in "field_group_image" field
- [ ] Create group and don't upload group image
- [ ] Save group and default image should be visible in group hero banner area.

## Release notes
<describe the release notes>

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -38,6 +38,7 @@ use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\social_group\GroupContentVisibilityUpdate;
 use Drupal\Core\Messenger\MessengerInterface;
+use Drupal\field\Entity\FieldConfig;
 
 /**
  * Implements hook_form_FORM_ID_alter().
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
@mohit-rocks Thanks for your PR.
When I tried to test this it didn't work for the default group url. 
It tries to grab the default image and render it in a certain image style but unfortunately this default image is not available in the imagestyle. Resulting in a URL not found.

See screenshots for some more information:

Screen Shot 2019-06-24 at 12 23 58
Screen Shot 2019-06-24 at 12 23 28
Screen Shot 2019-06-24 at 12 24 32

mohit-rocks:
Thanks for evaluating @ronaldtebrake 
I'll quickly check it again and will update the patch.

modules/social_features/social_group/social_group.module
```diff
@@ -235,23 +236,32 @@ function social_group_preprocess_group(array &$variables) {
...
-    // Fetch image style from field info.
-    $original_image_style = $variables['content']['field_group_image'][0]['#image_style'] ?? '';
-    $image_style = ImageStyle::load($original_image_style);
-
-    if ($image_style instanceof ImageStyle) {
+    $group_image = $group->get('field_group_image')->entity;
+  }
+  else {
+    // Provide default image from field configuration.
+    $field_info = FieldConfig::loadByName($group->getEntityTypeId(), $group->bundle(), 'field_group_image');
```
Kingdutch: Consider the following here:
```suggestion
    $default_image = $group->get('field_group_image')->getSetting('default_image');
```

Then we don't need to import FieldConfig and load it by name separately. The field item should already have this information :)

Kingdutch: Comes from a comment here: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Field%21FieldDefinitionInterface.php/function/FieldDefinitionInterface%3A%3AgetDefaultValue/8.2.x

jaapjan:
Closing PR since it's been open for a while. Please feel free to re-open after applying the fixes. Thanks in advance!

PR #1382, opened 2019-05-09
TB-1894 - Auto login links

royharink:
## Problem
New registered users get lost when they have signed up and had to change their password. There isn't a way to point them in the right direction.

## Solution
Added a module with a service and controller that can be implemented to create links. The link can be used to let the user log in automatically (using password reset mechanisms) and redirect the to the correct destination. This could be a group for example.

## Issue tracker
https://www.drupal.org/project/social/issues/3056361

## How to test

## Release notes
A new module has been added that allows developers to create one-time login links. These links do not force a password reset when a user already has password. This makes them useful in some situations where a password reset link is not appropriate.

### DIFF
--- a/modules/custom/social_magic_login/social_magic_login.module
+++ b/modules/custom/social_magic_login/social_magic_login.module
@@ -5,6 +5,7 @@
  * The Social Magic Login module.
  */
 
+use Drupal\Core\Url;
 use Drupal\user\UserInterface;
 
 /**
@@ -33,7 +34,6 @@ function social_magic_login_token_info() {
  * @throws \Drupal\Core\Entity\EntityMalformedException
  */
 function social_magic_login_tokens($type, $tokens, array $data, array $options, \Drupal\Core\Render\BubbleableMetadata $bubbleable_metadata) {
-  $token_service = \Drupal::token();
   $url_options = ['absolute' => TRUE];
   $replacements = [];
 
@@ -64,9 +64,9 @@ function social_magic_login_tokens($type, $tokens, array $data, array $options, 
 /**
  * Wrapper method for Service MagicUrlCreate.
  *
- * @see \Drupal\social_magic_login\Service\MagicUrlCreate::create
+ * @see \Drupal\social_magic_login\Service\MagicUrl::create
  */
-function social_magic_login_url_create(UserInterface $account, $destination, array $options) {
+function social_magic_login_url_create(UserInterface $account, string $destination, array $options) : ?Url {
   $magic_url_service = \Drupal::service('social_magic_login.create_url');
   return $magic_url_service->create($account, $destination, $options);
 }
--- a/modules/custom/social_magic_login/social_magic_login.module
+++ b/modules/custom/social_magic_login/social_magic_login.module
@@ -38,16 +38,17 @@ function social_magic_login_tokens($type, $tokens, array $data, array $options, 
 
     foreach ($tokens as $token => $original) {
       switch ($token) {
-        case 'magic-login-token':
+        case 'auto-login-token':
 
+          // Must be internal url.
           $group_url = $group->toUrl('canonical', [
             'absolute' => FALSE,
           ])->toString();
           if (!empty($group_url)) {
-            $url_options['destination'] = $group_url;
+            $destination = $group_url;
           }
 
-          $url = social_magic_login_url_create($user, $url_options);
+          $url = social_magic_login_url_create($user, $destination, $url_options);
 
           $replacements[$original] = $url->toString();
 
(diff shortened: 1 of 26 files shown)


modules/custom/social_magic_login/src/Service/MagicUrlCreate.php
```diff
@@ -0,0 +1,76 @@
...
+   *   The uri of the final destination.
+   * @param array $options
+   *   (optional) A keyed array of settings. Supported options are:
+   *   - langcode: A language code to be used when generating locale-sensitive
+   *    URLs. If langcode is NULL the users preferred language is used.
+   *
+   * @return \Drupal\Core\Url
+   *   An url based on the magic login route.
+   */
+  public function create(UserInterface $account, $destination, array $options) {
```
Kingdutch: Since this is a new service we can add parameter and return type hints here.

```suggestion
  public function create(UserInterface $account, string $destination, array $options) : ?Url {
```

modules/custom/social_magic_login/src/Service/MagicUrlCreate.php
```diff
@@ -0,0 +1,76 @@
...
+   * @param \Drupal\user\UserInterface $account
+   *   An object containing the user account.
+   * @param string $destination
+   *   The uri of the final destination.
+   * @param array $options
+   *   (optional) A keyed array of settings. Supported options are:
+   *   - langcode: A language code to be used when generating locale-sensitive
+   *    URLs. If langcode is NULL the users preferred language is used.
+   *
+   * @return \Drupal\Core\Url
```
Kingdutch: We should document the fact that invalid parameters can return a null type here.
```suggestion
   * @return \Drupal\Core\Url | NULL
```

modules/custom/social_magic_login/src/Service/MagicUrlCreate.php
```diff
@@ -0,0 +1,76 @@
...
+   *   An url based on the magic login route.
+   */
+  public function create(UserInterface $account, $destination, array $options) {
+    if (!isset($account, $destination)) {
```
Kingdutch: Is there a way we can allow login links without a destination? Just letting the default after-login routing do its things? Essentially making the `$destination` optional.

modules/custom/social_magic_login/src/Service/MagicUrlCreate.php
```diff
@@ -0,0 +1,76 @@
...
+namespace Drupal\social_magic_login\Service;
+
+use Drupal\Core\Path\PathValidatorInterface;
+use Drupal\Core\Url;
+use Drupal\user\UserInterface;
+
+/**
+ * Class MagicLoginCreate.
+ */
+class MagicUrlCreate {
```
Kingdutch: I think we should follow Drupal Core's example and create an interface of the MagicUrl service `MagicUrlInterface` which is implemented by `MagicUrl`. This would allow others to change the way this works by swapping out the service and controllers, leaving our tokens intact.

modules/custom/social_magic_login/social_magic_login.module
```diff
@@ -0,0 +1,87 @@
...
+          'absolute' => FALSE,
+        ])->toString();
+
+        $url = social_magic_login_url_create($account, $destination, $url_options);
+
+        $replacements[$original] = $url->toString();
+      }
+    }
+
+    if ($auto_login_tokens = $token_service->findWithPrefix($tokens, 'view-destination')) {
```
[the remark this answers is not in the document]
royharink: Wasn't entirely sure what to do with that. It's support for all the views tokens. I had that for groups as well, but it doesn't make sense as we derive an url from a group ID.
For views it doesn't really make sense to support all the tokens. I think we should discuss this tomorrow.

modules/custom/social_magic_login/src/Service/MagicUrlCreate.php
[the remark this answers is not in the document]
royharink: Sure.
Maybe better to use a more professional name as well? :P

modules/custom/social_magic_login/social_magic_login.module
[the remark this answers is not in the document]
royharink: We'll be removing the out of scope and a followup will be created.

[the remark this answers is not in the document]
Kingdutch: https://www.drupal.org/project/social/issues/3057348

modules/custom/social_magic_login/src/Service/MagicUrlCreate.php
Kingdutch: https://www.drupal.org/project/social/issues/3057348

Kingdutch:
Follow-ups have been created and the review has been completed.

PR #1384, opened 2019-05-09
Issue #3053668: Load all activity entities at once instead of one by one

bramtenhove:
## Problem
When a user marks notifications as read (automatically) this will load activity entities one by one.

## Solution
Load all activity entities at once.

## Issue tracker
https://www.drupal.org/project/social/issues/3053668

## How to test
- [x] Generate some notifications for a user
- [x] Log in as that user, hover over the notification centre
- [x] After a few seconds the notifications should be marked as read
- [x] Reload the page, check that they are no longer marked as read

## Release notes
By refactoring old code we improved the performance of the "mark as read" functionality of the notification centre.


jaapjan:
Merged to 5.x and 6.x

PR #1385, opened 2019-05-09
Issue #3053671 by robertragas: Check if the message thread exists bef…

robertragas:
## Problem
In the private message activity context it gets all the members from a certain thread so it can create the activity for all the correct people. This function is not checking correctly if a thread still exists, and when this threat does not exist it will return false, which crashes.

## Solution
Check if the thread is instanceof PrivateMessageThreadInterface

## Issue tracker
https://www.drupal.org/project/social/issues/3053671

## How to test
- [x] Send a private message and delete the thread before the activity is created.
- [x] Notice it will error because we are trying to use method from boolean
- [x] Checkout to this branch and try again

## Release notes
We have added better defensive checking to prevent errors if a private message thread is deleted.


jaapjan:
Merged to 5.x and 6.x

PR #1386, opened 2019-05-09
Improve event-max-enroll and topic-download-count test

jaapjan:
## Problem
1. When Sitemanager does not have permission to go to `admin/config/opensocial` the test is failing.
2. Topic download count test seems to fail if other humans.txt files were already uploaded.

## Solution
1. Let's go to the configuration page directly.
2. Make the check a bit smarter so we don't care if somebody uploads human 0.txt or humans 200.txt or humans.txt.

## Issue tracker
not needed

## How to test
- [x] Just check the code changes
- [x] Check if Travis and test in Moss is passing

## Release notes
not needed...


ronaldtebrake:
GREEEN TESSSTS WOOOOPWOOOOP!

ronaldtebrake:
Cherry picked!

PR #1394, opened 2019-05-13
Issue #2948988 by agami4: Add an option to show event creator on event teasers

agami4:
## Problem
On an event it's possible to see who created an event but this is not visible on teasers even though there's some room. It would be nice if the site administrator could configure to show this.

## Solution
Added an option to show event creator on event teasers

## Issue tracker
https://www.drupal.org/project/social/issues/2948988

## How to test
- [ ] Go to edit some event
- [ ] Enable the  'Display event creator on event teasers' checkbox
- [ ] Save this event
- [ ] Go to '/community-events' and find this node and you can see 'Event Creator'

## Release notes
You can show/hide event creator on event teasers


Kingdutch:
This needs some work. There are quite a lot of commits in this PR that don't belong here and are part of other work. This makes reviewing difficult (30 changed files).

Furthermore, a new field is added but I don't think we should do that. The story's description talks about adding a configuration setting that a site manager can control. This would mean that it is handled on platform level, not per event as happens in this PR.

agami4:
Hey @Kingdutch,
For 30 changed files - i don't know how those all files are preset in this PR. I remove this PR and create new.
Can you add more details to your idea and whether this is a FE task

Kingdutch:
Hi @agami4,

I think configuration is usually done by back-end developers. The Drupal Form API will generate any needed Front End for us and the Admin theme we use takes care of the rest :)

PR #1395, opened 2019-05-13
Issue #3054419 by chmez: Tour pop-up isn't hidable

ghost:
## Problem
Tour pop-up still visible after clicking on **Don't show tips like this anymore**.

## Solution
Clear cache related to tags based on tour types.

## Issue tracker
https://www.drupal.org/project/social/issues/3054419

## How to test
- [x] Enable **Social Tour** module.
- [x] Log in as **Normal Member**.
- [x] When you click **Don't show tips like this anymore** then you shouldn't see the **Welcome on the home page stream!** pop-up window after page reload.

## Release notes
Tour pop-up will not be showed next time when a user disabled it manually.

### DIFF
--- a/modules/custom/social_tour/config/install/tour.tour.social_event_overview.yml
+++ b/modules/custom/social_tour/config/install/tour.tour.social_event_overview.yml
@@ -4,7 +4,7 @@ translatable: true
 dependencies:
   module:
     - social_event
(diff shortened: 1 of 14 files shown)


jaapjan:
Only in 6.x is ok. Btw: this module is obsolete

PR #1396, opened 2019-05-14
Issue #3054601: VBO Javascript for Manage Members and Enrollees is attached too often

Kingdutch:
<h2>Problem</h2>
VBO Javascript for Manage Members and Enrollees is attached too often

<h2>Solution</h2>
Take context into account

The click handlers were added many times because the context of the
attach method was not taken into account.

Additionally the action button was before the temporary placeholder in
the DOM order which caused it to be unclickable.

## Issue tracker
https://www.drupal.org/project/social/issues/3054601

## How to test
- [x] Visit a group with 50 members
- [x] Select all members

<h2>Release notes</h2>
An over eager attaching of click handlers in the group member and event enrollee management interfaces caused some duplicate messages when all items were selected. We've had a chat with the click handler and it's no longer overly attached.

### DIFF
--- a/modules/social_features/social_event/modules/social_event_managers/js/frontUi.js
+++ b/modules/social_features/social_event/modules/social_event_managers/js/frontUi.js
@@ -9,20 +9,20 @@
 
   Drupal.behaviors.viewsBulkOperationsFrontUi = {
     attach: function (context, settings) {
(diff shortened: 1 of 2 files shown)


ronaldtebrake:
Also cherry picked in 8.x-5.x!

PR #1397, opened 2019-05-14
Issue #3054534 by robertragas: add patch to remove group machine name…

robertragas:
## Problem
When you have multilingual sites you are able to translate machine names of the group types. Translating this will break your website.

## Solution
Change type from text to string so they are not translatable anymore.

## Issue tracker
https://www.drupal.org/project/group/issues/3054534

## How to test
- [ ] Add a different language and go to translations and translate a group machine name such as closed_group
- [ ] Add closed group with content so it's present throughout the website.
- [ ] Check out to this branch and remove the source translation from the database
- [ ] It will not get imported anymore and not break the site.

## Release notes
We have added a fix for the group module removing the possibility to translate the machine names for group types which would break the site if translated.


ronaldtebrake:
Cherry picked in 4, 5.x

PR #1398, opened 2019-05-14
Issue #3054708: Add validation to check for unique nick names

bramtenhove:
## Problem
At this moment Open Social provides a nick name field via the Social Profile Fields module.

There are a few use cases where platforms would like the nick name to be unique. E.g. when it's meant as the primary "name" field. When in that case a lot of users have the same nick name (and often no image uploaded) it becomes very hard to distinguish between users.

## Solution
Add an option in the settings to apply a unique nick name validation constraint. By default it is not applied.

When applied it will validate the uniqueness of a nick name when the profile is saved through the profile edit form page.

## Issue tracker
https://drupal.org/project/social/issues/3054708

## How to test
First make sure you have the Social Profile Fields module enabled.

- [ ] Set nick name for profile X to `Pete`
- [ ] Set nick name for profile Y to `Pete`, should be no problem
- [ ] Go to /admin/config/opensocial/profile-fields, check the unique nick name validation option
- [ ] Go to profile Y, re-save the profile form, you'll get a validation error about the unique nick name
- [ ] Change your nick name to something else
- [ ] Re-save your profile without changing anything, no error should be thrown

## Release notes
TBD


bramtenhove:
Please be aware that there is *no* functionality added that updates users nicknames after the checkbox is turned on.

Also this still needs a Behat test!

robertragas:
I tested the steps above described by Bram and they work as expected.
As Bram said, a behat test is still needed, but also I see the moss branch failed, needs to be checked out (by Bananas)

jaapjan:
Who is going to write this behat test and work on Moss so it can be merged?

jaapjan:
Will be postponed to 6.1

bramtenhove:
Just changed the base branch to `7.x`. Behat test went fine on Travis although it doesn't seem to be kicked off again after I changed the base branch.

ronaldtebrake:
Rebased and merged in different PR :) 
#1516

PR #1281, opened 2019-03-28
Issue #2898087: Create "secret" group type/Private groups

OH17L:
## Problem
As a LU I want to be able to create a Secret Group

## Solution
1. Can't be found in the Search, not visible. 
2. Users can talk about campaign preparation that can't be visible before. 
3. Only possible for SM to create secret groups
4. Hidden on every place, but SM needs to find it in the Search
5. As a member of the Group you can find it in Search, and Group list of other members
6. Show it in My Groups
7. If you are a member this Group is no longer secret for you as a member
8. SM doesn't need to be able to change the Group type 
9. When a non-member visits the group we act as if the group never existed (404 not found) instead of access denied.

## Issue tracker
https://www.drupal.org/project/social/issues/2898087

## How to test
- [ ] Only SM+ can create a group.
- [ ] Nobody can change group type.
- [ ] A group and group content should be visible only for members and for CM+.
- [ ] Users without access should be redirected to page 404 instead of page 403 from group pages.

## Release notes
Create a new group type which accessible only for members and CM+.

PR #1399, opened 2019-05-16
#3055164 - Flexible Group Type

jaapjan:
## Problem

1. Allowed visibility for this group type is set to public & community by default.
2. When creating group post and having the "Group type permission" for "Access posts" to Anonymous it would always be visible.

## Background info + Solution:
Information

In order for the access to work properly we have to configure it as if it was a public/open group for seeing tabs.
This is due to the fact we still have a lot of checks that check if `$account->hasPermission('');`
So we need to allow it to not run in to access denied errors while it is allowed.
This means we have to treat our custom access to ensure route based access / entity based access checks for denials.

Therefor I have decided to create Services that based access for routes on flexible group custom options.
Next to that the I had to update entity access by field module, to take in to account that for flexible groups
Even if you have access to view a group, that is overridden by the group based option.

Also there is a new hook: `social_group_flexible_group_content_routes`
This hook allows all extensions that have group_content that will be added as content plugin to a flexible group
To grab the same content access data the other tabs rely on.

Do you have any other access requirement, see how we create a route subscriber with custom access check and do the same
with your logic.

## Issue tracker
https://www.drupal.org/project/social/issues/3055164

## How to test
- [ ] Enable the social flexible group module
- [ ] Create some flexible groups with different join / content visibility settings
- [ ] See that it works as expected, also the tabs access should work as expected.

## Release notes
Will Follow

## Change Record
Will Follow


jaapjan:
Note: some work is already done in the Secret Groups PR #1281 which might be useful for this as well:

- Alter `hook_social_group_allowed_visibilities` to change allowed visibilities per type
- Function `social_group_can_view_groups_of_type ` which calculates if someone is allowed to view the group.

I think both of these need to be made more flexible, since our logic is not based on the type, but it's based on the combination of the type and the settings within the group.

jaapjan:
Just updated this branch with latest changes from 6.x.

PR #1400, opened 2019-05-16
Fix Social module names for better consistency

ribel:
## Problem
I noticed that Social prefix is missing for **Max Event Enroll** module and second word for **Alternative frontpage** is in lowercase.

## Solution
Add Social prefix and use Uppercase to be consistent.

## Issue tracker
- None

## How to test
- [ ] Login as admin, go to the module list and see that it looks nice :)

## Release notes
I think it's not needed.


modules/social_features/social_event/modules/social_event_an_enroll/social_event_an_enroll.info.yml
```diff
@@ -1,4 +1,4 @@
-name: 'Social Event AN Enroll'
+name: 'Social Event Anonymous Enrollments'
```
Kingdutch:
```suggestion
name: 'Social Event Anonymous Enrolments'
```

See the `Social Event Anonymous Enrolments Export` as example.

PR #1406, opened 2019-05-23
#3056651 - Performance optimisation for counting group members

royharink:
## Problem
Counting group members originally had to load the user profiles of members to make sure the calculation matched the actual members within the group.

## Solution
Removed custom group membership calculation which isn't necessary any more as it has been fixed within groups. Also generated group operation urls based on routes instead of custom paths.

## Issue tracker
https://www.drupal.org/project/social/issues/3056651

## How to test
- [ ] As a logged in user go to the 'all groups' page and see that the membership count in the teasers for each group resembles the actual group count.
- [ ] Now go to the 'my group overview' page and see that the membership count in the teasers for each group resembles the actual group count.
- [ ] Go to a group from the list and see that the membership count in the hero region/header resembles the  actual group count.

## Release notes
<describe the release notes>

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -33,7 +33,6 @@ use Drupal\image\Entity\ImageStyle;
 use Drupal\Core\Block\BlockPluginInterface;
 use Drupal\user\Entity\Role;
 use Drupal\user\Entity\User;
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
Also added to 4.x :) !

PR #1408, opened 2019-05-24
Issue #3056832 by robertragas: set activity messages to queue instead…

robertragas:
## Problem
A lot of activity messages Open Social are using are send directly while they have no direct action such as creating a post on the activity stream.

## Solution
These message are only a notification and/or mail, so to increase the performance we could move these activities to the queue.

## Issue tracker
https://www.drupal.org/project/social/issues/3056832

## How to test
- [ ] Feature revert
- [ ] Test of one the message templates

Note: When sending an activity direct, it will create queue items for the emails right away. When using the queue for all, it will only create the activity creation in the queue, but when the cron is running it will create the emails also and send them in the same cronjob.

## Release notes
We have added specific message activities to be executed in the cronjob instead of directly. These message activities are related to activities where it would create a notification or email, so it will not delay any messages that are posted to the stream.
Using the queue for this increases performance.

These message templates are changed:

- create_comment_author_node_post
- create_comment_post_profile
- create_comment_reply
- create_content_in_joined_group
- join_to_group
- activity_on_events_im_organizing
- member_added_by_event_organiser
- create_like_node_or_post
- create_comment_reply_mention
- create_mention_comment
- create_mention_post


Kingdutch:
This PR should be merged together with #1410 as it may increase the risk of activities being processed after referenced content is deleted.

jaapjan:
Tests failing, who will continue work on this?

jaapjan:
Just restarted the test, let's see if they will be any better now.

Kingdutch:
Rebased on 8.x-7.x to fix merge conflicts.

jaapjan:
How can we move this PR forward? @robertragas

jaapjan:
Marking for 7.0 as this probably needs to be in major only.

Reason its on `needs work` is that the tests are still failing.

robertragas:
I tried running tests in the full suite locally, as in per test, but can't see why it's failing. 
But did run the tests locally with main branch merged so trying that one now and see how tests go.

What I did notice is because of the changes the demo content gets in the queue, so that's around 200 items in the queue and when the test starts it might time out on "I wait for the queue to be empty" although locally it goes quite fast.

Made a PR for that in https://github.com/goalgorilla/open_social_scripts/pull/29/files

jaapjan:
ok . merged and restarted the test.. like and post-stream test was failing I believe. Because notification bell did not have the notification yet. Did you check those tests locally as well? If not we should grep for all `notification` tests and run the queue before checking the 'bell'?

PR #1409, opened 2019-05-24
Issue #3056837 - Improve profile rendering

bramtenhove:
## Problem
Profiles are often rendered on the same page in the same way. However every time we load the profile entity again and render the same exact content as well.

## Solution
A new service is created that statically caches the loaded profile and the view mode render output.

## Issue tracker
https://www.drupal.org/project/social/issues/3056837

## How to test
- [ ] Browse through the site, the profiles should be rendered in the same way as before

## Release notes
<describe the release notes>

### DIFF
--- a/modules/social_features/social_profile/src/SocialProfileRenderService.php
+++ b/modules/social_features/social_profile/src/SocialProfileRenderService.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace Drupal\social_profile;
+
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\profile\Entity\ProfileInterface;
+use Drupal\user\UserInterface;
+
+/**
+ * Class SocialProfileRenderService.
+ *
+ * @package Drupal\social_profile
+ */
+class SocialProfileRenderService {
+
+  /**
+   * The Profile Entity Storage.
+   *
+   * @var \Drupal\Core\Entity\EntityStorageInterface
+   */
+  protected $profileStorage;
+
+  /**
+   * The Profile View Builder.
+   *
+   * @var \Drupal\Core\Entity\EntityViewBuilderInterface
+   */
+  protected $profileViewBuilder;
+
+  /**
+   * SocialProfileRenderService Constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The Entity Type Manager.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager) {
+    $this->profileStorage = $entity_type_manager->getStorage('profile');
+    $this->profileViewBuilder = $entity_type_manager->getViewBuilder('profile');
+  }
+
+  /**
+   * Renders the profile of a user
+   *
+   * @param UserInterface $account
+   *   The user account for which we render the profile.
+   * @param $view_mode
+   *   The view mode in which the profile gets rendered.
+   *
+   * @return array
+   *   A render array for the profile.
+   */
+  public function renderUserProfile(UserInterface $account, $view_mode) {
+    $profiles = &drupal_static(__FUNCTION__);
+
+    // Check if we already have rendered something previously.
+    if (!empty($profile[$account->id()]['view_mode'])) {
+      return $profile[$account->id()]['view_mode'];
+    }
+
+    // Try to load a cached users profile.
+    if ($profile[$account->id()]['profile'] instanceof ProfileInterface) {
+      $profile = $profile[$account->id()]['profile'];
+    }
+    else {
+      // Load the profile, we don't have anything cached at this moment..
+      $profile = $this->profileStorage->loadByUser($account, 'profile');
+
+      // Check if we have a valid profile.
+      if ($profile instanceof ProfileInterface) {
+        // Cache the loaded profile in the static variable.
+        $profile[$account->id()]['profile'] = $profile;
+      }
+      else {
+        // Return early, we have no valid profile to work with.
+        return [];
+      }
+    }
+
+    // Render the profile in the required view mode.
+    $profile_render = $this->profileViewBuilder->view($profile, $view_mode);
+
+    // Cache this in the static variable.
+    $profiles[$account->id()]['view_mode'] = $profile_render;
+
+    // Return the render array.
+    return $profile_render;
+  }
+
+}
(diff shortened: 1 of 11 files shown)


modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+use Drupal\profile\Entity\ProfileInterface;
+use Drupal\user\Entity\User;
+use Drupal\user\UserInterface;
+
+/**
+ * Class SocialProfileRenderService.
+ *
+ * @package Drupal\social_profile
+ */
+class SocialProfileRenderService {
```
Kingdutch: We should add and implement an interface for this service so that it can be used instead of the class. This allows projects building on Open Social to create alternative implementations if they need to.

See #1382 for an example of this. Specifically commit 538c6ca

modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager) {
+    $this->profileStorage = $entity_type_manager->getStorage('profile');
+    $this->profileViewBuilder = $entity_type_manager->getViewBuilder('profile');
+  }
+
+  /**
+   * Renders the profile of a user
+   *
+   * @param \Drupal\user\UserInterface|\Drupal\Core\Session\AccountInterface $account
```
Kingdutch: We should normalise this to a single interface type. 

Either choose `UserInterface` to let calling code handle the norrmalisation from an account to a user. User loading is cached within requests (similarly to what this service does so this shouldn't add additional overhead).

Alternatively choose `AccountInterface` and handle the loading of the user if `!($account instanceof UserInterface)` already.

modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+   *   The user account for which we render the profile.
+   * @param $view_mode
```
Kingdutch: View mode should always be a `string`

modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+   *   The view mode in which the profile gets rendered.
+   *
+   * @return array
+   *   A render array for the profile.
+   */
+  public function renderUserProfile($account, $view_mode) {
```
Kingdutch: This is a new service and method so lets add typehints (both parameter and return type) :D

modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+    // Build cache ID.
+    $cid = $account->id() . ':' . $view_mode;
+
+    // Check if we already have rendered something previously.
+    if (!empty($this->userProfileCache[$cid])) {
```
Kingdutch: Although I like guard clauses. Something can be said in this case to use Drupal's conventions for caching methods. This entails checking if the cache is empty, then filling it, and at the end always returning from the cache.

See `MediaRevisionAccessCheck::checkAccess` for an example.

Kingdutch: We should use `isset` here instead of `empty`. It could be that in an earlier call we've already determined that this would yield an invalid render and set the cache value to an empty array (`[]`). When using `empty` this would still result in recalculating the result.

modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+      return $this->userProfileCache[$cid];
+    }
+
+    // When we get an AccountInterface, we load the UserInterface.
+    if ($account instanceof AccountInterface) {
```
Kingdutch: This is always true. `UserInterface` implements `AccountInterface`. See my comment about choosing to implement a single interface in the function parameters and adjust/remove this accordingly.

modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+      $account = User::load($account->id());
+    }
+
+    // Load the profile, we don't have anything cached at this moment.
+    $profile = $this->profileStorage->loadByUser($account, 'profile');
+
+    // Check if we have a valid profile.
+    if (!empty($profile) && $profile instanceof ProfileInterface) {
```
Kingdutch: Why the extra `instanceof` here? We can simply check `$profile` against `NULL` to know whether we do or do not have a profile. This was fixed in https://www.drupal.org/project/profile/issues/3005113

modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+      // Cache the loaded profile in the static variable.
+      $user_profile = $profile;
+    }
+    else {
+      // Return early, we have no valid profile to work with.
+      return [];
```
Kingdutch: I think we should still cache this result.

modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+      $account = User::load($account->id());
+    }
+
+    // Load the profile, we don't have anything cached at this moment.
+    $profile = $this->profileStorage->loadByUser($account, 'profile');
+
+    // Check if we have a valid profile.
+    if (!empty($profile) && $profile instanceof ProfileInterface) {
+      // Cache the loaded profile in the static variable.
+      $user_profile = $profile;
```
Kingdutch: The comment here talks about a static variable but the `$user_profile` is actually not static. There's no benefit over this variable vs. `$profile` directly.

modules/social_features/social_profile/src/SocialProfileRenderService.php
```diff
@@ -0,0 +1,101 @@
...
+    else {
+      // Return early, we have no valid profile to work with.
+      return [];
+    }
+
+    // Render the profile in the required view mode.
+    $profile_render = $this->profileViewBuilder->view($user_profile, $view_mode);
+
+    // Cache this in the static variable.
+    $this->userProfileCache[$cid] = $profile_render;
```
Kingdutch: The intermediaire variable isn't needed here (also see my comment about line 69)

modules/social_features/social_profile/src/SocialProfileRenderService.php
Kingdutch: We may want to add a way to invalidate cache entries here. We never know if calling code knows better than we do that a profile has changed between renders (even if the case is unlikely).

jaapjan:
Is somebody going to implement the feedbback from @Kingdutch ?

royharink:
Hey @jaapjan,
I quickly checked with @bramtenhove and this PR still needs quite some work to have it finished.
While doing that we can take a look at the given feedback.

bramtenhove:
I removed the milestone, I think parts of this PR are done incorrectly. We need to revise quite a bit and the value is not that high.

ronaldtebrake:
I will close this one then. So we don't leave things open that don't have our attention, we can still create a new PR and grab info from this one.

PR #1410, opened 2019-05-24
#3066419 - Make the cron more stable by adding defensive checks

robertragas:
## Problem
On several occasions the cron fails when it's processing items related to activities.

The cause is in most cases that the referenced content in the queue item has since been deleted. The code assigned to processing the queue item doesn't properly check for this.

## Solution
Add checks to see if content was loaded before trying to perform any other actions on or with it.

## Issue tracker
https://www.drupal.org/project/social/issues/3066419

## How to test
- [ ] 

## Release notes
Checks have been added for processing queue items to make sure they don't fail on trying to load deleted referenced entities.


Kingdutch:
#1403 Was also opened to fix a related issue in another activity processor.

jaapjan:
@robertragas Could you add some more info to this PR, what is the prio? `low` correct?

robertragas:
I would define this as high, because without the code you cannot add more activities templates to the queue since they will break the cron otherwise.

jaapjan:
@robertragas ok could you please add more info this PR in terms of How to test etc? Issue number on drupal.org?

bramtenhove:
Cherry-picked to `8.x-5.x`

PR #1411, opened 2019-05-27
Issue #3065011 Lazy loading - Embedded media

ronaldtebrake:
## Problem
Uploading assets in text areas can cause pages to turn very slow. WIth social_embed also the ability to embed videos is enabled.

## Solution
This solution adds a new filter for all text formats.
It adds a hook to disable it on certain text formats (or add any new)
This filter will make sure all the videos get lazy loaded by using the Blazy library. Settings are provided through the lazy drupal.org module. 
We can also use this for Images in the future, a different story is created for that.

## Issue tracker
https://www.drupal.org/project/social/issues/3065011

## How to test
(Created demo environment here: 
https://feature-performance-wqvj2ny-dmio6k4roqn7k.eu.platform.sh/ )
See 
https://feature-performance-wqvj2ny-dmio6k4roqn7k.eu.platform.sh/stream
https://feature-performance-wqvj2ny-dmio6k4roqn7k.eu.platform.sh/node/195

- [ ] Enable social_lazy_loading and let it do it's magic
- [ ] Create or edit a node and add youtube videos with some lorem ipsum data
- [ ] Add youtube videos to the stream
- [ ] Add videos to the comments section

## Release notes
We have created a new module called social_lazy_loading, this module adds the ability to lazy load embedded media (iFrames in this case) that are added to all text filters. So Posts, content and comments. We will further extend this module in the future, but for now all assets should be lazy loaded when you enable this module!


modules/social_features/social_embed/src/SocialEmbedConfigOverride.php
```diff
@@ -84,8 +84,8 @@ protected function addFilterOverride($text_format, $convert_url, array &$overrid
     $config = $this->configFactory->getEditable($config_name);
     $filters = $config->get('filters');
 
-    $overrides = [];
-    $overrides[$config_name]['dependencies']['module'] = $config->get('dependencies.module');
+    $dependencies = $config->getOriginal('dependencies.module');
```
ronaldtebrake: This is necessary otherwise it will override all the embed configs and lazy won't work.

jaapjan:
Don't forget the .make file if possible :-)

modules/social_features/social_lazy_loading/src/SocialLazyLoadingTextFormatOverride.php
```diff
@@ -0,0 +1,128 @@
...
+  public function loadOverrides($names) {
+    $overrides = [];
+
+    $formats = [
+      'basic_html' => TRUE,
+      'full_html' => TRUE,
+      'plain_text' => TRUE,
+      'simple_text' => TRUE,
+      'restricted_html' => TRUE,
+    ];
```
jaapjan: Note that we are adding input formats this release (for e-mail and we might add comment) perhaps its better to just go through all the input formats programmatically

modules/social_features/social_lazy_loading/social_lazy_loading.info.yml
```diff
@@ -0,0 +1,7 @@
+name: 'Social Lazy Loading'
```
jaapjan: This is technically not a feature. Perhaps move to modules/custom instead?

composer.json
```diff
@@ -144,6 +148,7 @@
         "bower-asset/bootstrap": "v3.4.1",
         "bower-asset/select2": "~4.0.5",
         "bower-asset/autosize": "~4.0.2",
+        "bower-asset/blazy": "~1.8.2",
```
jaapjan: .make fie is missing. Which might be tricky for bower-assets.. :-)

modules/social_features/social_lazy_loading/social_lazy_loading.info.yml
ronaldtebrake: Good point! Done!

modules/social_features/social_lazy_loading/src/SocialLazyLoadingTextFormatOverride.php
ronaldtebrake: Thanks, thats an excellent point, i will see what I can do here!

ronaldtebrake:
Was green, just pushed the text format for mail_html from #1428 so when thats merged we already take that in to account :) 
Will merge it so we can release it in beta-1!

PR #1416, opened 2019-06-06
Issue #3059892: Added the static keyword to the statically called function

sjoerdvandervis:
## Problem
Deprecated function: Non-static method Drupal\social_profile_fields\SocialProfileFieldsBatch::performFlushFinishedCallback() should not be called statically in _batch_finished() (line 454 of core/includes/batch.inc)

## Solution
I added the static keyword to the callback function, because this function is always called statically.

## Issue tracker
https://www.drupal.org/project/social/issues/3059892

## How to test
- Go to `/admin/config/opensocial/profile-fields`;
- Deselect fields and save them;
- Flush the profiles. See that the deprecation warning is not shown in the logs.

## Release notes
Updated the static function call.


ronaldtebrake:
in 4.x and 5.x

PR #1417, opened 2019-06-07
#3060226 - The Data Policy module overrides the destination given in the auto login url

royharink:
## Problem
The Data Policy module overrides the destination given in the auto login url. This happens when a (new) user hasn't accepted the Data Policy and is automatically redirected to the Data Policy agreement form.

## Solution
Implement the newly created hook in [#3059927](https://github.com/goalgorilla/data_policy/pull/17) that makes it possible to change the destination in the scenarios as described in the problem.

## Issue tracker
https://www.drupal.org/project/social/issues/3060226

## How to test


jaapjan:
@royharink check coding standards and could you solve merge conflict as well:
https://travis-ci.org/goalgorilla/open_social/jobs/544357812

royharink:
@bramtenhove, PR has been updated with the coding standard fixes that @jaapjan mentioned.
Also the check on the optional data policy is now available.

modules/custom/social_magic_login/src/Service/MagicUrl.php
```diff
@@ -43,7 +43,7 @@ public function create(UserInterface $account, string $destination, array $optio
 
     // Get url options and prerequisites.
     $timestamp = \Drupal::time()->getRequestTime();
-    $lang_code = isset($options['langcode']) ? $options['langcode'] : $account->getPreferredLangcode();
+    $lang_code = $options['langcode'] ?? $account->getPreferredLangcode();
```
bramtenhove:
```suggestion
    $lang_code = isset($options['langcode']) ? $options['langcode'] : $account->getPreferredLangcode();
```

This is PHP 7.0 code that we cannot support yet in the `5.x` branch.
When we're using Drupal 8.7.0 we can start using PHP 7.0 code.

modules/custom/social_magic_login/src/Controller/MagicLoginController.php
```diff
@@ -70,7 +75,7 @@ public static function create(ContainerInterface $container) {
    *
    * @see \Drupal\user\Controller\UserController::resetPassLogin
    */
-  public function login($uid, $timestamp, $hash, $destination) {
+  public function login($uid, $timestamp, $hash, $destination): ?RedirectResponse {
```
bramtenhove:
```suggestion
  public function login($uid, $timestamp, $hash, $destination) {
```

We can't use PHP 7.0 code yet.

bramtenhove:
Does not need to be cherry-picked to `5.x`.

jaapjan:
@bramtenhove if it doesn't need to be cherry-picked to 5.6 could you change the mileestone in the future? I've now changed it to 6.0

bramtenhove:
Check, missed that one. Thanks!

PR #1418, opened 2019-06-11
Issue #3060876 by chmez: All Day Event Time

ghost:
## Problem
User from **not UTC** timezone can see a time in the date field on the event overview page when this event has turned on the **All day** option.

## Solution
Check only minutes value for detecting the **All day** mode.

## Issue tracker
- https://www.drupal.org/project/social/issues/3060876
- https://www.drupal.org/project/social/issues/3046037
- https://jira.goalgorilla.com/browse/YANG-1207

## How to test
- [ ] Log in and navigate to the **Settings** user page.
- [ ] Switch timezone from **UTC** to **New York**.
- [ ] Create a new event with checked the **All day** option.
- [ ] You should see **00:01** time on the event overview page.

## Release notes
Improve detecting the "All day" event mode for a user not from UTC timezone.


jaapjan:
I can merge this one, but I think the solution for all day events is not very sustainable this way. I propose we refactor this by adding a new checkbox via a new field or improving the field type.

jaapjan:
Cherry-picked to 5.x as well.

PR #1424, opened 2019-06-17
Issue #3061121 by Kalavan, albertoalaejos: Warning: in_array() expect…

albertoalaejos:
Warning: in_array() expects parameter 2 to be array, null given in Drupal\social_core\Entity\EntityAutocompleteMatcher->getMatches()

## Problem
When user uses autocomplete this two warning appearing in status report.

1) Warning: in_array() expects parameter 2 to be array, null given in Drupal\social_core\Entity\EntityAutocompleteMatcher->getMatches() (line 40 of /var/www/html/docroot/profiles/contrib/open_social/modules/social_features/social_core/src/Entity/EntityAutocompleteMatcher.php)

2) Notice: Undefined index: skip_entity in Drupal\social_core\Entity\EntityAutocompleteMatcher->getMatches()

## Solution
Defensive code in an if condition

## Issue tracker
https://www.drupal.org/project/social/issues/3061121

## How to test
- [x] Select a group where the user is an admin
- [x] Click on manage members
- [ ] Search so that it triggers autocomplete
- [ ] Check the Recent log messages


ronaldtebrake:
Cherry-picked back in to 5.x! :D

PR #1427, opened 2019-06-18
Issue #3065016 issue with node add/edit/delete pages and pathauto

ronaldtebrake:
## Problem
After enabling social_path_manager some issues arrised
- Redirect not enabled
- On pages with edit or add or delete in the URL the hero was gone

## Solution
Fixed legacy code + added dependency :) should be a quick fix.

## Issue tracker
- https://www.drupal.org/project/social/issues/3065016

## How to test
- [ ] Create a page with a custom alias that containt /add or /delete or /edit as part of the URL.
- [ ] See that the redirect module is also enabled with social_path_manager

## Release notes
On pages with add/edit/delete in the custom URL alias the hero was removed. This should now be fixed. Also we made sure the Redirect module is enabled by default.


jaapjan:
Marking as needs work, see coding standards:
https://travis-ci.org/goalgorilla/open_social/jobs/547683085

jaapjan:
Since we add a new dependency this should actually be 6.0

jaapjan:
When reviewing this one, I suggest to include https://github.com/goalgorilla/open_social/pull/1436 as well :-)

jaapjan:
I reviewed #1436 already and is working as expected. However I found a bug in this feature.

Reproduce by:

- [ ] Adding topic with title "Edit pages".
- [ ] On node view you can see two heros instead of only one.

robertragas:
Working like designed.

robertragas:
Saw Jaapie his comment too late and didn't test that scenario, so still working on it.

PR #1428, opened 2019-06-18
Issue #3062477 by chmez: Allow rich text formatting when sending bulk emails to group members or event enrollees

OH17L:
## Problem
As as CM I want some basic formatting for emails I send via the person bulk actions

## Solution
- it should be possible to add 
  - formatting (bold, italic, underlined)
  - hyperlinks 
- should work for custom emails send by GM, EM and CM+ via
  - Manage enrollments tab
  - Manage groupmemebers (also challenges and courses)
  - People overview

## Issue tracker
- https://www.drupal.org/project/social/issues/3062477
- https://jira.goalgorilla.com/browse/FOR-244

## How to test
- [x] Log in as **SM+**.
- [x] Open the **/admin/people** page and select one or more users.
- [x] Select the **Send email** item in the **Actions** filter and press **Apply to selected items**.
- [x] You should see CKEditor for the **Message** field.
- [x] You should see the following buttons on the CKEditor panel: **Bold**, **Italic**, **Underline**, **Link**, **Unlink**.
- [x] You should not see the **Text format** field.
- [x] Open **/all-groups** page and click on the title of any group with members.
- [x] Click **Manage members**.
- [x] Select one or more members.
- [x] Click **Actions** and select the **Email selected members**.
- [x] You should see CKEditor for the **Message** field.
- [x] You should see the following buttons on the CKEditor panel: **Bold**, **Italic**, **Underline**, **Link**, **Unlink**.
- [x] You should not see the **Text format** field.
- [x] Enable the **Social Event Organisers** (social_event_managers) module.
- [x] Open **/community-events** page and click on the title of any event with enrollments.
- [x] Click **Manage enrollments**.
- [x] Select one or more enrollments.
- [x] Click **Actions** and select the **Email selected enrollees**.
- [x] You should see CKEditor for the **Message** field.
- [x] You should see the following buttons on the CKEditor panel: **Bold**, **Italic**, **Underline**, **Link**, **Unlink**.
- [x] You should not see the **Text format** field.

## Release notes
Allow applying text formatting for a message of bulk operations of the following types: **users**, **group members**, **event enrollees**.


ribel:
Hi @OH17L,
On a functional level, all works excellent!

I am only not sure if we need to add one more text format to the distro.
I see that we have filter format `restricted_html`, but looks like it's not used anywhere.

Maybe we can just create editor config for it with the same buttons that you add to the Mail (bold, italic, underline, link) 

@jaapjan, @ronaldtebrake What do you think about that? Do you remember the purpose of the Restricted HTML text format and can we use it for emails? 
See: original commit: https://github.com/goalgorilla/open_social/blob/07813b6890ca06d10194297432d86b9cb1824ebe/modules/social_features/social_editor/config/install/filter.format.restricted_html.yml

ronaldtebrake:
Hi @ribel,

It's just the one that Core offers and I don't believe we have altered it from core itself that much (if any). 
I wouldn't mind using that one for mail as long as it stays close to what core does. The problem, if we go overboard with the amount of configuration options, is that if users already implemented `restricted_html` somewhere else in the Open Social Distro we will now override those options for them there as well. 

So for me adding a new one could outweigh any BC issues. But that's just my two cents. It seems like for now we are staying pretty close to `restricted_html`.

jaapjan:
There is a merge conflict I see, marking as `needs work`

denysyukm:
@OH17L, looks good, just fix merge conflicts

ribel:
@ronaldtebrake @jaapjan This one is ready for your final review and merge.

modules/social_features/social_core/social_core.install
```diff
@@ -892,3 +893,11 @@ function social_core_update_8501() {
   $config->save();
 
 }
+
+/**
+ * Allow using the "Mail HTML" text format by the logged user.
+ */
+function social_core_update_8601() {
+  // Set some default permissions.
+  _social_core_set_permissions();
```
ronaldtebrake: This could potentially enable more permissions if distro users wanted to disable them through the UI they will be enabled again. Next time just singularly add one permission that would be better :).

ronaldtebrake:
Hi guys, this seems to cause some regression.

Screen Shot 2019-07-14 at 16 16 06

Now on for example a group add page users have multiple filters they can choose (this is for a LU). 
Also SM etc have this and the biggest issue is the link for the "about text formats" doesn't really feel usefull. 

Guess this will be for 7.x :(

JohannesHartmann:
Hey guys,
 it would be a great feature if also group members would have the possibility to send attached files (pdf, zip) via CKeditor messages

modules/social_features/social_core/social_core.install
OH17L: Done.

xinyumaGG:
Hi guys, 

When testing on an environment with this pull request enabled, I found the text field for sending emails disabled. 
Because there are other pull requests also enabled in this environment, I can not be sure if it is caused by this pull request. Would you mind to take a look?
Screenshot 2019-09-03 at 10 25 26

ribel:
@OH17L could you take a look to this?

jaapjan:
I don't understand anymore what's going on with this format. It is being defined in social-editor. We are giving permissions in social_swiftmail_install hook and now we are changing the permission in social-core update hook. Shouldn't this be more consistent? Or at least change the permission in the install hook of the social_swiftmail?

jaapjan:
I changed the title since we don't allow (inline) images, right?

jaapjan:
@OH17L 

Most of it is working, but I'm requesting two changes:

- [ ] Make social_swiftmail.install so that the permissions are set correctly and same way as in social_core_update_8601.
- [ ] The mail html is now available for everyone and visible on all the fields as a selector, also when you create content or group. Please take a look in `social_core_filter_process_format` and add `mail_html` format there, so it is not accessible in those fields. (perhaps more for enterprise sites as well, not sure).

OH17L:
@jaapjan I already fixed two issues from your last comment. Please review it again.

jaapjan:
Thanks for the PR, works great!

jaapjan:
@OH17L FYI I found a critical issue caused by this and solved it like this: https://github.com/goalgorilla/open_social/commit/7e3f11dc3723e40a94c8f3b2127165ff91fd266a Please let me know if its not ok.

PR #1429, opened 2019-06-19
Issue by ribel: Fix failing Landing Page behat tests on enterprise sites

ribel:
## Problem
Two behat tests for Landing pages are failing on all enterprise sites on step: _And I fill in "URL alias"_ .

## Solution
Fix those behat tests in distro by adding step before: _And I uncheck "Generate automatic URL alias"_

## Issue tracker
None

## How to test
- [x] Behat test should pass

## Release notes
Improve Landing Page behat tests


ronaldtebrake:
Hi @ribel,

I believe the social_path_manager module is the culprit but that isn't enabled by default in the Distro therefor it's not an issue here. Maybe we can make it more generic by adding a custom step to see if the checkbox is enabled in that case we need to uncheck it otherwise we can just fill in the URL Alias.

Or am I missing something?

jaapjan:
Marking as `needs work` since test is failing. See comment by Ronald

ribel:
@jaapjan @ronaldtebrake I have made those improvements that were proposed.

jaapjan:
Cherry picked to 6.x as well :-)

PR #1431, opened 2019-06-19
Issue #2948979 - Improve search relevancy (node type, creation date and title boost)

jaapjan:
## Problem
Search relevancy sorting should be improved.

## Solution
Let's work within the limitation of database search and add the following:
* Fallback for creation date (descending = newest on top) after relevancy sort.
* Boost the title 21 times
* Add content type to the index to enable searching for `Pizza event` when searching for events about Pizza.

We also wanted to include Prioritising intact phrases over separate words (e.g. when searching for "Intro call" any exact matches with "Intro call" should prioritise over "Let's plan a call for an introduction.") However that is not possible it seems. We should start with this and in case we want to add this prioritisation later we should consider using Apache SOLR.

## Issue tracker
https://www.drupal.org/project/social/issues/2948979

## How to test
- [ ] Create multiple nodes and test the search.
- [ ] Make sure to test for `Introduction call` and `Intro call` and `Monthly welcome` (e.g. all the monthly welcome nodes in one node with title `welcome 2019` should now be lower than nodes with 
title `monthly welcome march 2019`) and `Pizza event`, etc.
- [ ] See if tests pass

## Release notes
Will follow

## Change Record
Maybe needed.

### DIFF
--- a/modules/social_features/social_search/config/install/search_api.index.social_all.yml
+++ b/modules/social_features/social_search/config/install/search_api.index.social_all.yml
@@ -2,8 +2,8 @@ langcode: en
 status: true
 dependencies:
   module:
-    - node
     - group
+    - node
     - profile
     - user
     - search_api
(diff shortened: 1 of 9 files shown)


ronaldtebrake:
Works like a charm, merging this for the 6.0 testing day, would be great to see some results on other platforms as well 👍

royharink:
I tested this on a pac fixture db and it works pretty well.
The search results are as expected from the explanation.

PR #1432, opened 2019-06-20
Issue #2983330 by kavbiswa, Eitisha: Rename "Swiftmailer settings" to…

Kingdutch:
… "E-mail settings"

<h3>Problem</h3>
Users don’t care what Swiftmail is or that we use it. If we ever decide that we no longer use the swiftmailer module because something else is now better, the name is no longer applicable.

<h3>Solution</h3>
Rename "Swiftmail Settings" to "E-mail settings".

Ideally we'd rename the module as well to decouple it from our usage of the swiftmailer module but that might break things (we could do it in a major release with a Change Record).

## Issue tracker
https://www.drupal.org/project/social/issues/2983330

## How to test
- [ ] Check the menu item in the admin menu

## Release notes
The name of the menu item for site builders that allows you to remove Open Social branding from notification e-mails is now called "E-mail settings". This was previously named after the module used to send the e-mails which may be inaccurate if that module is not used.

### DIFF
--- a/modules/social_features/social_swiftmail/social_swiftmail.links.menu.yml
+++ b/modules/social_features/social_swiftmail/social_swiftmail.links.menu.yml
@@ -1,5 +1,5 @@
 social_swiftmail.settings:
-  title: 'Swiftmail settings'
+  title: 'E-mail settings'
   description: 'Settings for Social Swiftmail.'
   route_name: social_swiftmail.settings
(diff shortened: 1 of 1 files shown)


robertragas:
During testing day I noticed this one and think we should then also change the description that still says "settings for swiftmailer"

PR #1434, opened 2019-06-25
Issue #3063996: add WYSIWYG to comments.

jaapjan:
## Problem
Some people want to add WYSIWYG to comments to add links or add some markup to their comment. This is not possible currently.

## Solution
Let's add a simple WYSIWYG to make it configurable.

This change includes:

- New format (comment_text)
- New editor (comment_text)
- Settings for Comments (including previous enabled settings and the option to enable WYSIWYG)
- Code which loads the correct editor when the WYSIWYG config is enabled and the user has the correct permission.
- Removing the social_comment.comment_settings.yml from `config/install` directory to prevent accidental feature overrides during feature revert.

## Issue tracker
https://www.drupal.org/project/social/issues/3063996

## How to test
- [ ] Check code changes.
- [ ] Test re-install and update from previous version.
- [ ] Disable and enable the WYSIWYG in the comment settings.

## Release notes
It is now possible for community members to use a WYSIWYG editor when authoring comments. A site managers can configure this setting on the Comment settings section in Open Social Settings.

## Change Record
might follow

### DIFF
--- a/modules/social_features/social_comment/config/install/editor.editor.comment_text.yml
+++ b/modules/social_features/social_comment/config/install/editor.editor.comment_text.yml
@@ -1,0 +1,34 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - filter.format.comment_text
+  module:
+    - ckeditor
+format: comment_text
+editor: ckeditor
+settings:
+  toolbar:
+    rows:
+      -
+        -
+          name: Formatting
+          items:
+            - Bold
+            - Italic
+            - Underline
+        -
+          name: Links
+          items:
+            - DrupalLink
+            - DrupalUnlink
+        -
+          name: Lists
+          items:
+            - BulletedList
+            - NumberedList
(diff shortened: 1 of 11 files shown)


ronaldtebrake:
@jaapjan 
1. As a LU / CM / SM after doing an update on this branch I don't get the WYSIWYG editor
2. What do you think, should we do something with the height here of that area? 
It does feel like it gives stream pages way toooo much extra space for the benefits users get from this.

jaapjan:
1 Yes i think you need to enable it first. Perhaps for new installs we can
just enable it but existing we shouldn’t imho. 🤔

2 Good point, but how to change the height? 🤔

Op wo 26 jun. 2019 om 17:30 schreef ronaldtebrake <notifications@github.com>

jaapjan:
As discussed this won't make it into 6.0 as the design implications are currently not acceptable

PR #1435, opened 2019-06-26
issue #2882629 by albertoalaejos sort comments in ascending order

albertoalaejos:
## Problem
Currently the newest comments are shown at the top. This is quite jarring because it's different from how almost all other social networks or (threaded) conversations on the web are shown.

An argument that was made for why this was the case was that this ensures people don't have to scroll to see the newest comments. I would like to make the following two counter arguments to this for moving to the conventional "oldest at the top".

1. If users are coming to the site for the new comments, they're probably navigating there through a notification, in which case they won't have to scroll by using an anchor.
2. If I'm browsing and encountering the content for the first time, the only way for me to figure out what the conversation is about and to join in is to realise the order is 'reversed' and scroll to the bottom.

## Solution
Create a config Override that loads the comments in ascending order, some logic to make sure that the two comments below the post are the latest ones. Use of order and weight attributes to place the comment form under the comments.

## Issue tracker
https://www.drupal.org/project/social/issues/2882629

## How to test
- [ ] Create post
- [ ] Make few comments on the post
- [ ] Check in post/{post_id} and home that the last comments are below
- [ ] The comment form (write a comment) should be under the comments
- [ ] The last two comments will be shown


modules/social_features/social_post/src/Plugin/Field/FieldFormatter/CommentPostActivityFormatter.php
```diff
@@ -0,0 +1,74 @@
...
+      ->addMetaData('base_table', 'comment')
+      ->addMetaData('entity', $entity)
+      ->addMetaData('field_name', $field_name);
+
+
+    if (!$this->currentUser->hasPermission('administer comments')) {
+      $query->condition('c.status', CommentInterface::PUBLISHED);
+    }
+
+    $query->orderBy('c.cid', 'DESC');
```
albertoalaejos: Needed to get the last 2

modules/social_features/social_post/src/Plugin/Field/FieldFormatter/CommentPostActivityFormatter.php
```diff
@@ -0,0 +1,74 @@
...
+    // Limit The number of results.
+    if ($comments_per_page) {
+      $query->range(0, $comments_per_page);
+    }
+
+    $cids = $query->execute()->fetchCol();
+
+    $comments = [];
+    if ($cids) {
+      krsort($cids);
```
albertoalaejos: Needed to make order ASC again (you want last 2 comments from database, but you want to show them ascending)

jaapjan:
As discussed there is still a bit of work to be done, mainly to theme the comments in the post in the stream. The margins are not correct anymore.

jaapjan:
Just fixed the margins. Let's wait for tests again 🤞

jaapjan:
Should only be in 6.x

PR #1438, opened 2019-07-02
Issue #2983330: rename description for e-mail settings (Remove Swiftmail reference)

jaapjan:
Related to PR: #1432 a small textual change.

I don't think release note changes or How to test is needed here. If automated tests pass let's merge.


jaapjan:
Should only be in 6.0. 5.x is not necessary.

PR #1439, opened 2019-07-04
#3063713 - bulk actions on all pages fails for event enrollments

robertragas:
## Problem
When selecting all entities of a view (event enrollments or group memberships) and performing a bulk action on them you end up on a 404 or 403 page.

## Solution
By adding an extra check in the custom batch callback function for views that are executed, rather than just checking for the final batch, we can make it work. We have to also add the custom batch callback function for this second batch job after the first pre-populating one is finished.

## Issue tracker
https://www.drupal.org/project/social/issues/3063713

## How to test
- [x] Create an event and enroll enough people so you have multiple pages, either by enrolling a lot or changing view pager to 1 item per page
- [x] Select all the members across all pages by using the VBO
- [x] Notice it ends up in a 404/403 page, but after this fix not anymore

## Release notes
We have created a fix for the views bulk operations module which is used to execute actions on multiple people throughout multiple pages such as exporting event enrollees. Normally when selecting all members across al pages you would end up on a 404/403 page when trying to export, but this has been solved.


bramtenhove:
Cherry-picked to 5.x

PR #1441, opened 2019-07-08
Issue #3019458 by royharink: Issue with Custom email header footer settings page

royharink:
## Problems
Multiple problems occurred with the custom e-mail settings.

1. The applied text format settings in the template configuration of the Social Swiftmail are not saved correctly.
2. Custom headers and/or footer were not always displayed when set.
3. The setting for removal of the Open Social branding in e-mails wasn't respected.
4. Permissions for the custom e-mail template settings aren't set.

## Solution
1. The text format settings are added to the form submit and configuration. The settings are now also added to the form and checked if they are set for a before applying the default when they aren't. The settings is also checked when building the e-mail markup.
2. Both the header and footer configuration settings are now checked in the _social_swiftmail.module_ if they are set and the header/footer are added to the render array when applicable.
3. Make sure the configuration check is a boolean like it's set in the schema.
4. Set the permissions for the configuration route and the Mail HTML path.

## Issue tracker
Bugs introduced with: https://www.drupal.org/project/social/issues/3019458

### No. 1
- [ ] Go to the Social Swiftmail settings, _/admin/config/opensocial/swiftmail_.
- [ ] Set the Text format to any other then the default (Mail HTML) and save the form.
- [ ] When returning to the configuration form, see that the Text format is set to the new setting.

### No. 2
- [ ] Using the settings from no. 1 with a custom header/footer.
- [ ] Test an email via the Swift email tester and see that the header/footer is applied.
- [ ] Interact with a users content and see that the header/footer is applied in the activity emails.

### No. 3
- [ ]  Go to the Social Swiftmail settings, _/admin/config/opensocial/swiftmail_ and set the option to remove the branding.
- [ ] Send some emails and see that the branding of Open Social is removed.

### No. 4
- [ ] Go to the Social Swiftmail settings, _/admin/config/opensocial/swiftmail_ and see that you can access it as a site manager.
- [ ] Also not that you can select the Mail HTML text format.

### DIFF
--- a/modules/custom/activity_send/modules/activity_send_email/activity_send_email.module
+++ b/modules/custom/activity_send/modules/activity_send_email/activity_send_email.module
@@ -88,7 +88,6 @@ function activity_send_email_theme() {
       'variables' => [
         'notification_count' => NULL,
         'notifications' => NULL,
-        'notification_footer' => NULL,
         'notification_settings' => NULL,
       ],
     ],
@@ -96,7 +95,6 @@ function activity_send_email_theme() {
       'template' => 'directmail',
       'variables' => [
         'notification' => NULL,
-        'notification_footer' => NULL,
         'notification_settings' => NULL,
       ],
     ],
--- a/modules/custom/activity_send/modules/activity_send_email/src/Plugin/EmailFrequency/Immediately.php
+++ b/modules/custom/activity_send/modules/activity_send_email/src/Plugin/EmailFrequency/Immediately.php
@@ -67,15 +67,6 @@ class Immediately extends EmailFrequencyBase {
       ],
       ['langcode' => $langcode]),
     ];
-
-    // Get Footer settings.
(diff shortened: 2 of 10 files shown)


ronaldtebrake:
Amazing, works pretty well :)!

Small PR: Update
1. Update CS with minor fix.
2. Added allowed formats, if we don't do that SM can decide to change the format to basic html, which allows for images to be added. Which is all nice. But that doesn't work very well yet, due to the message not rendering absolute URL's. So that might be a nice improvement for later. 
Not sure if that matches your AC criteria :) 

Screen Shot 2019-07-14 at 17 05 49

Let me know if you agree so we can merge it before 6.0 @robertragas @royharink!

ronaldtebrake:
@royharink @robertragas - Checked this with Bram and agreed this is a good solution for now. Thanks!

PR #1446, opened 2019-07-09
Issue #3066600 make sure EO can add enrollees to their event

ronaldtebrake:
## Problem
Event Organisers get an access denied on adding enrollees, also the Add Enrollee button is gone.

## Solution
I've created a custom access check that doens't look at Roles & Permissions alone but uses the Service that already was available that checks whether or not a user is able to add enrollees to an event.

## Issue tracker
https://www.drupal.org/project/social/issues/3066600

## How to test
- [x] Enable the social_event_managers module
- [x] Create a new event as chrishall
- [x] Add yourself and Robert Andrews as Organisers
- [x] See that after creation you see the manage enrollees tab
- [x] See that when you visit that tab you can now add enrollees to your event
- [x] See that when you login as Robert Andrews and you follow the same steps it works the same
- [x] See that when you login as SM/CM you are also able to add people to the event
- [x] See that when you login as a different LU you can't add people to the event

## Release notes
There was an issue where Event Organisers weren't able to add people to their events. This is now solved so next to Site Managers & Content Managers, Event Organisers can fully manage their events as well.

### DIFF
--- a/modules/social_features/social_event/modules/social_event_managers/social_event_managers.routing.yml
+++ b/modules/social_features/social_event/modules/social_event_managers/social_event_managers.routing.yml
@@ -12,7 +12,7 @@ social_event_managers.add_enrollees:
     _form: '\Drupal\social_event_managers\Form\SocialEventManagersAddEnrolleeForm'
     _title: 'Add enrollees'
   requirements:
-    _permission: 'manage everything enrollments'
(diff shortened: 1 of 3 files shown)


ronaldtebrake:
Thanks @albertoalaejos , I will merge it so we can put it in the 5.6 release now the tests are passing :)

PR #1448, opened 2019-07-09
#3063713 - bulk actions on all pages fails for Group Members

ronaldtebrake:
## Problem
1. When selecting all entities of a view (event enrollments or group memberships) and performing a bulk action on them you end up on a 404 or 403 page.

## Solution
By adding an extra check in the custom batch callback function for views that are executed, rather than just checking for the final batch, we can make it work. We have to also add the custom batch callback function for this second batch job after the first pre-populating one is finished.

## Issue tracker
https://www.drupal.org/project/social/issues/3063713
https://jira.goalgorilla.com/browse/TB-2351

## How to test
- [ ] Login as CM/SM and or Group Manager
- [ ] Perform group management actions (Change role / Send mail) on a group with at least 11 members
- [ ] Select all members by pressing the select members across all pages
- [ ] See that it works as expected right now :) 

## Release notes
We have created a fix for the views bulk operations module which is used to execute actions on multiple people throughout multiple pages such as sending an email to all group members. 
Normally when selecting all members across al pages you would end up on a 404/403 page when trying to export, but this has been solved.


Kingdutch:
The code looks good although I don't completely understand what everything does.

I've tested this as admin. When I try to send an e-mail to all group members then I end-up on `/views-bulk-operations/execute/group_manage_members/page_group_manage_members` with an Access Denied page after the batch is executed.

It appears no e-mails were actually sent.

ronaldtebrake:
Pair reviewed this with @robertragas, will merge it. Thanks!

PR #1451, opened 2019-07-09
Issue #3066730 by robertragas, kingdutch: Hide language switcher when…

robertragas:
## Problem
When you would use a module to unset certain languages, the language switcher still shows up. To make this language switcher more flexible we should hide it when only 1 language is available.

## Solution
Create an access check to disallow access to the language switcher if only 1 language has been found.

## Issue tracker
https://www.drupal.org/node/3066730

## How to test
- [ ] Enable multiple language
- [ ] unset a language with a hook like the access language has to disable language for specific roles.
- [ ] Notice that the language switcher is gone now where it would normally show

`/**
 * Implements hook_language_switch_links_alter().
 */
function language_access_language_switch_links_alter(array &$links, $type, Url $url) {
  $languages = \Drupal::languageManager()->getLanguages();
  foreach ($languages as $language) {
    if (!\Drupal::currentUser()->hasPermission('access language ' . $language->getId())) {
      if (isset($links[$language->getId()])) {
        unset($links[$language->getId()]);
      }
    }
  }
}`

## Release notes
When unsetting a language because you don't want a user to have access to it, it still show in the language switcher. We have added an access check to hide this switcher when only 1 language is active for the specific role.

### DIFF
--- a/modules/custom/social_language/src/Plugin/Block/LanguageSwitcherBlock.php
+++ b/modules/custom/social_language/src/Plugin/Block/LanguageSwitcherBlock.php
@@ -2,6 +2,8 @@
 
 namespace Drupal\social_language\Plugin\Block;
 
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Session\AccountInterface;
 use Drupal\language\Plugin\Block\LanguageBlock;
 use Drupal\Core\Url;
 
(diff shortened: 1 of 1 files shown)


modules/custom/social_language/src/Plugin/Block/LanguageSwitcherBlock.php
```diff
@@ -22,6 +24,27 @@
  */
 class LanguageSwitcherBlock extends LanguageBlock {
 
+  /**
```
Kingdutch: We may want to move these new methods below the `build` method. I think the `build` method will be most interesting to people visiting this file.

robertragas: Shouldn't the blockAccess always be above the build?

royharink:
Fixed warning:
`Warning: count(): Parameter must be an array or an object that implements Countable in Drupal\social_language\Plugin\Block\LanguageSwitcherBlock->blockAccess() (line 32 of profiles/contrib/social/modules/custom/social_language/src/Plugin/Block/LanguageSwitcherBlock.php).`

jaapjan:
What is the status of this pull request?

Kingdutch:
Roy's bugfix is actually a workaround for https://www.drupal.org/project/drupal/issues/2940121

I think we should postpone this issue until that is fixed so we can create a coherent implementation.

Kingdutch:
This PR is no longer needed when #1626 lands because it deprecates the LanguageSwitcherBlock of Open Social and allows the Drupal core language switcher to work in 5ab6578946d8f6183df910b915d7cb4ab4712134

That switcher should already properly hide if only one language is available (otherwise it would be a Drupal core bug)

PR #1456, opened 2019-07-15
Issue #3065169 by agami4: Header menu hidden when PhotoSwipe is active/open

agami4:
## Problem
After click open an image in comment attachments, the lightbox is shown behind the header menu as shown in the image above. This is inconsistent with how post image is shown and it might cover some action options at the top, such as the close button.

## Solution
The comment image is shown above the header menu.

## Issue tracker
https://www.drupal.org/project/social/issues/3065169

## How to test
- [x] Attach image on comment form on stream or detail pages
- [x] Post comment
- [ ] Click on image

## Release notes
When you open image via PhotoSwipe, you can see header menu under comment image


Kingdutch:
It looks like this same issue can possibly be fixed by removing the `z-index: 0` from `.main-container` in `layout.scss`.

That would allow the browser to do its natural layering and allows the `z-index: 1050` of pwsop to properly function.

agami4:
@Kingdutch,
I updated this PR.

jaapjan:
@agami4 cc @Kingdutch 

Marking as needs work as it has conflicting file. Also it needs a rebase to 7.x (as do most of your PRs, so might be worth to plan some time to do this)

Kingdutch:
Hey @agami4,

Merging in 8.x-7.x is not the same as rebasing.

We now run the risk that changes that have been made only on 6.x will end up in 7.x. Alternatively it could be that something was changed later on 6.x and not on 7.x that could undo work done on 7.x.

This will need to be cleaned up and force pushed in this PR.

jaapjan:
It works, but indeed moving to here.. https://github.com/goalgorilla/open_social/pull/1496 Will merge that one after Travis tests are done

PR #1458, opened 2019-07-16
DS-6666 As a LU I can filter events by date and users by registration date

ronaldtebrake:
## Problem
1. LU can filter events within specific date ranges of event: before / after / between.
2. LU can filter users by registraiton date on the search users page

## Solution
Created a new plugin for Search API that shows before / after between, we can reuse this for other date based filters in the future. And added these for the above problems.

## Issue tracker
https://jira.goalgorilla.com/browse/DS-6666

## How to test
- [x] Install on this branch
- [x] Go to /search/content
- [x] See that nothing changed untill you click Event
- [ ] See that you can now select an Event Date based on Before / After / Between dates
- [ ] And it works ;)

- [ ] Go to /search/users
- [ ] See that you can now select a Registration Date based on Before / After / Between dates
- [ ] And it works ;)

## Release notes
We have added the possibility to filter the Search content tab based on the Event Date and the Registration date on the Search users tab!


ronaldtebrake:
Oops was merged in 6.x cherry picking to 7.x did work :) so its in both 7 and 6

PR #1461, opened 2019-07-23
Issue #3019458 by royharink: Skip group type selection page when creating a group for users that who can only add one group type

royharink:
## Motivation
With the introduction of the flexible group type it makes sense to disable the creation of other group types for certain user roles.
When disabling all but one group type, it's not user friendly to be sent to a page where there is just one option.

## Solution
When a user can only add one type of group, we adjust the route of the menu item "New group" from group add page to the group add group type page.

## Issue tracker
https://www.drupal.org/project/social/issues/3069607

## How to test
- [x] As an admin disable all but one group types to create.
- [x] Log in as a normal user and see in your create content menu that the "Add group" route is going to the single available group type.

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -12,6 +12,7 @@ use Drupal\Core\Ajax\AjaxResponse;
 use Drupal\Core\Ajax\HtmlCommand;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Link;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\group\Entity\Group;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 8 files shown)


modules/social_features/social_group/social_group.module
```diff
@@ -1699,8 +1699,43 @@ function social_group_user_cancel_methods_alter(&$methods) {
  * Implements hook_social_user_account_header_create_links().
  *
  * Adds the "Create Group" link to the content creation menu.
+ * When the user can only create groups of one type, the user skips the
+ * group type selection page.
  */
 function social_group_social_user_account_header_create_links($context) {
+  // We can't create a "Add Groups" link without a user to determine permission.
+  if (empty($context['user'])) {
```
Kingdutch: This is not entirely true because the previous method already worked without a user. In this case we could just default to the normal group type choice page.

bramtenhove: Agreed.

bramtenhove:
Cherry-picked to `8.x-6.x`!

PR #1463, opened 2019-07-26
Issue #3067525 search autocomplete separator comma

ronaldtebrake:
## Problem
If you have more than 1000 users
and if you search for a user who has nid 1006
the search result displays /user/1,006

## Solution
Make sure there is no separator. 

## Issue tracker
https://www.drupal.org/project/social/issues/3067525
## How to test
- [ ] Test the results for search autocomplete for a user group or node that has an id higher than 1000 and see there is no comma anymore

## Release notes
If you have more than 1000 users, groups or pieces of content the search autocomplete has a comma in the URL which causes a wrong link to the piece of content.


Kingdutch:
Hey @ronaldtebrake,

For new post update hooks, please keep in mind to add numbering like we do for normal update hooks. This makes sure that the order is preserved (order of execution is alphabetic).

Perhaps we can still change this if it's not yet released? : )

PR #1464, opened 2019-07-29
Issue #3071035 by Kingdutch: Geolocation Maps Markers use incorrect c…

Kingdutch:
…olor for title text

<h2>Problem</h2>

The pop-ups of the markers use the navigation link text which can conflict with the card background color that is used. Instead the default link color should be used because it's meant to be used with the cards.

<h2>Solution</h2>
Change the title color to use the default link-color in brand.scss

## Issue tracker
https://www.drupal.org/project/social/issues/3071035

## How to test
- [ ] Enable the Social Geolocation Maps module
- [ ] Create an event with an address
- [ ] Visit /community-evens
- [ ] See that the title of the pop-up of the marker is now visible

## Release notes
The color used for the title on location markers on the event map was sometimes the same as the card background color which caused links to be invisible. This color now utilises the default link color which should be configured to work correctly in multiple places.

### DIFF
--- a/themes/socialblue/assets/css/badge.css
+++ b/themes/socialblue/assets/css/badge.css
@@ -14,7 +14,6 @@
 .badge-default {
   background-color: #e6e6e6;
   color: #4d4d4d;
-  -webkit-transition: 0.3s;
   transition: 0.3s;
 }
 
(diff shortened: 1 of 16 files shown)


ronaldtebrake:
Cherry-picked :)

Kingdutch:
This was closed instead of merged 😅

PR #1470, opened 2019-08-02
#3065010 - Lazy load images

ronaldtebrake:
## Problem
Images that aren't necessarily shown to a user are loaded on a full page load. We can also load them after the page is shown. Here comes the lazy loading!

## Solution
We use the lazy loading module to override our image fields and add the lazy loading capability to it.

## Issue tracker
- [ ] https://www.drupal.org/node/3065010

## How to test
- [x] Enable social_lazy_loading and see that for the following it should work:
- [x] Events, pages, books, topics, user & profiles, groups.
- [x] Check all overviews
- [x] Check streams
- [x] Check search

For all of the above the images should be lazy loaded with exception of the images in your viewport or those that are background urls (like the hero).

## Release notes
We have a new addition to our lazy loading module, images that are provided by Open Social can be lazy loaded now by enabling the social_lazy_loading_images module.

### DIFF
--- a/modules/custom/social_lazy_loading/README.txt
+++ b/modules/custom/social_lazy_loading/README.txt
@@ -1,0 +1,5 @@
+We use the lazy module for:
+- Inline images & iFrames in the WYSIWYG editor. This works by adding processing functions to a text format.
+
+We use the Blazy module for:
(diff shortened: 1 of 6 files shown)


ronaldtebrake:
Cherry picked to 8.x-6.x

PR #1474, opened 2019-08-06
Issue #3072888: Sorting on search content

jaapjan:
## Problem
Currently it is not possible to sort the search results. Especially on the content tab this would be very valuable and would help with finding what you are looking for.

Proposal
Add sorting possibility on the search pages.

- created date
- title (alphabetically)
- relevancy
- nice to have: updated date
- It's implemented on all the tabs (preferably: but content and users is highest prio).

Default sort: relevancy.

## Issue tracker
https://www.drupal.org/project/social/issues/3072888

## How to test
- [ ] Run update hooks and feature reverts.
- [ ] Go to the content search and notice the new sorting option is there on the top.
- [ ] Design is reflected in the implementation
- [ ] Note we renamed Filter to "Apply" to make clearer that the submit is for both filtering and sorting.
- [ ] Make sure it works as expected on all the devices.

## Release notes
It is now possible to sort search results on the "Search content" page. This will help users find what they are looking!


Kingdutch:
@jaapjan I've completed the styling (see Slack pictures). I think if you make the changes I requested then this is ready to be committed? :)

jaapjan:
Great! Thanks again @Kingdutch 

Will only be in 7.0 so no backport needed.

PR #1479, opened 2019-08-13
Feature/3056821 ajaxify notification centre

navneet0693:
## Problem
The notification centre loads a view on every page, while it's not immediately visible to the user. Therefor the page load performance drops significantly.

## Solution
1. Ajaxified the notification view to render via controller callback.
2. Added a **load_link** element in `social_activity_social_user_account_header_items`
3. Made changes in notification.js to click the link to trigger ajax callback and show the notifications
4. Added cache max-age 0 to prevent from caching.

## Issue tracker
https://www.drupal.org/project/social/issues/3056821

## How to test
- Login via two users
- Try to create notifications for any of the user.
- Click the notification bell icon
- It should load via notifications via ajax
- Hide/show should work
- Notifications should be mark as seen
- Subsequent page load should not load the render html from activity notification view until notification bell icon is clicked (Check in developer tools for present of view's html)

## Release notes
Deprecated `social/modules/custom/activity_creator/src/Controller/NotificationsController.php::readNotificationCallback()`


modules/custom/activity_creator/js/notifications.js
```diff
@@ -15,27 +15,18 @@
     },
 
     _updateNotificationCount: function () {
-      var $notificationCount = $('.notification-bell .badge');
 
       // We won't proceed if the notification count is 0 or the dropdown is
       // already open.
       // @todo: Refactor to use data-* attributes and `jQuery.data()` to store
```
Kingdutch: We can remove this TODO because it's no longer applicable and should probably update the comment on the live above :)

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -116,10 +120,12 @@ public function markEntityAsRead(AccountInterface $account, EntityBase $entity)
    * @param \Drupal\activity_creator\Entity\Activity $activity
    *   Activity object.
    * @param int $status
-   *   See: activity_creator_field_activity_status_allowed_values().
+   *   See: activity_creator_field_activity_status_allowed_values()
    *
-   * @return \Drupal\activity_creator\Entity\Activity
+   * @return int
```
Kingdutch: Is this correct? Even if it's not the Activity object I would expect a `bool` instead of `int`.

modules/custom/activity_creator/src/Controller/NotificationsController.php
```diff
@@ -2,34 +2,98 @@
...
 
+  /**
+   * Ajax callback to mark notifications as read.
+   *
+   * @deprecated in social:8.x-7.0 and is removed from social:8.x-8.0. Use NotificationsController::getNotificationListCallback()::getNotificationListCallback() instead.
+   * @see https://www.drupal.org/project/social/issues/3056821
+   */
+  public function readNotificationCallback(): AjaxResponse {
     // Create AJAX Response object.
     // @todo: Implement a Ajax command instead and call via addCommand().
```
Kingdutch: This Todo has actually been implemented :D so we can probably remove it.

modules/custom/activity_creator/src/Controller/NotificationsController.php
```diff
@@ -2,34 +2,98 @@
...
+    return new static(
+      $container->get('activity_creator.activity_notifications'),
+      $container->get('current_user')
+    );
+  }
 
+  /**
+   * Ajax callback to mark notifications as read.
+   *
+   * @deprecated in social:8.x-7.0 and is removed from social:8.x-8.0. Use NotificationsController::getNotificationListCallback()::getNotificationListCallback() instead.
```
Kingdutch: @jaapjan @ronaldtebrake Do you agree here? : )

modules/social_features/social_activity/social_activity.module
```diff
@@ -62,7 +43,30 @@ function social_activity_social_user_account_header_items(array $context) {
         '#wrapper_attributes' => ['class' => 'divider'],
         '#markup' => '',
       ],
-      'notifications' => $notifications_view,
+      // Element to contain the link for ajaxified view for notifications.
+      'load_link' => [
```
Kingdutch: We should create a follow-up to change the throbber here to our spinner that's currently the contents of `notifications`. After this we can clear the markup of the `notifications` element. 

This can be a drupal.org issue for Open Social : )

navneet0693: Created: https://www.drupal.org/project/social/issues/3074740

modules/custom/activity_creator/src/ActivityNotifications.php
navneet0693: Updated.

modules/custom/activity_creator/js/notifications.js
navneet0693: Updated.

albertoalaejos:
@Kingdutch I have tested it, and it works, except for the last point (view for notification), that I don't know the exact view to expect. Can you tell me the html item that I should see?

Kingdutch:
I'm not sure @albertoalaejos, I'm going to redirect your question to @navneet0693 who wrote the testing steps : )

modules/custom/activity_creator/src/Controller/NotificationsController.php
ronaldtebrake: Yes!

navneet0693:
@albertoalaejos Sorry for my typo, I have updated the comment. Earlier the render html of activity notification view are loaded on every page, but after ajaxification the render html from view should only be present in the DOM after you click the notification bell.

Kingdutch:
@navneet0693 The `load_link` on line 47 does not have a `#title` attribute (see https://github.com/goalgorilla/open_social/pull/1479/files#diff-00bd33aa5b37f56da46702f29b225d52R47 ). Can you add this (an empty string `''` as value should be fine). That should resolve the warning that @albertoalaejos was seeing :)

navneet0693:
@albertoalaejos @Kingdutch  Just pushed another commit adding the #title property. It should resolve the warning now.

PR #1481, opened 2019-08-18
Issue #3075687 by mdeny: Social registration fields module

denysyukm:
## Problem
As a SM I want to see profile fields in the signup form

## Solution
Implement module which adds the possibility to display profile fields on the sign up form

## Issue tracker
https://jira.goalgorilla.com/browse/FOR-256
https://www.drupal.org/project/social/issues/3075687

## How to test
- [ ] Enable social_registration_fields module
- [ ] Go to /admin/config/people/social-registration-fields
- [ ] Set First name and Last name fields Show on sign-up form. Set First name field as a mandatory. Save the form
- [ ] Go to the sign up form
- [ ] You should see First name and Last name fields with First name field as a mandatory

## Release notes
Adds the possibility to display profile fields on the sign up form.

### DIFF
--- a/modules/social_features/social_profile/modules/social_registration_fields/social_registration_fields.features.yml
+++ b/modules/social_features/social_profile/modules/social_registration_fields/social_registration_fields.features.yml
@@ -1,0 +1,2 @@
+bundle: social
+required: true
--- a/modules/social_features/social_profile/modules/social_registration_fields/social_registration_fields.info.yml
+++ b/modules/social_features/social_profile/modules/social_registration_fields/social_registration_fields.info.yml
@@ -1,0 +1,9 @@
+name: 'Social registration fields'
+description: 'Provides possibility to add profile fields to registration form.'
+type: module
+core: 8.x
+configure: social_registration_fields.settings
+dependencies:
+  - social:social_profile
+  - drupal:user
+package: Social
--- a/modules/social_features/social_profile/modules/social_registration_fields/social_registration_fields.links.menu.yml
+++ b/modules/social_features/social_profile/modules/social_registration_fields/social_registration_fields.links.menu.yml
@@ -1,0 +1,5 @@
+social_registration_fields.settings:
+ title: 'Registration fields settings'
+ description: 'Registration fields settings for Profile module.'
+ route_name: social_registration_fields.settings
+ parent: user.admin_index
--- a/modules/social_features/social_profile/modules/social_registration_fields/social_registration_fields.module
+++ b/modules/social_features/social_profile/modules/social_registration_fields/social_registration_fields.module
@@ -1,0 +1,136 @@
+<?php
+
+/**
+ * @file
+ * The Social registration fields module file.
+ */
+
+use Drupal\Core\Entity\Entity\EntityFormDisplay;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Render\Element;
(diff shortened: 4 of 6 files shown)


xinyumaGG:
Hi @denysyukm and @ribel 
When testing this pull request, I found 3 issues:
1. The link to /admin/config/people/social-registration-fields is only reachable for the admin role, not for site manager role.
2. After adding a few fields to the sign up form, then save, it works. But when I want to enable more fields, it does not work anymore. 
3. For the address field, we might need to adjust the display of fields a bit. Please find the design proposal attached. 
now
should be

bramtenhove:
Making the Address look nicer as @xinyumaGG proposed sounds like a good idea. I saw a few of the forms with the Address field and it looks confusing.

ronaldtebrake:
Hi @denysyukm and @ribel,

I would like to ask for one more change next to the ones above.
After setting the field to be mandatory you can still change it on your profile afterwards.
Imagine setting first and last name as mandatory fields, registering and then getting your one time login link, filling in a password, you get redirect to your edit profile and you can remove everything again because there it's optional fields. 

The mandatory setting isn't really persistent in that way.

Would you be able to fix this as well today?

ribel:
@ronaldtebrake So would you recommend to make fields required also in edit profile form if they configured like these for registration?
If yes - I think we can fix this today.

ronaldtebrake:
> @ronaldtebrake So would you recommend to make fields required also in edit profile form if they configured like these for registration?
> If yes - I think we can fix this today.

Yes that exactly what I'm thinking about! Otherwise users can still register without having these fields filled in. Thanks.

denysyukm:
@ronaldtebrake, @ribel, @xinyumaGG 
I made fields required also in the edit profile form. Also, I moved some new code from bitbucket and made reachable config form for the site manager role.

About issue that Xinyu described: "After adding a few fields to the sign up form, then save, it works. But when I want to enable more fields, it does not work anymore." I can say that it happens because of caching signup form for anonymous. You need just clear cache. So I discussed this with Taras a bit and we propose create followups to fix this + to adjust the display of address fields.

bramtenhove:
With the latest changes affecting the edit profile form as well, it might make sense to also rename the module 🤔

ronaldtebrake:
Hi there,

Are the todo's mentioned in your sprint? It feels like were now adding a feature that needs improvements right away. 

Also instead of adding different `hook_FORM_ID_form_alters` it would be easier if we could just use or alter a form view/display mode for this instead of using separate form alters.
 
I will add this to 7.1 so we have some more time for the reviewing and fixing.

ribel:
@ronaldtebrake 7.1 sounds fine for us. 
Then I think we can also add those address and caching improvements.

And regarding module name - to what would you like to change, @bramtenhove?
Maybe `social_profile_required_fields` ?

ribel:
@bramtenhove, @navneet0693,
could you also plan some work on fixing those TODOs in your team, since they were introduced in this commit:

https://bitbucket.org/goalgorilla/social_registration_fields/commits/69b158a5606bae5b3e1991f2a050624b831f0e25

ronaldtebrake:
This still needs work, removed the needs maintainer input label. I think all our input is there? (or let me know what you want to know from us)

ronaldtebrake:
After discussing during roadmap meeting this will be an extension.

Kingdutch:

PR #1484, opened 2019-08-19
Issue #3075854 by jaapjan: defensive coding in the activity token handling

jaapjan:
## Problem
Error in notification window when group or groupcontent is missing (still to be determined what exact root cause is).

## Solution
Let's make the token handling a bit more defensive.

## Issue tracker
https://www.drupal.org/project/social/issues/3075854

## How to test
- [ ] Open a site with this error.
- [ ] Apply the patch and verify that the error is solved.
- [ ] Dive some more into the root issue. Is this good enough? Why is this happening? Do we need to fix the root cause as well?

## Release notes
If there is a notification in the notification bar for a group or group content item which does not exist anymore the token replacement code should not crash. The code is now more defensive and thus will prevent the site from breaking on the menu bar resulting in missing content and blocks.

## Change Record
not needed

### DIFF
--- a/modules/social_features/social_activity/social_activity.tokens.inc
+++ b/modules/social_features/social_activity/social_activity.tokens.inc
@@ -49,57 +49,59 @@ function social_activity_tokens($type, $tokens, array $data, array $options, Bub
             $entity = \Drupal::entityTypeManager()->getStorage($target_type)
               ->load($target_id);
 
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
Restarted travis :)

ronaldtebrake:
Cherry-picked :)

PR #1487, opened 2019-08-23
issue #3069371 by albertoalaejos: The mention in a email notification…

albertoalaejos:
## Problem
The new email notifications show a piece of the post/comment/reply. When mentioning someone in the post, it is shown as a number (e.g. [~45]) in the email notification.

## Solution
Replace the number (which is the profile id) with its related username before the email is sent.

## Issue tracker
https://www.drupal.org/project/social/issues/3069371

## How to test
- [ ] Login as LU
- [ ] Post in a group including mentions in the post
- [ ] Check that email notifications are sent (social_mailcatcher)
- [ ] Make sure that the mention does not show as a number anymore

### DIFF
--- a/modules/social_features/social_activity/social_activity.tokens.inc
+++ b/modules/social_features/social_activity/social_activity.tokens.inc
@@ -69,12 +69,12 @@ function social_activity_tokens($type, $tokens, array $data, array $options, Bub
                   ['absolute' => TRUE]
                 )->toString();
 
-                $string_parts = preg_split("/(\[\~\d+\])/", $summary, -1, PREG_SPLIT_DELIM_CAPTURE  | PREG_SPLIT_NO_EMPTY);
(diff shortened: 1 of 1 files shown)


modules/social_features/social_activity/social_activity.tokens.inc
```diff
@@ -67,6 +69,28 @@ function social_activity_tokens($type, $tokens, array $data, array $options, Bub
                   ['absolute' => TRUE]
                 )->toString();
 
+                $string_parts = preg_split("/(\[\~\d+\])/", $summary, -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);
```
jaapjan: @albertoalaejos cc @navneet0693 

I don't think this is the proper way of dealing with this issue. most likely the mentions module will replace this (with input formats). So I don't think tokens is used for that.

jaapjan:
As discussed per Slack this needs a different solution. Let's leverage the Input formats instead.

PR #1488, opened 2019-08-27
Added drupal_static function to optimise permission checks in social_comment_comment_links_alter.

navneet0693:
## Problem
When the entity isn't a post comment "extra checks" take place. These extra checks are potentially going to load all the Group Content of the entity which can be a lot, then it's going to loop on all that loaded group content to check permissions. 

## Solution
We have used drupal_static function to optimise code.

## Issue tracker
https://www.drupal.org/project/social/issues/3077355

## How to test
User should not be able to post comments if doesn't have access to do that.

## Release notes
N.A

## Change Record
N.A


bramtenhove:
@navneet0693, some coding standard issues were found here: https://travis-ci.org/goalgorilla/open_social/jobs/577244925#L1510.
Could you take a look at that?

Kingdutch:
Hey @navneet0693,

I think we need to go a step back here. The optimisation that you made doesn't actually perform the alter anymore. I think some more refactoring here will be needed than wrapping the entire thing in a Drupal static function.

I think we may need to take a step back to figure out what in this function is actually expensive (whether it's the permission checks at all) and what the impact is of changing this on certain scenario's. The base links array that we get could be changed each time so we can't simply return a cached version of our own links array, we will have to act upon the input we get.

Is the `Problem` description in the Drupal issue the one originally written down by Roy and myself? Looking at this again it could be that we can't really optimise this any further other than adding `break;` after line 350.

navneet0693:
Thanks @Kingdutch and @bramtenhove for review. 

@bramtenhove I have fixed the coding standard issue.

@Kingdutch Reverted the use of drupal_static function and added a break statement to stop altering at first encounter of permission conflict for a group.

jaapjan:
Make sure to set the story to fixed after merging @albertoalaejos 
I've done it for now
https://www.drupal.org/project/social/issues/3077355

PR #1497, opened 2019-09-10
Issue #3080318 by chmez: Lazy Loading Image module installation

ghost:
## Problem
When enabling the **Social Lazy Loading** module then the **Social Lazy Loading Image** module is still disabled.

## Solution
Fix function name of the hook_install of the **Social Lazy Loading** module.

## Issue tracker
- https://www.drupal.org/project/social/issues/3080318
- https://getopensocial.atlassian.net/browse/YANG-1283

## How to test
- [x] Log in as administrator and navigate to **Extend** page.
- [x] When you enable the **Social Lazy Loading** module then you should see that the **Social Lazy Loading Image** module is still disabled.

## Release notes
Fix installation of the **Social Lazy Loading Image** module when installing the **Social Lazy Loading** module.


jaapjan:
Just wondering, why do we install this in a install hook? It can also be a dependency right?

OH17L:
@ronaldtebrake ^^^

ronaldtebrake:
> Just wondering, why do we install this in a install hook? It can also be a dependency right?

So you can easily disable the modules as well.

jaapjan:
Cherry-picked to 6.x

PR #1498, opened 2019-09-11
VBO - Update to 3.2

ronaldtebrake:
## Problem
Update VBO to 3.x and make sure our plugins work with the new major version. This should hopefully solve a lot of the issues we have been encountering with selections across multiple pages and performing operations on them.

## Solution
Update from 2.x to 3.2.

## Issue tracker
TBD

## How to test
- [ ] TBD

## Release notes
We have updated VBO too 3.x
- This means we also have removed the select all button and made it a select box instead


ronaldtebrake:
Jeej tests are green after only 4 retries on private message 🚀

PR #1500, opened 2019-09-16
Issue #3081555 by jaapjan: add BubbleableMetadata.

jaapjan:
## Problem
When using the JSON API you could encounter some error messages like:
```
The controller result claims to be providing relevant cache metadata, but leaked metadata was detected. Please ensure you are not rendering content too early.
This has to do with early rendering. Most likely in the toUrl->toString calls on entities.
```

## Solution
Let's rewrite the code.

## Issue tracker
https://www.drupal.org/project/social/issues/3081555

## How to test
- [ ] Test JSON API POST event enrolment and there should not be any error.
- [ ] See https://www.lullabot.com/articles/early-rendering-a-lesson-in-debugging-drupal-8

## Release notes
In the token handling of event enrolments there was some leaked cacheable metadata. This has now been solved.

## Change Record
n/a

### DIFF
--- a/modules/custom/activity_creator/src/ActivityFactory.php
+++ b/modules/custom/activity_creator/src/ActivityFactory.php
@@ -5,6 +5,7 @@ namespace Drupal\activity_creator;
 use Drupal\activity_creator\Entity\Activity;
 use Drupal\Component\Render\FormattableMarkup;
 use Drupal\Core\Controller\ControllerBase;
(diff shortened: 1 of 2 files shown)


jaapjan:
Let's put it in 7.x for now only. Also: it solves issues with the Event enrollments in this case. It is likely more of these types of changes are needed in the future.

PR #1503, opened 2019-09-20
fix private messages sorting

albertoalaejos:
## Problem
The private messages are not in a chronological classification.
Private messages are sorted based on last updated time, but the private messages inbox view show the creation time, which does not always match the updated time.

## Solution
Show the last updated time instead of creation time.

## Issue tracker
https://getopensocial.atlassian.net/browse/TB-2919

## How to test
- [ ] Go to /user/inbox
- [ ] Make sure that private messages are sorted in the correct order.

### DIFF
--- a/themes/socialbase/templates/private_message/private-message-thread.html.twig
+++ b/themes/socialbase/templates/private_message/private-message-thread.html.twig
@@ -32,7 +32,9 @@
         </div>
         <div class="media-body message__inbox-body">
           {{ content.membernames }} <span class="read-indicator"></span>
-          {{ content.last_message }}
+          <div class="message__updated">
+            {{ content.updated }} {{ 'ago' }}
+          </div>
         </div>
       </div>
     </div>
--- a/modules/social_features/social_private_message/social_private_message.module
+++ b/modules/social_features/social_private_message/social_private_message.module
@@ -414,6 +414,28 @@ function social_private_message_preprocess_field(&$variables) {
 }
 
 /**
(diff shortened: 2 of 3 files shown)


themes/socialbase/templates/private_message/private-message-thread.html.twig
```diff
@@ -32,7 +32,9 @@
         </div>
         <div class="media-body message__inbox-body">
           {{ content.membernames }} <span class="read-indicator"></span>
-          {{ content.last_message }}
+          <div class="message__updated">
+            {{ content.updated }} {{ 'ago' }}
```
jaapjan: @albertoalaejos should be translatable 😛

jaapjan:
I saw that there was an (unrelated) failure in Travis install_update test which is solved in 8.x-7.x already so merged that into this branch. :crossed_fingers:

themes/socialbase/templates/private_message/private-message-thread.html.twig
```diff
@@ -32,7 +32,9 @@
         </div>
         <div class="media-body message__inbox-body">
           {{ content.membernames }} <span class="read-indicator"></span>
-          {{ content.last_message }}
```
bramtenhove: You're removing the last message here. How come?

bramtenhove:
It turns out that we thought the issue we saw with private messages ordering had to be fixed in the distro. This is not the case, I'll close the PR.

PR #1504, opened 2019-09-20
Issue #3082600 by robertragas: replace html escape with xss filter as…

robertragas:
## Problem
Currently the search input escapes html which make it impossible to search for content in language that use special characters like ', which is commonly used in French.

## Solution
Replace the html escape with a Xss:filter as this also protects the input against attacks.

## Issue tracker
https://www.drupal.org/project/social/issues/3082600

## How to test
- [x]  Create content that has quotes in the title (h'allo)
- [x]  Search for this content and see it will cut off after the h and not able to find the content
- [x]  Try with this solution and see it can be found now
- [x]  Try to execute xss within the input box and url bar.

I used the cheatsheet here
https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

## Release notes
We have fixed and issue where the search input would get trimmed due to special characters like an apostrophe resulting in not being able to find the right content. Now these special characters are usable in search words.

### DIFF
--- a/modules/social_features/social_search/src/Form/SearchContentForm.php
+++ b/modules/social_features/social_search/src/Form/SearchContentForm.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\social_search\Form;
 
-use Drupal\Component\Utility\Html;
+use Drupal\Component\Utility\Xss;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
 use Drupal\Core\Form\FormBase;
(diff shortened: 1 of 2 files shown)


jaapjan:
I saw that there was an (unrelated) failure in Travis install_update test which is solved in 8.x-7.x already so merged that into this branch. 🤞

jaapjan:
Merged in 6.x and will be in 7.0 as well.

PR #1506, opened 2019-09-23
issue #3026830 by albertoalaejos: language switcher not switching

albertoalaejos:
## Problem
Language switcher does not switch languages

## Solution
Update drupal core version to 8.7.7

## Issue tracker
https://www.drupal.org/project/social/issues/3026830

## How to test
- [ ] enable social language module
- [ ] add a second language
- [ ] switch language in the header tab


albertoalaejos:
I thought it fixed the issue, but when testing it again I saw that it remains, so better to close the PR. I will do that now

PR #1508, opened 2019-09-26
Issue #3083874 by ribel, agami4: As a SM I want to switch to different flavors of my  Social Blue theme

ribel:
## Problem
Open Social comes with two related themes: socialbase and socialblue. The visual style of those themes was not really changed from the beginning. It is also hard for maintenance and support to keep multiple themes in a distro. That's why we decide to create a mechanism for changing the visual style of a theme, without switching a theme. 

## Solution

1. Add new style selector to theme settings of socialblue theme. 
2. Use this style for template suggestions.
3. Use this style as a variable for in preprocess functions.
4. Use this style as a body class for CSS selectors.

image

## Issue tracker
- https://www.drupal.org/project/social/issues/3083874
- https://getopensocial.atlassian.net/browse/FLAV-1
- https://marvelapp.com/4c5ch9a/screen/52141569

## How to test
- [ ] Go to **Social blue** theme settings
- [ ] You should see a new tab called **Style**
- [ ] Select **Sky (new)** style
- [ ] You should see class `socialblue--sky` in body
- [ ] You should be able to use new template suggestions with the suffix `--sky`
- [ ] You should see cards in a new style
- [ ] You should see **Read more** links in node teasers in a button style
- [ ] You should see **See all** link with an arrow in the end 
image

## Release notes
We created a mechanism for changing the visual style (flavour) of a theme, without switching a theme. You can switch to a new **Sky** flavour in **Style** settings of the **Social Blue** theme.


agami4:
@Kingdutch, thank you for review this PR :)
1. We have `autoprefixer` option with the same conditions as in the `.browserslistrc` file. 
Whether it makes sense to attach `.browserslistrc` file?
2. I agree with you. Can I use `style-sky` variable to attach libraries into templates or need to attach all libraries via preprocess? 
3. I'll ask Xinyu about it.
4. What do you mean about hiding the dropdown menu for SM?  Hide the dropdown menu on the main menu?

xinyumaGG:
@agami4 
If I am not mistaken, point 3 is about using the same colour set (primary, secondary and accent ) of the old theme in the new theme.

Kingdutch:
@agami4 you're welcome :D 

1. Interesting, I believe the `-webkit-` prefixes that were added in this PR are not needed anymore in the browsers that we support. Perhaps we can restrict the `.browserslistrc` file a bit further 🤔 
2. I'm not sure what you mean? Do you mean something like 
  ```
  {% if style-sky %}
  {{ attach_library('lib') }}
  {% endif %}
  ``` 
  I'm not sure how performant that is and how well it works with caching 🤔 My thoughts were to use the [`hook_library_info_alter`](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Render!theme.api.php/function/hook_library_info_alter/8.7.x) function to extend the libraries offered by SocialBlue with the sky specific styling files. This makes it easy to see in a single function what changes the sky flavour makes and even allows us to disable some non-sky CSS if that makes theming easier.

3. 👍 
4. In the screenshot for this PR there's the "Style" option in the Open Social settings page. I mean hiding this dropdown for now. This way we can merge these features for internal use without exposing them to the wider Drupal community. Once the theme is ready to use for everyone and we're ready to provide support to everyone then we can add the dropdown again :)

agami4:
@Kingdutch,
1. I agree with you about this prefix, but, as practice shows, many users use old browsers. If this prefix no needed, I'll remove `autoprefixer` form gulp and use `.browserslistrc`.
2. Yes, I wrote about it)
3. Xinyu answered my question and added a comment about the color scheme.
4. Ok, this point for BE side)

Kingdutch:
1. Aah I missed that our gulpfile had a browserlist hardcoded. Newer versions of `autoprefixer` use the `.browserlistrc` as configuration automatically. Then just leave it as is for now. I'll try to get https://github.com/goalgorilla/sweet-pea working in the coming months to fix that :) 
2. I think the `hook_library_info_alter` implementation is more predictable in behaviour than conditional attachment of styles to templates (and probably works better with Behat). So this may be something you can implement with some help of a back-end developer. We will have to clear the library cache when the value of the flavour configuration is changed.
3. Okay, so the conclusion is indeed to use the same colors in both flavours? :)
4. 👍

agami4:
1. Can I help you with the https://github.com/goalgorilla/sweet-pea?)
2. 👍
3. Yep :)

agami4:
3. 'Social Blue Sky' color scheme was removed

ribel:
@Kingdutch, @agami4 I have implemented `socialblue_library_info_alter()` for adding extra CSS files if style is chosen.
See commit: bfe13ce

PR #1509, opened 2019-09-26
Issue #3083976 by chmez: Push notifications provider

ghost:
## Problem
When I want to use some custom provider of push notifications together with (or separate) PWA provider from then [Social PWA](https://www.drupal.org/project/social_pwa) module then I can not create some settings wrapper on the user settings page.

## Solution
Create a new module which provides a new plugin for creating instances for each provider of push notifications as PWA, Firebase, etc.

## Issue tracker
- https://www.drupal.org/project/social/issues/3083976
- https://getopensocial.atlassian.net/browse/ECI-921


jaapjan:
Could you take a look at the naming of this module? 

It looks confusing as we have almost the same module name here: https://github.com/goalgorilla/social_pwa

What happens with the one in social_pwa if this one is added? Can we use the same provider for both Apps and PWA?

jaapjan:
> What is your suggestion for the module name? We thought to use the activity_send_push name but it already using for the submodule of the Social PWA module.

Yes, indeed. What about `activity_send_push_firebase` or does it not make sense?

And good to know that the other PR also need to be merged when this is going to be merged.

OH17L:
> Yes, indeed. What about activity_send_push_firebase or does it not make sense?

Yes, it does not make sense because this module provide a new plugin for different push notification providers (not only for **Firebase**).

ribel (approved):
The idea behind this refactoring and code looks good to me.
Just we need to be sure that it is merged and released together with https://github.com/goalgorilla/social_pwa/pull/9
And please note that we update Message module to stable version within this PR

ronaldtebrake:
Thanks guys, looks good to me. As discussed this will be backported to 7.x and will land in 7.2 
the PR here: https://github.com/goalgorilla/social_pwa/pull/9#issuecomment-553455245 
Will be updated with this minimal dependency of 7.2 and higher :) so we can tag and release it separately.

ronaldtebrake:
P.S. Travis was green before but due to core patch isnt as we speak. :) Will merge anyway.

ronaldtebrake:
Cherry-picked in 7.x :)

PR #1510, opened 2019-09-27
Issue #3087229: Add roles field to paragraphs to control visibility

albertoalaejos:
## Problem
As CM we want to be able to select which information will be visible in the landing page for each role.

## Solution
Add a roles field in section paragraphs and hide the paragraph for those not selected.

## Issue tracker
https://getopensocial.atlassian.net/browse/TB-3077
https://www.drupal.org/project/social/issues/3087229

## How to test
- [ ] enable social landing page module
- [ ] make sure that cache is enabled
- [ ] go to /node/add/landing_page
- [ ] click on "add section"
- [ ] choose a type
- [ ] choose CM+ roles
- [ ] save
- [ ] visit the new landing page as CM+, the paragraph should be visible
- [ ] visit the new landing page as LU, the paragraph should be hidden

## Release notes
We added a roles field for section paragraphs in landing pages where you can select the type of user that has permission to see it.


ronaldtebrake:
FYI - Make sure you add the correct milestone when merging otherwise we don't know in which releases this should end up and they might get lost in filters. 🚀

Kingdutch:
Hi @bramtenhove,

@albertoalaejos and I had this discussion last week. Not doing this dynamically was a conscious decision with a focus on site managers. We would already have to do some work to not add this field to the `section` paragraph because that would cause this field to be nested. Additionally we don't know in what way people are using paragraphs on their site (perhaps they're using it for product pages in a commerce set-up). Additionally, in the future we may want to use paragraphs for a feature where role control doesn't make sense. We would have to perform additional work to fix issues arising from (site builders) not wanting this field in their custom paragraphs. 

With this in mind, it's safest to just add this manually to all the paragraphs that we control, even if this means a bit more work up front.

Having said the above @albertoalaejos: Did we consider adding this **only** to the section paragraph and no other paragraphs? I believe that paragraph wraps all other available types for landing pages already so it may be best of both worlds. It would add this to all paragraph types available for landing pages, but no others.

albertoalaejos:
@ronaldtebrake I changed the milestone to 7.0. Thank you for the tip :)

albertoalaejos:
@bramtenhove and @Kingdutch
I think that adding the roles field only to section paragraph would be indeed the best solution. I will do that ;)

bramtenhove:
@Kingdutch, I see your point. There is an obvious risk in doing it like that, but it also reduces the maintenance burden. Now, when a new Paragraph type is created, it also needs to be added to this list. And we can all foresee this will be forgotten at some point.

I'm OK with doing it like this, but then still this part of my previous message needs an answer and acted on:
> If not, please make sure you add tasks to create the field for all modules we control. As well as documentation on D.org how to use the feature for custom development.

bramtenhove:
This PR is ready to be merged, but since we have a feature freeze it will have to wait until 7.1.

jaapjan:
This still needs a drupal.org ticket. could you create one and remove the label `needs drupal.org ticket`?

albertoalaejos:
> This still needs a drupal.org ticket. could you create one and remove the label `needs drupal.org ticket`?

Done

Kingdutch:
@bramtenhove @ronaldtebrake I'm hesitant to merge this. It makes a pretty sweeping change for all people that manage landing pages. Are we sure we want this functionality on all rows and not make this something that people can opt-in to? The field is pretty intrusive from a UX perspective.

If we do want to make this optional on platforms, work will be needed to move it to a separate platform. Otherwise this is good to go for 8.0.

ronaldtebrake:
We won't be supporting this in the Distro, we will move towards the layout builder. This will have added complexity and will be a blocker for migrating landing pages towards the new layout builder initiative. 
Instead we opt for making this a Drupal contrib module, on paragraphs level this could have added benefits for a lot of Drupal users. 

cc @Kingdutch

modules/social_features/social_landing_page/src/SocialLandingPageConfigOverride.php
```diff
@@ -32,6 +51,29 @@ public function loadOverrides($names) {
...
+      'core.entity_form_display.paragraph.section.default',
+    ];
+
+    foreach ($config_names as $config_name) {
+      if (in_array($config_name, $names)) {
+        // Grab current configuration and push the new values.
+        $config = $this->configFactory->getEditable($config_name);
+        // We have to add config dependencies to field storage.
+        $dependencies = $config->getOriginal('dependencies', FALSE)['config'];
+        $dependencies[] = 'field.field.paragraph.field_roles';
```
navneet0693: Should be `field.field.paragraph.section.field_roles`

modules/social_features/social_landing_page/src/SocialLandingPageConfigOverride.php
```diff
@@ -32,6 +51,29 @@ public function loadOverrides($names) {
...
+        $overrides[$config_name]['dependencies']['config'] = $dependencies;
```
navneet0693: Not necessary can update `core.entity_form_display.paragraph.section.default.yml`

royharink:
@ronaldtebrake, @bramtenhove this code has been moved to a private repo.
You can delete this branch if you need.

PR #1513, opened 2019-09-27
Issue #3084268 by robertragas: Optimize group membership count

robertragas:
## Problem
Currently there are two places in Open Social where a count is being done for the amount of members in a group. This shows up in the group overview, but also on user profiles or wherever a group teaser shows up.
The other one is in the admin page of groups.

What it does is get all the member entities of a group and then does a count on this. When a group has over 30k members this will cause a timeout.

## Solution
Query the database, only getting the information you need instead of all the entities and then counting them which is very heavy.

## Issue tracker
https://www.drupal.org/project/social/issues/3084268

## How to test
- [x] Clear the cache
- [x] Create a group with a lot members, I tested it with 35k
- [x] Go to the explore overview of groups and the admin overview of groups
- [x] The request will time out every single time
- [x] Check out to this branch and check again

## Release notes
We changed the way we count the group members making it more performant. Because groups are used in a lot of places of Open Social the difference should be noticeable for everybody, but especially for Open Social installations with a lot of users in groups.

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -319,7 +319,8 @@ function social_group_preprocess_group(array &$variables) {
   }
 
   // Count number of group members.
-  $variables['group_members'] = count($group->getMembers());
+  $group_members_count = \Drupal::service('social_group.group_members_count');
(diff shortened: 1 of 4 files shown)


jaapjan:
Cherry-picked to 6.x

PR #1514, opened 2019-09-29
#3077289 - Secret group doesnt have defaults for visbility settings

ronaldtebrake:
## Problem & Solution
Secret group doenst have defaults for visibility settings, also added a defensive check for when other group types, it needs to have the field available on the entity for us to check it.

## Issue tracker
https://www.drupal.org/project/social/issues/3077289

## How to test
- [x] Enable secret groups, see that for the Post form it has group members as only option.

## Release notes
We have made sure there are sane defaults for the secret group which is now available in Open Social.


jaapjan:
Cherry-picked to 6.x as well

PR #1515, opened 2019-09-30
Issue #3083877 by ribel, agami4: As a LU I want to have group and profile layout in new sky flavor

ribel:
## Problem
<describe the problem you're trying to solve>

## Solution
<describe the solution you're created>

## Issue tracker
https://www.drupal.org/project/social/issues/3083877

## How to test

## Release notes
<describe the release notes>

## Change Record


modules/social_features/social_group/social_group.module
```diff
@@ -317,10 +317,20 @@ function social_group_preprocess_group(array &$variables) {
...
-  $variables['group_members'] = $group_members_count->getGroupMemberCount($group);
+  $variables['group_members'] = count($group->getMembers());
+
+  // Prepare variables for statistic block.
+  if ($group->getGroupType()->hasContentPlugin('group_node:event')) {
+    $variables['group_events'] = count($group->getContent('group_node:event'));
+  }
+  if ($group->getGroupType()->hasContentPlugin('group_node:topic')) {
+    $variables['group_topics'] = count($group->getContent('group_node:topic'));
+  }
```
Kingdutch: ❗️ This is a serious performance regression. We're basically loading all the group content here before counting it. 

The code that was removed was more efficient. We should probably create a `GroupStatistics` service in the `social_group` module and create method for each type there. Then we can deprecate `SocialGroupMembersCount` and route the `SocialGroupMembersCount::getGroupMemberCount` call to `GroupStatistics::memberCount`.

modules/social_features/social_group/social_group.module
```diff
@@ -317,10 +317,20 @@ function social_group_preprocess_group(array &$variables) {
...
+  if ($group->getGroupType()->id() !== 'closed_group' || $group->getMember($account) || $group->hasPermission('manage all groups', $account)) {
+    $variables['group_stream_url'] = Url::fromUserInput('/group/' . $group->id() . '/stream');
```
Kingdutch: We shouldn't be using `fromUserInput` but instead use the routename with the proper argument. This ensures that we don't create issues if the URL is changed.

modules/social_features/social_group/social_group.module
```diff
@@ -1325,6 +1336,12 @@ function social_group_block_access(Block $block, $operation, AccountInterface $a
     }
     return $access;
   }
+  // Show statistic block only when new style is enabled.
+  if ($operation == 'view' && $block->getPluginId() === 'group_statistic_block') {
+    if (theme_get_setting('style') !== 'sky') {
+      return AccessResult::forbidden();
+    }
+  }
```
Kingdutch: We control the block so we should probably just add this condition to the block.

Additionally, this creates a dependency of the block on the theme, which shouldn't happen. So we should find a way around this.

modules/social_features/social_landing_page/social_landing_page.module
```diff
@@ -62,10 +62,16 @@ function social_landing_page_theme() {
   $theme_templates['group__featured'] = [
     'base hook' => 'group',
   ];
+  $theme_templates['group__featured__sky'] = [
+    'base hook' => 'group',
+  ];
```
Kingdutch: We're creating a new template/theme couple here. We probably want to adjust the template suggestions for the `group__featured` template instead.

modules/social_features/social_landing_page/social_landing_page.module
```diff
@@ -62,10 +62,16 @@ function social_landing_page_theme() {
...
   // Profile.
   $theme_templates['profile__featured'] = [
     'base hook' => 'profile',
   ];
+  $theme_templates['profile__featured__sky'] = [
+    'base hook' => 'profile',
+  ];
```
Kingdutch: We're creating a new template/theme couple here. We probably want to adjust the template suggestions for the `profile__featured` template instead.

modules/social_features/social_landing_page/templates/group--featured--sky.html.twig
```diff
@@ -0,0 +1,65 @@
+{% extends 'group--featured.html.twig' %}
```
Kingdutch: I feel like this file has a lot in common with changes we make in #1511 Perhaps we can deduplicate this somehow?

Additionally, we shouldn't have `--sky` things in modules since this is a theme thing.

modules/social_features/social_landing_page/templates/node--featured--sky.html.twig
```diff
@@ -1,17 +1,18 @@
 {% extends "node--featured.html.twig" %}
```
Kingdutch: I missed where this file was introduced, but I think we should keep `--sky` templates in the `socialblue` template because they're a specialcase of that theme and not part of this module.

modules/social_features/social_landing_page/templates/profile--featured--sky.html.twig
```diff
@@ -0,0 +1,21 @@
+{% extends 'profile--featured.html.twig' %}
```
Kingdutch: I think we should keep `--sky` templates in the `socialblue` theme because they're a special case of that theme and not part of this module.

modules/social_features/social_profile/social_profile.module
```diff
@@ -377,6 +379,18 @@ function social_profile_preprocess_profile(array &$variables) {
...
+  $storage = \Drupal::entityTypeManager()->getStorage('node');
+  $variables['profile_topics'] = count($storage->loadByProperties([
+    'uid' => $user->id(),
+    'type' => 'topic',
+  ]));
+  $variables['profile_events'] = count($storage->loadByProperties([
+    'uid' => $user->id(),
+    'type' => 'event',
+  ]));
+  $variables['profile_groups'] = count(\Drupal::service('social_group.helper_service')->getAllGroupsForUser($user->id()));
```
Kingdutch: ❗️  This is a performance red flag because we're loading a lot of content here to count it. We should instead just create an entity count query. Perhaps create a `UserStatistics` service much like `GroupStatistics`

modules/social_features/social_profile/social_profile.module
```diff
@@ -660,24 +674,24 @@ function social_profile_social_user_account_header_account_links(array $context)
...
+        '#type' => 'link',
+        '#attributes' => [
+          'title' => new TranslatableMarkup("Edit profile"),
+        ],
+        '#title' => new TranslatableMarkup("Edit profile"),
+        '#weight' => 1300,
+      ] + Url::fromRoute('entity.profile.type.user_profile_form', [
+        'user' => $context['user']->id(),
+        'profile_type' => 'profile',
+      ])->toRenderArray(),
```
Kingdutch: Did the coding standards change?

modules/social_features/social_profile/social_profile.module
```diff
@@ -711,3 +725,17 @@ function social_profile_social_user_account_header_items_alter(array &$menu_item
...
+function social_profile_block_access(Block $block, $operation, AccountInterface $account) {
+  // Show statistic block only when new style is enabled.
+  if ($operation == 'view' && $block->getPluginId() === 'profile_statistic_block') {
+    if (theme_get_setting('style') !== 'sky') {
+      return AccessResult::forbidden();
+    }
+  }
+  // No opinion for other situations.
+  return AccessResult::neutral();
+}
```
Kingdutch: We own the block, so we don't need an external access handler. Additionally, we should decouple this from the theme in some way.

modules/social_features/social_profile/src/Plugin/Block/ProfileStatisticBlock.php
```diff
@@ -0,0 +1,106 @@
...
+
+    $storage = $this->entityTypeManager->getStorage('profile');
+    $profile = $storage->loadByUser($account, 'profile');
+
+    if ($profile) {
+      $build['content'] = $this->entityTypeManager
+        ->getViewBuilder('profile')
+        ->view($profile, 'statistic');
+      $build['content']['#cache']['tags'] = $this->getCacheTags();
+      $build['content']['#cache']['contexts'] = $this->getCacheContexts();
```
Kingdutch: I like all the code above! However, I don't think we need these tags and contexts because they duplicate the block lower in the tree so this has no real effect other than some bookkeeping overhead.

modules/social_features/social_profile/src/Plugin/Block/ProfileStatisticBlock.php
```diff
@@ -0,0 +1,106 @@
...
+   */
+  public function build() {
+    $build = [];
+    $account = $this->getContextValue('user');
+
+    $storage = $this->entityTypeManager->getStorage('profile');
+    $profile = $storage->loadByUser($account, 'profile');
+
+    if ($profile) {
+      $build['content'] = $this->entityTypeManager
```
Kingdutch: I think we can just set $build to equal the result of this and don't need to nest this in `content`.

themes/socialbase/templates/layout/page.html.twig
```diff
@@ -39,16 +39,20 @@
 
 <main id="content" class="main-container" role="main">
 
-  {% if page.breadcrumb %}
-    {{ page.breadcrumb }}
+  {% if not style == 'sky' %}
+    {% if page.breadcrumb %}
+      {{ page.breadcrumb }}
+    {% endif %}
```
Kingdutch: Socialbase shouldn't be aware of the `sky` flavour. Perhaps we can remove the breadcrumbs through a preprocess hook in the `socialblue` theme. 

@xinyuma should also probably think about how we want to use breadcrumbs throughout all flavours, I think this is still something we're inconsistent with even though they can be useful.

themes/socialbase/templates/layout/page.html.twig
```diff
@@ -39,16 +39,20 @@
...
   {% if page.hero %}
     {{ page.hero }}
   {% endif %}
 
-  {% if page.secondary_navigation %}
-    {{ page.secondary_navigation }}
+  {% if not style == 'sky' %}
+    {% if page.secondary_navigation %}
+      {{ page.secondary_navigation }}
+    {% endif %}
```
Kingdutch: Socialbase shouldn't be aware of the sky flavour

themes/socialbase/templates/layout/page.html.twig
```diff
@@ -63,14 +67,22 @@
         {{ page.title }}
       {% endif %}
 
+      {% if style == 'sky' %}
```
Kingdutch: Socialbase shouldn't be aware of the sky flavour

themes/socialbase/templates/layout/page.html.twig
```diff
@@ -63,14 +67,22 @@
...
+
       {% if page.complementary_top or page.complementary_bottom %}
         <aside class="region--complementary" role="complementary">
-          {% if page.complementary_top %}
-            {{ page.complementary_top }}
-          {% endif %}
-          {% if page.complementary_bottom %}
-            {{ page.complementary_bottom }}
-          {% endif %}
+          <div class="complementary--content-merged card">
```
Kingdutch: Does this `div` truly add value here? Can we not just style the `region--complementary` instead?

themes/socialblue/components/03-molecules/navigation/navbar/navbar-secondary--sky.js
```diff
@@ -0,0 +1,65 @@
+(function ($) {
+
+  Drupal.behaviors.navbarSecondaryScrollable = {
```
Kingdutch: Please add a top level comment to this function to explain what it's used for so we still know this in 6 months time.

Additionally, please add comments throughout the function to make it easier for other people than the author to follow what's going on. Comments that document **why** something happens are more useful than **what** happens because the latter can be seen by reading the code.

themes/socialblue/socialblue.info.yml
```diff
@@ -10,6 +10,7 @@ core: 8.x
 libraries:
   - socialblue/brand
   - socialblue/site-footer
+  - socialblue/nav-secondary--sky
```
Kingdutch: This means we're adding the library always, even when the sky flavour isn't used.

themes/socialblue/src/Plugin/Preprocess/Page.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Drupal\socialblue\Plugin\Preprocess;
+
+use Drupal\socialbase\Plugin\Preprocess\Page as PageBase;
```
Kingdutch: We should check if we should overwrite the Socialbase Page Preprocessor here because I think they are all executed by Bootstrap, this would case the SocialBase one to be executed twice. Use the Bootstrap PreprocessBase instead.

themes/socialblue/src/Plugin/Preprocess/Page.php
```diff
@@ -0,0 +1,45 @@
...
+class Page extends PageBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function preprocess(array &$variables, $hook, array $info) {
+    parent::preprocess($variables, $hook, $info);
+
+    $style = theme_get_setting('style');
+    if ($style && $style === 'sky') {
```
Kingdutch: The first check is unneccessary here I think. Any value that would return false there would also not `=== 'sky'`.

themes/socialblue/templates/group/group--statistic--sky.html.twig
```diff
@@ -0,0 +1,40 @@
...
+    </li>
+    <li>
+      <span class="card__counter-quantity">{{ group_topics }}</span>
+      <span class="card__counter-text">{% trans %}topic{% plural group_topics %}topics{% endtrans %}</span>
+    </li>
+    <li>
+      <span class="card__counter-quantity">{{ group_members }}</span>
+      <span class="card__counter-text">{% trans %}member{% plural group_members %}members{% endtrans %}</span>
+    </li>
+  </ul>
```
Kingdutch: Can we in some way make this more generic or fieldable? It think we're using this structure more often going forward and we could set this up in such a way that we don't need to adjust a template when we want to add a statistic (or have an extension add a statistic or change what's shown).

themes/socialblue/templates/group/group--statistic--sky.html.twig
```diff
@@ -0,0 +1,40 @@
...
+          <li><a href="{{ group_operations_url }}">{% trans %}Leave group{% endtrans %}</a></li>
+        </ul>
+      {% elseif closed_group %}
+        <a href="{{ group_operations_url }}" class="btn btn-accent disabled" title="{{ cta }}">{{ cta }}</a>
+      {% else %}
+        <a href="{{ group_operations_url }}" class="btn btn-accent" title="{% trans %}Join{% endtrans %}">{% trans %}Join{% endtrans %}</a>
+      {% endif %}
+    </div>
+  </div>
+{% endif %}
```
Kingdutch: This feels like a lot of logic that probably shouldn't be in a template file but simply be created with a render array.

themes/socialblue/templates/profile/profile--profile--statistic--sky.html.twig
```diff
@@ -0,0 +1,51 @@
+{{ attach_library('socialbase/hero') }}
```
Kingdutch: The same questions about making this more generic exist as for `themes/socialblue/templates/group/group--statistic--sky.html.twig`.

modules/social_features/social_group/social_group.module
rollki: @Kingdutch 
I agree with you. I already fixed this performance issue [here](https://github.com/goalgorilla/open_social/pull/1515/commits/cd9ededd5a30a0f2fcfb176a63bc12b5972d5a7f).
I created a new service `GroupStatistics` which has 2 methods:
* `nodeCount` for an event and topic types
* `memberCount` for the `group_membership`
Also I removed old service `SocialGroupMembersCount` and used Dependency Injection in `SocialGroupListBuilder` for the `groupStatistics` service instead of `\Drupal->service()`

modules/social_features/social_group/social_group.module
rollki: @Kingdutch
Fixed for all cases in this module.

modules/social_features/social_group/social_group.module
rollki: @Kingdutch 
Added to the `blockAccess`.

modules/social_features/social_profile/social_profile.module
rollki: @Kingdutch 
Fixed. Created the UserStatistics service.

modules/social_features/social_profile/social_profile.module
rollki: Fixed.

modules/social_features/social_landing_page/social_landing_page.module
ribel: Templates were moved to socialblue theme (the same as for content).

modules/social_features/social_landing_page/templates/group--featured--sky.html.twig
ribel: A template was refactored and simplified.

modules/social_features/social_landing_page/templates/node--featured--sky.html.twig
ribel: Fixed and removed from here.

modules/social_features/social_landing_page/social_landing_page.module
ribel: This Template was moved to socialblue theme (the same as for content).

modules/social_features/social_landing_page/templates/profile--featured--sky.html.twig
ribel: Fixed.

themes/socialblue/src/Plugin/Preprocess/Page.php
ribel: In this case, we need to use `$variables['content_attributes']`, which was defined only in socialbase, but not exists in bootstrap. It gives error when I try to extend it directly from bootstrap, so we need to leave it like this.

ribel:
Hi @Kingdutch, we fixed all the things after your initial review and fixed issues found by Xinyu.

The only thing which is left from your general comment to not have a condition about `sky` in modules are left in of Group and Profile Statistic block access. 
I've researched but looks like there are no real options in Drupal to avoid it. We can create custom ConditionPlugin, but this will not change too much since anyway this code will be located in the module. So I would propose to leave it like it is for now, and maybe later We can just display the same block in the default style as well.

themes/socialblue/components/03-molecules/navigation/navbar/navbar-secondary--sky.js
agami4: The comment was added to the `navbar-secondary--sky.js`

modules/social_features/social_group/src/Plugin/Block/GroupStatisticBlock.php
```diff
@@ -0,0 +1,110 @@
...
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $build = [];
+
+    $group = _social_group_get_current_group();
```
Kingdutch: This should use the block context instead. The block context should probably be set to `required`. You may need to set the default configuration for this block to pick the group from the URL.

An example can be seen in https://bitbucket.org/goalgorilla/social_dashboard/src/master/src/Plugin/Block/DashboardNavigationBlock.php

modules/social_features/social_landing_page/config/install/core.entity_view_display.profile.profile.featured.yml
```diff
@@ -16,7 +16,7 @@ dependencies:
     - field.field.profile.profile.field_profile_profile_tag
     - field.field.profile.profile.field_profile_self_introduction
     - field.field.profile.profile.field_profile_show_email
-    - image.style.social_featured
+    - image.style.social_large
```
Kingdutch: The `social_featured` image style was created for this view mode, so if we need to change it because we're changing how landing pages look then we should probably just change the image style itself.

Additionally this will also impact "old-style" landing pages.

modules/social_features/social_profile/social_profile.module
```diff
@@ -376,7 +376,15 @@ function social_profile_preprocess_profile(array &$variables) {
...
-
+  // Profile information URL.
+  $variables['profile_info_url'] = Url::fromRoute('view.user_information.user_information', ['user' => $profile->getOwnerId()]);
+
+  // Prepare variables for statistic block.
+  /** @var \Drupal\social_profile\UserStatistics $user_statistics */
+  $user_statistics = \Drupal::service('social_profile.user_statistics');
+  $variables['profile_topics'] = $user_statistics->nodeCount($user->id(), 'topic');
+  $variables['profile_events'] = $user_statistics->nodeCount($user->id(), 'event');
+  $variables['profile_groups'] = count(\Drupal::service('social_group.helper_service')->getAllGroupsForUser($user->id()));
```
Kingdutch: ❗️  This now runs for all view modes and can have a negative effect on performance on user listings (where this data is not used).

themes/socialblue/color/color.inc
```diff
@@ -5,6 +5,16 @@
  * Lists available colors and color schemes for the Social Blue theme.
  */
 
+$brand_css = [
+  'assets/css/brand.css',
+];
+// Add brand css files for new Sky style.
+$theme_path = drupal_get_path('theme', 'socialblue');
+$sky_brand_css = 'assets/css/brand--sky.css';
+if (file_exists($theme_path . '/' . $sky_brand_css)) {
```
Kingdutch: Isn't this always true?

themes/socialblue/components/03-molecules/dropdown/dropdown--sky.scss
```diff
@@ -0,0 +1,17 @@
+// Behaviors arrow on the dropdown menu
+
+.socialblue--sky {
+  .dropdown.open .caret,
+  .enroll-action-form.open .caret,
+  .hero-footer__cta .open .caret,
+  .social-gtranslate-list--desktop.open > a:after {
+    transform: rotate(180deg);
+  }
```
Kingdutch: I think this is a change we can make in both flavours.

Additionally the following can make it look at bit more friendly:
```
.dropdown .caret {
    transition: transform 0.1s;
}
```

themes/socialblue/components/04-organisms/comment/comment--sky.scss
```diff
@@ -0,0 +1,62 @@
+@import "settings";
+
+.socialblue--sky {
+  [id='section-comments'] {
```
Kingdutch:
```suggestion
  #section-comments {
```

themes/socialblue/package-lock.json
```diff
@@ -0,0 +1,9873 @@
+{
```
Kingdutch: Please remember that we use `yarn` and not `npm` for our theme package management.

themes/socialblue/socialblue.theme
```diff
@@ -32,5 +32,34 @@ function socialblue_library_info_alter(&$libraries, $extension) {
         $libraries[$key]['css']['theme'][$style_css]['weight'] = 600;
       }
     }
+
+    // Add extra JS for secondary navigation in Sky style.
+    $navbar_js = 'assets/js/navbar-secondary--sky.min.js';
+    if (file_exists($theme_path . '/' . $navbar_js)) {
```
Kingdutch: Isn't this always true and shouldn't this depend on whether the Sky flavour is enabled?

modules/social_features/social_landing_page/config/install/core.entity_view_display.profile.profile.featured.yml
ribel: We need to change image style on featured teasers only for profiles (for both old and sky - this was agreed with Xinyu).
But we still use `social_featured` image style for featured content and groups.
So in this way, it makes the most sense to change image style to ones that fit with a new design.
Also, it shouldn't break anything, since CSS for both old and new styles were updated as well

themes/socialblue/components/04-organisms/comment/comment--sky.scss
agami4: Fixed

themes/socialblue/components/03-molecules/dropdown/dropdown--sky.scss
agami4: @Kingdutch, I've moved these styles in the SocialBase theme and add `transition`.

themes/socialblue/package-lock.json
agami4: The `package-lock.json` was removed

modules/social_features/social_group/social_group.services.yml
```diff
@@ -32,6 +32,6 @@ services:
   social_group.hero_image:
     class: Drupal\social_group\SocialGroupHero
     arguments: ['@config.factory']
-  social_group.group_members_count:
-    class: Drupal\social_group\SocialGroupMembersCount
+  social_group.group_statistics:
```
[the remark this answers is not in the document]
ribel: This was fixed in PR #1680

modules/social_features/social_group/src/GroupStatistics.php
```diff
@@ -36,14 +36,44 @@ public function __construct(Connection $connection) {
    *   The group entity.
    *
    * @return int
-   *   Number of members in a group.
+   *   The number of members.
    */
-  public function getGroupMemberCount(GroupInterface $group) {
+  public function memberCount(GroupInterface $group) {
```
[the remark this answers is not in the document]
ribel: This was fixed in PR #1680 - I reverted the old method name

modules/social_features/social_group/src/Plugin/Block/GroupStatisticBlock.php
ribel: This was fixed in PR #1680

modules/social_features/social_profile/social_profile.module
ribel: This was fixed in PR #1680

themes/socialblue/color/color.inc
ribel: Good catch - I simplified it and fixed in PR #1680

themes/socialblue/socialblue.theme
ribel: Good catch - I simplified it and fixed in PR #1680

themes/socialblue/package-lock.json
ribel: It was fixed in PR #1680

themes/socialblue/components/04-organisms/comment/comment--sky.scss
ribel: It was fixed in PR #1680

themes/socialblue/components/03-molecules/dropdown/dropdown--sky.scss
ribel: It was fixed in PR #1680

PR #1519, opened 2019-09-30
Issue #3030233 Don't use existing configuration on Private text editor override class

jaapjan:
## Problem
Configuration service is injected via Drupal service class.

## Solution
Is injected via dependency injection now.

## Issue tracker
https://www.drupal.org/project/social/issues/3030233

## How to test
- [x] Upload file via WYSIWYG
- [x] Note it is actually uploaded in private files.

## Release notes
will follow

### DIFF
--- a/modules/custom/social_file_private/social_file_private.services.yml
+++ b/modules/custom/social_file_private/social_file_private.services.yml
@@ -5,5 +5,6 @@ services:
       - {name: config.factory.override, priority: 5}
   social_file_private_text_editor.overrider:
     class: \Drupal\social_file_private\SocialFilePrivateTextEditorConfigOverride
+    arguments: ['@config.factory']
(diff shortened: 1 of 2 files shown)


jaapjan:
@Kingdutch what did you want to do with this? Code works perfectly,  but is this what you expected when you created the issue?

I thought you wanted to remove the entire config check right?

Kingdutch:
Hey @jaapjan, 

The title and the work done indeed don't match. I think it's a good improvement to commit. However, the ticket should just be renamed to something like `Use dependency injection in <service>` and decoupled from its parent.

Removing the config check here is difficult because theoretically someone may use a streamwrapper such as `cdn://` to serve things from an external source.

jaapjan:
Also in 6.x

PR #1524, opened 2019-09-30
Issue #3036356 by jyoti.singh: On-going events should be still shown in event overviews

jaapjan:
## Problem
Events that started are not visible in event overviews.

## Solution
Filter on event end date instead of start date.

## Issue tracker
https://www.drupal.org/project/social/issues/3036356

## How to test
- [ ] Create multi day event with startdate before "now".
- [ ] Make sure the event is in the overview.

## Release notes
Multi-day event dates are now shown in event overviews.


jaapjan:
Marking as needs work since a lot of the tests are failing..

PR #1526, opened 2019-10-01
#2994248 Make sure on install SM + CM also need to consent

ronaldtebrake:
## Problem
SM + CM are not asked to give consent to data policy

## Solution
On install of the module both CM + SM get the "without consent" permission so they don't need to consent. I have removed that so it's now only for admin and anon. 
I decided this is only for 7.0 and newer installs so current CM + SM aren't disrupted, we can always do it for clients themselves in our own update hook.

## Issue tracker
- [ ] https://www.drupal.org/project/social/issues/2994248

## How to test
- [x] Enable social_gdpr
- [x] See that CM and SM need to consent now

## Release notes
SM and CM now need to consent for GDPR as well, if you want this on your existing platforms make sure CM and SM do not have the "without consent" permission.


jaapjan:
Marking as needs work since Behat test need to be updated still (since CM/SM now need to give consent 👍 )

https://travis-ci.org/goalgorilla/open_social/jobs/591925204

PR #1527, opened 2019-10-01
#3087400 - TB-3068 - Add optional support for advanced images module

royharink:
## Motivation
This adds support for the optional module advanced images. It gives the possibility to add a copyright attribute to images.

## Solution
Optional checks for the module need to be implemented so the correct attribute is added to the theme.

## Issue tracker
- [D.org #3087400](https://www.drupal.org/project/social/issues/3087400)

## How to test
- [ ] Make sure the optional module is installed.
- [ ] Create entities and upload an image, see the copyright field is available.
- [ ] See you entity with the image still can be saved. This together with all the existing settings like the image style and cropping.
- [ ] Go to the entity, when the image is in the hero region, see that it has the data attribute copyright.
- [ ] When hovering on the hero image see it shows the copyright on the bottom left.

- [ ] In a new or the earlier created entity, add an image to the body field.
- [ ] Now right click and see that in the contextual menu there is an option copyright.
- [ ] A popup should open with the copyright field.
- [ ] Fill this in and click save.
- [ ] Go to the entity, see that the image has the data attribute copyright.
- [ ] When hovering on the inline body image see it shows the copyright on the bottom left.


modules/custom/social_advanced_image/social_advanced_image.install
```diff
@@ -0,0 +1,48 @@
...
+}
+
+
+/**
+ * Implements hook_uninstall().
+ *
+ * Perform actions on uninstall related to the image copyright fields.
+ */
+function social_advanced_image_uninstall() {
+
```
bramtenhove: In here I think we need to remove the fields from the node table schemas still.

modules/custom/social_advanced_image/social_advanced_image.services.yml
```diff
@@ -0,0 +1,10 @@
+services:
+  plugin.manager.image_copyright_attributes:
```
bramtenhove: As part of the renaming effort, don't forget to rename this to something with `advanced_image` instead of `image_copyright`.

modules/custom/social_advanced_image/social_advanced_image.services.yml
```diff
@@ -0,0 +1,10 @@
...
+    class: Drupal\social_advanced_image\AdvancedImageAttributesManager
+    arguments: ['@module_handler', '@cache.discovery']
+
+  image_copyright.overrider:
```
bramtenhove: As part of the renaming effort, don't forget to rename this to something with `advanced_image` instead of `image_copyright`.

modules/custom/social_advanced_image/social_advanced_image.install
royharink: Yeah I know, it's on the todo of this PR.

jaapjan:
This still needs a drupal.org ticket. could you create one and remove the label `needs drupal.org ticket`?

ronaldtebrake:
This seems nice to me, what about the TODO list.. will these be planned soon? Can imagine the CKEditor being a follow-up but the rest needs to be solved before final review & merge.

royharink:
The testing of this has been postponed but it's also highly likely that we split up the story.

bramtenhove:
During the roadmap meeting we discussed that we need to see if we can add this feature in a different repository.

We should think about what should actually be in the distro to make it work well.

royharink:
> During the roadmap meeting we discussed that we need to see if we can add this feature in a different repository.
> 
> We should think about what should actually be in the distro to make it work well.

I've split the code and made the feature within a separate repository.

ronaldtebrake:
Will be dealt with by @royharink separately

PR #1530, opened 2019-10-02
Issue #3085183 by makishima: Remove superfluous functionality in the social_profile module that sets the default country for users

VolodymyrMartsynenko:
## Problem
When creating a new user, the country field in the profile has a default value that is equal to the default country of the site, but should be set to "None" value.

## Solution
Remove superfluous functionality in the social_profile module that sets the default country for users from the default country on the site.

## Issue tracker
https://www.drupal.org/project/social/issues/3085183
https://getopensocial.atlassian.net/browse/YANG-1451

## How to test
- [ ] Create new user
- [ ] Go to user profile, check the Country field, make sure the value is set to  "- None -"

### DIFF
--- a/modules/social_features/social_profile/social_profile.module
+++ b/modules/social_features/social_profile/social_profile.module
@@ -425,14 +425,6 @@ function social_profile_user_insert(UserInterface $account) {
         'uid' => $account->id(),
       ];
 
-      // Get all field instances for the profile entity and check if the address
-      // field exists.
-      $instances = \Drupal::service('entity_field.manager')->getFieldDefinitions('profile', $profile_type->id());
-      if (array_key_exists('field_profile_address', $instances)) {
-        // Set the users default country to the site default country.
-        $default_values['field_profile_address'][0]['country_code'] = \Drupal::config('system.date')->get('country.default');
-      }
-
       // Create a profile.
       $profile = Profile::create($default_values);
       $profile->save();


ribel:
@bramtenhove, @ronaldtebrake, @jaapjan 
I see this functionality was added in PR 528, but our client reported it as a bug since even if the site could be located in one country users can be international. And by default, they all receive one country. Client ask to set to none by default, and we agree.
What do you think?

bramtenhove:
I think just removing it is a bad idea, also why wouldn't you do it for the content as well then. There are a lot of sites where it is also very logical to have it set to the default site country, basically any community that is not international.

And next to that removing it might raise the opposite bug for the distro again.

Wouldn't it make more sense to make it configurable to set the default country to the site country? By default we should keep it the site country as it is now imho.

_Related to this we should in any way keep in mind that we're building a product for *a lot* of use cases. One client may want this, another may want this. The middle ground (and only one we can support in the long run) is usually in saying *no* or making it configurable in the distro._

ribel:
@bramtenhove Thanks for the clarification and input! 
Then let's make it configurable. 

I propose to add new checkbox under Default country setting, like this:

image

I think it has the highest prio for profiles. Not sure if we also need to add checkboxes for events and groups.

Kingdutch:
@ribel I'm not sure I understand the problem entirely.

Why does setting the "Default Country" to `- None -` not achieve the desired result? The checkbox to me reads like `If I specify a default country, then I specify a default country`.

ribel:
@Kingdutch I thought that setting Default country still might be useful in cases of some Drupal Core behaviours or other contrib modules. Like for example, this setting is used in `html/core/lib/Drupal/Core/Datetime/DateFormatter.php`.
But the setting is a default country for all users profiles may be questionable, as in our case.

ribel:
@Kingdutch You actually right. We can just set the default country as none.

I am closing this PR as not needed atm.

PR #1532, opened 2019-10-03
Issue #3085504: Improve performance of ConfigOverrides in Social Lazy Load

jaapjan:
## Problem
The `SocialLazyLoadingImageDisplayOverride` and `SocialLazyLoadingTextFormatOverride` are checking for existing config. This is adding an extra few 100ms to a page request as this code is executed via `lazy_is_enabled()` in `lazy.module`.

## Solution
Let's remove the configfactory injects inside these overrides and simplify the code a bit.

blackfire-comparison

## Issue tracker
https://www.drupal.org/project/social/issues/3085504

## How to test
- [ ] Profile `lazy_is_enabled`
- [ ] Check the difference
- [ ] Make sure lazy loading is still working as expected.

## Release notes
The configuration overrides in Social Lazy Load are optimised by removing existing configuration checks.

### DIFF
--- a/modules/custom/social_lazy_loading/social_lazy_loading_images/src/SocialLazyLoadingImageDisplayOverride.php
+++ b/modules/custom/social_lazy_loading/social_lazy_loading_images/src/SocialLazyLoadingImageDisplayOverride.php
@@ -3,7 +3,6 @@
 namespace Drupal\social_lazy_loading_images;
 
 use Drupal\Core\Cache\CacheableMetadata;
-use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Config\ConfigFactoryOverrideInterface;
 use Drupal\Core\Config\StorageInterface;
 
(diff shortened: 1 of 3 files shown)


jaapjan:
We need to test this thoroughly, I have some white screens.. not sure if it's related to the probing or this code. My guess is the probing!

modules/custom/social_lazy_loading/social_lazy_loading_images/src/SocialLazyLoadingImageDisplayOverride.php
```diff
@@ -82,40 +64,15 @@ public function loadOverrides($names) {
     ];
 
     foreach ($config_fields as $config_name => $field_name) {
-      $this->addImageOverride($config_name, $field_name, $overrides);
+      if (in_array($config_name, $names)) {
```
jaapjan: @ronaldtebrake I've added this extra check as well. So we only override the right config 🙈 Please verify if this is implemented correctly 🤞

ronaldtebrake:
I have tested this pretty thorougly this morning and couldn't see any problem without probing, so i'm fairly certain it's the probing doing this.

ronaldtebrake:
Also in 6.x :) !

PR #1533, opened 2019-10-04
Added conditional check to account object to be null.

navneet0693:
## Problem
In social_profile_privacy_form_alter(), $account is not checked for being NULL when the parameter user is not found in URL. This blocks any user to edit profiles.

## Solution
Add a null check.

## Issue tracker
https://www.drupal.org/project/social/issues/3085803

## How to test
- [x] Enabled social_profile_privacy module.
- [x] Login in as SM or admin
- [x] Go to profile listing
- [x] Try to edit any profile.

## Release notes
Introduces an empty check in social_profile_privacy_form_alter in social_profile_privacy module.

## Change Record
N.A


ronaldtebrake:
Also in 6.4

PR #1535, opened 2019-10-07
Issue #3086229: Download count: Invalidate file cache tags only when needed

jaapjan:
## Problem
In download count module we implement hook_ENTITY_TYPE_access for files which adds a file download count for every downloaded file. However this code is also triggered for comment attachments. Depending on the settings or user permissions some files are excluded and the download count won't be adjusted. The file cache tags are invalidated for every file download.

## Solution
Let's do the invalidating of the cache tags only when the file download count is actually changed.

## Issue tracker
https://www.drupal.org/project/social/issues/3086229

## How to test
- [ ] Check code changes and logic explanation
- [ ] Enable social_download_count and dependencies
- [ ] Login as LU
- [ ] Create topic with multiple file attachments.
- [ ] Add a comment (with attachment) image and make sure to check with breakpoints when the code is actually executed (for all of them)
- [ ] Verify the settings (Flood + Exclude extensions) and make sure the feature is working as expected.

## Release notes
The download count should only invalidate the file cache for a file download if the file download usage is updated.

### DIFF
--- a/modules/custom/download_count/download_count.module
+++ b/modules/custom/download_count/download_count.module
@@ -29,15 +29,11 @@ function download_count_help($route_name, RouteMatchInterface $route_match) {
  * Implements hook_ENTITY_TYPE_access().
  */
 function download_count_file_access(EntityInterface $entity, $operation, AccountInterface $account) {
-  if ($operation == 'download') {
+  if ($operation === 'download') {
     $flood = \Drupal::flood();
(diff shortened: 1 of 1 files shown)


jaapjan:
@albertoalaejos I think this is your browser cache kicking in sometimes.. I've tried multiple times and it works. but sometimes when you click super fast it is just browser caching serving you the .pdf directly.

PR #1536, opened 2019-10-07
by robertragas: replace notification email templates div with tables …

robertragas:
…to avoid illegal translation characters and make emails more compatible with old browsers

## Problem
One of the new features of Open Social is getting previews in the email notifications. On the default english version they are working correct, but when having multiple languages you need to translate them.

The problem here is that the template contains <div> tags which are not allowed in translations.

In our case in addition have some trouble with the fallback not being picked up, resulting in empty emails.

## Solution
Replace the <div> with table,td,tr tags as they are also more supported in emails.
The styling has been done on classes so it does not matter if you use other tags.

## Issue tracker
https://www.drupal.org/project/social/issues/3086239

## How to test
- [ ] Try to translate the email notification preview, it does not work. 
- [ ] Apply patch and try again, also notice that the look of it should not differ

## Release notes
We have fixed a translation issue with the email notification previews. At first it would only be possible to have them in English, but now supports multilanguage.


jaapjan:
@robertragas could you update the branch since there are conflicts now?

robertragas:
> LGTM

TYFM, HVJ WJNMK

robertragas:
> @robertragas could you update the branch since there are conflicts now?

Merge conflict solved.

jaapjan:
Will be in 6.x as well. @robertragas will cherry-pick

jaapjan:
see https://github.com/goalgorilla/open_social/commit/a13a6d32947883c14d56f83e5f23fe40e9e52438

PR #1537, opened 2019-10-07
#3091551 by ronaldtebrake: Implement Ajax for adding and deleting comments

ronaldtebrake:
## Problem
As a LU we want to place comments without losing the position of the screen and without a page refresh. This goes for the stream but also on nodes, for adding and deleting.

## Solution
Use Ajax to place, edit, delete and unpublish comments.

## Issue tracker
https://www.drupal.org/project/social/issues/3091551

## How to test
- [ ] enable social_ajax_comments
- [ ] See that as a LU you can now place / delete a comment in the stream
- [ ] See that it doesn't take a full page refresh
- [ ] See that it nicely adds a new number towards the count if more than 2 comments on a post
- [ ] See that deletion works and also doesn't redirect and takes into account the amount of comments on a post
- [ ] Similar check out the node page, see that also reply works there with Ajax

## Release notes
Adding or deleting a comment is now being done with the help of Ajax. It won't take a page load anymore for you to add or delete a comment. This ensures you don't leave the position and context of the comment you just created.


themes/socialbase/templates/comment/field--comment.html.twig
```diff
@@ -30,7 +30,8 @@
 {# extra condition to make sure the card is not shown when there are no comments and comments are closed #}
 {% if comments or comment_form %}
 
-  <section{{ attributes.setAttribute('id', 'section-comments') }}>
+  <a id="section-comments"></a>
+  <section{{ attributes }}>
```
Kingdutch: This could negatively affect styling of our own code or extending projects. It's also not needed if this is used to focus the user's viewport. `id` attributes on any element can be used for this :)

modules/custom/social_ajax_comments/social_ajax_comments.module
```diff
@@ -0,0 +1,138 @@
...
+
+use Drupal\ajax_comments\Utility;
+use Drupal\comment\CommentInterface;
+use Drupal\Component\Utility\Html;
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Url;
+use Drupal\node\Entity\Node;
+use Drupal\social_post\Entity\Post;
+
+function social_ajax_comments_entity_display_build_alter(&$build, $context) {
```
Kingdutch: This needs an `implements hook` comment.

modules/custom/social_ajax_comments/social_ajax_comments.module
```diff
@@ -0,0 +1,138 @@
...
+ */
+function social_ajax_comments_comment_links_alter(array &$links, CommentInterface $entity, array &$context) {
+  $bundle = $entity->bundle();
+  $commented_entity = $entity->getCommentedEntity();
+  $account = \Drupal::currentUser();
+
+  if ($entity->isPublished() && $account->hasPermission('administer comments')) {
+    // For now we don't support edit ajax actions nor unpublishing.
+    $links['comment']['#links']['comment-edit']['attributes']['class'] = [];
+    $links['comment']['#links']['comment-edit']['url'] = Url::fromRoute('entity.comment.edit_form', ['comment' => $entity->id()]);
```
Kingdutch: Did you test whether simply adding the `Url::fromRoute()->toRenderable` already included the access checks that you need? Then perhaps the `$entity->isPublished` and `hasPermission` check are taken care of for you :)

See the implementation of the "Add content" links for an example.

modules/social_features/social_post/social_post.module
```diff
@@ -78,16 +78,18 @@ function social_post_form_comment_post_comment_form_alter(&$form, FormStateInter
...
-  // Store in a hidden field.
-  $form['redirect_after_save'] = [
-    '#type' => 'hidden',
-    '#title' => t('Redir'),
-    '#default_value' => $uri,
-  ];
-
-  // Submit function to retrieve the action uri and redirect to it.
-  $form['actions']['submit']['#submit'][] = '_social_post_comment_post_comment_form_submit';
+  if (!\Drupal::moduleHandler()->moduleExists('ajax_comments')) {
```
Kingdutch: Lets add a comment here on why this is and move the existing comment of what is happening into the if statement :)

modules/custom/social_ajax_comments/social_ajax_comments.module
ronaldtebrake: Yes, unfortunately we need the custom checks to override what Ajax comments does for us. Mainly because of UX flaws in the edit form that opens up. (Out of scope for now) Will add this in the code doc for explanation purposes. Good question.

themes/socialbase/templates/comment/field--comment.html.twig
ronaldtebrake: We need this to allow this section to have it's own id based on the drupal render array. That's how Ajax comments works and knows what section to replace after a comment is being made. So the entire comment field is updates including the counts etc.

Adding

ronaldtebrake:
Closing this in favor of a new fresh rebase: https://github.com/goalgorilla/open_social/pull/1665

Kingdutch:

PR #1540, opened 2019-10-08
DS-6910 by robertragas: remove styling of <em> so its default italic,…

robertragas:
… and change heading to add <strong>

## Problem
When using the WYSIWYG and using italic. The text is not made italic in the e-mail while it is visible in the WYSIWYG. 

## Solution
Remove the <em> styling so browsers will give the default styling which is italic.
Also replace all the message templates <em> with <strong>

Needed to add the <strong> tag in the t() function to be able to get this markup.

## Issue tracker

## How to test
- [ ] Cache clear
- [ ] Do a bulk email send and add italic letters
- [ ] See it works, and the Hi Displayname is still bold

## Release notes
We have made a fix for sending out bulk emails with the ckeditor where the italic style would not apply.

## Change Record
We have altered the translation of the heading of the email notifications from `Hi %profilename` to `Hi <strong>@profilename</strong>`. This requires you to translate this header again.


jaapjan:
@Kingdutch what happens if we merge this to 7.x only, but not to 6.x? Will weblate still handle translations well? 🤞

PR #1541, opened 2019-10-08
Issue #3081723 by agami4: Add anchor navigation for the secondary navigation

agami4:
## Problem
On mobile, when navigating to another tab of a profile or a group, users still see the hero region of the profile and group first. This makes the navigation inefficient.

## Solution
When going to another tab, the screen should scroll automatically to the main content of this tab.

## Issue tracker
https://www.drupal.org/project/social/issues/3081723

## How to test
- [ ] If you want to have anchor navigation for the secondary navigation you should add more information for a page(for example Profile page)
- [ ] Go to this profile page on the mobile and you can see the  secondary navigation
- [ ] If you are on the page with the first active element, scrolling to the secondary navigation doest not work
- [ ] Please click to the `Event` tab and if the height of the window(x2) will be less than the height of the page, you can see scrolling to the secondary navigation

## Release notes
Added anchor scrolling to the secondary navigation on the mobile devices

## Change Record


Kingdutch:
@agami4 Do you see a way to fix the two issues I raised? Otherwise we may want to close this and revisit it in the future.

ronaldtebrake:
I think we will close this for now, will revisit this in the future with React. Feel free to re-open if the impact is very high now, but it seems the implication we still have and with the other priorities this is low prio.

PR #1546, opened 2019-10-10
Issue #3087026 by mdeny: Social Group Welcome Message

denysyukm:
## Problem
Private message when successfully joining the group

## Solution
Implement Social Group Welcome Message module

## Issue tracker
https://www.drupal.org/project/social/issues/3087026
https://getopensocial.atlassian.net/browse/YANG-1463

## How to test
- [ ] log in as group manager
- [ ] go the the group you manage
- [ ] go to the edit form of the group
- [ ] set "Send a private message people joined"
- [ ] write Welcome message
- [ ] join this group in any way as another user
- [ ] check private messages

## Release notes
Implements possibility to send private message when successfully joining the group

### DIFF
--- a/modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
+++ b/modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
@@ -1,0 +1,151 @@
+<?php
+
+/**
+ * @file
+ * Primary module hooks for social_group_welcome_message module.
+ */
+
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Field\BaseFieldDefinition;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\group\Entity\GroupContentInterface;
+use Drupal\private_message\Entity\PrivateMessage;
+
+/**
+ * Implements hook_entity_base_field_info().
+ */
+function social_group_welcome_message_entity_base_field_info(EntityTypeInterface $entity_type) {
+  $fields = [];
+
+  // Add a 'Welcome message' base fields to all group types.
+  if ($entity_type->id() === 'group') {
+    $fields['private_message_send'] = BaseFieldDefinition::create('boolean')
+      ->setLabel(t('Send a private message people joined'))
+      ->setRevisionable(TRUE)
+      ->setTranslatable(TRUE)
+      ->setDefaultValue(FALSE)
+      ->setDisplayOptions('form', [
+        'type' => 'boolean_checkbox',
+        'settings' => [
+          'display_label' => TRUE,
+        ],
+        'weight' => 99,
+      ])
+      ->setDisplayConfigurable('form', TRUE);
+
+    $fields['private_message_body'] = BaseFieldDefinition::create('text_long')
+      ->setLabel(t('Message'))
+      ->setRevisionable(TRUE)
+      ->setTranslatable(TRUE)
+      ->setDisplayOptions('form', [
+        'type' => 'text_textfield',
+        'weight' => 100,
+      ])
+      ->setDisplayConfigurable('form', TRUE);
+  }
+
+  return $fields;
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function social_group_welcome_message_form_alter(array &$form, FormStateInterface $form_state, $form_id) {
+  $social_group_types = [
+    'open_group',
+    'closed_group',
+    'public_group',
+  ];
+
+  \Drupal::moduleHandler()->alter('social_group_types', $social_group_types);
+
+  $group_forms = [];
+  foreach ($social_group_types as $social_group_type) {
+    $group_forms[] = "group_{$social_group_type}_add_form";
+    $group_forms[] = "group_{$social_group_type}_edit_form";
+  }
+
+  if (
+    in_array($form_id, $group_forms) &&
+    isset($form['private_message_body']) &&
+    isset($form['private_message_send'])
+  ) {
+    $form['private_message_body']['#states'] = [
+      'visible' => [
+        ':input[name="private_message_send[value]"]' => ['checked' => TRUE],
+      ],
+    ];
+  }
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_insert().
+ */
+function social_group_welcome_message_group_content_insert(GroupContentInterface $group_content) {
+  $type = $group_content->getContentPlugin()->getPluginId();
+  $group = $group_content->getGroup();
+
+  // If new membership is created.
+  if (
+    $type === 'group_membership' &&
+    !$group->get('private_message_send')->isEmpty() &&
+    $group->get('private_message_send')->value == 1 &&
+    !$group->get('private_message_body')->isEmpty()
+  ) {
+    $g_type_id = $group->getGroupType()->id();
+
+    // Get group admins entities.
+    $g_admins = $group->getMembers([$g_type_id . '-group_admin']);
+    $g_admins_users = [];
+    /** @var \Drupal\group\GroupMembership $member */
+    foreach ($g_admins as $key => $member) {
+      $g_admins_users[$key] = $member->getUser();
+    }
+
+    // Get group managers entities.
+    $g_managers = $group->getMembers([$g_type_id . '-group_manager']);
+    $g_managers_users = [];
+    /** @var \Drupal\group\GroupMembership $member */
+    foreach ($g_managers as $key => $member) {
+      $g_managers_users[$key] = $member->getUser();
+    }
+
+    // Set sender (owner) of the pm.
+    if (count($g_admins_users) === 1) {
+      list($owner) = $g_admins_users;
+    }
+    elseif (count($g_managers_users) > 0) {
+      list($owner) = $g_managers_users;
+    }
+    else {
+      $owner = $group->getOwner();
+    }
+
+    // Merge all managers of the group to managers array.
+    $managers = array_merge($g_admins_users, $g_managers_users, [$group->getOwner()]);
+
+    // Load joined user.
+    $user = $group_content->getEntity();
+
+    // Add the joined user to the already existing managers array.
+    $managers[] = $user;
+
+    /** @var \Drupal\private_message\Service\PrivateMessageServiceInterface $private_message_service */
+    $private_message_service = \Drupal::service('private_message.service');
+
+    // Create a pm thread between these users.
+    $thread = $private_message_service->getThreadForMembers($managers);
+
+    // Get body of pm.
+    $private_message_body = $group->get('private_message_body')->first()->getValue();
+
+    // Create a single message with the pm body.
+    $private_message = PrivateMessage::create([
+      'owner' => $owner,
+      'message' => $private_message_body,
+    ]);
+
+    $private_message->save();
+    $thread->addMessage($private_message)->save();
+  }
+}
--- a/modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
+++ b/modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
@@ -95,7 +95,7 @@ function social_group_welcome_message_group_content_insert(GroupContentInterface
     $g_type_id = $group->getGroupType()->id();
 
     // Get group admins entities.
-    $g_admins = $group->getMembers([$g_type_id . '-group_admin']);
+    $g_admins = $group->getMembers($g_type_id . '-group_admin');
     $g_admins_users = [];
     /** @var \Drupal\group\GroupMembership $member */
     foreach ($g_admins as $key => $member) {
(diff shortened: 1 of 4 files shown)


modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  return $fields;
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function social_group_welcome_message_form_alter(array &$form, FormStateInterface $form_state, $form_id) {
+  $social_group_types = [
```
bramtenhove: How is this going to work for flexible groups and/or secret groups?

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -0,0 +1,151 @@
...
+  if (
+    $type === 'group_membership' &&
+    !$group->get('private_message_send')->isEmpty() &&
+    $group->get('private_message_send')->value == 1 &&
+    !$group->get('private_message_body')->isEmpty()
+  ) {
+    $g_type_id = $group->getGroupType()->id();
+
+    // Get group admins entities.
+    $g_admins = $group->getMembers([$g_type_id . '-group_admin']);
```
OH17L:
```suggestion
    $g_admins = $group->getMembers($g_type_id . '-group_admin');
```

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -0,0 +1,151 @@
...
+    $g_admins_users = [];
+    /** @var \Drupal\group\GroupMembership $member */
+    foreach ($g_admins as $key => $member) {
+      $g_admins_users[$key] = $member->getUser();
+    }
+
+    // Get group managers entities.
+    $g_managers = $group->getMembers([$g_type_id . '-group_manager']);
```
OH17L:
```suggestion
    $g_managers = $group->getMembers($g_type_id . '-group_manager');
```

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -0,0 +1,151 @@
...
+    $private_message_service = \Drupal::service('private_message.service');
+
+    // Create a pm thread between these users.
+    $thread = $private_message_service->getThreadForMembers($managers);
+
+    // Get body of pm.
+    $private_message_body = $group->get('private_message_body')->first()->getValue();
+
+    // Create a single message with the pm body.
+    $private_message = PrivateMessage::create([
```
OH17L:
```suggestion
    $private_message = \Drupal::entityTypeManager()->getStorage('private_message')->create([
```

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.services.yml
```diff
@@ -0,0 +1,8 @@
+services:
+
+  # Defines service for overriding configs
+  social_group_welcome_message.overrider:
+    class: Drupal\social_group_welcome_message\SocialGroupWelcomeMessageConfigOverride
+    arguments: ['@config.factory', '@module_handler']
```
OH17L: Add the **string_translation** service if it is possible.

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
denysyukm: Changed

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
denysyukm: Changed

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
denysyukm: Changed

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.services.yml
denysyukm: Added

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
bramtenhove: I see, thanks @OH17L.

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -0,0 +1,150 @@
...
+
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Field\BaseFieldDefinition;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\group\Entity\GroupContentInterface;
+
+/**
+ * Implements hook_entity_base_field_info().
+ */
+function social_group_welcome_message_entity_base_field_info(EntityTypeInterface $entity_type) {
```
Kingdutch: Hey @OH17L ,

Is there a reason we're adding this as a base field here? Doesn't it make more sense to add this as a Field API field instead? That'll make it easier to modify per bundle if this is ever needed (e.g. for other group like types such as courses).

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -0,0 +1,150 @@
...
+      ->setRevisionable(TRUE)
+      ->setTranslatable(TRUE)
+      ->setDisplayOptions('form', [
+        'type' => 'text_textfield',
+        'weight' => 100,
+      ])
+      ->setDisplayConfigurable('form', TRUE);
+  }
+
+  return $fields;
```
Kingdutch: Do these need to be two dependant fields? I think we can get away with a single welcome message field here that can be nullable.

If no text is filled in then no message will be sent. If you want the user experience of having a checkbox with a hideable text field then this can be done in a Field Widget instead but does not need to be stored. That also simplifies your form alter logic below.

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -0,0 +1,150 @@
...
+ * Implements hook_form_FORM_ID_alter().
+ */
+function social_group_welcome_message_form_alter(array &$form, FormStateInterface $form_state, $form_id) {
+  $social_group_types = [
+    'open_group',
+    'closed_group',
+    'public_group',
+  ];
+
+  \Drupal::moduleHandler()->alter('social_group_types', $social_group_types);
```
Kingdutch: Do we have a default implementation for this already? It feels a bit weird to define the default group types and alter them.

We should probably just have a service that caches this per request instead.

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -0,0 +1,150 @@
...
+    }
+    elseif (count($g_managers_users) > 0) {
+      list($owner) = $g_managers_users;
+    }
+    else {
+      $owner = $group->getOwner();
+    }
+
+    // Merge all managers of the group to managers array.
+    $managers = array_merge($g_admins_users, $g_managers_users, [$group->getOwner()]);
```
Kingdutch: Rename `$managers` to `$recipients` instead since the user joining is added later (not necessarily a manager)

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -0,0 +1,150 @@
...
+    $thread = $private_message_service->getThreadForMembers($managers);
+
+    // Get body of pm.
+    $private_message_body = $group->get('private_message_body')->first()->getValue();
+
+    // Create a single message with the pm body.
+    $private_message = \Drupal::entityTypeManager()->getStorage('private_message')->create([
+      'owner' => $owner,
+      'message' => $private_message_body,
+    ]);
```
Kingdutch: Since we're not using the text format here, perhaps we should not create a field with a text format but use the `string_long` field type instead (this will remove the WYSIWYG though, but we'll now have bugs here either way).

bramtenhove:
@ribel @denysyukm can you update the PR with fixes for the raised issues?

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
denysyukm: Yeah, the reason to add this as a base field is that we're implementing welcome message to all group types and here (in hook_base_field_info) I'm able to do this with just one condition:
`if ($entity_type->id() === 'group')` (line 20 here)

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
denysyukm: In my opinion better will be to keep these fields as they are because I think it's not good to check text field if it's empty or not and also GM can prepare body of welcome message, exclude checkbox and just store text without implementing base functionality of welcoming message.

But if you really want to adjust this then we can

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
denysyukm: Good point but it should be a general fix of approach getting all social group types because a lot of modules use the the same approach as I implemented here. Thats why I think if we agree to improve this it shouldn't be a part of this story.

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
denysyukm: Will adjust

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
denysyukm: Very good point. If I'm not mistaken private messages support some text format (maybe `full html`, maybe just `basic html`). So we really should check this and adjust if needed (I mean if private messages support `full html` text then we can leave this as it is, agree?)

denysyukm:
@bramtenhove , @Kingdutch , @ribel, I created the story with changes which should be done to make this PR ready for merging to Distro

https://getopensocial.atlassian.net/browse/YANG-2052

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
Kingdutch: I discussed this with Ronald and we agree to keep it like this.

This is not because it's technically easier than adding fields to multiple bundles. The thought behind doing it this way is that the welcome message is something that fits the model of the group module.

This means that if something builds on the group entity that is not supposed to have this because it's not a group (for example a course) then that functionality should be implemented differently.

👍

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
Kingdutch: I think we should be careful here. By default on sites I think only site managers (and maybe content managers) have access to `full html`. However, normal users can also make groups. So if this is a group setting, then this should probably be adjusted also to what the user creating the welcome message has access to.

However, I suspect that private messages only allows `basic html` or `plain text` (since any user can create a private message), so we should use that, because that'll also be used for rendering (i.e. allowing SM to use `full text` would not render correctly).

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
Kingdutch: Given the time constraints, adding this as a follow-up on Drupal.org is fine. I think in the future it would be acceptable to include such improvements in a pull request. Following the girl-/boy-scouts motto of "Leave the code better than you found it" :D 

Of course this always depends on how big the required changes are ☺️

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
Kingdutch: Okay

denysyukm:
@ronaldtebrake , @bramtenhove , @Kingdutch , I created[ new PR](https://github.com/goalgorilla/open_social/pull/1673) of these 4 commits to do merge into 8.x-8.x. Also, I added some changes there in [last commit](https://github.com/goalgorilla/open_social/pull/1673/commits/5e1bada6c839cd8d9f305ee9abae4fbd61d407aa): "Renamed $managers to $recipients where it needs. Limited a text format of Welcome message body field to basic_html"

PR #1547, opened 2019-10-10
Issue #3087083: One to many notification activities

navneet0693:
## Problem
Currently, multiple activities are created for multiple users even for a single activity. This is an overload on the database.

## Solution
 Allowing an activity to be attached to many users instead of just one.

## Issue tracker
https://www.drupal.org/project/social/issues/3087083

## How to test
- [ ] drush -y fra --bundle=social
- [ ] drush updb -y
- [ ] Create notifications for a group by posting a comment or topic which will trigger notification to multiple users.
- [ ] Log in as at least 2-3 users in different browser at the same time.
- [ ] Read notifications from users.
- [ ] Every users notification should remain intact.
- [ ] If there is a notification in the group, reading by one user/member of the group should not mark notification as read for other users.

## Release notes
- Only one activity bundle will be created for each notification trigger targetting multiple instead of multiple activities for multiple users.
- Notification will no longer be marked as read.
- **readNotificationCallback** and **markEntityAsRead** are deprecated.
- **field_activity_recipient_user** has been changed from one to unlimited references.

## Change Record
- **field_activity_recipient_user** has cardinality -1.
- Notifications status will be stored in **activity_notification_status** table.
- **Mark as seen** functionality will have a new behaviour. **getNotificationListCallback()** will now mark the notification as seen once user clicks the bell icon.
- The field **field_activity_status** will no longer be used to denote the status of any activity.
- Activities will have no status as **ACTIVITY_STATUS_READ** 
- **Mark as read** status for notification has been completely removed.
- **ActivityNotifications** services are updated to provide more robust functions.


bramtenhove:
@navneet0693 can you check what is going on with Travis?

navneet0693:
> @navneet0693 can you check what is going on with Travis?

Yes. I pushed a commit to fix it.

modules/custom/activity_creator/activity_creator.install
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Migrate all the activity information to new table.
+ * This is necessary as we have changed the logic of reading notifications
+ * and marking them as seen. So, we have migrate the existing activity entries
+ * to new table so as to avoid any missing notifications by users.
+ *
+ * @throws \Exception
+ */
+function activity_creator_update_8002() {
```
bramtenhove: @navneet0693 this is a nice one to do in a post update hook.

Rule of thumb is that we run data migrations in post update hooks instead of normal update hooks.

ronaldtebrake:
Hey guys, 
Heard you wanted to get this in 🚀 
It seems fairly heavy as a "last minute" contribution so lets make sure the quality is tested very very thorough.
Some questions that pop in to my mind already:
- Should users disable the cron when updating from 6.x? 
- Should the hook_update be done in batches? 
- Will the feature revert handle the field cardinality changes when there is content inside already? 

I'm mainly worried about the update path on this one. Just want to make sure we don't rush this in.

bramtenhove:
>Will the feature revert handle the field cardinality changes when there is content inside already?
Yes, this is not a problem. We're extending from 1 to unlimited, so we'll not get issues here.

modules/custom/activity_creator/activity_creator.install
navneet0693: Ok, I will convert it to batch and move to post_update_hook.

navneet0693:
> Hey guys,
> Heard you wanted to get this in 🚀
> It seems fairly heavy as a "last minute" contribution so lets make sure the quality is tested very very thorough.
> Some questions that pop in to my mind already:
> 
> * Should users disable the cron when updating from 6.x?
> * Should the hook_update be done in batches?
> * Will the feature revert handle the field cardinality changes when there is content inside already?
> 
> I'm mainly worried about the update path on this one. Just want to make sure we don't rush this in.

1. Yes, I will recommend implementing an upgrade after the queue is clean. But this shouldn't have an adverse effect on existing items.
2. I will add the batch operation.
3. Yes, feature revert would handle changing the cardinality from Limited to Unlimited, but the other way round is not possible.

navneet0693:
> @navneet0693 can you check out my and Ronald's comments?

Sure.

navneet0693:
> @navneet0693 did you consider what to do with the `field_activity_status` Activity entity field?
> 
> I think that's a field we are no longer using anywhere and can get rid of right?
> 
> If so, we probably need to do this in a next version. Can you think about how to do this and create follow-ups for this where needed (also on d.org)?
> _We also need to consider in this case what happens when a user does two updates at once._

@bramtenhove Added: https://www.drupal.org/project/social/issues/3087398

modules/custom/activity_creator/activity_creator.post_update.php
```diff
@@ -0,0 +1,71 @@
...
+
+  // Insert the information in activity_notification_status table.
+  foreach ($results as $result) {
+    \Drupal::database()->insert('activity_notification_status')
+      ->fields([
+        'aid',
+        'uid',
+        'status',
+      ], [
+        $result->{'entity_id'},
```
bramtenhove: @navneet0693, wouldn't you be able to just access it with `$result->entity_id`? Same for the other 2 lines.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -41,11 +66,14 @@ public function getNotifications(AccountInterface $account, array $status = [ACT
...
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
-  public function getNotificationsActivities(AccountInterface $account, array $status = [ACTIVITY_STATUS_RECEIVED]) {
+  public function getNotificationsActivities(AccountInterface $account, array $status = [ACTIVITY_STATUS_RECEIVED]): array {
     $ids = $this->getNotificationIds($account, $status);
 
-    return entity_load_multiple('activity', $ids);
+    return \Drupal::entityTypeManager()->getStorage('activity')->loadMultiple($ids);
```
bramtenhove: This still needs dependency injection.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -100,38 +125,36 @@ public function markEntityNotificationsAsRead(AccountInterface $account, EntityB
    * @param \Drupal\Core\Entity\EntityBase $entity
    *   Entity object.
    *
-   * @throws \Drupal\Core\Entity\EntityStorageException
+   * @deprecated in opensocial:8.x-7.0 and is removed from opensocial:8.x-7.1. Use
```
bramtenhove: Are we sure this is removed in 8.x-7.1? Do we have another open PR to do this or is it scheduled?

modules/custom/activity_send/src/Plugin/SendActivityDestinationBase.php
```diff
@@ -75,6 +76,28 @@ public static function getSendTargetUser($activity) {
...
+   * @return \Drupal\Core\Entity\EntityInterface[]
+   *   Returns an array of target user accounts.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  public static function getSendTargetUsers(Activity $activity) {
+    $targets = [];
+    if (isset($activity->field_activity_recipient_user) && !empty($activity->field_activity_recipient_user)) {
+      $targets = \Drupal::entityTypeManager()
```
bramtenhove: Needs dependency injection.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -56,18 +84,11 @@ public function getNotificationsActivities(AccountInterface $account, array $sta
...
-
+  public function markAllNotificationsAsSeen(AccountInterface $account): int {
     // Retrieve all the activities referring this entity for this account.
-    $ids = $this->getNotificationIds($account, [ACTIVITY_STATUS_RECEIVED]);
-
-    $activities = Activity::loadMultiple($ids);
-    foreach ($activities as $activity) {
-      $this->changeStatusOfActivity($activity, ACTIVITY_STATUS_SEEN);
-    }
+    $ids = $this->getNotificationIds($account);
```
bramtenhove: We should limit it by getting just the IDs that were `received` and not all the IDs ever. We'd otherwise be updating a ton of items each time.

modules/custom/activity_creator/src/ActivityNotifications.php
bramtenhove: There's another `@deprecated` in this PR, same goes for that I suppose.

If we say something like that we need to schedule the work as well. I couldn't really figure out if that was the case.

bramtenhove:
@navneet0693, I also see that, although I merged in `8.x-7.x` the Behat tests are still failing. Also something you can take a look at.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -79,16 +100,20 @@ public function markAllNotificationsAsSeen(AccountInterface $account) {
...
    *   Entity object.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
   public function markEntityNotificationsAsRead(AccountInterface $account, EntityBase $entity) {
 
     // Retrieve all the activities referring this entity for this account.
-    $ids = $this->getNotificationIds($account, [ACTIVITY_STATUS_RECEIVED, ACTIVITY_STATUS_SEEN], $entity);
+    $entity_query = $this->entityTypeManager->getStorage('activity')->getQuery();
```
bramtenhove: When I install the branch locally I get an immediate fatal error on this line. The `entityTypeManager` is a dependency but not injected. Something we forgot to do?

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -56,18 +84,11 @@ public function getNotificationsActivities(AccountInterface $account, array $sta
...
+  public function markAllNotificationsAsSeen(AccountInterface $account): int {
     // Retrieve all the activities referring this entity for this account.
-    $ids = $this->getNotificationIds($account, [ACTIVITY_STATUS_RECEIVED]);
-
-    $activities = Activity::loadMultiple($ids);
-    foreach ($activities as $activity) {
-      $this->changeStatusOfActivity($activity, ACTIVITY_STATUS_SEEN);
-    }
+    $ids = $this->getNotificationIds($account);
+    $this->changeStatusOfActivity($ids, ACTIVITY_STATUS_SEEN);
```
bramtenhove: We should check if we have any IDs to parse here.

If we have no IDs the `changeStatusOfActivity` function will break.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -79,16 +100,20 @@ public function markAllNotificationsAsSeen(AccountInterface $account) {
...
+    $entity_query->condition('field_activity_recipient_user', $account->id(), '=');
+    $entity_query->condition('field_activity_destinations', ['notifications'], 'IN');
+    $ids = $entity_query->execute();
 
-    $activities = Activity::loadMultiple($ids);
-    foreach ($activities as $activity) {
-      $this->changeStatusOfActivity($activity, ACTIVITY_STATUS_READ);
-    }
+    // Change the status of this entity.
+    $this->changeStatusOfActivity($ids, ACTIVITY_STATUS_READ);
```
bramtenhove: Same as above. We should check if the `$ids` are not empty.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -100,38 +125,36 @@ public function markEntityNotificationsAsRead(AccountInterface $account, EntityB
...
   public function markEntityAsRead(AccountInterface $account, EntityBase $entity) {
 
     // Retrieve all the activities referring this entity for this account.
     $ids = $this->getNotificationIds($account, [ACTIVITY_STATUS_RECEIVED, ACTIVITY_STATUS_SEEN], $entity);
-
-    $activities = Activity::loadMultiple($ids);
-    foreach ($activities as $activity) {
-      $this->changeStatusOfActivity($activity, ACTIVITY_STATUS_READ);
-    }
+    $this->changeStatusOfActivity($ids, ACTIVITY_STATUS_READ);
```
bramtenhove: Same as above. We should check if the `$ids` are not empty.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -100,38 +125,36 @@ public function markEntityNotificationsAsRead(AccountInterface $account, EntityB
...
-   *   SAVED_NEW or SAVED_UPDATED is returned depending on the operation
-   *   performed.
-   *
-   * @throws \Drupal\Core\Entity\EntityStorageException
    */
-  public function changeStatusOfActivity(Activity $activity, $status = ACTIVITY_STATUS_RECEIVED) {
-    $activity->set('field_activity_status', $status);
-
-    return $activity->save();
+  public function changeStatusOfActivity(array $ids, $status = ACTIVITY_STATUS_RECEIVED) {
```
bramtenhove: We may want to put this update in a try catch.

In the catch we could log our failed attempt.

bramtenhove: Additionally you may want to check for an empty array of `$ids` here.

modules/custom/activity_creator/activity_creator.post_update.php
navneet0693: @bramtenhove Yes, we can. I will revert.

modules/custom/activity_creator/src/ActivityNotifications.php
navneet0693: I will use entityManager() from ControllerBase class.

modules/custom/activity_creator/src/ActivityNotifications.php
navneet0693: It should have come from controllerbase abstracter class. I will replace with entityTypeManager() getter function.

modules/custom/activity_creator/src/ActivityNotifications.php
navneet0693: Agreed.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -100,38 +125,36 @@ public function markEntityNotificationsAsRead(AccountInterface $account, EntityB
...
+    $this->database->update('activity_notification_status')
+      ->fields([
+        'status' => $status,
+      ])
+      ->condition('aid', $ids, 'IN')
```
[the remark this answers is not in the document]
navneet0693: We are fetching `$ids` every time for a particular user account only, which is done in getNotificationIds() by adding a condition: https://github.com/goalgorilla/open_social/pull/1547/files#diff-4a8db74dc26a2a7c39078682842180daR181.

modules/custom/activity_creator/src/ActivityNotifications.php
navneet0693: Agreed.

modules/custom/activity_creator/src/ActivityNotifications.php
navneet0693: Also the removal-version should match standard: drupal:n.n.n or project:n.x-n.n

modules/custom/activity_creator/src/ActivityNotifications.php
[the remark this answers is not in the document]
bramtenhove: I understand your point, but these IDs are also used for other users. The whole point of this change is that they can be shared across multiple users :)

So when we change the status of one or multiple activities we need to do that for a particular user account only.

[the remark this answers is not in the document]
navneet0693: @bramtenhove I got your point now.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -82,38 +88,58 @@ public function getNotificationsActivities(AccountInterface $account, array $sta
...
-    $this->changeStatusOfActivity($ids, ACTIVITY_STATUS_SEEN);
+    if (!empty($ids = $this->getNotificationIds($account, [ACTIVITY_STATUS_RECEIVED]))) {
+      return $this->changeStatusOfActivity($ids, $account, ACTIVITY_STATUS_SEEN);
+    }
 
-    return 0;
+    // Log as notice.
+    $this->loggerFactory->get('activity_creator')->notice('There are no notification to be marked as seen for user @id', [
+      '@id' => $account->id(),
+    ]);
```
navneet0693: @bramtenhove I update the service to be completely independent as a service should literally be as it should just manipulate information or provide us with the information we asked for.

changeStatusOfActivity() will return TRUE or FALSE depending upon the execution of update query. 

In this case, markAllNotificationsAsSeen()  is returning whatever changeStatusOfActivity() returns but it also checks ids() for being empty. So, is it a good idea to log a notice and return FALSE or just return FALSE after if statement.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -82,38 +88,58 @@ public function getNotificationsActivities(AccountInterface $account, array $sta
...
-
-    // Retrieve all the activities referring this entity for this account.
-    $entity_query = $this->entityTypeManager->getStorage('activity')->getQuery();
-    $entity_query->condition('field_activity_recipient_user', $account->id(), '=');
-    $entity_query->condition('field_activity_destinations', ['notifications'], 'IN');
-    $ids = $entity_query->execute();
-
-    // Change the status of this entity.
-    $this->changeStatusOfActivity($ids, ACTIVITY_STATUS_READ);
+    try {
```
navneet0693: @bramtenhove Please check this try-catch block if it looks good to you?

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -125,36 +151,92 @@ public function markEntityNotificationsAsRead(AccountInterface $account, EntityB
...
+   * @return bool
+   *   TRUE or FALSE depending upon update status.
+   */
+  public function markAllNotificationAsRead(AccountInterface $account) {
+    // Retrieve all the activities referring this entity for this account.
+    if (!empty($ids = $this->getNotificationIds($account, [ACTIVITY_STATUS_RECEIVED, ACTIVITY_STATUS_SEEN]))) {
+      return $this->changeStatusOfActivity($ids, $account, ACTIVITY_STATUS_READ);
+    }
+
+    $this->loggerFactory->get('activity_creator')->notice('There are no notification to be marked as read for user @id', [
```
navneet0693: @bramtenhove I am wondering that logging after a negative response from a service's method is good idea or it should be done by party whose calling the service w.r.t to the response from service?

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -140,36 +213,63 @@ public function changeStatusOfActivity(Activity $activity, $status = ACTIVITY_ST
...
+      try {
+        $this->database->delete('activity_notification_status')
+          ->condition('aid', $activity_ids)
+          ->execute();
+      }
+      catch (\Exception $exception) {
+        // Something went wrong somewhere, so roll back now.
+        $txn->rollBack();
+        // Log the exception to watchdog.
+        $this->loggerFactory->get('default')->error($exception->getMessage());
```
bramtenhove: @navneet0693, all the Behat tests are failing. Looks like it's because of this line.

https://travis-ci.org/goalgorilla/open_social/builds/600691656

navneet0693: @bramtenhove I am on it.

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -40,56 +66,86 @@ public function getNotifications(AccountInterface $account, array $status = [ACT
...
+        $entity_query = $this->entityTypeManager()->getStorage('activity')->getQuery();
+        $entity_query->condition('field_activity_recipient_' . $entity_type, $entity_id, '=');
+        $ids = $entity_query->execute();
+        break;
 
-    // Retrieve all the activities referring this entity for this account.
-    $ids = $this->getNotificationIds($account, [ACTIVITY_STATUS_RECEIVED]);
+      case 'group_content':
+        /** @var \Drupal\group\Entity\GroupContent $group_content */
+        $group_content = $entity;
```
bramtenhove: This variable seems to be obsolete right?

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -40,56 +66,86 @@ public function getNotifications(AccountInterface $account, array $status = [ACT
...
+      case 'group':
+        $entity_query = $this->entityTypeManager()->getStorage('activity')->getQuery();
+        $entity_query->condition('field_activity_recipient_' . $entity_type, $entity_id, '=');
+        $ids = $entity_query->execute();
+        break;
 
-    // Retrieve all the activities referring this entity for this account.
-    $ids = $this->getNotificationIds($account, [ACTIVITY_STATUS_RECEIVED]);
+      case 'group_content':
+        /** @var \Drupal\group\Entity\GroupContent $group_content */
```
bramtenhove:
```suggestion
        /** @var \Drupal\group\Entity\GroupContent $entity */
```

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -40,56 +66,86 @@ public function getNotifications(AccountInterface $account, array $status = [ACT
...
+        $group_content = $entity;
+        $linked_entity = $group_content->getEntity();
```
bramtenhove:
```suggestion
        $linked_entity = $entity->getEntity();
```

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -40,56 +66,86 @@ public function getNotifications(AccountInterface $account, array $status = [ACT
...
+        $group = $group_content->getGroup();
```
bramtenhove:
```suggestion
        $group = $entity->getGroup();
```

bramtenhove:
I changed the base branch to `8.x-8.x`. Now there a bunch of conflicting files.

@navneet0693, can you take a look at it?

navneet0693:
> I changed the base branch to `8.x-8.x`. Now there a bunch of conflicting files.
> 
> @navneet0693, can you take a look at it?

But this PR is targetted for 7.1 milestone?

navneet0693:
> I changed the base branch to `8.x-8.x`. Now there a bunch of conflicting files.
> 
> @navneet0693, can you take a look at it?

I did it :-)

bramtenhove:
@navneet0693, Travis is failing, can you look into it?

Afterwards this can be merged in `8.x-8.x` and the merge cherry-picked into `8.x-7.x`.

navneet0693:
> @navneet0693, Travis is failing, can you look into it?
> 
> Afterwards this can be merged in `8.x-8.x` and the merge cherry-picked into `8.x-7.x`.

It's because of https://www.drupal.org/project/social/issues/3093583

modules/custom/activity_creator/src/ActivityNotifications.php
```diff
@@ -100,38 +156,55 @@ public function markEntityNotificationsAsRead(AccountInterface $account, EntityB
...
   public function markEntityAsRead(AccountInterface $account, EntityBase $entity) {
-
     // Retrieve all the activities referring this entity for this account.
     $ids = $this->getNotificationIds($account, [ACTIVITY_STATUS_RECEIVED, ACTIVITY_STATUS_SEEN], $entity);
-
-    $activities = Activity::loadMultiple($ids);
-    foreach ($activities as $activity) {
-      $this->changeStatusOfActivity($activity, ACTIVITY_STATUS_READ);
-    }
+    $this->changeStatusOfActivity($ids, $account, ACTIVITY_STATUS_READ);
```
ronaldtebrake: I thought we removed the `ACTIVITY_STATUS_READ`? 
If I look at the code in `modules/custom/activity_creator/activity_creator.module` that is.

Also we might have to look in to changing any current activities that have the status `ACTIVITY_STATUS_READ` if we fully remove this. Otherwise we might end up with data in the database that get lost in our flow.
Imagine having to write an update hook in the future and forgetting about the fact thare are still activities with this status.

navneet0693: Oh yes @ronaldtebrake I will remove it. For the latter, we are migrating all the notifications in a new table here: https://github.com/goalgorilla/open_social/pull/1547/commits/cc493a55517cf488775044cef697f745a752f761#diff-88b7e01399cd0601d630de05185566e3 , do you think this will solve the problem ?

ronaldtebrake: I don't think so, I think we will still have rogue data that uses this `ACTIVITY_STATUS_READ` value. I'd rather see this being updated to the other status.

ronaldtebrake:
After discussing with @navneet0693 we changed the milestone to 8.0, we're changing cardinality, also deprecating the ACTIVITY_STATUS_READ so it's better of in 8.x

modules/custom/activity_creator/src/ActivityNotifications.php
navneet0693: @ronaldtebrake This should solve it now: https://github.com/goalgorilla/open_social/pull/1547/commits/cb2ecb27a085da7c3eb6a54be6e591b99dd18f01

bramtenhove:
@navneet0693, cc @ronaldtebrake, this works and can be merged when Travis says OK.

I understood Travis is failing because of a patch, is that already fixed and can we possibly merge that in this branch?

navneet0693:
> @navneet0693, cc @ronaldtebrake, this works and can be merged when Travis says OK.
> 
> I understood Travis is failing because of a patch, is that already fixed and can we possibly merge that in this branch?

@bramtenhove Unfortunately https://www.drupal.org/project/social/issues/3093583 is not fixed yet.

navneet0693:
As https://github.com/goalgorilla/open_social/pull/1606 has been merged, restarting the travis builds again.

navneet0693:
The build is failing as explained by @ronaldtebrake here: https://github.com/goalgorilla/open_social/pull/1606#issuecomment-554763424

PR #1549, opened 2019-10-10
issue #3087042 by albertoalaejos: email button for comment on comment…

albertoalaejos:
Reply to this comment button bring you to the commented entity

## Problem
Previously we solved the issue for reply links in e-mails regarding comments on posts and nodes. It still needs to be solved for comments on comments.

## Solution
Change the link in the email to reply to the entity instead of the comment on the post.

## Issue tracker
https://www.drupal.org/project/social/issues/3087042

## How to test
- [x] Login as LU X
- [x] Create a topic of type Blog
- [x] Login as LU Y
- [x] Comment on the topic
- [x] Login as LU Z
- [x] Reply to the comment
- [x] Check the emails and click on the emails buttons called "Reply to this comment"
- [x] It should bring you to the commented entity

## Release notes
We solved the issue for reply links in e-mails regarding comments on comments, now it brings you to the commented entity.


modules/social_features/social_follow_content/config/install/message.template.create_comment_following_node.yml
```diff
@@ -26,7 +26,7 @@ text:
...
+    value: "<p><a href=\"[message:author:url:absolute]\">[message:author:display-name]</a> commented on [social_comment:parent_entity_author]'s [social_comment:commented_content_type] <a href=\"[message:field_message_related_object:entity:url:absolute]\">[message:node-title]</a> you are following</p>\r\n\r\n<table class=\"comments\">\r\n<tr>\r\n<td class=\"post\"><strong>[social_comment:comment_author]</strong> <small>• [social_comment:comment_date]</small>\r\n<p>[social_comment:comment_text]</p>\r\n</td>\r\n</tr>\r\n</table>\r\n<table class=\"btn-wrapp\">\r\n\t<tbody>\r\n\t\t<tr>\r\n\t\t\t<td class=\"align-center\"><a class=\"btn-link btn-link-bg btn-link-one\" href=\"[social_comment:comment_reply_link]\">Reply to this comment </a></td>\r\n\t\t</tr>\r\n\t</tbody>\r\n</table>"
```
jaapjan: Shouldn't `social_comment:comment_reply_link ` be `social_comment:commented_entity_link`? 🤔

jaapjan:
Also tested the comment on a following node and that works as well.

I've one (big) remark about it. We are now linking to `/node/x` but then you would still need to scroll to the comment. This can be tedious. I think it would be a lot better to link to `/comment/x` then the redirect mechanism in Open Social will do the job for you already. Would that be easy to achieve?

jaapjan:
Scratch that, I've created a follow-up for that: https://www.drupal.org/project/social/issues/3087340

PR #1550, opened 2019-10-11
Issue #3087351 Make sure lazy loading works on accordion items

ronaldtebrake:
## Problem
When opening an accordeon item on a landing page the lazy loading doesn't retrigger the images automatically. 

## Solution
Make sure we reactive the lazy loading after opening an landing page item.

## Issue tracker
https://www.drupal.org/project/social/issues/3087351

## How to test
- [ ] Add a landing page with accordion
- [ ] Add image to the CKEditor in the accordion item
- [ ] See that it works :) 

## Release notes
When adding an image to a landing page accordion, if lazy loading is enabled it won't render the image correctly, this should now be solved.


modules/custom/social_lazy_loading/social_lazy_loading.module
```diff
@@ -4,3 +4,16 @@
...
+ */
+function social_lazy_loading_preprocess_node(&$variables) {
+  /** @var \Drupal\node\Entity\Node $node */
+  $node = $variables['node'];
+  if ($node->getType() === 'landing_page') {
+    // Add lazy loading to landing pages.
+    // Make sure it will be reactivated.
+    $variables['content']['#attached']['library'][] = 'social_lazy_loading/lazy_load_reactivate';
+  }
+}
```
Kingdutch: Landing page already has a library definition that's attached. Perhaps we should implement a `library_alter` and add our library/CSS there.

This way we don't introduce a hook that's called for every node :D

modules/custom/social_lazy_loading/js/lazy_load_reactivate.js
```diff
@@ -0,0 +1,20 @@
...
+
+      $accordion.each(function () {
+        var $currentAccord = $(this);
+        var $accordItem = $currentAccord.find('.paragraph--type--accordion-item');
+        $accordItem.find('.card__title-accord').on('click', function () {
+            var bLazy = new Blazy();
+            bLazy.revalidate();
+        });
+      });
+    }
```
Kingdutch: I think we can just collapse the searching done here into a single query:
```
$('.paragraph--type--accordion .paragraph--type--accordion-item .card__title-accord', context).on('click', function () {
  var bLazy = new Blazy();
  bLazy.revalidate();
});
```

:)

ronaldtebrake:
Cherry picked as well

PR #1552, opened 2019-10-14
Issue #3087741: Color scheme should apply to ideation module

agami4:
## Problem
I want to change the color scheme but it doesn't affect some particular parts of the ideation module

## Solution
Add particular parts of the ideation module to the brand file

## Issue tracker
https://www.drupal.org/project/social/issues/3087741

## How to test
- [ ] https://getopensocial.atlassian.net/browse/YANG-1611

## Release notes
The ideation extension now properly respects theme color settings set by sitemanagers.

**Maintainer note:** The ideation extension is not currently public so this probably doesn't affect distribution users.


Kingdutch:
Cherry-picked to d8bfd9ec2

PR #1555, opened 2019-10-14
Issue #3087870 by ribel: Add routes for redirect to the user groups and topics overview pages

ribel:
## Problem
Currently, User Groups and Topics pages URLs can be accessed only with URLs in the format: `/user/ID/groups`, `/user/ID/topics`, in this way it is not possible to place them on landing pages or menu links.

## Solution
Add redirect URLs: `/my-groups`, `/my-topics`, similar like we have for `/my-events`

## Issue tracker
- https://www.drupal.org/project/social/issues/3087870

## How to test
- Login as a normal user
- [ ] Go to URL /my-groups and you should be redirected to your groups page
- [ ] Go to URL /my-topics and you should be redirected to your topics page

## Release notes
Add redirect URLs: `/my-groups`, `/my-topics` for redirect to the user groups and topics overview pages.


ronaldtebrake:
Hi @ribel this works perfect, like the solution!
Could you ensure the conflicts are solved, after we can merge it :D

social.info.yml
```diff
@@ -2,7 +2,7 @@ name: Social
 type: profile
 description: 'Open Social install profile.'
 core: '8.x'
-version: '8.x-7.x'
+version: '8.x-7.0'
```
ronaldtebrake: Probably due to changing base branch.

bramtenhove:
Good idea @ribel. 

Should we make this a standard approach for custom modules as well? E.g. `my-topics`, `my-challenges`, etc.

ribel:
@bramtenhove Here we also add a route for `/my-topics`, and I think we can do the same in our extensions.
@ronaldtebrake I re-created another PR to avoid conflicts after rebase: #1607.

PR #1557, opened 2019-10-15
Issue #2929368: hide create content button if user do not have permission

albertoalaejos:
hide create content button if user do not have permissions to create content

## Problem
When permissions to create page, book, group, topic and event are disabled, the "+" icon should not be displayed in the header.

## Solution
hide create content button if user do not have permissions to create content

## Issue tracker
https://www.drupal.org/project/social/issues/2929368

## How to test
- [ ] Login as LU
- [ ] Make sure that the "+" button is visible
- [ ] Login as Admin
- [ ] Remove permissions to create content from LU
- [ ] Login as LU
- [ ] Make sure that the "+" button is not visible

## Release notes
Hide create content button if user do not have permissions to create content


Kingdutch:
This has been superseded by #1633 which only performs access checks instead of trying to render everything for the same result. This should be more performant.

Additionally, that PR caches the results of the access checks so they only need to happen once.

PR #1558, opened 2019-10-15
Issue #2977781: create post using ctrl + enter

albertoalaejos:
## Problem
When using Ctrl + Enter to send a post, the page just reloads with the content not posted.

## Solution
Trigger post and comments creation on (ctrl+enter) key.

## Issue tracker
https://www.drupal.org/project/social/issues/2977781

## How to test
- [ ] Try to post and comment by using ctrl+enter

## Release notes
Trigger posts and comments creation on (ctrl+enter) key.


modules/social_features/social_post/js/keycode-submit.js
```diff
@@ -10,29 +10,24 @@
...
+      .on('keydown', onTextAreaKeyDown);
     }
   };
+  function onTextAreaKeyDown(e) {
+    var textarea = $(this);
+    var submit = textarea.closest('form').find('.form-submit');
+    if ($.trim(textarea.val()) != '') {
+      if ((e.keyCode === 13 && e.ctrlKey) || (e.keyCode === 13 && e.metaKey)) {
+        e.preventDefault();
+        submit.trigger('click');
```
Kingdutch: Hey @albertoalaejos,

What is the reason that we click the submit button here instead of just calling `$form.submit()` on the form the textarea belongs to? I think the `.form-submit` class is mostly added for styling in our own theme, which means this Javascript will depend on our theme to function.

If there is a specific reason then we should probably add a comment to document this : )

modules/social_features/social_post/js/keycode-submit.js
```diff
@@ -10,29 +10,24 @@
...
+        $('.comment-form .form-textarea').on('keydown')
+      .off('keydown', onTextAreaKeyDown)
+      .on('keydown', onTextAreaKeyDown);
     }
   };
+  function onTextAreaKeyDown(e) {
+    var textarea = $(this);
+    var submit = textarea.closest('form').find('.form-submit');
+    if ($.trim(textarea.val()) != '') {
+      if ((e.keyCode === 13 && e.ctrlKey) || (e.keyCode === 13 && e.metaKey)) {
```
Kingdutch: Perhaps we can move this if statement to the top of this function, that way we do the quickest checks first (checking properties of the event object). This way we can avoid doing more intensive checks on each keystroke (DOM lookup and jQuery object creation). This keeps the post field more responsive.

modules/social_features/social_post/js/keycode-submit.js
```diff
@@ -10,29 +10,24 @@
...
-        keySubmit($(this), $('.form-textarea', this), $('.form-submit', this));
-      });
-
+        $('#social-post-entity-form .form-textarea')
+      .off('keydown', onTextAreaKeyDown)
+      .on('keydown', onTextAreaKeyDown);
+      // comment form
+        $('.comment-form .form-textarea').on('keydown')
+      .off('keydown', onTextAreaKeyDown)
+      .on('keydown', onTextAreaKeyDown);
```
Kingdutch: Could you explain why you used `off` here? If it is to ensure that we don't add the event handler multiple times, then we can also add the `context` variable as second parameter to our search (`$`) call to only search in the elements that were added when this function was triggered.

It also looks like the `.on('keydown')` for the comment form text area is a stray trigger.

ronaldtebrake:
Marking this as needs work, some improvements can be made in the JS see comments by @Kingdutch

ronaldtebrake:
Closing this for now, it's not high on our radar currently and quite some work is needed. Feel free to re-open in the future.

Kingdutch:
I've fixed the remarks that I had.

ronaldtebrake:
Thanks, update looks and works well :)

PR #1563, opened 2019-10-16
Issue #3088371: Update views_bulk_operations from 3.2 to 3.3

jaapjan:
## Problem
We are running drupal/views_bulk_operations 8.x-3.2 while 3.3 is available.

Not that many changes, only these:
https://www.drupal.org/project/views_bulk_operations/releases/8.x-3.3

## Solution
Changed the version in composer.json and drupal-org.make. There were 3 patches missing for drupal-org.make. But all of these are included in 3.3.

## Issue tracker
https://www.drupal.org/project/social/issues/3088371

## How to test
- [ ] Test VBO functionalities.
- [ ] Actually there should not be any code change between "our 3.2" and 3.3. But checking a diff between those would be good.
- [ ] Check patches and versioning in composer.json and .make file.

## Release notes
Updated drupal/views_bulk_operations from 3.2 to 3.3.


ronaldtebrake:
Also in 7.x

PR #1566, opened 2019-10-16
issue #3087651 - see all results button translatable on autocomplete

albertoalaejos:
## Problem
On the autocomplete search "See all results" is not translated.

## Solution
Make "See all results" translatable in Javascript file

## Issue tracker
https://www.drupal.org/project/social/issues/3087651

## How to test
- [ ] Change the language
- [ ] Autocomplete search
- [ ] The button text should be translated

## Release notes
On the autocomplete search "See all results" is now translated.

### DIFF
--- a/modules/social_features/social_search/modules/social_search_autocomplete/js/src/components/SearchSuggestions.jsx
+++ b/modules/social_features/social_search/modules/social_search_autocomplete/js/src/components/SearchSuggestions.jsx
@@ -23,7 +23,7 @@ function SearchSuggestions(props) {
       <div className="search-suggestions">{results}</div>
       <div className="search-suggestions__all">
         <a href={searchUrl} className="btn btn-default btn-raised">
-          See all results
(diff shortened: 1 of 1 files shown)


Kingdutch:
Hey @albertoalaejos ,

Simply changing the `jsx` sourcecode won't actually change something. You'll need to compile the code to have changes :) Let me know if you need help with this.

I also don't think coupling the component to Drupal in this way is a good idea. It'd prevent re-use in other applications.

Instead we can probably pass this string into the `SearchSuggestions` element as a property instead. The default can still be an untranslated `See all results`. That way the translation can be put in the code that glues the component together with Drupal in `index.jsx` or even moved to a PHP file where the `drupalSettings` element is set.

This approach shouldn't be too much additional work. It has the added benefit that tests also remain functional without adding the entire Drupal framework : )

Kingdutch:
Fixed in 98c2f29

PR #1573, opened 2019-10-24
YANG-1689: Fix theming for social birthday block

agami4:
## Problem
The birthday small block has the wrong styles

## Solution
Update styles and layout for the birthday small block

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-1689

## How to test
- [ ] Display the birthday small block on the sidebar
- [ ] You can see the birthday small block on the home page

## Release notes
Updated layout and display the birthday small block on the sidebar

### DIFF
--- a/modules/social_features/social_core/social_core.module
+++ b/modules/social_features/social_core/social_core.module
@@ -243,6 +243,13 @@ function social_core_preprocess_block(&$variables) {
       $link->setOption('use_more', FALSE);
       break;
 
+    case 'birthday-block_upcoming_birthdays':
+      $variables['subtitle'] = t('in the community');
+      $variables['view_all_path'] = Url::fromRoute('view.birthday.upcoming_birthdays');
+      $variables['button_text'] = $more_link;
+      $link->setOption('use_more', FALSE);
+      break;
+
     case 'events-block_events_on_profile':
       if ($account instanceof UserInterface) {
         $variables['subtitle'] = t('for this user');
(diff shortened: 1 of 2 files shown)


bramtenhove:
Just wondering, the Birthday module is not in Open Social core right?

In that case I'm not sure if we should just add this to the code. I think it would be nicer if we change it maybe to, for example, an alter hook and let the Birthday module implement that.

_Also, if you don't have a change record you can remove the default text._

ribel:
@bramtenhove You are correct, but in this case, as we know there are plans to add Birthday module to the distro, is that correct @ronaldtebrake?
So we just fixed the issue here that Xinyu noticed as in advance.

ronaldtebrake:
> @bramtenhove You are correct, but in this case, as we know there are plans to add Birthday module to the distro, is that correct @ronaldtebrake?
> So we just fixed the issue here that Xinyu noticed as in advance.

@ribel //cc @bramtenhove 
It's not going to be in the Distro but will still be used as an extension.
The alter option would definitely be very interesting (also for other use cases we will have with extensions in the future) is that something you can still take a look at?

ribel:
@ronaldtebrake Got it now.
Unfortunately, I can not guarantee that we will have time to work on it in ECI team, also since this is not the highest priority for us at the moment.
So If distro team can take this over I would appreciate it.

Kingdutch:
Hey @ribel @ronaldtebrake ,

This PR is adding an implementation for `social_core_preprocess_block`. This can easily be copied to the `social_birthday` module without any changes in Open Social right? This is an alter hook that's already supported by Drupal core.

That means we can just close this PR and any associated Drupal.org ticket as not needed.

Kingdutch:

PR #1575, opened 2019-10-24
Issue #3089936: Select all of VBO and the selected member count in gr…

royharink:
…oup manage members overlaps.

## Problem
Select all of VBO and the selected member count in group manage members overlaps when a language with long words like German is used and is therefor unreadable and not easy to use.

## Solution
The fields are aligned next to eachother. The fields should be stacked so longer the translated texts in languages that have a longer word count fit.

## Issue tracker
https://www.drupal.org/project/social/issues/3089936


ronaldtebrake:
Also in 7.x :)

PR #1576, opened 2019-10-24
Issue #3089968 by chmez: Entity types update

ghost:
## Problem
Need to have functionality for update entity types after deleting the same functionality from the **drush entup** command.

## Solution
Create a new service in the **Social Core** module.

## Issue tracker
- https://www.drupal.org/project/social/issues/3089968
- https://getopensocial.atlassian.net/browse/YANG-1679

### DIFF
--- a/modules/social_features/social_core/social_core.services.yml
+++ b/modules/social_features/social_core/social_core.services.yml
@@ -1,4 +1,14 @@
 services:
+  social_core.helper:
+    class: Drupal\social_core\Service\SocialCoreHelper
+    arguments:
+      - '@entity.definition_update_manager'
+      - '@entity_type.manager'
+      - '@entity_field.manager'
+      - '@entity.last_installed_schema.repository'
+      - '@entity_type.listener'
+      - '@field_storage_definition.listener'
+
   social_core.route_subscriber:
     class: \Drupal\social_core\Routing\RouteSubscriber
     tags:
--- a/modules/social_features/social_core/social_core.services.yml
+++ b/modules/social_features/social_core/social_core.services.yml
@@ -10,6 +10,6 @@ services:
       - '@field_storage_definition.listener'
 
   social_core.route_subscriber:
-    class: \Drupal\social_core\Routing\RouteSubscriber
+    class: Drupal\social_core\Routing\RouteSubscriber
     tags:
       - { name: event_subscriber }
--- a/modules/social_features/social_core/src/Service/SocialCoreHelper.php
+++ b/modules/social_features/social_core/src/Service/SocialCoreHelper.php
@@ -1,0 +1,188 @@
+<?php
+
+namespace Drupal\social_core\Service;
+
+use Drupal\Core\Entity\EntityDefinitionUpdateManagerInterface;
+use Drupal\Core\Entity\EntityFieldManagerInterface;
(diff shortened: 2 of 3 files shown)


Kingdutch:
Hi @OH17L,

To me it seems like we're trying to rebuild the functionality of the `drush entup` command within Open Social. This isn't a good idea, because we'll still suffer from the same issues that caused the commands to be removed in the first place. Instead, entity updates should be made explicit in update or post_update hooks.

For an explanation of why the method was removed and how to create these entity update hooks instead, please see: https://www.drupal.org/node/3034742

The biggest issue with these automated updates is that they're non-deterministic (different things happen based on different circumstances). For example, with the way that you used the above service in the Course Quiz extension, the behaviour is different whether I would run this update hook now, or in a few releases where more changes may have been made. 

The proper way to make this change is to update the schema and field data specifically in the update hook, based on the current circumstances, instead of an automated diff. This means that code we add in the future has a well defined schema to work with.

OH17L:
Hi @Kingdutch, it works for our case for adding new base fields. Also, we using it in the update hooks as you said. Do you have an example of code which you propose to use in the hook update?

Kingdutch:
I can understand that it works for this use-case, in this scenario. Because you're basically recreating what `drush entup` used to do, with all the same problems. See also the links in the Change Record for more information on what could go wrong. The below quote from the [EntityDefinitionUpdateManagerInterface](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityDefinitionUpdateManagerInterface.php/interface/EntityDefinitionUpdateManagerInterface/8.7.x) sums it up nicely:

> This typically happens in Update API functions, which need to take the system from a known state to another known state. Relying on the definitions living in code might prevent this, as the system might transition directly to the last available state, and thus skipping the intermediate steps. Manipulating the definitions in state allows to avoid this and ensures that the various steps of the update process are predictable and repeatable.

The Change Record that I linked to already has some code examples for you.

From what I can find, you're adding the `created` and `changed` fields (although this is not clear from the update hook). So adapting an example from the Change Record, your code would look something like this (untested):

```
/**
 * Changeme
 */
function example_update_8000(&$sandbox) {
  $definition_update_manager = \Drupal::entityDefinitionUpdateManager();
  /** @var \Drupal\Core\Entity\EntityLastInstalledSchemaRepositoryInterface $last_installed_schema_repository */
  $last_installed_schema_repository = \Drupal::service('entity.last_installed_schema.repository');

  $entity_type = $definition_update_manager->getEntityType('attempt');
  $field_storage_definitions = $last_installed_schema_repository->getLastInstalledFieldStorageDefinitions('attempt');


  $field_storage_definitions['created'] = BaseFieldDefinition::create('created')
      ->setLabel(t('Created'))
      ->setDescription(t('The time that the attempt was created.'));

  $field_storage_definitions['changed'] = BaseFieldDefinition::create('changed')
      ->setLabel(t('Changed'))
      ->setDescription(t('The time that the attempt was last edited.'));

  $definition_update_manager->updateFieldableEntityType($entity_type, $field_storage_definitions, $sandbox);

  return t('Added created and changed field to Attempt entity type.');
}
```

Basically you mimick the changes that you make in the entity's field definitions and make any data changes if needed.

OH17L:
Ok, I'll try to use your solution.

PR #1578, opened 2019-10-25
Inconsistent and confusing threaded reply on reply

slowflyer:
## Problem
When commenting users "cant" reply on a reply. This second reply will be handled as mention to the first. This is currently a large issue in my communities.
f.e. We start a discussion, a user answers with a comment. The moderator replies to this comment and asks for more details. Now users can't answer in the same way they used to do. And if there are a lot comments in the discussion, they jump to the end of page to create a comment with a mention. But thats not clear to the user.

## Solution
To keep things as they are for communities where deeper threaded comments are not needed, I made it configurable to define how many levels should be possible for normal threaded comment behaiviour.

## Issue tracker
https://www.drupal.org/project/social/issues/2907511
https://www.drupal.org/project/social/issues/3033987

## How to test
- [x] install the patch, clear cache 
- [x] goto /admin/config/opensocial/socialcomment an specify more than one level
- [x] reply on comments and see, that more threaded levels up to the specified threshold are allowed

## Release notes
Make it up to the administrator of the community to allow more threaded levels when commenting.


slowflyer:
Will create a follow up, compatible with ajax comments

PR #1585, opened 2019-10-30
#3091105: Prevent fatal error by checking if the commented entity still exists

bramtenhove:
## Problem
On the admin comment overview page we show all the comments.

We also try to render a "translate" operations link if the content translation is turned on. However when a content item is deleted, but the reference for some reason stays behind in the comment an access check for the translate link will fail.

## Solution
An extra check is added to make sure the commented entity still exists.

## Issue tracker
https://www.drupal.org/project/social/issues/3091105

## How to test
- [ ] Place a comment on a node
- [ ] Remove the node, but keep the reference to it in a comment
- [ ] Observe that /admin/content/comment works

## Release notes
In a rare occasion where the commented entity no longer exists, but the reference to it in the comment does, a fatal error might occur on the comment admin overview page.


ronaldtebrake:
Cool, nice find :)

ronaldtebrake:
Was already the case in 6.x so cherry picked to 6 and 7.x

PR #1586, opened 2019-10-30
Issue #3091190 by chmez: Unpublished content in the notifications center

ghost:
## Problem
Group members without access for viewing unpublished content receiving messages in the notification center about creating unpublished content in a group.

## Solution
Check access for viewing an entity when a user adding as a recipient of the notification.

## Issue tracker
- https://www.drupal.org/project/social/issues/3091190
- https://getopensocial.atlassian.net/browse/YANG-842

## How to test
- [ ] Log in as **User 1** and create the **Group 1** open group.
- [ ] Log in as **User 2** and join to the **Group 1** group.
- [ ] Log in as **User 1** and create the **Topic 1** unpublished topic in the **Group 1** group.
- [ ] Log in as **User 2**.
- [ ] **Notification center** should not contain message about creating the **Topic 1** topic in the **Group 1** group by **User 1**.

## Release notes
Group members will not receive notification about creating unpublished content.


ribel (approved):
I tested and it works well as for me.
The only thing I noticed - if you create Unpublished Topic, and after editing change status to Published - it still doesn't appear in the notification centre. I think it would be good to fix it, but not sure how complex it is.
Anyway, I think it is already better than display notifications about Unpublished content.

ronaldtebrake:
I can consider that out of scope for this fix, so will accept it as it is know :)

ronaldtebrake:
Also backported.

PR #1587, opened 2019-10-30
Feature/social poll module

albertoalaejos:
## Problem
As a CM I want to be able to start a poll

## Solution
Give Content managers the option to create polls so that LU can vote by using an existing module called Poll: https://www.drupal.org/project/poll

## Issue tracker
https://getopensocial.atlassian.net/browse/DS-6940

## How to test
- [ ] Enable the module
- [ ] Login as CM+
- [ ] Click on Content and create a Poll
- [ ] Create a Landing Page
- [ ] Add a Poll Item Section
- [ ] Select the created poll

## Release notes
Content managers now have the option to create polls so that LU can vote.


modules/social_features/social_poll/social_poll.install
```diff
@@ -0,0 +1,17 @@
...
+
+use Drupal\user\Entity\Role;
+
+/**
+ * Implements hook_install().
+ *
+ * Perform actions related to the installation of social_page.
+ */
+function social_page_install() {
+
```
ronaldtebrake: empty? :D

modules/social_features/social_poll/social_poll.module
```diff
@@ -0,0 +1,6 @@
+<?php
+
+/**
+ * @file
```
ronaldtebrake: Empty file? :D

modules/social_features/social_poll/social_poll.install
albertoalaejos: empty file removed ;)

modules/social_features/social_poll/social_poll.module
albertoalaejos: empty file removed ;)

modules/social_features/social_poll/social_poll.info.yml
```diff
@@ -0,0 +1,10 @@
+name: 'Social Poll'
+description: 'Social Poll'
+type: module
+core: 8.x
+dependencies:
+  - drupal:poll
+  - drupal:block
+  - drupal:field
+  - social:social_landing_page
```
ronaldtebrake: i don't think we want this hard dependency, poll should be able to work without social_landing_pages.

modules/social_features/social_poll/src/SocialPollConfigOverride.php
```diff
@@ -0,0 +1,75 @@
...
+   * Load overrides.
+   */
+  public function loadOverrides($names) {
+    $overrides = [];
+    $config_factory = \Drupal::service('config.factory');
+
+    $config_names = [
+      'field.field.paragraph.section.field_section_paragraph',
+    ];
+    foreach ($config_names as $config_name) {
```
ronaldtebrake: why a foreach? we only have one config name to check

Kingdutch:

PR #1593, opened 2019-11-07
issue #3092891 by albertoalaejos: Remove features from Social User mo…

albertoalaejos:

## Problem

## Solution
- Remove the social_user.features.yml file
- Move the default editable config from hook_install() into a config/install/*.yml file.
- Remove any _module_set_defaults function

## Issue tracker
https://www.drupal.org/project/social/issues/3092891

## How to test
- [ ] Enable the module
- [ ] Change the login help text
- [ ] Revert features
- [ ] The text should keep changed

## Release notes
The social_user module is no longer managed by features. The fixes the issue where login help text edited by Site Managers would be overwritten.

### DIFF
--- a/modules/social_features/social_user/config/features_removal/core.entity_form_display.user.user.default.yml
+++ b/modules/social_features/social_user/config/features_removal/core.entity_form_display.user.user.default.yml
@@ -1,0 +1,52 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - field_group
+    - user
+third_party_settings:
+  field_group:
+    group_information:
+      children:
+        - account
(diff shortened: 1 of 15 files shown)


ronaldtebrake:
social_user has in the config/install folder these two files:

`core.entity_form_display.user.user.register.yml`
`core.entity_view_display.user.user.full.yml`

But these are already provided (and exactly the same) in the Drupal user module which is a dependency from social_user.

We can therefore safely remove them since they are the same. 
If they weren’t the same, we can move it to the social_profile config install folder, Drupal will take care of the rest.

PR #1601, opened 2019-11-14
Issue #3071081 by agami4: Make field label visible

agami4:
## Problem
Field labels are not visible, even if a field label is configured to be visible in a view mode. This happens because the label is never printed in socialbase templates.

## Solution
Update twig templates and add the field label to the output if the view mode is configured to have a label.

## Issue tracker
https://www.drupal.org/project/social/issues/3071081

## How to test
- [ ] Create new CT or open 'Topic' menage display page
- [ ] Select 'Above' or 'Inline' for the label
- [ ] Add content or open page with content and you can see label

## Release notes
Field labels will now be properly displayed if they are configured to show in a view mode.


themes/socialbase/templates/field/field.html.twig
```diff
@@ -41,31 +41,41 @@
...
+      {% endif %}
     {% else %}
-      {{ item.content }}
+      <div{{ title_attributes.addClass(title_classes) }}>{{ label }}</div>
+
+      {% if entity_type == "node" and field_name == "body" and part_of_teaser %}
+        {{ item.content|render|striptags }}
+      {% else %}
+        {{ item.content }}
+      {% endif %}
```
Kingdutch: This duplicates quite a bit of code. We should just render the label if it's not hidden and deduplicate the rest.

themes/socialbase/templates/group/group--default.html.twig
```diff
@@ -41,8 +41,22 @@
...
+
     <div{{ content_attributes.addClass('card__body body-text') }}>
-      {{ content.field_group_description }}
+      {% if label_hidden %}
+        {{ content.field_group_description }}
+      {% else %}
+        <div{{ title_attributes.addClass(title_classes) }}>{{ label }}</div>
+
+        {{ content.field_group_description }}
+      {% endif %}
```
Kingdutch: We should just render the label if it's not hidden and not duplicate the `{{ content.field_group_description }}`

themes/socialbase/templates/node/field--node--body.html.twig
```diff
@@ -6,10 +6,27 @@
...
+    </div>
+  {% else %}
+    <div{{ title_attributes.addClass(title_classes) }}>{{ label }}</div>
+
+    <div class="body-text clearfix">
+      {% for item in items %}
+        {{ item.content }}
+      {% endfor %}
+    </div>
+  {% endif %}
```
Kingdutch: Just render the label without duplicating the rest please :)

themes/socialbase/templates/field/field.html.twig
agami4: Fixed

themes/socialbase/templates/group/group--default.html.twig
agami4: Fixed

themes/socialbase/templates/node/field--node--body.html.twig
agami4: Fixed

ribel:
@Kingdutch, I see that with this PR we also introduced a lot of new visual issues  :see_no_evil:
Should we revert this PR, or fix field label visibility in configs all over the distro and enterprise modules? 
image
// cc @ronaldtebrake @agami4

Kingdutch:
Hey @ribel ,

Yeah we were aware that this would happen when merging. We expect that we'll find 80-90% of these during the testing day. This configuration is one of the things we should improve anyway so I think fixing this before the 8.0 release is the best way to go :D

ronaldtebrake:
https://www.drupal.org/project/social/issues/3108296
Created issue here

PR #1604, opened 2019-11-14
Fix undefined index notice from user entity reference selection plugin

drclaw:
## Problem
The plugin definition for the UserSelection Entity Reference Selection (ERS) plugin is causing an undefined index notice on user reference field settings forms.
```
Notice: Undefined index: base_plugin_label in Drupal\Core\Field\Plugin\Field\FieldType\EntityReferenceItem->fieldSettingsForm() (line 388 of core/lib/Drupal/Core/Field/Plugin/Field/FieldType/EntityReferenceItem.php)
```

## Solution
Currently there are a couple undocumented formats in which you must structure your Entity Reference Selection plugins in order for them to appear in the list on the field settings page. The format we're using in social is technically meant for derivatives where the deriver inserts a `base_plugin_label` into the derivative definition. For backwards compatibility with existing open social config, we'll need to keep using this derivative method which means we need to add the `base_plugin_label` into our plugin definition.

## Issue tracker
https://www.drupal.org/project/social/issues/3094529

## How to test
- [ ] Create a user reference field
- [ ] Go to the field settings form, notice should be logged but may also display on-screen depending on error reporting setting

### DIFF
--- a/modules/social_features/social_profile/src/Plugin/EntityReferenceSelection/UserSelection.php
+++ b/modules/social_features/social_profile/src/Plugin/EntityReferenceSelection/UserSelection.php
@@ -9,12 +9,11 @@ use Drupal\user\Plugin\EntityReferenceSelection\UserSelection as UserSelectionBa
  * Provides specific access control for the user entity type.
  *
  * @EntityReferenceSelection(
- *   id = "social:user",
+ *   id = "social_user",
  *   label = @Translation("Social user selection"),
  *   entity_types = {"user"},
- *   group = "social",
- *   weight = 1,
- *   base_plugin_label = @Translation("Social user selection")
+ *   group = "social_user",
+ *   weight = 1
  * )
  */
 class UserSelection extends UserSelectionBase {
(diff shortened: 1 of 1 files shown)


drclaw:
Okay change of plans, initial pull request changed the plugin id which was breaking existing fields using the selection plugin (duh). So instead, lets just include the `base_plugin_label` directly on our plugin definition (PR updated)

Kingdutch:
@drclaw Thanks for your contribution :) 

Am I correct in assuming that the issue is the `:` in the plugin ID? If we can change this in all Open Social code I'd be okay with changing this going forward given the following.

1. Change the plugin definition for `UserSelection` to `id => 'social_user_selection'`
2. Change all code that references this plugin to use `social_user_selection` in Open Social
3. Create new class `UserSelectionTransition extends UserSelect` with the old plugin definition that includes the `base_plugin_label`
4. Mark the `UserSelectionTransition` class as deprecated (can be removed in 9.0 or 10.0).

This should allow existing implementation to remain functional while fixing the issue. It's a bit more cumbersome but also prevents weird behaviour from Drupal thinking this is some sort of child-plugin in the future.

We should create a change notice to create any implementing code from `social:user` to `social_user_selection`.

navneet0693:
This is quite old. Please open a new PR, if required.

PR #1608, opened 2019-11-19
Add untranslatable strings in a static PHP file so it can be imported…

albertoalaejos:
Add untranslatable strings in a static PHP file so it can be imported in Weblate

## Problem
Strings can't be found in Weblate

## Solution
Strings that are Confirmed and can't be found in Weblate should be added in a static PHP file.

## Issue tracker
https://getopensocial.atlassian.net/browse/TB-3454

## How to test
- [ ] Check that the Strings added are translatable


Kingdutch:
Hey there!

Can we add some comments in the file to document why the strings are there? Linking to where the string came from and what issue (number) needs to be fixed to make this happen ensures we don't have a giant list with which we don't know what to do in 6 months.

bramtenhove:
Fair enough @Kingdutch, but we expect all these to be fixed as part of the larger efforts for Features removal.

It's too much work to identify what the issues are in this specific ticket. After Features removal we can use some time to see if all these strings work or not.

Kingdutch:
@bramtenhove That's fine :) However, at least a top level comment like the following will ensure other people know why this list is there and not just randomly add others to it.
```
// These strings have been added because they were not being picked
// up by the POTX tool. This usually indicates an issue with configuration
// schema or a string not passed through `TranslatableMarkup`.
// These should be removed once the underlying issue is identified and fixed.
```

bramtenhove:
Hehe sure, already did that ;)

PR #1609, opened 2019-11-19
Travis issues occured due to core patch not working - Make it green again!

ronaldtebrake:
## Problem
We don't lock a core version, which means when core releases updates our patches on 6.0.0 might fail. Which usually isn't an issue because we can control it ourselves in updates, but with travis we test the update path of Open Social and do this from a tagged version. 
Therefor it will always fail on the composer install with the latest core version.

## Solution
Test the update path from 8.x-6.x-dev, allowing us to backport any core patches when necessary.

## Issue tracker
- not needed

## How to test
- [ ] See that travis turns green.

## Release notes
Not needed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ before_install:
   # Installing hirak/prestissimo allows composer to parallelise downloads which speeds up the composer install.
   - composer global require hirak/prestissimo
   - echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, PR=$PR, BRANCH=$BRANCH, COMMIT=$COMMIT"
-  - if [ "$TEST_SUITE" = "install_update" ]; then composer require goalgorilla/open_social:6.0.0 --prefer-dist; fi
+  - if [ "$TEST_SUITE" = "install_update" ]; then composer require goalgorilla/open_social:dev-8.x-6.x --prefer-dist; fi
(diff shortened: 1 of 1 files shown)


Kingdutch:
Hey @ronaldtebrake ,

This does change the migration path that we test because we no longer test 6.0 -> 6.x.

That said! Perhaps we should think about how we test this anyway. Because we currently only test the "major version" upgrades (e.g. cherry-picks to a previous minor branch for minor-branch.0 -> minor-branch.y do not get tested).

Maybe we should just include a tests/composer.json file that sets up a known previous state of Open Social to test from? That way patches would also not be applied.

Especially since testing from 6.x with certain patches applied can cause a different update path than if those patches were only applied when reaching 8.x. The patches could also include update hooks that are then being ignored after initial patched 6.x install.

ronaldtebrake:
> Hey @ronaldtebrake ,
> 
> This does change the migration path that we test because we no longer test 6.0 -> 6.x.
> 
> That said! Perhaps we should think about how we test this anyway. Because we currently only test the "major version" upgrades (e.g. cherry-picks to a previous minor branch for minor-branch.0 -> minor-branch.y do not get tested).
> 
> Maybe we should just include a tests/composer.json file that sets up a known previous state of Open Social to test from? That way patches would also not be applied.
> 
> Especially since testing from 6.x with certain patches applied can cause a different update path than if those patches were only applied when reaching 8.x. The patches could also include update hooks that are then being ignored after initial patched 6.x install.

Yes, that is obviously known ;) @Kingdutch 
But that doesn't make our travis test green right now, so i'd rather go for a green state and get confidence on what people create versus not having anything green and having 50+ PR's open.

In paralel we can think about how we test, what we want to test and how to fix this.

Kingdutch:
Concious changes are good changed! Go for it :D

PR #1615, opened 2019-11-25
Issue #3096602 by agami4: Update position of css style for menu on the mobile

agami4:
## Problem
When the admin toolbar is visible the z-index of the main menu is changed. This is done to ensure the admin toolbar is shown on top of the main menu. However, this also causes the main menu to appear under certain page elements.

This happens in all browsers

## Solution
Don't try to change the z-index of the main menu when the admin bar is present, but instead properly update the admin toolbar's z-index.

## Issue tracker
https://www.drupal.org/project/social/issues/3096602
https://getopensocial.atlassian.net/browse/YANG-1776

Also affected https://www.drupal.org/node/3106782 where a slightly different fix was previously applied.

## How to test
- [ ] Login as user with access to the admin toolbar
- [ ] Open the menu on a mobile device
- [ ] See that the menu is under certain elements on the page.

## Release notes
The main menu no longer appear under page content on mobile devices when a user has access to the administrative menus.

## Screenshots
afbeelding

afbeelding

### DIFF
--- a/themes/socialbase/assets/css/admin-toolbar.css
+++ b/themes/socialbase/assets/css/admin-toolbar.css
@@ -10,9 +10,14 @@
   transform: translateY(39px) translateX(0);
 }
 
+.toolbar-vertical .navbar-default,
 .toolbar-fixed .navbar-default {
   position: relative;
-  z-index: 100;
+}
+
+.toolbar-vertical .toolbar .toolbar-bar,
+.toolbar-fixed .toolbar .toolbar-bar {
+  z-index: 1080;
 }
 
 .toolbar-fixed .main-container,
@@ -26,12 +31,5 @@
   }
   .toolbar-vertical .off-canvas-right.is-open {
     transform: none;
-  }
-}
-
-@media (max-width: 599px) {
-  .toolbar-vertical .navbar-default {
-    position: relative;
-    z-index: 1;
   }
 }
--- a/themes/socialbase/components/00-config/_variables.scss
+++ b/themes/socialbase/components/00-config/_variables.scss
@@ -7,6 +7,7 @@ $zindex-modal-background:  1040;
 $zindex-modal:             1050;
 $zindex-tour-backdrop:     1060;
 $zindex-tour:              1070;
+$zindex-admin-menu:        1080;
 
 @import 'variables/colors';
 @import 'variables/components';
--- a/themes/socialbase/components/04-organisms/admin-toolbar/admin-toolbar.scss
+++ b/themes/socialbase/components/04-organisms/admin-toolbar/admin-toolbar.scss
@@ -22,19 +22,22 @@
 
 }
 
-// make sure we can see both menu's
-.toolbar-fixed .navbar-default {
-  position: relative;
-  z-index: 100;
+.toolbar-vertical,
+.toolbar-fixed {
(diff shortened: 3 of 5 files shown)


Kingdutch:
With the changes outlined above we'll also have to remove the 'slide-open' animation from the menu on mobile devices.

agami4:
Hey @Kingdutch, I have some questions about your comments above.
- mobile main menu and mobile profile and menu "Create new CT's" - those menus should have the same styles and behaviors of JS and markup? If this question is correct - the mobile main menu should be like a profile menu?
- when the main menu(for example) is opened and the user clicks on the profile menu, the main menu must close? (This behavior for all mobile menus)

Kingdutch:
Hey @agami4 ,

Yes, the create content menu and the navigation menu should have the same behaviour as the profile menu currently does (I discussed this briefly with Xinyu). So we probably need to change some styling and markup to achieve this.

Only one menu should be open at a time. So indeed if a user is viewing the main menu and clicks the profile menu then the main menu should close. If a user then presses the "Create Content" menu, the profile menu should close.

agami4:
@Kingdutch, I have a proposal about new updates for all mobile menus. Maybe I create new issue on the Distro and add those updates for mobile menus and add `z-index: 3` for `.toolbar-vertical .navbar-default` and change top position from `top: 100` to `top: initial` for `.navbar-collapse .dropdown-menu` on this PR? What do you think about it? :)

Kingdutch:
Hi @agami4 ,

I'd prefer to fix this in the right way now instead of using a workaround first. Especially since we'll probably forget to remove the unneeded workaround later, cluttering our styles.

ronaldtebrake:
This will need some love from @Kingdutch, timewise we probably won't be able to make it in the release. But assigned accordingly so we know Shipsters will need time to work on the feedback.

Kingdutch:
I've fixed the styling, hopefully this fixes it once and for all :D I've also moved the button to a different template where I think it makes more sense.

I've not done the larger refactoring that I outlined in my previous comment. The reason for this is that it requires quite some interruptive changes and the menu moving the main content is only an issue for people who have access to the admin toolbar. So although we see this ourselves quite often, this bug actually doesn't impact a large part of Open Social users.

Kingdutch:
@agami4 Could you review/test this to verify that it also still fixes the issue you were having? :)

@ronaldtebrake I've described what I've done in my previous comment and updated the PR summary as well as the drupal.org issue. If you have some time to merge and cherry-pick this then that fixes an issue for SIM :)

agami4:
@Kingdutch, I've tested your updates and all looks good for me)

frankgraave:
This has been ported back to 8.x-7.x and 8.x-8.x.

**8.x-7.x:**
https://github.com/goalgorilla/open_social/commit/e4ba40eb1f1dc39317ce3134838cb3af39b274db
https://github.com/goalgorilla/open_social/commit/693d3c3a32a5d146cc48525fcf991b666ea4659a

**8.x-8.x:**
https://github.com/goalgorilla/open_social/commit/a5fdc526d161bd0510abd6bafcac942bffd2b196
https://github.com/goalgorilla/open_social/commit/af79b2d0396d8eb666ba1b49cb5bd1b7b9be1433

PR #1618, opened 2019-11-26
Improve landing page performance and code quality

Kingdutch:
## Problem
The landing page module performed quite a bit of preprocessing that could be handled by Drupal instead. Simplifying these methods or removing them altogether reduces the amount of code that is created and simplifies caching by producing less data in more centralised locations.

### Remove config load from LandingPageConfigOverride

We now know how to use the NestedArray::mergeDeep properly to add new items to an array without loading existing configuration. This speeds up config overrides slightly.

### Use variables instead of route parsing
    
The current node is already present in the `$variables` that is passed to the preprocess hook. It's cleaner (and more stable and performant) to use this value instead of trying to parse the value from the URL and load it manually.

### Replace preprocess hook by display mode change
    
The Drupal display modes have the possibility to show fields as URL. This mode still allows you to select a image mode. Using this method removes to need to add a custom variable to a template through a preprocess hook. This method is also more easily customised.
   
 **BREAKING CHANGE**: Because a value is removed from a template.

### Remove node_edit_url from paragraph hero's
    
 The variable isn't actually ever set for paragraphs and is probably a copy paste error from another hero template.
 
 **BREAKING CHANGE**: In case the variable was used in an extending project

### Change node_edit_link to render array
    
 The preprocess hook does not need to check what page it's on because it's not called for the edit or delete page. 

 This commit also changes the way the URL is rendered to use a render array. This causes less operations in the twig template (if statement, string append etc.) and allows the node edit URL to do its own access checking instead of doing this in a preprocess hook.

### Use NodeInterface instead of Node for argument types
    
Something can be created that implements the NodeInterface type, which we can process, but is not a Node.
    
Additionally, the `instanceof` check is not needed because this is guaranteed by our function argument type.

### Add landing page hero image field
    
This field will be used to store the image that is used for featured landing pages. This allows the featured image to be calculated when the node is saved instead of when it is viewed. This should speed up viewing performance (which is the majority of the use-cases).

### Replace _social_landing_page_get_hero_image with pre-calculated image field
    
This commit uses the newly added field_landing_page_image to provide a teaser image for landing pages. This improves the view performance by moving the calculations for which image to display into the saving of the entity instead of during every display.
    
The added benefit of using a standard name of field_{node_type}_image allows code to be simplified by removing special casings that were added for landing pages where this field was not yet available.

## Issue tracker
https://www.drupal.org/project/social/issues/3100302

## How to test
- [ ] Create a landing page with a hero section
- [ ] Edit the landing page to add itself as a featured item and save the node
- [ ] Observe that the hero's image is used as image for the featured item
- [ ] Add a new small hero above the previous hero and save the node
- [ ] Observe that the featured image has changed to use the image of the new hero
- [ ] Change the image for the small hero at the top and save the node
- [ ] Observe that the new image is shown as featured image for the landing page

Additionally 
- [ ] Play around with the landing page to see how these changes impact performance.

## Release notes
Various improvements have been made to the Social Landing Page module. These improve the overall quality of the module and prefer pre-calculated values during save over calculating values on page load. These improvements make some changes to how the landing page module renders values, if you've overridden this, please look at the issue to see what has changed in detail.

### DIFF
--- a/modules/social_features/social_landing_page/config/install/core.entity_form_display.node.landing_page.default.yml
+++ b/modules/social_features/social_landing_page/config/install/core.entity_form_display.node.landing_page.default.yml
@@ -3,6 +3,7 @@ status: true
 dependencies:
   config:
     - field.field.node.landing_page.field_content_visibility
+    - field.field.node.landing_page.field_landing_page_image
     - field.field.node.landing_page.field_landing_page_section
     - node.type.landing_page
   module:
@@ -38,6 +39,8 @@ third_party_settings:
         id: sections
         classes: card
       label: Sections
+_core:
+  default_config_hash: ZmNvrc200HUdqH2M5CVzebvr-UKPAQWa7mq4XhhrYNU
 id: node.landing_page.default
 targetEntityType: node
 bundle: landing_page
@@ -98,6 +101,7 @@ content:
     region: content
     third_party_settings: {  }
 hidden:
+  field_landing_page_image: true
   groups: true
   promote: true
   sticky: true
--- a/modules/social_features/social_landing_page/config/install/core.entity_form_display.node.landing_page.default.yml
+++ b/modules/social_features/social_landing_page/config/install/core.entity_form_display.node.landing_page.default.yml
@@ -39,8 +39,6 @@ third_party_settings:
         id: sections
         classes: card
       label: Sections
-_core:
-  default_config_hash: ZmNvrc200HUdqH2M5CVzebvr-UKPAQWa7mq4XhhrYNU
 id: node.landing_page.default
 targetEntityType: node
 bundle: landing_page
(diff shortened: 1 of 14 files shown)


Kingdutch:
I don't think a change record is needed here. The breaking changes are low-risk and should be clear from the issue for anyone that runs into them anyway.

Kingdutch:
Rebased based on 8.x-9.x to fix merge conflicts. This now needs work to provide an update hook due to the configuration changes.

This also needs new testing because https://github.com/goalgorilla/open_social/commit/5ae7982f973f7882778aafe9b5090c2ec1615339 changed how the hero image was determined and it's a bit unclear whether that commit may have broken featured images for landing pages.

From the merge conflicts that were solved during the rebase it doesn't look like that commit necessarily interferes with this PR but it still touches a bit of the same code.

ronaldtebrake:
Closing this for now, needs work but seems like there is no actual priority given. We can always re-open and go from the last comment.

PR #1619, opened 2019-11-27
Replace features with config management for Social Content Report

albertoalaejos:

## Problem

## Solution
Remove the social_content_report.features.yml file
Move the default editable config from hook_install() into a config/install/*.yml file.
Remove any _module_set_defaults function

## Issue tracker
https://www.drupal.org/project/social/issues/3096748
https://getopensocial.atlassian.net/browse/TB-3460

## How to test
- [ ]  Enable the module
- [ ]  Enable email notifications
- [ ]  Revert features
- [ ]  Emails notifications should keep enabled

## Release notes
The social_content_report module is no longer managed by features.

### DIFF
--- a/modules/social_features/social_content_report/config/features_removal/core.entity_form_display.flagging.report_comment.default.yml
+++ b/modules/social_features/social_content_report/config/features_removal/core.entity_form_display.flagging.report_comment.default.yml
@@ -1,0 +1,32 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.flagging.report_comment.field_other_reason
+    - field.field.flagging.report_comment.field_reason
+    - field.field.flagging.report_comment.field_status
+    - flag.flag.report_comment
+  enforced:
+    module:
+      - social_content_report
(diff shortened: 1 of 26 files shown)


Kingdutch:
Hey @albertoalaejos,

We should use this moment to add the schema file for the module. I can see that it at least has a settings.yml file for which we can do this.

The settings file is here: https://github.com/goalgorilla/open_social/blob/8.x-8.x/modules/social_features/social_content_report/config/install/social_content_report.settings.yml

You can see in the SettingsForm what they should have as description, additionally the settings form sets some configuration that's not in the settings.yml file that also needs a schema here: https://github.com/goalgorilla/open_social/blob/8.x-8.x/modules/social_features/social_content_report/src/Form/ContentReportSettingsForm.php

You can find an example schema file in another features removal PR: https://github.com/goalgorilla/open_social/pull/1614/files#diff-041917e3a907f12a604385fab1f887fd

Kingdutch:
Looks like this already has a schema 🤷‍♂

PR #1623, opened 2019-11-29
Replace features with config management for Social Like

albertoalaejos:

## Problem

## Solution
Remove the social_like.features.yml file
Move the default editable config from hook_install() into a config/install/*.yml file.
Remove any _module_set_defaults function

## Issue tracker
https://www.drupal.org/project/social/issues/3097638
https://getopensocial.atlassian.net/browse/TB-3523

## How to test
- [ ]  Enable the module
- [ ]  Enable email notifications
- [ ]  Revert features
- [ ]  Emails notifications should keep enabled

## Release notes
The social_like module is no longer managed by features.


Kingdutch:
The first two issues have been fixed. The third issue of an overwritten view mode turned out to be slightly bigger.

The view mode in social_like is identical to the one in social_post. However, the one in social_post actually depends on social_like (even though this is not defined anywhere). I've opened the following issue for this: https://www.drupal.org/project/social/issues/3099988

Before this PR is merged we should find a way to properly add the like_and_dislike field to Posts and other entities (probably just removing it from our config and lettings the like_and_dislike module's hooks take care of it). Once that's done we can remove the file from this module and merge this.

Kingdutch:
This can be resolved for now by removing the duplicate view mode and declaring a dependency from Social Post to Social Like. Social Like is already a non-optional module in Open Social so this doesn't change any behaviour, it just makes the relation more explicit if we ever want to decouple the modules properly.

PR #1626, opened 2019-12-02
Issue #3098046 by Kingdutch, robertragas: Replace SocialLanguageMetadataBubblingUrlGenerator by Entity::toUrl implementations

Kingdutch:
<h2>Problem</h2>
The short summary is that we've overwritten link generation to fix the stream always switching the link to English. This overwritten link generation breaks links that are supposed to switch language (e.g. Language Switcher and Content Translation overview).

<h2>Solution</h2>
Remove the <code>SocialLanguageMetadataBubblingUrlGenerator</code> as it is too heavy handed.

Instead, implement the <code>toUrl</code> methods for the entities that we use. In this method we check if a language is explicitly requested and otherwise change the language of the link to the one for the current content language. This relies on Drupal's behaviour to show the default language for an entity if no translation is available. With this behaviour it preserves the language that the user is currently viewing the site in instead of changing the language abruptly.

## Issue tracker
https://www.drupal.org/project/social/issues/3098046

## How to test
- [ ] Enable the Social Language Content Translation module
- [ ] Add a second language
- [ ] View the site in the second language
- [ ] Observe that the stream content does not change your shown language when navigated to
- [ ] Observe that Drupal core's language switcher block works as expected
- [ ] Observe that the links on content translation overview pages lead to the correct language

**Note: ** the default demo content shipped with Open Social does not declare a language. The Content Translation module does not allow translation of LANGUAGE_NONE content. This is a bug in the demo content. Use fresh content for testing instead.

## Release notes
We've reworked the way we ensure that a user's shown language doesn't change when dealing with stream content. The new fine grained approach introduces less new issues and should make switching languages more consistent.

## Change Record
https://www.drupal.org/node/3098144
https://www.drupal.org/node/3098145
https://www.drupal.org/node/3098146
https://www.drupal.org/node/3098148


robertragas:
Unfortunately we still have some issues.
- **Groups overview page** leads to the group without the url. Once in the group the subcontent does get the correct url (also admin group overview, but same issue)
- **Events overview** read more link
- **Topics overview** read more link
- The **search autocomplete** suggested links miss the dash as in /nlnode3/
- **Search results** read more link (but off course related to the content-type thingy like above.

And there is 1 tricky part with the language detection, once you put user detection on top you cannot really switch languages anymore, so the language switcher should then still be disabled for logged in users if you chose to have user detection on top.

robertragas:
Approved but kicking off the travis again, seems like a false positive.

PR #1629, opened 2019-12-03
#3098456: Add language as a column to the user export

bramtenhove:
## Problem
The user export does not contain the users (preferred) language. Specially for multilingual platforms this is useful information.

## Solution
Added the preferred language of the user to the export. Defaults to the site default when it is not known.

## Issue tracker
https://www.drupal.org/project/social/issues/3098456

## How to test
- [ ] Enable Social Language, don't add any other language
- [ ] Export user(s)
- [ ] Observe the user(s) have the site default language as value in the "Language" column
- [ ] Add a language, assign it to one or more users
- [ ] Export user(s)
- [ ] Observe the user(s) have their preferred language or otherwise thesite default language as value in the "Language" column

## Release notes
It's important to know which language your users prefer in your communication to them. We now add the language, if known, to the user export.

## Change Record
n/a

## Translations
n/a

### DIFF
--- a/modules/custom/social_language/src/Plugin/UserExportPlugin/UserLanguage.php
+++ b/modules/custom/social_language/src/Plugin/UserExportPlugin/UserLanguage.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace Drupal\social_language\Plugin\UserExportPlugin;
+
+use Drupal\Core\Database\Connection;
+use Drupal\Core\Datetime\DateFormatterInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Language\LanguageManagerInterface;
+use Drupal\social_user_export\Plugin\UserExportPluginBase;
+use Drupal\user\UserInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'UserLanguage' user export row.
+ *
+ * @UserExportPlugin(
+ *  id = "user_language",
+ *  label = @Translation("Language"),
+ *  weight = -340,
+ * )
+ */
+class UserLanguage extends UserExportPluginBase {
+
+  /**
+   * The language manager service.
+   *
+   * @var \Drupal\Core\Language\LanguageManagerInterface
+   */
+  public $languageManager;
+
+  /**
+   * UserExportPluginBase constructor.
+   *
+   * @param array $configuration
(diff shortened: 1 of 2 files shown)


modules/custom/social_language/src/Plugin/UserExportPlugin/UserLanguage.php
```diff
@@ -0,0 +1,96 @@
...
+      $container->get('database'),
+      $container->get('language_manager')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getHeader() {
+    return $this->t('Language');
```
Kingdutch: I don't think we should translate file headers. This means that if a user with a different language tries to download this file and uploads it into an automated external system that system may break.

Do we do this anywhere else? Then we should probably strip this out everywhere (or make it configurable).

modules/custom/social_language/src/Plugin/UserExportPlugin/UserLanguage.php
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getValue(UserInterface $entity):string {
+    return $this->languageManager->getLanguage($entity->getPreferredLangcode())->getName();
```
Kingdutch: Should we output the langcode instead? This may be slightly less friendly to human consumers but makes automation massively easier.

modules/custom/social_language/src/Plugin/UserExportPlugin/UserLanguage.php
bramtenhove: We do this everywhere. Maybe not part of this scope?

modules/custom/social_language/src/Plugin/UserExportPlugin/UserLanguage.php
bramtenhove: I thought about this, but the language code itself is not very readable. I'll add another column for the language code.

modules/custom/social_language/src/Plugin/UserExportPlugin/UserLanguage.php
bramtenhove: We should also think if we want to do this. There are pros and cons on both sides.

In any case it should not be part of the scope of this ticket.

robertragas:
Approved, merged and cherry picked to 8.x-7.x

PR #1632, opened 2019-12-09
Issue #3097392 - Email batch operations to queue process

royharink:
## Problem
At this point we're sending emails through the interface. In large communities this can take up to 1 hour per 1000 users.
In some cases it might not even finish because it broke down before.

## Solution
Moving the email sending execution from the batch operation to the queue would mean emails are sent in the background.
A user can leave the batch interface and background processes take care of the execution.

## Issue tracker
https://www.drupal.org/project/social/issues/3097392

## How to test
The email actions are at multiple places:
### admin people overview
- [ ] Go to _/admin/people_.
- [ ] Select X amount of users, then select action _send email_ and click apply to selected items.
- [ ] Write an email and confirm you want to send an email.
- [ ] See that the batch process is much faster.
- [ ] Run the queue worker `drush queue-run user_email_queue` and see that emails are send.
- [ ] When all items are finished, check your the notifications and see that there is a new notification informing you about the finished process.

### group manage members
- [ ] Go to the manage members page of a group.
- [ ] Repeat the test steps from the admin people overview test.

### Event enrolments
- [ ] Go to the enrolment overview of an event.
- [ ] Repeat the test steps from the admin people overview test.
- [ ] See that anonymous 'guest' enrolments are also added. The main difference here is that they have a separate queue item and that these contain their email address instead of an UID.

## Release notes
When sending emails to users via a user overview page like people overview or group manage members. The bulk action is no longer sending the emails directly but creates queue items to be processed by a dedicated queue worker.
The bulk action to create the queue items is much faster then immediately sending the emails. The queue worker sends emails in batches in the background which is faster as well.
When the queue worker is finished sending emails the sender will receive a notification that notifies the user that the process has been completed in the background.

## Change Record
When developing an bulk operation action to send out emails, make sure it creates queue items instead of immediately sending the emails. This by making use of the `SocialSendEmail` class with the goal to add user id's to the queue items.

## Translations
- [x] Changed or removed source strings are added to the `translations.php` file.


royharink:
> Our conclusion is that this is something we want to have enabled by default. We'll need to install it by default and also add a post update hook.

Fixed this by generating a notification for the user with a custom message.

royharink:
> One other thing we discussed as that generating the Entity also added a lot of useless routes and functionality. We now have forms and entity view possibilities which we don't really want.
> 
> This entity is just for some backend storage, we shouldn't for example be able to add it manually.

The entity routes have been cleaned up. Some remain like, view edit and delete for the entity itself. This is only visible for administrator not for a site manager role.

modules/custom/social_queue_storage/social_queue_storage.install
```diff
@@ -0,0 +1,37 @@
...
+ * @file
+ * Install, update and uninstall functions for the social_queue_storage module.
+ */
+
+use Drupal\Core\Field\BaseFieldDefinition;
+
+/**
+ * Add finished field to the "queue_storage_entity" entity.
+ */
+function social_queue_storage_update_8001(&$sandbox) {
```
ronaldtebrake: Don't we want this as an install hook considering this is new? 🤔

modules/custom/social_queue_storage/social_queue_storage.install
```diff
@@ -0,0 +1,37 @@
...
+  $field_storage_definitions['finished'] = BaseFieldDefinition::create('boolean')
+    ->setName('finished')
+    ->setTargetEntityTypeId('queue_storage_entity')
+    ->setLabel(t('Finished'))
+    ->setDescription(t('Is the entity considered finished by the background task it was used in?'))
+    ->setInitialValue(FALSE)
+    ->setDefaultValue(FALSE);
+
+  $definition_update_manager->updateFieldableEntityType($entity_type, $field_storage_definitions, $sandbox);
+}
```
ronaldtebrake: 🆕

modules/custom/social_queue_storage/src/Form/QueueStorageEntityForm.php
```diff
@@ -0,0 +1,66 @@
...
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function save(array $form, FormStateInterface $form_state) {
+    $entity = $this->entity;
+
```
ronaldtebrake: some documentation would be helpfull here, what do we expect this status to contain considering we do a switch statement with only 1 case other than default. (For someone extending this could be nice)

modules/custom/social_queue_storage/src/Form/QueueStorageEntitySettingsForm.php
```diff
@@ -0,0 +1,53 @@
...
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   *
+   * @return array
+   *   Form definition array.
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $form['queuestorageentity_settings']['#markup'] = 'Settings form for Queue storage entity entities. Manage field settings here.';
```
ronaldtebrake: t('');

modules/social_features/social_event/modules/social_event_an_enroll/src/Plugin/Action/SocialEventAnEnrollSendEmail.php
```diff
@@ -103,31 +116,58 @@ public static function create(ContainerInterface $container, array $configuratio
...
+          'email_address' => $entity->field_email->value,
+          'display_name' => $this->getDisplayName($entity),
+        ];
       }
+    }
 
-      $this->configuration['display_name'] = $display_name;
+    if (!empty($guests)) {
+      // Create some chunks and then make queue items.
+      $chunks = array_chunk($guests, 10);
```
ronaldtebrake: Could be nice to make this configurable / or even just a setting in settings.php if we were to find a better sweet spot. But still, it's nais.

modules/custom/social_queue_storage/social_queue_storage.install
royharink: This was mainly there for @bramtenhove.
It could indeed be removed.

modules/social_features/social_event/modules/social_event_an_enroll/src/Plugin/Action/SocialEventAnEnrollSendEmail.php
royharink: The parent class: \Drupal\social_user\Plugin\Action\SocialSendEmail has todo's for this.
Currently out of scope.

modules/custom/social_queue_storage/src/Plugin/ActivityEntityCondition/QueueStorageEntitySendActivityCondition.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+
```
bramtenhove:
```suggestion
```
Remove the extra empty line.

bramtenhove:
@ronaldtebrake, do you want to do a final check? Can we merge it afterwards?

There's at least one change needed still: https://github.com/goalgorilla/open_social/pull/1632#discussion_r362838995.

ronaldtebrake:
Looks good to me, just wondering if this is something we need to have enabled by default or maybe add as an optional module to the social.profile optional module list. So members can enable it on install through the UI.

royharink:
It's probably best to enable this by default as the vbo actions are overridden and it won't work anymore.

PR #1642, opened 2019-12-16
Issue #3101326 by robertragas: Update the config to include notificat…

robertragas:
…ions for creating content in public, secret and flexible groups

## Problem
When creating an event or topic within public, secret or flexible group we are not getting any notification in the notification centre or even emails.

This is being caused by the message template not being triggered.

## Solution
Add the correct bundles to the config and for groups that are optional include them in the config override.

## Issue tracker
https://www.drupal.org/project/social/issues/3101326

## How to test
- [x] Create a public, secret or flexible group and add 1 additional member to it
- [x] Create a topic and run the cron
- [x] Note that there is no email notification or notification in the account menu. You can also see the in the queue table or message table that there are no new entries.
- [x] Apply patch and try again

## Release notes
We solved an issue with the message template item "A person created a post, event or topic in a group I joined" where it would not trigger for secret, flexible and public groups resulting in no notifications being sent when someone would create content in the group.

### DIFF
--- a/modules/social_features/social_activity/config/install/message.template.create_content_in_joined_group.yml
+++ b/modules/social_features/social_activity/config/install/message.template.create_content_in_joined_group.yml
@@ -10,6 +10,8 @@ third_party_settings:
       group_content-closed_group-group_node-topic: group_content-closed_group-group_node-topic
       group_content-open_group-group_node-event: group_content-open_group-group_node-event
       group_content-open_group-group_node-topic: group_content-open_group-group_node-topic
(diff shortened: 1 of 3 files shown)


navneet0693:
Merged and Cherry Picked on 0245999d80a938b3976039d32fcf16bec9c10c05

Kingdutch:
@navneet0693 When cherry-picking, please make sure the milestone is correct. This has been missed in the 7.x release notes.

Also after cherry-picking please change `status: needs backport` into `verified-cherry-picks` so we know the cherry-pick has been done : )

cc @ronaldtebrake

PR #1646, opened 2019-12-18
Issue #3101792 by agami4: Update private messages on mobile

agami4:
## Problem
The private messages are not on the full width on mobile devices

## Solution
Update width for the private messages on mobile devices

## Issue tracker
https://www.drupal.org/project/social/issues/3101792

## How to test
- [ ] Need to send a private message to any user

## Release notes
The private messages have full width (width: 100%) on the mobile devices

### DIFF
--- a/themes/socialbase/assets/css/message.css
+++ b/themes/socialbase/assets/css/message.css
@@ -83,6 +83,7 @@
 }
 
 .private-message-full {
+  width: 100%;
   margin-bottom: 2rem;
 }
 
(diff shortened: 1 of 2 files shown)


Kingdutch:
Hi @agami4 ,

In this case I think it's better to set a `max-width` than a `width` for the reasons mentioned. That should solve the issue for long messages while not affecting short ones.

In a follow-up we can look at improving the overall look for distinguishing what's sent "from" and what's sent "to" the user : )

Kingdutch:
Cherry picked to 8.x-7.x in d46cab4ec

PR #1657, opened 2020-01-08
Issue #2905325: Added conditions to consider 'edit current entity' permission for current user to view unpublished nodes not owned by them.

navneet0693:
## Problem
A content manager loses access to nodes which are not authored by him/.her, if he/she un-publishes it.

## Solution
Added check for considering permission to edit any entity.

## Issue tracker
https://www.drupal.org/project/social/issues/2905325

## How to test
- [ ] Login as Content Manager/Site Manager
- [ ] Unpubslih a node which not authored by the current user
- [ ] Current user should be able to access the unpublished node.

## Screenshots
N.A.
## Release notes
N.A.

## Change Record
N.A.

## Translations
N.A.


modules/custom/entity_access_by_field/src/EntityAccessHelper.php
```diff
@@ -21,19 +21,30 @@ class EntityAccessHelper {
    * @return array
    *   An array containing a list of values to ignore.
    */
-  public static function getIgnoredValues() {
+  public static function getIgnoredValues(): array {
     return [];
```
sjoerdvandervis: Hi @navneet0693 ,

Already discussed this a bit with @ronaldtebrake , can you think of any reason why we need this helper function as it merely returns an empty array?

navneet0693: @sjoerdvandervis I am not very sure on this and haven't touched the internal code. The changes were introduced in https://github.com/goalgorilla/open_social/pull/422 which was specifically changed in this commit: https://github.com/goalgorilla/open_social/pull/422/commits/09e071c7d6b939471a686272a89134ed8c53a983. Also I see a TODO added https://github.com/goalgorilla/open_social/blob/8.x-8.x/modules/custom/entity_access_by_field/src/EntityAccessHelper.php#L19

ronaldtebrake:
Moved to 7.3 as per comment by Roy

ronaldtebrake:
Updated this to 7.7 as it requires work still would be great to have it in the next release though :)

ronaldtebrake:
Outdated, can always be re-opened

PR #1659, opened 2020-01-09
#3105285 by ronaldtebrake: Add support for the layout builder in Open Social

ronaldtebrake:
## Problem
Right now we have Social Landingpages built on top of paragraphs. This works well but causes quite some maintenance burden. We need to support the layout builder in the theme Open Social ships with.

## Solution
We propose to move in to using the layout builder from Drupal Core. So users can create their own entities with layout builder configuration. Dashboards, landingpages all of this will become available.

## TODO
- [ ] Make sure theme supports the right margins / padding / card classes etc. 

## Issue tracker
- [ ] https://www.drupal.org/project/social/issues/3105285

## How to test
- [ ] Create a entity
- [ ] Add layout builder configuration on form displays
- [ ] See that the grid / layout works like a charm.

## Release notes
Open Social and its themes now comes with support for the layout builder. This enables you to use the layout builder on your custom entities and create beautiful flexible pages.

## Change Records
https://www.drupal.org/node/3107940


modules/social_features/social_core/src/Service/LayoutService.php
```diff
@@ -0,0 +1,34 @@
...
+use Drupal\layout_builder\LayoutEntityHelperTrait;
+
+/**
+ * Class LayoutService.
+ *
+ * @package Drupal\social_core\Service
+ */
+class LayoutService {
+
+  use LayoutEntityHelperTrait;
```
Kingdutch: @ronaldtebrake Unfortunately this breaks if layout_builder is not enabled because Drupal's autoloading is dependant on enabled modules.

An example of a similar scenario that broke is here: https://www.drupal.org/project/social/issues/3102067

Kingdutch: Sorry, I did not yet see that we only enabled this service when the layout_builder is enabled. However, that probably means we can remove the snippet on line 27.
```
if (!\Drupal::moduleHandler()->moduleExists('layout_builder')) {
      return FALSE;
    }
```

This will never run because `moduleExists('layout_builder')` will always be true if this service is registered :)

ronaldtebrake: Yeah was merily a double measurement to avoid any other things registering it in the wrong time and or place. Not sure if that ever happens but felt like a good better safe than sorry thingy.

ronaldtebrake:
Leaving this open for now, we need to fix the modals 
Screen Shot 2020-01-23 at 14 50 05

Screen Shot 2020-01-23 at 14 42 37
Screen Shot 2020-01-23 at 14 49 59

ronaldtebrake:
All green and cherry-picked and merged!

PR #1666, opened 2020-01-17
Issue #3080369 by robertragas: add defensive coding to stop crashing …

robertragas:
…the cronjob when sending activity emails with old deleted data

## Problem
Currently when running the cron, it will go through the queue and based on that create activities with different kind of tokens.

In our case it seems like an enrollment or person got deleted, but the cronjob didn't run yet, so now when running the cronjob it tries to get an enrollment from a non-existing entity. This results in an error message and causes the cronjob to not run, building up huge amounts.
This also resulting in more of the same errors because if the cron hasn't run for a while, the possibilities of more things that got deleted are bigger.

## Solution
Solution would be to be more defensive in coding.

## Issue tracker
https://www.drupal.org/project/social/issues/3080369

## How to test
- [ ] Create enrollments for events so that the queue gets filled with email notifications
- [ ] Delete the source (users/events) of the enrollments so it cannot retrieve the name anymore.

## Release notes
We added more defensive coding when generating email notifications for the activities. In some cases it crashes because the source has been deleted.

### DIFF
--- a/modules/social_features/social_activity/social_activity.tokens.inc
+++ b/modules/social_features/social_activity/social_activity.tokens.inc
@@ -7,6 +7,7 @@
 
 use Drupal\Core\Render\BubbleableMetadata;
 use Drupal\Core\Url;
+use Drupal\group\Entity\GroupContentInterface;
 
 /**
  * Implements hook_token_info().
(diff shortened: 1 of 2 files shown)


ronaldtebrake:
Shouldn't we backport this still to 6.x? 🤔

ronaldtebrake:
Also cherry picked

PR #1667, opened 2020-01-20
#3102763 - Add new Views Filter for flexible group content 

ronaldtebrake:
## Problem
Filter our content placed in Flexible groups correctly based on visibility settings, group memberships etc.

## Solution
Add a new views filter, we can add these correctly to the views we need.

## Issue tracker
https://www.drupal.org/node/3102763

## How to test
- [ ] Will do pair programming with Robert 

## Release notes
We have added a new views filter for Flexible group content


robertragas:
@ronaldtebrake 

I added the filter as a config override to the views that needed it. So only still need to create some additional behat tests, and/or fix ones that are now broken. Not sure if the flexible groups are even in a test scenario currently.

robertragas:
Travis now running, Changed the following:
- Had to add the permission "administer content" as a exception otherwise CM will not see the content correct
- Added the behat test similar to creating a public group, but added some extra steps to check out visibility on public and group members only on the all-topics overview and search.

robertragas:
Merged and cherry picked to 8 and 7

PR #1689, opened 2020-01-28
Issue #3109219 by robertragas: Update the user data of the recipient …

robertragas:
…after receiving a welcome message to show it as unread in the account header menu

## Problem
After joining a group with the group welcome message enabled, the PM inbox doesn't show there is a new PM in the inbox, is the cache cleared correctly?

## Solution
Update the user data for the recipient for the concerned thread to set the last access time of the created private message to 0 so it shows up as unread in the account header menu.

## Issue tracker
https://www.drupal.org/project/social/issues/3109191

## How to test
- [ ] Create a group and set a welcome message.
- [ ] Create a new authenticated user and join the group
- [ ] Notice that you don't see the unread counter in the account header menu, but the mail does show as unread in the detail view.
- [ ] Check out to this branch and repeat, see it now shows the unread counter correctly.
- [ ] Try out multiple scenario's, join other groups, or whatever you think of.

## Release notes
Social_welcome_message not released yet.


bramtenhove:
@robertragas, before I start testing this I have a quick question.

Does it also mean that the user does not get a notification that they received a private message? Or is it just that the unread counter is not updated?

robertragas:
> @robertragas, before I start testing this I have a quick question.
> 
> Does it also mean that the user does not get a notification that they received a private message? Or is it just that the unread counter is not updated?
> 
> -- edit
> Another question, isn't this something we should consider solving in the Private Message module? What are your thoughts?

AFAIK only the read counter is not updated. Private messages don't get a notification in the centre currently I believe, because the private message message template only has the trigger to send out an email.

There are a few ways of solving it, I took this way as it was sufficient for now.
1. Fix it in the private message module, but couldn't really find related code to that private message in the 1.x version. But we could also consider using private message 2.x with that patch, saves us time fixing a version that will be succeeded anyway.
2. Also normally when someone would send a private message it does not create this users_data for the receiver. If we can just stop this from creating it would also solve the problem, but likely also a problem in the contrib module. Also don't saw any methods to prevent this from happening when making the thread/private message.
3. Change our own method updateUnreadCount to keep this created value thing in my mind, but wouldn't advise this if it becomes a lot heavier since this method loads quite often.

robertragas:
Are you sure you didn't had more private messages? I can't reproduce this problem.

bramtenhove:
I'm not sure anymore. @robertragas, if you can show me later we can merge it.

robertragas:
I changed the base branch to 9.x 

@bramtenhove  I also made a Loom video where I show where the notification disappears after reading.
https://www.loom.com/share/e92dc393aa734c949da1f254c441239e

ronaldtebrake:
Also in 8.x cherry picked

PR #1692, opened 2020-01-28
Issue #310855 by sjoerdvandervis: Quickpost with ctrl/cmd + enter

sjoerdvandervis:
## Problem
Since the introduction of Ajax comments, the quickpost functionality was not working anymore (Cmd / Ctrl + Enter).

## Solution
The `click()` function is not possible to be used as it seems, as it triggers an ajax call. When using a `mousedown()` function, it does trigger the quickpost action.

## Issue tracker
https://www.drupal.org/project/social/issues/3108558

## How to test
- [ ] Make sure you have `social_ajax_comments` enabled;
- [ ] Respond with a comment to a post in your stream. See that the Ajax post is triggered now.
- [ ] Create a new post, use cmd/ctrl + enter and see this still works
- [ ] Mention a user in a comment on a post to see that still works
- [ ] Add a comment to a topic or event to see that it still works
- [ ] Disable `social_ajax_comments` and repeat the previous to see you didnt cause regression

### DIFF
--- a/modules/social_features/social_post/js/keycode-submit.js
+++ b/modules/social_features/social_post/js/keycode-submit.js
@@ -24,7 +24,7 @@
           // post forms this has some side effects that are required for a
           // successful submission.
           if ($submit.length) {
-            $submit.click();
+            $submit.mousedown();
           }
           // If a submit button isn't found we fall back to submitting the form
           // outright.
--- a/modules/social_features/social_post/js/keycode-submit.js
+++ b/modules/social_features/social_post/js/keycode-submit.js
@@ -24,7 +24,16 @@
           // post forms this has some side effects that are required for a
           // successful submission.
           if ($submit.length) {
-            $submit.mousedown();
+            // If it's a submit button, used by Ajax comments
+            // the click event is canceled by ajax_comments, see
+            // e.preventDefault() in ajax_comments.js.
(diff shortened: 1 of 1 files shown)


modules/social_features/social_post/js/keycode-submit.js
```diff
@@ -24,7 +24,7 @@
           // post forms this has some side effects that are required for a
           // successful submission.
           if ($submit.length) {
-            $submit.click();
+            $submit.mousedown();
```
ronaldtebrake: Hi @sjoerdvandervis 
Have you checked if this actually doesn't cause any regression without Ajax comments enabled? 
Click doesn't work probably due to the ajax_comments module stopping propagation, I'm not sure why but it sounds like there was a thought behind it. 

Just curious :)

ronaldtebrake:
Unfortunately this causes regression in the post form. When using the short key submit in there, the post isn't created. Instead you see an ajax throbber and nothing happens.
Putting it back to needs work.

ronaldtebrake:
@bramtenhove //cc @sjoerdvandervis 
I have updated the PR and the test steps to check for regression I found retest whenever you have time :)

bramtenhove:
Thanks for test steps @ronaldtebrake, I'll thoroughly test it this afternoon.

PR #1701, opened 2020-01-31
Issue #3110341 by royharink: Embedded Vimeo videos are sometimes blocked when hosted on cloud hosting

royharink:
## Problem
When Open Social is hosted on cloud hosting providers it sometimes happens that Vimeo blocks requests for embedding videos. This happens because some of these providers are on a blacklist at Vimeo, see https://github.com/oscarotero/Embed/issues/326#issuecomment-531739948.
Because the requests are blocked, the videos aren't loaded and therefor not visible.

## Solution
Within OpenSocial the url_embed module is used to embed media such as Vimeo. The module makes us an older version of the Embed library which is v2.7.8) while a newer 3.x.x version is released. This newer version has better support for Vimeo but requires the url_embed module to be refactored.

Since updating the contrib module isn't a solution as we are on the latest version and there isn't any progress on a newer version.
The solution is to add support for Vimeo to the Embed library.

## Issue tracker
https://www.drupal.org/project/social/issues/3110341

## How to test
- [ ] Create or edit a node.
- [ ] Add a Vimeo url via embed.
- [ ] When loading the the node, see that the video loads.
- [ ] Do the same on cloud hosting such as Pantheon.

## Translations
- _None_


robertragas:
Fix works correctly, but think for formal reasons, we should also post this patch to the embed module, in case other people in the community face the same issue.

PR #1705, opened 2020-02-03
Issue #3110778 by rolki: Fixed an issue that not set the type for the Post entity

rollki:
## Problem
The `setType` method doesn't work correctly for the Post entity. Try to set a new Post type to some entity and you will get the previous type

## Solution
In the method setType set type from parameter instead of current bundle

## Issue tracker
https://www.drupal.org/project/social/issues/3110778

## How to test
Try to set a new Post type to some entity and you should get the new type

### DIFF
--- a/modules/social_features/social_post/src/Entity/Post.php
+++ b/modules/social_features/social_post/src/Entity/Post.php
@@ -144,7 +144,7 @@ class Post extends ContentEntityBase implements PostInterface {
    * {@inheritdoc}
    */
   public function setType($type) {
-    $this->set('type', $this->bundle());
+    $this->set('type', $type);
     return $this;
   }
 


ronaldtebrake:
@rollki @ribel 
Why is there a maintainer input tag on this very small minor issue. I don't see why we need to provide input on this if you have tested this. 
Feel free to merge it once its reviewed, it's on the corect base branch, and cherry-picked back to 7.x 

Thanks.

ribel:
@ronaldtebrake, You are actually right. I have merged this PR and cherry-picked to 8.x-8.x.
Should I also cherry-pick this to 8.x-7.x? And to what status should I change issue on drupal.org?

ronaldtebrake:
Hi @ribel 
Thanks :) !
Yes please, since this issue is also in 7.x and that's our latest supported release we also cherry-pick this back to 7.x
Afterwards you can put the ticket on fixed ^^

PR #1673, opened 2020-01-24
Issue #3087026 by mdeny: Social Group Welcome Message

denysyukm:
## Problem
Private message when successfully joining the group

## Solution
Implement Social Group Welcome Message module

## Issue tracker
https://www.drupal.org/project/social/issues/3087026
https://getopensocial.atlassian.net/browse/YANG-1463

## How to test
Enable social_group_welcome_message
- [ ] log in as group manager
- [ ] go the the group you manage
- [ ] go to the edit form of the group
- [ ] set "Send a private message people joined"
- [ ] write Welcome message
- [ ] join this group in any way as another user
- [ ] check private messages

## Release notes
Implements possibility to send private message when successfully joining the group

PR #1708, opened 2020-02-05
Issue #3098031 by Kingdutch, ronaldtebrake: Evaluate which modules sh…

Kingdutch:
…ould be marked as experimental

## Problem
There are quite a few modules that we ship with Open Social which we've marked as experimental. Some of these have been experimental for quite a few versions and can probably be marked as stable and supported at this point.

Before the 8.x release we should evaluate the list.

## Solution
Mark modules as stable (see release notes)

## Issue tracker
https://www.drupal.org/project/social/issues/3098031

## How to test
- [ ] No functional change

## Release notes
The following modules are no longer considered experimental: Social Embed, Social Private Message, Social Profile Fields, Social Profile Privacy, Social Profile Organization Tag, Social Search Autocomplete.

### DIFF
--- a/modules/custom/social_ajax_comments/social_ajax_comments.info.yml
+++ b/modules/custom/social_ajax_comments/social_ajax_comments.info.yml
@@ -2,6 +2,6 @@ name: Social Ajax Comments
 type: module
 description: Integration with Ajax Comments.
 core: 8.x
-package: Social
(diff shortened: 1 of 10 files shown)


bramtenhove:
I marked the module added in #1673 as experimental. There are still some open bugs.

PR #1709, opened 2020-02-05
Issue #3111353 - Ensure we run test against all optional modules

ronaldtebrake:
Merge after: https://github.com/goalgorilla/open_social_scripts/pull/30/files

# Problem 
We want to run tests using all the optional modules by default, to ensure we encounter issues in our stability tests with everything enabled, but also ensure we test at least stability-1 using no optional modules. To test scenario's that might have misplaced config etc.

## Solution
1. Updated the scripts in https://github.com/goalgorilla/open_social_scripts/pull/30/files
That will give us the flag --with-optional
2. This is now used for all the stability tests
3. Added a new test suite, this one will run the normal install_script without the flag to ensure we run the same tests as install_update

## Issue tracker
https://www.drupal.org/project/social/issues/3111353

## How to test
- [ ] Make sure travis runs!

## Release notes
We now also test all optional modules that have a `module.installer_options.yml`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ matrix:
   - env: TEST_SUITE=install_stability_3
   - env: TEST_SUITE=install_stability_4
   - env: TEST_SUITE=install_update
+  - env: TEST_SUITE=install_without_optional
   - env: TEST_SUITE=drush_make
 #  - env: TEST_SUITE=accessibility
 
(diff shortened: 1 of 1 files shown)


Kingdutch:
@ronaldtebrake @bramtenhove This is actually green and can be merged. Unfortunately, the drupal security advisory breaks our upgrade path testing.

As of last night VBO 2.5.0 is marked as insecure so it won't be installed 🤦‍♂

PR #1712, opened 2020-02-06
Issue #3089112 by royharink: Registration field specific validation error messages

royharink:
## Problem
When a user tries to create an account using an e-mail address and username.
The process returns the wrong error message when the username already exists.
Currently, the message states that there is an error with the e-mail address, while it's actually the username that gives the problem.

## Solution
A better distinction has been made in the field that didn't pass validation. The message is kept generic by (still) removing the original Drupal message, trying not to disclose information about the existence of the user.
The generic message is left in place to inform about the information disclosure but hasn't been assigned to a field but to the form.

## Issue tracker
https://www.drupal.org/project/social/issues/3089112

## How to test
- [ ] To to the registration page
- [ ] Try registering with a non-existing email address and an existing username.
- [ ] See that the username is validated wrong, the message is generic and that the information disclosure error is still appearing.

## Screenshots
image

## Release notes
Registration form now have more distinct messages.

## Translations
Updated strings are now added to translation file.

### DIFF
--- a/modules/social_features/social_user/social_user.module
+++ b/modules/social_features/social_user/social_user.module
@@ -236,21 +236,38 @@ function social_user_form_user_register_form_alter(&$form, FormStateInterface $f
  * Validate function for the user register form.
  */
 function social_user_register_validate(&$form, FormStateInterface $form_state) {
-
   // Fetch input.
   $input = $form_state->getValues();
+  // Check if user data with the input can be found.
+  $mail_exists = user_load_by_mail($input['mail']);
+  $name_exists = user_load_by_name($input['name']);
+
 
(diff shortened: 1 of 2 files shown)


bramtenhove:
I think Roy made a good first step here, although it needs some work before it's finished.

@robertragas, @sjoerdvandervis, @navneet0693, @frankgraave, can you finish the work?
Please check with someone about the language we use here. Do we still want to say `Oops, ...`?

navneet0693:
@bramtenhove I will do it today ;-)

navneet0693:
Cherry-picked on 004df29fd3cdafc01870779835efcab3ba89b676

PR #1714, opened 2020-02-07
Issue #3111959 by chmez: As a CM I want to customise filtering options in the custom content block

OH17L:
## Problem
As a CM I want to customise filtering options, such as CoP, in the custom content block

## Solution
- Should be possible to select one or multiple taxonomies
- All entities applicable tagged with selected taxonomy should show up in the results block

## Issue tracker
- https://www.drupal.org/project/social/issues/3111959
- https://getopensocial.atlassian.net/browse/YANG-1999

## How to test
- [x] When you open **Add Custom content list block custom block** page (_/block/add/custom_content_list_) then you should see that these page doesn't use an administrative theme.
- [x] **Block description** and **Subtitle** fields should be in the **Information** section.
- [x] **Content** section should contain the following fields: **Type of content** (mandatory), **Content selection**, **Topic type**, **Content tags**, **Group**.
- [x] **Type of content** field should be locked.
- [x] You should see the **Topic** option in the **Type of content** field.
- [x] You should see the **All** option in the **Content selection** field.
- [x] When you select the **Topic type** option in the **Content selection** field then you shouldn't see **Content tags** and **Group** fields.
- [x] **Sorting** section should contain the following mandatory fields: **Sorting**, **Number of items**.
- [x] **Read more link** section should contain the following fields: **URL**, **Link text**.
- [x] Values from the hidden fields shouldn't be applied to filtering topics in the block.

## Release notes
Extend configuration form of the **Custom content list block** custom block by adding new fields for selecting filters based on the new plugin


modules/social_features/social_content_block/social_content_block.install
```diff
@@ -0,0 +1,95 @@
...
+
+    $sandbox['total'] = count($sandbox['configs']);
+    $sandbox['current'] = 0;
+    $sandbox['path'] = drupal_get_path('module', 'social_content_block') . '/' . InstallStorage::CONFIG_INSTALL_DIRECTORY;
+  }
+
+  $source = new FileStorage($sandbox['path']);
+  $config_storage = \Drupal::service('config.storage');
+  $config = $sandbox['configs'][$sandbox['current']++];
+  $config_storage->write($config, $source->read($config));
```
Kingdutch: We shouldn't read configuration from the folder because it may change. This means that running this update hook now can be very different from running this update hook in a year. The update hook should always do the same, no matter when it's run.

modules/social_features/social_content_block/social_content_block.install
```diff
@@ -0,0 +1,95 @@
...
+
+  $sandbox['#finished'] = $sandbox['current'] / $sandbox['total'];
+}
+
+/**
+ * Fill in fields for plugin ID and plugin fields in existing blocks.
+ */
+function social_content_block_update_8002(&$sandbox) {
```
Kingdutch: Not sure if this should be ah update hook or a post_update hook @ronaldtebrake any ideas?

modules/social_features/social_content_block/social_content_block.install
```diff
@@ -0,0 +1,95 @@
...
+  if (!isset($sandbox['total'])) {
+    $sandbox['total'] = \Drupal::entityQuery('block_content')
+      ->condition('type', 'custom_content_list')
+      ->count()
+      ->execute();
```
Kingdutch: We should probably add a condition here to check for the thing we're setting being empty so that if this gets run later it won't cause any issues.

modules/social_features/social_content_block/social_content_block.install
```diff
@@ -0,0 +1,95 @@
...
+    if (!$entity->get($field)->isEmpty()) {
+      $fields[] = $field;
+    }
+  }
+
+  if (count($fields) === 1) {
+    $entity->field_plugin_field->value = reset($fields);
+  }
+
+  $entity->save();
```
Kingdutch: We should probably do more than one entity per batch loop, otherwise running this through the UI would cause a lot of network requests. We have some other update hooks that use a batch_size setting for this and that have the loop logic, you can use those as examples.

modules/social_features/social_content_block/social_content_block.module
```diff
@@ -56,25 +56,50 @@ function _social_content_block_is_custom_content_block(BlockPluginInterface $blo
 }
 
 /**
- * Implements hook_form_BASE_FORM_ID_alter().
+ * Implements hook_form_alter().
  */
-function social_content_block_form_block_content_custom_content_list_form_alter(&$form, FormStateInterface $form_state, $form_id) {
-  // Add general description above first field to explain what the filters do.
-  $form['field_topic_type'] += [
-    '#prefix' => t('To make the list of topics more specific you can additionally configure more filters such as topic types, content tags and groups.'),
```
Kingdutch: This removed translation should be added to the translations.php file with a proper description.

Additionally, this description is not brought back so should we add something somewhere to show the user what should be done?

modules/social_features/social_content_block/social_content_block.services.yml
```diff
@@ -8,6 +11,12 @@ services:
     class: Drupal\social_content_block\ContentBuilder
     arguments:
       - '@entity_type.manager'
-      - '@database'
       - '@module_handler'
       - '@string_translation'
+      - '@plugin.manager.content_block'
+      - '@database'
```
Kingdutch: Changing the argument of services is a breaking change because it breaks extending services. Adding an argument may be possible in a minor version if there's a fallback for when the service is not present. Changing the order is always a breaking change due to mismatched types.

modules/social_features/social_content_block/src/ContentBlockBase.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Drupal\social_content_block;
+
+use Drupal\Component\Plugin\PluginBase;
+
+/**
+ * Class ContentBlockBase.
+ *
+ * @package Drupal\social_content_block\Plugin\ContentBlock
+ */
```
Kingdutch: Any comment on what a ContentBlock is? How they should be used? What makes a ContentBlock different from the normal Drupal Blocks?

modules/social_features/social_content_block/src/ContentBlockPluginInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Drupal\social_content_block;
+
+use Drupal\Core\Database\Query\SelectInterface;
```
Kingdutch: Do we need to consider making this a SelectInterface that's not database specific? Perhaps we want to use these blocks for a different data source (like search, or the API of another OS platform) in the future?

cc @ronaldtebrake

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -36,109 +35,117 @@ class ContentBuilder implements ContentBuilderInterface {
...
-      $query->condition('tt.field_topic_type_target_id', $topic_types, 'IN');
+  protected function getEntities(BlockContentInterface $block_content) {
+    $plugin_id = $block_content->field_plugin_id->value;
+    $definition = $this->contentBlockManager->getDefinition($plugin_id);
+
+    if ($block_content->field_plugin_field->isEmpty()) {
+      $field_names = $definition['fields'];
+    }
+    else {
+      $field_names = [$block_content->field_plugin_field->value];
```
Kingdutch: Maybe some comments on what the difference is? When would the field stored on the content block differ from the default field set of the plugin? Can the user select a subset of fields?

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -36,109 +35,117 @@ class ContentBuilder implements ContentBuilderInterface {
...
     }
 
+    $plugin->query($query, $fields);
+
     // Allow other modules to change the query to add additions.
     $this->moduleHandler->alter('social_content_block_query', $query, $block_content);
 
     // Add sorting.
-    $query->orderBy('n.' . $block_content->getFieldValue('field_sorting', 'value'));
+    $query->orderBy('base_table.' . $block_content->field_sorting->value);
```
Kingdutch: SQL Injection opportunity?

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -36,109 +35,117 @@ class ContentBuilder implements ContentBuilderInterface {
...
+    /** @var \Drupal\Core\Entity\EntityTypeInterface $entity_type */
+    $entity_type = $this->entityTypeManager->getDefinition($definition['entityTypeId']);
+
+    /** @var \Drupal\Core\Database\Query\SelectInterface $query */
+    $query = $this->connection->select($entity_type->getBaseTable(), 'base_table')
+      ->fields('base_table', [$entity_type->getKey('id')]);
+
+    if ($definition['bundle']) {
+      $query->condition(
+        'base_table.' . $entity_type->getKey('bundle'),
```
Kingdutch: Is the table name of condition sanitized? 🤔

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -188,15 +195,18 @@ public function build($entity_type_id, $entity_id) {
...
       return [];
     }
 
     $data = [
       '#theme' => 'social_content_block',
       '#title' => $block_content->label(),
-      '#subtitle' => $block_content->getFieldValue('field_subtitle', 'value'),
-      '#topics' => $this->getTopics($block_content),
+      '#subtitle' => $block_content->field_subtitle->value,
+      '#topics' => $this->getEntities($block_content),
```
Kingdutch: If entities are changed from topics, should we change the key name too?

modules/social_features/social_content_block/social_content_block.services.yml
OH17L: Fixed

modules/social_features/social_content_block/social_content_block.install
OH17L: Fixed

modules/social_features/social_content_block/social_content_block.install
OH17L: Fixed

modules/social_features/social_content_block/social_content_block.install
OH17L: Fixed

modules/social_features/social_content_block/social_content_block.module
OH17L: According to the new design it doesn't need anymore. So I added these phrases to the translations.php file

modules/social_features/social_content_block/src/ContentBuilder.php
OH17L: What do you mean?

modules/social_features/social_content_block/src/ContentBuilder.php
OH17L: What do you mean?

modules/social_features/social_content_block/src/ContentBlockBase.php
OH17L: Done

modules/social_features/social_content_block/src/ContentBuilder.php
OH17L: > Maybe some comments on what the difference is?

Done

> When would the field stored on the content block differ from the default field set of the plugin?

A stored field should be added to one from plugins

> Can the user select a subset of fields?

Yes, it's possible when select **All** option in the **Content selection** field

modules/social_features/social_content_block/src/ContentBuilder.php
OH17L: I think no because the **field_sorting** field has a fixed list of allowed values.

modules/social_features/social_content_block/src/ContentBlockPluginInterface.php
OH17L: I think that we can do it in the future when we'll entirely sure that it's really needed

Kingdutch:
The block wouldn't render on layout builder pages because the correct
hook was not implemented.

Additionally, implementing that hook exposed some bugs. When no fields
were selected, the `empty` value of the field selector would not
actually be empty but would be interpreted as a special field. This is
fixed by checking for that field name, although this will break if a
field named `all` is every added. So it's probably better to just make
the field be actually empty.

Additionally a query alter would fail if there were no filters selected
because it attempted to do a join with an `IN` clause without and
values. This is not allowed in a SQL database.

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -36,109 +35,130 @@ class ContentBuilder implements ContentBuilderInterface {
...
+    else {
+      $field_names = [$block_content->field_plugin_field->value];
+    }
+
+    $fields = [];
+
+    foreach ($field_names as $field_name) {
+      $field = $block_content->get($field_name);
+
+      if ($field->isEmpty()) {
```
Kingdutch:
```suggestion
      if (!$field->isEmpty()) {
```

modules/social_features/social_content_block/src/ContentBlockPluginInterface.php
Kingdutch: Well, it's not something that can be done in the future because it would break the interface :) But maybe it's not worth the time now.

modules/social_features/social_content_block/src/ContentBuilder.php
OH17L: Fixed

modules/social_features/social_content_block/src/SocialContentBlockOverride.php
```diff
@@ -32,6 +39,10 @@ class SocialContentBlockOverride implements ConfigFactoryOverrideInterface {
    */
   public function __construct(ConfigFactoryInterface $config_factory) {
     $this->configFactory = $config_factory;
+
+    if (\Drupal::hasService('plugin.manager.content_block')) {
```
ribel: @OH17L Is this condition needed since this service is implemented in the same module?
Also, can we use Dependency Injection here?

modules/social_features/social_event/config/install/core.entity_view_display.node.event.small_teaser.yml
```diff
@@ -13,50 +13,42 @@ dependencies:
     - field.field.node.event.field_event_image
     - field.field.node.event.field_event_location
     - field.field.node.event.field_files
+    - image.style.social_medium
     - node.type.event
   module:
     - datetime
+    - image
     - user
-_core:
```
ribel: @OH17L I don't think we need to remove `default_config_hash` with a new config management flow.

modules/social_features/social_group/config/install/core.entity_view_display.group.closed_group.small_teaser.yml
```diff
@@ -28,7 +38,7 @@ hidden:
   created: true
   field_group_address: true
   field_group_description: true
-  field_group_image: true
   field_group_location: true
   search_api_excerpt: true
+  social_tagging: true
```
ribel: @OH17L Do we need to add this, since `social_tagging` is optional?

ribel:
@OH17L,
I have tested functionality and checked the code and in general, it looks good and solid to me.
Just please do a few more following things: 

1. Title and problem here don't really reflect those big changes that we made in this PR. So please update it with explanation what was really done here, for the distro. As I can see you made two main things here: a) Made custom content block pluggable, b) Improve UI of the custom content block form.

2. Please check my comments regarding removing `default_config_hash` and new dependencies from `social_tagging` in small_teaser configs. Also please check open comments from @Kingdutch.

@Kingdutch, I am not really sure if this task should be blocked by tests, since this is not new functionality, but existing, and we just improve it here. Also, I think even these changes that we made was not in the original part of our story, so, unfortunately, I am afraid that we don't have more time to work on new behat tests.

modules/social_features/social_content_block/src/SocialContentBlockOverride.php
OH17L: @ribel it's needed because the config overrides service executing before the process for creating a new service. Also, this is the reason why we can not inject the service

modules/social_features/social_group/config/install/core.entity_view_display.group.closed_group.small_teaser.yml
OH17L: @Kingdutch I see that you did it [here](https://github.com/goalgorilla/open_social/blame/feature/3111959-extend-content-block/modules/social_features/social_group/config/install/core.entity_view_display.group.closed_group.small_teaser.yml#L43). Can you explain why?

modules/social_features/social_event/config/install/core.entity_view_display.node.event.small_teaser.yml
OH17L: @Kingdutch I see that you did it [here](https://github.com/goalgorilla/open_social/pull/1714/commits/e249005fe9a2e3432cde0ef8cfa3e588d61d8ef8#diff-bfe5a2c29641f5ddd592c5706b6cb912L20). Can you explain why?

modules/social_features/social_event/social_event.install
```diff
@@ -167,3 +179,31 @@ function social_event_update_8801() {
...
+  $updateHelper->executeUpdate('social_event', 'social_event_update_8802');
+
+  // Output logged messages to related channel of update execution.
+  return $updateHelper->logger()->output();
+}
+
+/**
+ * Update view mode for new small teaser style.
+ */
+function social_event_update_8803() {
```
ronaldtebrake: Probably these update helper update_hooks need to have a dependency on social_core 8805 which enables the update helper module otherwise we're not sure if these are fired in the correct order.

OH17L: We have dependency between previous update hooks (**social_event_update_8802** will be executed after **social_core_update_8802**) so it should implement your suggestion

ronaldtebrake: Yes, but update helper is only enabled in 8805, not 8802. That's where features is removed.
So the dependency for using update helper should be on 8805. 

Have you tried updating from a previous version? 

You will see everything is nicely waiting for 8802 but failing due to the fact 8805 hasn't run yet.

modules/social_features/social_content_block/social_content_block.module
```diff
@@ -25,56 +28,136 @@ function social_content_block_preprocess_block(&$variables) {
...
+  }
 
+  /** @var \Drupal\social_content_block\ContentBlockManagerInterface $content_block_manager */
+  $content_block_manager = \Drupal::service('plugin.manager.content_block');
+
+  foreach ($content_block_manager->getDefinitions() as $plugin_id => $plugin_definition) {
+    $fields = &$form['field_plugin_field']['widget'][0][$plugin_id]['#options'];
+
+    foreach ($fields as $field_name => $field_title) {
+      if ($field_name === $field_title) {
```
Kingdutch: This looks like magic and could use a comment to explain why we need to do something in this case.

modules/social_features/social_content_block/social_content_block.module
```diff
@@ -25,56 +28,136 @@ function social_content_block_preprocess_block(&$variables) {
...
+
+  $form['#attributes']['class'][] = 'form--default';
+
+  foreach (Element::children($form['actions']) as $key) {
+    if ($key !== 'submit') {
+      $form['actions'][$key]['#attributes']['class'][] = 'btn';
+      $form['actions'][$key]['#attributes']['class'][] = 'btn-flat';
+    }
+  }
+}
```
Kingdutch: I know it's annoying but this should happen in one of the themes.

modules/social_features/social_event/config/install/core.entity_view_display.node.event.small_teaser.yml
Kingdutch: I recently learned that we should remove the `default_config_hash` key as it will be calculated by the config installer when the config is imported. So it shouldn't be in the .yml files.

Previously I thought we should add it, but I was wrong.

modules/social_features/social_group/config/install/core.entity_view_display.group.closed_group.small_teaser.yml
Kingdutch: @OH17L I suspect that was an error in that commit so it can be removed as Taras says

OH17L: Fixed

modules/social_features/social_content_block/social_content_block.module
OH17L: Done

modules/social_features/social_content_block/social_content_block.module
OH17L: Done

ribel:
@OH17L, I see you implemented last fixes, but added `uuid` to new config files in b7e5f55. 
But I am not sure if we need to do this. @Kingdutch could you clarify?

Kingdutch:
The `uuid` is site specific so this is indeed not desired. Sorry if I gave the impression that this was needed 🙈

Kingdutch:
Could not cherry-pick a merge commit so used `git diff 8b1ac1c^! | git apply -` to apply the diff of the squashed commit to 8.x-8.x. This is committed in 099622d9e

PR #1715, opened 2020-02-10
Issue #3112434 Behat test for GDPR inform block fails

ronaldtebrake:
## Problem
On our 8.x and 9.x branches the behat test for /gdpr/inform-block.feature fails, due to the fact we have enabled optional modules the order of blocks has changed.

## Solution
Ensure we directly place the block from the machine name link which doesnt change, the order can change due to more or less blocks being available.

## Issue tracker
https://www.drupal.org/project/social/issues/3112434

## How to test
- [ ] Travis behat stability step 4 should now work.

## Release notes
We have ensured our travis test are running for 8 and 9.


ronaldtebrake:
Also in 8.0

PR #1716, opened 2020-02-11
Issue #3111945 by frankgraave: SM should not be able to grant/revoke the administrator role

frankgraave:
## Problem
With the introduction of Views Bulk Operations the "Add to user" actions have been replaced with a single "Add role(s) to user" action. This different action does not prevent the Site Manager from adding the Administrator role.

## Solution
We should add the https://www.drupal.org/project/role_delegation module to Open Social. This module allows us to select which roles can be added per role. This allows us to select "Site Manager" and "Content Manager" to be added by the Site Manager role and no other.

An update hook should be written to assign the new "Add Site Manager Role" and "Add Content Manager Role" to the roles with the "Administer roles" permission. Furthermore if a site has an administrator role configured they should be granted the "Add all roles" permission.

This will also provide us flexibility to hide custom system roles that should not be administered by Site Managers.

After the module is added and the permissions are configured we can remove any custom code that hides the administrator role for users. This includes form alters in the social_user module as well as configuration alters to the add role actions in core.

## Issue tracker
https://www.drupal.org/project/social/issues/3111945

## How to test
- [ ] Log in as an SM
- [ ] Go to admin/people and select yourself (the SM)
- [ ] From the actions, select "Add a role to the selected users"
- [ ] See that you no longer can assign the administrator role
- [ ] Repeat the steps with the "Remove a role from the selected users" action and notice you can't revoke the administrator role

## Release notes
The sitemanager role can no longer grant/revoke the administrator role.

## Change Record
There's a `UserRoleHelper.php` helper class added which has a method `alterAddOrRemoveRoleOptions` that can alter the options for when a certain role grants/revokes roles from other users.


modules/social_features/social_user/src/Plugin/Action/SocialRemoveRoleUser.php
```diff
@@ -0,0 +1,54 @@
...
+      $account->save();
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    /** @var \Drupal\social_user\UserRoleHelper $roles */
+    $roles = UserRoleHelper::alterAddOrRemoveRoleOptions(user_role_names(TRUE), 'sitemanager', ['administrator']);
```
Kingdutch: @frankgraave This looks like we're hardcoding something for sitemanagers but I'm having a hard time figuring out what. Could you explain? : ) What is the helper function doing? Why do we pass it site manager and not content manager? Does this still work of the "Assign X" permissions are different or if there are different roles on the platform?

It may also be clearer if there's some code duplication in the `SocialRemoveRoleUser`/`SocialAddRoleUser` classes instead of a helper that's only used twice.

Kingdutch:
There's still a question open here: https://github.com/goalgorilla/open_social/pull/1716#discussion_r387660491

Additionally, I think this doesn't have an update hook to enable the new role delegation module, but this module is added as a new dependency to the social user module. This would bring dependencies in an unwanted state.

@ronaldtebrake Should this go in a minor release or should this be kept back for 9.0 ? We are fixing a bug, but also adding a new module and enabling it by default.

ronaldtebrake:
Usually the answer would be 9.0, or we dont enable it by default.
But if the module itself is contained for JUST fixing that bug and not opening up features as of now that would be ok to backport.
At least make it consistent for enabling it by default for everybody and not just for new users in that case, it's that important.

ronaldtebrake:
@frankgraave could you answer the last open question somewhere in the coming days, would be great to see this finished but can imagine the helper function could use some improvements.

modules/social_features/social_user/src/Plugin/Action/SocialRemoveRoleUser.php
frankgraave: Hi @Kingdutch SM's were able to grant the administrator role to themself. Therefore, as described in the doc comment of the helper function, the 'sitemanager' role is used as an (flexible) check to determine if certain options should or should not be removed from an options list. Open Social always comes with a sitemanager - as a trusted role - if I'm correct. However, there was a scenario where multiple sitemanagers were granting themself the administrator role, which caused a lot of trouble in the platform.

At first, there are indeed only two places where this needs to be applied, but I wanted to keep it DRY. The idea is that if there were other places that need to filter out certain options by role, this helper could be used.

ronaldtebrake: It just feels like this helper function is a bit too specifically unspecific.
I can imagine you want to re-use code to say, ok you know the administrator role is something nobody should ever be able to give themselves or to others if you are not one yourself.  
At least my understanding that is what is what we are trying to achieve.

What I would opt for is to not perse do this with a helper function in the current format, i doubt anyone would reuse this and it might be not flexible enough or not clear enough how to make it flexible. 
Alternatively you can create permissions that are saying "allow to assign or revoke the X role". 
So administrator can assign or revoke this always, 
Site manager can assign content manager and SM roles
and you can ship with this by default and check for those permissions?

Kingdutch: Ah good argument against introducing the abstraction here is AHA Programming: "Avoid Hasty Abstractions".

This is an interesting read by the person who coined the term: https://kentcdodds.com/blog/aha-programming

navneet0693:
Cherry-picked on 8.8.x - fd7885f0d99f515ae8646e4fb40a694d3e3bc0b2

Cherry-picked on 8.7.x - eb5d3f531d0cc992d89b631921b2ee9e1ac24d5f

PR #1720, opened 2020-02-17
Feature/tb 2922 datepicker for birtdate field

sjoerdvandervis:
## Problem
The datepicker that is included with Open Social is stepping month by month. If a user wants to enter a birthdate for example, he/she needs to click a lot of times to get to the correct month and year.

## Solution
We use the native datepicker from the browser instead of the custom datepicker that we ship. We only do this for fields which have stated in the drupalSettings that they want the native datepicker instead of the one shipped with OS.

These changes only work when the field is defined in the javascript settings. For the social_birthday extension, see this PR: https://bitbucket.org/goalgorilla/social-birthday/pull-requests/10

## How to test
- [ ] Enable the social_birthday extension;
- [ ] Go to the page where the field is shown, for example on the profile;
- [ ] See that you get the native datepicker from the browser instead of the custom one that's shipped with OS, if the browser supports a native datepicker. Safari on MacOS does not support this, this should still show the custom datepicker.


sjoerdvandervis:
Closing this one as we had a lot of merge conflicts and it was based on 8.x . Re-opening a new PR.

PR #1724, opened 2020-02-19
YANG-2458: Update template and styles for the information block

agami4:
## Problem
Some elements have wrong styles on the information block on the profile page

## Solution
Update template and styles 

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-2458

## How to test
- [ ] Go to the profile information page

## Screenshots
before: image-20200217-141708

after: Screenshot at Feb 20 09-51-52

## Release notes
Fixed layout and update styles for the information block on the profile page


Kingdutch:
@agami4 That screenshot looks like it still needs a bite more work?

agami4:
@Kingdutch, Hey) I've updated screenshots.) We had a problem with badge labels.

Kingdutch:
@agami4 Ah, now it's clear what the difference is :D 

Theming wise this looks good to me. However, I do see that the sky templates have different labels than the original templates. Another problems in this is that it always says "Your ..." even when someone else would be viewing the profile. (cc @ribel )

Is there already an issue for that or do you want to quickly include that in this PR?

agami4:
@Kingdutch, I've removed 'Your ...' from the templates. And I've added small fixes for the profile teaser on the landing page.
Before: Screenshot at Feb 20 12-39-38
After: Screenshot at Feb 20 12-41-12

agami4:
@Kingdutch, thank you for the review!

ronaldtebrake:
Also in 8.x :)

PR #1727, opened 2020-02-20
Issue #3113150 by robertragas: since the feature removal failed for t…

robertragas:
…he social_private_message we manually import the configs in an update hook

## Problem
The final config revert from the feature removal script for social_private_message does not get the correct config yml from the features_removal folder, but from the contrib config/install. So our changes are not reflected.

## Solution
Create an update hook that will manually import the changes into the config.

## Issue tracker
https://www.drupal.org/project/social/issues/3113150

## How to test

## Screenshots

## Release notes
The features removal script during our 8.0 release skipped the folder for the social_private_message resulting in broken views. We have reimported the config so it will work again.


Kingdutch:
@robertragas Please add a check to see if the Social Private Message module is enabled as it's an optional module : ) The features revert update hook (social_core_update_8802) does this automatically by requesting only the enabled modules

Other than that it looks good to me from a code perspective.

modules/social_features/social_private_message/social_private_message.install
```diff
@@ -210,3 +210,31 @@ function social_private_message_update_8002() {
...
   }
 }
+
+/**
+ * Config import social_private_message default configs.
+ *
+ * The feature removal script did not succeed for social_private_message.
+ * Import the configs in an update hook.
+ */
+function social_private_message_update_8003() {
```
ronaldtebrake: We probably need to use hook_update_dependencies so it runs after social_core features removal

ronaldtebrake:
Also in 8.x :)

PR #1729, opened 2020-02-24
Issue #3115540 by sjoerdvandervis: Keep language when using search filters

sjoerdvandervis:
## Problem
When a user uses the search filters, the chosen language in a multi language environment gets lost and the search filter goes back to the system default language.

## Solution
Keep the language that has been chosen instead of going back to the default language. We need to use the chosen language prefix in the action of the filters, so that when a filter is applied, the language code is added in the path.

## Issue tracker
https://www.drupal.org/project/social/issues/3115540

## How to test
- [ ] In a multi language environment, select a different language and search for something;
- [ ] See that the language is still active when you land on the search results;
- [ ] Apply a search filter and see that the language you selected is still active.


robertragas:
@sjoerdvandervis I made some code changes, because in some cases it can be that the default language also has a prefix, so we then would like to add it.

So instead of that check I changed it to apply it on prefixes and then it wouldn't matter anymore if it would be a default language. Also I think it would otherwise fail is someone would have the prefix "nederlands" when the lancode is "nl".

navneet0693:
Cherry-picked on aa8744e0769762e5af4c56fb951de666eed4ec2c

PR #1734, opened 2020-02-27
Issue #3116229 by SV, ribel:  Exclude duplicated fields from content section for Events

volodymyr-sydor-lemberg:
## Problem
The number of enrollments and enrolled status displays in the content sections of Events.

## Solution
Exclude not needed fields from the full event template.

## Screenshots
image

image

## Issue tracker
- https://www.drupal.org/project/social/issues/3116229
- https://getopensocial.atlassian.net/browse/ECI-1002

## How to test

- [ ] Enroll to event

- [ ] Go to the Event page and you should not see the number of enrollments and status above the description.


ronaldtebrake:
Hi guys, didnt we want to update the display configuration rather than removing them from the content render array? This way we can also remove it from the socialbase template. But this also ensure our users can actually show it in the content if they want to.

ribel:
Hi, @ronaldtebrake,
In general, you are correct but in this case, we need to display `field_event_location` and `field_event_address` in the default style in full view mode, but we need to hide it in the sky, since it is displayed in-hero view mode there.
So the most logical way to fix it - override in the event full template for sky style.
The only thing we can hide in full view display is maybe `enrollments_count` 
@volodymyr-sydor-lemberg could you check it and set `enrollments_count`  as hidden by default if so?

ronaldtebrake:
Ah right, that seems valid. Thanks for your explanation!

themes/socialblue/templates/node/event/node--event--full--sky.html.twig
```diff
@@ -1,2 +1,6 @@
- {# This file overrides node--event--full.html.twig #}
+{# This file overrides node--event--full.html.twig #}
 {% extends "node--full--sky.html.twig" %}
+
+{% block nodefull_body %}
+  {{ content|without('flag_follow_content', 'like_and_dislike', 'field_event_location', 'field_event_address', 'enrollments_count', 'links') }}
```
Kingdutch: Could we add a comment here that these are displayed in the hero instead? This way if that changes in the future we know quicker why this was done :)

ribel: Done

ribel: @Kingdutch, @ronaldtebrake I have checked that address fields was fixed recently by @bramtenhove in a commit: 89b1ae0. But event enrollments and event status still displaying there (see the screenshot from normal heme).
We can't hide those fields in the config, since they are added `social_event_node_view_alter`, so I think we just need to hide them here as well. 
Please check again.

ribel:
@Kingdutch, @ronaldtebrake I have checked that address fields was fixed recently by @bramtenhove in a commit: 89b1ae0. But event enrollments and event status still displaying there (see the screenshot from normal heme).
We can't hide those fields in the config, since they are added social_event_node_view_alter, so I think we just need to hide them here as well.
Please check again.

robertragas:
I have added the field enrollment status as we also have this with the event invite feature etc.
for the rest working like expected, approved and merging + cherrypicking.

PR #1737, opened 2020-02-28
YANG-1004: Add condition for the group and profile pages

agami4:
## Problem
The page title is present on the group and profile pages on the sky theme

## Solution
Add condition for the hide page title on the group and profile pages on the sky theme

## Issue tracker
https://getopensocial.atlassian.net/browse/ECI-1004
https://www.drupal.org/project/social/issues/3116659

## How to test
- [ ] Go to the profile or group page

## Screenshots
before(red border)/after(green border):
image-20200225-144902

## Release notes
The page title does not preset on the profile and group pages on the sky theme

### DIFF
--- a/themes/socialblue/src/Plugin/Preprocess/Page.php
+++ b/themes/socialblue/src/Plugin/Preprocess/Page.php
@@ -25,7 +25,6 @@ class Page extends PageBase {
       $route_match = \Drupal::routeMatch();
       if ($route_match->getParameter('user') && $route_match->getRouteName() !== 'entity.profile.type.user_profile_form') {
         $variables['content_attributes']->addClass('sidebar-left', 'content-merged--sky');
-//        dump($variables);
 
         // If page has title.
         if ($variables['page']['title']) {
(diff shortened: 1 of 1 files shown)


themes/socialblue/src/Plugin/Preprocess/Page.php
```diff
@@ -25,11 +25,21 @@ public function preprocess(array &$variables, $hook, array $info) {
       $route_match = \Drupal::routeMatch();
       if ($route_match->getParameter('user') && $route_match->getRouteName() !== 'entity.profile.type.user_profile_form') {
         $variables['content_attributes']->addClass('sidebar-left', 'content-merged--sky');
+
+        // If page has title.
```
ribel: @agami4 Please add more clear comment, which explains why you need to hide a title here. See example from socialbase:

```
        // If there is a title and node type is excluded remove class.
        if (in_array($node->bundle(), $page_to_exclude, TRUE)) {
          $attributes->removeClass('with-title-region');
          $variables['display_page_title'] = FALSE;
        }
```
And I think this condition is not neded here `if ($variables['page']['title']) {`
But should we also add line `$attributes->removeClass('with-title-region'); ` here?

ribel:
This issue was fixed in Challenge and Discussions modules.

PR #1741, opened 2020-03-02
TB-2922 - Use native date picker on specified fields.

sjoerdvandervis:
### Problem
The datepicker that is included with Open Social is stepping month by month. If a user wants to enter a birthdate for example, he/she needs to click a lot of times to get to the correct month and year.

### Solution
We use the native datepicker from the browser instead of the custom datepicker that we ship. We only do this for fields which have stated in the drupalSettings that they want the native datepicker instead of the one shipped with OS.

These changes only work when the field is defined in the javascript settings. For the social_birthday extension, see this PR: https://bitbucket.org/goalgorilla/social-birthday/pull-requests/10

### How to test
- [x] Enable the social_birthday extension;
- [x] Go to the page where the field is shown, for example on the profile;
- [x] See that you get the native datepicker from the browser instead of the custom one that's shipped with OS, if the browser supports a native datepicker. Safari on MacOS does not support this, this should still show the custom datepicker.

### Release notes
From now on, the birthday field will not show our custom date picker anymore, as this is not really usable (it scrolls month by month, which is not ideal when you want to select a date a few years ago). From now on, the birthday field which is added by the extension social_birthday will show the native picker from the browser, if the browser supports a native date picker. If not, it will still show our own custom date picker.


bramtenhove:
@sjoerdvandervis please make sure to add the correct labels and milestone. I did it for the labels now, you still need to fix the milestone.

Also make sure to **always** add release notes describing what has changed so we can use it for the release notes on D.org as well as others.

If this should be added to `7.x` the milestone should be the next `7.x` release. If just for `8.x`, it's the next `8.x` milestone, etc.

After it is merged the PR may need to be cherry-picked to for example the `8.x` and `7.x` branches depending on the milestone.

sjoerdvandervis:
Cherry-picked to 8.x

PR #1744, opened 2020-03-02
Move overwritten book config to config/optional

Kingdutch:
This ensures that the config can properly be installed when the book
module is enabled instead of giving an error during features revert or
module install.

Features removal ensured we have removed social_book as part of features here https://github.com/goalgorilla/open_social/pull/1617/files
However a lot of the configuration is already provided by the core book module, resulting in 

<code>Drupal\Core\Config\PreExistingConfigException: Configuration objects (core.base_field_override.node.book.promote, core.entity_form_display.node.book.default, core.entity_view_display.node.book.default, core.entity_view_display.node.book.teaser, field.field.node.book.body, node.type.book) provided by    [error]
social_book already exist in active configuration in core/lib/Drupal/Core/Config/PreExistingConfigException.php:65</code>

Ensure we remove the obsolete configuration that is already provided by Book and ensure we use install process to update the configuration after installing social_book. 

## Issue tracker
https://www.drupal.org/project/social/issues/3117047

## How to test
- [ ] Normal install without book should succeed
- [ ] Check none of the optional config was created
- [ ] Install the Social Book module
- [ ] This should be doable without errors.

## Release notes
Duplicate configuration from the Social Book module has been moved into the install profile to ensure that the module can be installed properly.


ronaldtebrake:
Cherry-picked to 8.x-8.x

PR #1752, opened 2020-03-10
Remove features removal code

Kingdutch:
## Problem
New updates should be properly written not to use this code and 
users should upgrade to 8.x before moving on to 9.x

## Solution
This commit removes the code that is used in the 8.0 update to remove
features. It's no longer needed after the 8.0 update. 

## Issue tracker
https://www.drupal.org/project/social/issues/3151942

## How to test
- [ ] Check that updates to 9.x keep working

## Release notes
You must update to Open Social 8.x before updating to Open Social 9


Kingdutch:
Travis should fail :D I believe it upgrades from 7.x and the requirements check at that point should fail.

Kingdutch:
Force push fixed a coding standards issue. Travis was green otherwise.

The update did properly show the requirements error. However, running `drush updb -y` just blazes past that :( So this doesn't actually cause Travis to fail (which may be okay?)

https://travis-ci.org/goalgorilla/open_social/jobs/660698734#L1788

ronaldtebrake:
Screen Shot 2020-07-20 at 11 01 08
Update from 8.x -> 9.x seems fine from what I can see. Only issues I encounter are related to known issues.

PR #1753, opened 2020-03-10
Issue #3118987 by chmez: As a CM I want to customise Events and Groups for the custom content block according to my specific use case

OH17L:
## Problem
As a CM I want to customise Events and Groups for the custom content block according to my specific use case

## Solution
Create the content block plugin for events and groups

## Issue tracker
- https://www.drupal.org/project/social/issues/3118987
- https://getopensocial.atlassian.net/browse/YANG-1913

- https://www.drupal.org/project/social/issues/3122081

## How to test
- [ ] When you open **Add Custom content list block custom block** page (_/block/add/custom_content_list_) then you should see the **Number of items** field in the **Content** section and you shouldn't see the same field in the **Sorting** section
- [ ] Enable the **Social Event Content Block** module
- [ ] When you open **Add Custom content list block custom block** page then you should see the **Event** item in the drop-down list of the **Type of content** field
- [ ] When you select the **Event** item in the **Type of content** field then you should see the following fields in the **Content** section: **Content tags**, **Groups**
- [ ] You shouldn't see the **Event types** field
- [ ] Enable the **Social Event Type** module
- [ ] When you open **Add Custom content list block custom block** page and select the **Event** item in the **Type of content** field then you should see the **Event types** field
- [ ] Enable the **Social Group Content Block** module
- [ ] When you open **Add Custom content list block custom block** page then you should see the **Group** item in the drop-down list of the **Type of content** field
- [ ] When you select the **Group** item in the **Type of content** field then you should see the following fields in the **Content** section: **Group types**, **Content tags**

## Release notes
Add support of events and groups in the **Custom content list block** custom block


Kingdutch:
@OH17L @ribel Is it possible to clean this branch up and pick out the commits relevant for this change? We can't merge 130 commits (including stuff from 8.x-8.x) into 8.x-9.x.

If needed a cleaned up branch can be force-pushed to the branch of this PR (so no need to make a new PR).

ribel:
@Kingdutch, could you maybe explain why it happens all the time, and how can we avoid it or fix?

Kingdutch:
Hey @ribel ,

This happens when a branch is created on 8.x-8.x and is then moved to base on 8.x-9.x. Unfortunately those branches diverge which causes this behaviour.

It's made more difficult when the old (or new) branch is merged in because it mixes commits. One option is to rebase on the new branch once this happens (given that the branch doesn't contain any merge commits yet).

Another option when the base branch needs to change is to create a new branch from the new base branch and cherry-pick the commits, then force push that to the old branch before changing the base branch on GitHub. This is essentially manually doing a rebase. 

Either can be made easier by doing this before changing the base in the PR and merging in the new branch because it's clearer in GitHub which commits belong to the PR itself and which are commits that are out of sync between the version branches.

Unfortunately between 8.x-8.x and 8.x-9.x this was a bit more visible than in our other versions as some things were committed to 8.x-8.x and cherry-pick to 8.x-9.x instead of the other way around when the 8.x-9.x branch was first created.

Kingdutch:
Looking at the diff for a code review. I notice two things:

1. There is now a `GroupContentBlock` and an `EventContentBlock` but there is not `TopicContentBlock`. Should that last one exist as well? If not, why is it special?

2. The content blocks are submodules in the `social_event` and `social_group` modules. I wonder if it's better to just make those submodules in the `social_content_block` since they are probably going to change together (I do like that they are decoupled like this!) (@ronaldtebrake any thoughts?)

OH17L:
@Kingdutch, the **TopicContentBlock** plugin exists. You can see it [here](https://github.com/goalgorilla/open_social/blob/feature/3118987-events-and-groups-content-block/modules/social_features/social_content_block/src/Plugin/ContentBlock/TopicContentBlock.php)

Kingdutch:
@OH17L Oh I see, thanks! Should this be moved to a `social_topic_content_block` submodule too, to be in line with the rest?

OH17L:
@Kingdutch, it's a good idea! For which module should the **social_topic_content_block** module be submodule - **social_content_block** or **social_topic**?

Kingdutch:
@OH17L Lets make `social_topic_content_block` a submodule of `social_content_block` and move the other submodules there as well. This way all `content block` related functionality is in one place. :)

Kingdutch:
It appears that there is a caching issue either in the form or in the plugin definition registry.

When I enabled the `social_event_content_block` module it did not show up until I manually cleared all caches. Additionally, after uninstalling the `social_group_content_block` module, the `group` type did not disappear from the form after a page refresh. Clearing all caches did make it disappear.

----- 

It also appears that configuration is not properly cleaned up when the module is uninstalled. i.e. after the above, I can no longer install the `social_group_content_block` module.

```
The following extensions will be enabled: social_group_content_block
Do you really want to continue? (y/n): y
Drupal\Core\Config\PreExistingConfigException: Configuration objects [error]
(field.field.block_content.custom_content_list.field_group_content_tag,
field.field.block_content.custom_content_list.field_group_type,
field.storage.block_content.field_group_content_tag,
field.storage.block_content.field_group_type) provided by
social_group_content_block already exist in active configuration in
/var/www/html/core/lib/Drupal/Core/Config/PreExistingConfigException.php:65
Stack trace:
#0
/var/www/html/core/lib/Drupal/Core/Config/ConfigInstaller.php(520):
Drupal\Core\Config\PreExistingConfigException::create('social_group_co...',
Array)

```

I think this'll be the same for `social_event_content_block` too.

OH17L:
@Kingdutch I don't think that it's Drupal way - store all plugin instances in the module which provide this plugin type. If you looking at Drupal core that you can see that, for example, all blocks are not located in the **Block** module but they located in different modules as **User**, **Search**, etc.

Kingdutch:
I agree with your analogy but I don't think it really applies here.

In the *_content_block modules we're still adding fields to the same `field.field.block_content.custom_content_list.*` So it's not actually that separate (e.g. like bundles). They also don't actually extend anything for the node bundle of the `social_{event,topic,group}` module.

If I look at how a developer would work on this then it would probably be with multiple of those modules open at the same time. For example the `ContentBuilder` class has sorting implemented for all types. 

Putting it all under the `social_content_block/modules` makes it easier to navigate for developers and allows us to easily spin this out to a separate module (like `social_geolocation`/`social_course`) if needed later.

OH17L:
@Kingdutch ok, we'll move those modules to the **social_content_block/modules** directory. Should we rename **social_event_content_block** to **social_content_block_event** and **social_group_content_block** to **social_content_block_group**?

Kingdutch:
👍 
I have no strong opinion on renaming those modules, either works for me. If you can rename without issues for existing platforms then feel free to do so :)

Kingdutch:
I think the following fix is needed to prevent errors when certain fields don't exist (e.g. when the `social_event_type` module is not enabled).

```
diff --git a/modules/social_features/social_content_block/src/ContentBuilder.php b/modules/social_features/social_content_block/src/ContentBuilder.php
index 20deb1277..d2318d763 100644
--- a/modules/social_features/social_content_block/src/ContentBuilder.php
+++ b/modules/social_features/social_content_block/src/ContentBuilder.php
@@ -99,7 +99,14 @@ class ContentBuilder implements ContentBuilderInterface {
     // then the block base query will be built based on all filled filterable
     // fields.
     if ($block_content->field_plugin_field->isEmpty()) {
-      $field_names = $definition['fields'];
+      // It could be that the plugin supports more fields than are currently
+      // available, those are removed.
+      $field_names = array_filter(
+        $definition['fields'],
+        static function ($field_name) use ($block_content) {
+          return $block_content->hasField($field_name);
+        }
+      );
     }
     // When the user selected some filter in the "Content selection" field then
     // only condition based on this filter field will be added to the block base

```

Kingdutch:
@OH17L @ribel I fixed the bug we discussed, unfortunately I don't have time to move the modules around. If it's easier, feel free to do that on the branch from #1778 instead of this PR when you get to it. Then we can merge both PRs at the same time.

OH17L:
@Kingdutch, the pull request not needed anymore because #1778 contains all changes from this pull request. Also, when this pull request will be merged then be created the same modules in different places because the plugin modules were moved in the #1778

Kingdutch:
@OH17L I'd prefer doing it in 2 pull requests still so it's clearer what work was done where. Git should also be smart enough to know the commits are the same and take care of this for us.

Kingdutch:
@OH17L That doesn't mean this needs work! I think we can merge this. Then the other PR will update itself right?

PR #1754, opened 2020-03-11
Issue #3117047 run update hook to ensure config of books is re-imported

ronaldtebrake:
## Problem
After features removal the configuration for books didn't go that well. 
Screen Shot 2020-03-11 at 10 05 24

## Solution
Run another update hook, after features removal has taken place to re-import config one last time.

Screen Shot 2020-03-11 at 10 05 47

## Issue tracker
http://drupal.org/node/3117047

## How to test
- [ ] Ensure after update it still works as expected, for example the attachments were lost on the book node form.

## Release notes
None needed.

### DIFF
--- a/modules/social_features/social_book/social_book.install
+++ b/modules/social_features/social_book/social_book.install
@@ -6,6 +6,7 @@
  */
 
 use Drupal\user\Entity\Role;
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
Also in 8.x

PR #1756, opened 2020-03-11
[DRAFT] Feature/3118222 request event enrollment

robertragas:
## Problem
<describe the problem you're trying to solve>

## Solution
<describe the solution you're created>

## Issue tracker

## How to test

## Screenshots

## Release notes
<describe the release notes>

## Change Record

## Translations


modules/social_features/social_event/social_event.module
```diff
@@ -519,6 +548,61 @@ function social_event_form_alter(&$form, FormStateInterface $form_state, $form_i
...
+            ->getStorage('event_enrollment')
+            ->getQuery();
+          $condition_group = $query->andConditionGroup()
+            ->condition('field_event.target_id', $node->id())
+            ->condition('field_request_status.value', 'pending')
+            ->condition('field_enrollment_status.value', '0');
+          $query->condition($condition_group);
+          $pending_requests = $query->execute();
+          if (!empty($pending_requests)) {
+            $link = Link::createFromRoute(t('(@amount) new requests', ['@amount' => count($pending_requests)]), 'view.event_manage_enrollment_requests.page_manage_enrollment_requests', ['node' => $node->id()])->toString();
```
bramtenhove: I know this might be tricky here, but it would be really nice if we can do the plural string translation here.

E.g. "There is 1 new request" and "There are 15 new requests".

You could use the `@amount` to create the plural string translations for both line `590` and `592`.

frankgraave: Ah good one! I'll check it out, thanks 😄

modules/social_features/social_event/src/Controller/EnrollRequestDialogController.php
```diff
@@ -0,0 +1,100 @@
...
+   * Helper method so we can have consistent dialog options.
+   *
+   * @return string[]
+   *   An array of jQuery UI elements to pass on to our dialog form.
+   */
+  protected static function getDataDialogOptions() {
+    return [
+      'dialogClass' => 'form--default',
+      'closeOnEscape' => TRUE,
+      'width' => '400',
```
bramtenhove: We should see if we can make this wider (as per design) and also have it scale nicely on mobile. Ask Alexander if needed.

modules/social_features/social_event/config/install/field.storage.node.field_enroll_method.yml
```diff
@@ -0,0 +1,27 @@
...
+    - node
+    - options
+id: node.field_enroll_method
+field_name: field_enroll_method
+entity_type: node
+type: list_string
+settings:
+  allowed_values:
+    -
+      value: '1'
```
bramtenhove: I think we should step away from using `strings` here, but rather use `integers`. Imho it's cleaner to do this with integers, also as these are simply numbers?
This would mean that the checks we added should also be changed to check on integers instead of the string value.

I know that for the `field_enrollment_status` we also use `strings`, but I think we should change that at a different moment.

modules/social_features/social_event/social_event.module
```diff
@@ -638,3 +724,153 @@ function social_event_get_current_event() {
...
+ */
+function social_event_owner_or_organizer(NodeInterface $node = NULL, AccountInterface $current_user = NULL) {
+  if (!$current_user) {
+    $current_user = \Drupal::currentUser();
+  }
+  if (!$node) {
+    $node = social_event_get_current_event();
+  }
+  // Get the event owner/author and it's organisers (if any).
+  $event_accountables['owner'] = $node->getOwnerId();
```
ronaldtebrake: This is also fired on admin/people, which means there is no `$node` available. Resulting in a
`Error: Call to a member function getOwnerId() on null in social_event_owner_or_organizer`

ronaldtebrake:
Could you also make sure this is based from branch 8.x-9.x, thanks!

ronaldtebrake:
As a EM when I approve a request I see the following message which is meant for LU:
Screen Shot 2020-03-27 at 10 50 09

modules/social_features/social_event/social_event.module
bramtenhove: I may have fixed this in ab2b10afedae20ace50af658c379a49fdc6ef828

modules/social_features/social_event/src/Form/EventSettingsForm.php
```diff
@@ -33,14 +33,14 @@ public function getFormId() {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state) {
-    $group_type_ids = $this->configFactory->getEditable('social_event.settings')
-      ->get('enroll');
+    // Get the config.
+    $social_event_config = $this->configFactory->getEditable('social_event.settings');
```
bramtenhove: We don't need `getEditable()` here, we're not editing anything right? We can use `get()`.

modules/social_features/social_event/templates/event-request-enroll-dialog.html.twig
```diff
@@ -0,0 +1,6 @@
+<div>
+  <div class="col-md">
+    <h5>{% trans %}Request enrollment{% endtrans %}</h5>
+    <p>Tell me whyyyyy</p>
```
bramtenhove: Hahaha, what is this?

modules/social_features/social_event/src/Form/EventSettingsForm.php
Kingdutch: In forms we should always use `getEditable` otherwise config overrides might accidentally get moved into database configuration because it'll be put into the form.

🏎 💨

modules/social_features/social_event/src/Routing/EnrollRequestsOverviewSubscriber.php
```diff
@@ -0,0 +1,54 @@
...
+    return $events;
+  }
+
+  /**
+   * Check if the user is allowed to view this overview.
+   *
+   * @param \Symfony\Component\HttpKernel\Event\GetResponseEvent $event
+   *   The event.
+   */
+  public function checkAccessToEnrollRequestsOverview(GetResponseEvent $event) {
```
bramtenhove: @frankgraave for this one you can better use a custom `views access` plugin as you did earlier on for the Event Invite feature :)

modules/social_features/social_event/social_event.module
```diff
@@ -638,3 +724,161 @@ function social_event_get_current_event() {
...
+ *
+ * @param \Drupal\node\NodeInterface|null $node
+ *   The (optional) event node.
+ * @param \Drupal\Core\Session\AccountInterface|null $current_user
+ *   The (optional) user.
+ *
+ * @return bool
+ *   Return TRUE if the user is the owner or organizer of this event.
+ */
+function social_event_owner_or_organizer(NodeInterface $node = NULL, AccountInterface $current_user = NULL) {
```
bramtenhove: @frankgraave this is better suited for a `service` instead of in a module file.

frankgraave: Alright, will move this towards a service. 👍

bramtenhove:
`Uncaught TypeError: this._handles.disableSelection is not a function` is thrown in console when I click on Request to enroll as LU.

Screenshot 2020-03-27 at 11 38 12

modules/social_features/social_event/social_event.module
bramtenhove: Ok, `admin/people` part is fixed.

modules/social_features/social_event/templates/event-request-enroll-dialog.html.twig
frankgraave: This is already gone in the invites branch!

modules/social_features/social_event/config/install/field.storage.node.field_enroll_method.yml
bramtenhove: We should also make this constants with a readable format.

Example:
```
class GroupInvitation extends GroupContentEnablerBase {
  /**
   * Invitation created and waiting for user's response.
   */
  const INVITATION_PENDING = 0;
  /**
   * Invitation accepted by user.
   */
  const INVITATION_ACCEPTED = 1;
  /**
   * Invitation rejected by user.
   */
  const INVITATION_REJECTED = 2;
```

bramtenhove:
Closing the PR. The work continues in #1774

PR #1759, opened 2020-03-12
Issue #3118044 by rolki: As a CM I want to sort content for the custom content block according to my specific use case

rollki:
## Problem
As a CM I want to sort content for the custom content block according to my specific use case

## Solution
- Add sorting by most commented
- Add sorting by most liked

## Issue tracker
- https://www.drupal.org/project/social/issues/3118044
- https://getopensocial.atlassian.net/browse/YANG-2000

## How to test
- When you open Add Custom content list block custom block page (/block/add/custom_content_list) then you should see that these page doesn't use an administrative theme
- You should see two extra sorting options (most commented and most liked)
- Use a new specific sorting
- Paste this block to the Complementary top region

## Screenshots
image
image

### DIFF
--- a/modules/social_features/social_content_block/src/ContentBuilder.php
+++ b/modules/social_features/social_content_block/src/ContentBuilder.php
@@ -268,9 +268,13 @@ class ContentBuilder implements ContentBuilderInterface {
           $query = $this->connection->select('comment_field_data', 'cfd');
           $query->condition('cfd.entity_id', $entities, 'IN');
           $query->condition('cfd.created', $start_time, '>');
-          $query->addField('cfd', 'entity_id');
+          $query->condition('cfd.status', 1, '=');
+          // Only nodes, without posts or other entities.
+          $query->innerJoin('node_field_data', 'nfd', 'nfd.nid = cfd.entity_id');
+
+          $query->addField('nfd', 'nid');
           $query->addExpression('COUNT(cfd.entity_id)', 'count');
-          $query->groupBy('cfd.entity_id');
+          $query->groupBy('nfd.nid');
         }
         $query->orderBy('count','DESC');
         break;
@@ -295,9 +299,12 @@ class ContentBuilder implements ContentBuilderInterface {
           $query->condition('vv.entity_id', $entities, 'IN');
           $query->condition('vv.entity_type', $entity_type, '=');
           $query->condition('vv.timestamp', $start_time, '>');
-          $query->addField('vv', 'entity_id');
+          // Only nodes, without posts or other entities.
+          $query->innerJoin('node_field_data', 'nfd', 'nfd.nid = vv.entity_id');
+
+          $query->addField('nfd', 'nid');
           $query->addExpression('COUNT(vv.entity_id)', 'count');
-          $query->groupBy('vv.entity_id');
+          $query->groupBy('nfd.nid');
         }
         $query->orderBy('count','DESC');
         break;
@@ -339,6 +346,7 @@ class ContentBuilder implements ContentBuilderInterface {
           // Last interacted for all the time.
           $query = $this->connection->select('node_field_data', 'nfd');
           $query->condition('nfd.nid', $entities, 'IN');
+          $query->condition('nfd.status', 1, '=');
           // Comment entity.
           $query->leftjoin('comment_field_data', 'cfd', 'nfd.nid = cfd.entity_id');
           // Like node.
--- a/modules/social_features/social_content_block/src/ContentBuilder.php
+++ b/modules/social_features/social_content_block/src/ContentBuilder.php
@@ -285,9 +285,9 @@ class ContentBuilder implements ContentBuilderInterface {
           $query = $this->connection->select('group_content_field_data', 'gfd');
           $query->condition('gfd.gid', $entities, 'IN');
 
-          $query->leftJoin('votingapi_vote', 'vv', 'gfd.entity_id = %alias.entity_id');
-          $query->leftjoin('comment_field_data', 'cfd', 'gfd.entity_id = %alias.entity_id');
-          $query->leftjoin('node_field_data', 'nfd', 'gfd.entity_id = %alias.nid');
+          $query->leftJoin('votingapi_vote', 'vv', 'gfd.entity_id = vv.entity_id');
+          $query->leftjoin('comment_field_data', 'cfd', 'gfd.entity_id = cfd.entity_id');
+          $query->leftjoin('node_field_data', 'nfd', 'gfd.entity_id = nfd.nid');
 
           // Create or update post.
           $query->leftjoin('post__field_recipient_group', 'pst', 'gfd.gid = pst.field_recipient_group_target_id');
@@ -317,11 +317,11 @@ class ContentBuilder implements ContentBuilderInterface {
           $query = $this->connection->select('node_field_data', 'nfd');
           $query->condition('nfd.nid', $entities, 'IN');
           // Comment entity.
-          $cfd_alias = $query->leftjoin('comment_field_data', 'cfd', 'nfd.nid = %alias.entity_id');
+          $query->leftjoin('comment_field_data', 'cfd', 'nfd.nid = cfd.entity_id');
           // Like node.
-          $query->leftjoin('votingapi_vote', 'vv', 'nfd.nid = %alias.entity_id');
+          $query->leftjoin('votingapi_vote', 'vv', 'nfd.nid = vv.entity_id');
           // Like comment related to node.
-          $query->leftjoin('votingapi_vote', 'vvn', "{$cfd_alias}.cid = %alias.entity_id");
+          $query->leftjoin('votingapi_vote', 'vvn', 'cfd.cid = vvn.entity_id');
 
           $query->addField('nfd', 'nid');
           $query->addExpression('GREATEST(COALESCE(MAX(vv.timestamp), 0),
--- a/modules/social_features/social_content_block/src/ContentBuilder.php
+++ b/modules/social_features/social_content_block/src/ContentBuilder.php
@@ -234,6 +234,8 @@ class ContentBuilder implements ContentBuilderInterface {
   }
 
   /**
+   * Sorting and range logic by specific case.
+   *
    * @param \Drupal\block_content\BlockContentInterface $block_content
    *   The block content where we get the settings from.
    * @param \Drupal\Core\Database\Query\SelectInterface $query
@@ -250,13 +252,7 @@ class ContentBuilder implements ContentBuilderInterface {
     switch ($block_content->field_sorting->value) {
       case 'most_commented':
         if ($entity_type === 'group') {
-          $query = $this->connection->select('group_content_field_data', 'gfd');
-          $query->condition('gfd.gid', $entities, 'IN');
-          $cfd_alias = $query->innerJoin('comment_field_data', 'cfd', 'gfd.entity_id = %alias.entity_id');
-          $query->condition("{$cfd_alias}.created", $start_time, '>');
-          $query->addField('gfd', 'gid');
-          $query->addExpression("COUNT({$cfd_alias}.entity_id)", 'count');
-          $query->groupBy("gfd.gid");
+          $this->groupSorting($query, $entities, $start_time, 'comment_field_data', 'created');
         }
         else {
           $query = $this->connection->select('comment_field_data', 'cfd');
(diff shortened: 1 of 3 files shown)


modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -284,4 +279,148 @@ public static function processBlockForm(array $element, FormStateInterface $form
...
+          $query->leftJoin('group_content_field_data', 'gfd', 'cfd.entity_id = gfd.entity_id');
+
+          $query->addExpression('COALESCE(gfd.gid, pfrg.field_recipient_group_target_id)', 'group_id');
+          $query->addExpression('COUNT(cfd.cid)', 'count');
+          $query->where("COALESCE(gfd.gid, pfrg.field_recipient_group_target_id) IN ($entities)");
+          $query->groupBy('group_id');
+        }
+        else {
+          $query = $this->connection->select('comment_field_data', 'cfd');
+          $query->condition('cfd.entity_id', $entities, 'IN');
```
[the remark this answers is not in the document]
rollki: Added:
`    if (empty($entities)) {
      $query->orderBy('base_table.' . $block_content->field_sorting->value);
      $query->range(0, $block_content->field_item_amount->value);
      return $query->execute()->fetchAllKeyed(0, 0);
    }
`

rollki:
Hey @Kingdutch 
Could you please take a look at this PR?

Kingdutch:
As an example, the following query adds the comment count with a join to a query that has `node` as the base table.

`SELECT n.nid, n.vid, COUNT(n.nid) as comment_count FROM node n JOIN comment_field_data cfd ON n.nid=cfd.entity_id AND cfd.entity_type='node' GROUP BY nid;`

This can be verified by performing a similar query on just the comment_field_data (which was similar to what was done as a separate query).

```
MariaDB [social]> SELECT n.nid, n.vid, COUNT(n.nid) as comment_count FROM node n JOIN comment_field_data cfd ON n.nid=cfd.entity_id AND cfd.entity_type='node' GROUP BY nid;
+-----+------+---------------+
| nid | vid  | comment_count |
+-----+------+---------------+
|   1 |    1 |             1 |
|   3 |    3 |             5 |
|   4 |    4 |             1 |
|   5 |    5 |             1 |
|   6 |    6 |             1 |
|   7 |    7 |             1 |
|   9 |    9 |             2 |
|  10 |   10 |             1 |
|  11 |   11 |             1 |
|  12 |   12 |             3 |
|  14 |   14 |             1 |
|  16 |   16 |             2 |
|  17 |   17 |             1 |
|  18 |   18 |             2 |
|  19 |   19 |             1 |
|  20 |   20 |             2 |
|  21 |   21 |             1 |
|  22 |   22 |             2 |
+-----+------+---------------+
18 rows in set (0.01 sec)

MariaDB [social]> SELECT entity_id, COUNT(cid) FROM comment_field_data WHERE entity_type='node' GROUP BY entity_id;
+-----------+------------+
| entity_id | COUNT(cid) |
+-----------+------------+
|         1 |          1 |
|         3 |          5 |
|         4 |          1 |
|         5 |          1 |
|         6 |          1 |
|         7 |          1 |
|         9 |          2 |
|        10 |          1 |
|        11 |          1 |
|        12 |          3 |
|        14 |          1 |
|        16 |          2 |
|        17 |          1 |
|        18 |          2 |
|        19 |          1 |
|        20 |          2 |
|        21 |          1 |
|        22 |          2 |
+-----------+------------+
18 rows in set (0.00 sec)
```

Kingdutch:
I've simplified the code in 4a9fefee6e75f8349cf8c543a3a8414d7a34dfa0

I've done this only by testing the query that should be constructed directly in the database. I haven't actually functionally tested this in an Open Social website, that's why I've committed it to a separate branch.

Please verify that it works and include it in this PR's branch if it does. If you spot errors, please sent me a message. After that this is good to go :)

rollki:
@Kingdutch 
I tested your changes, added to the PR and made some improvements there, please take a look :) 
https://github.com/goalgorilla/open_social/pull/1759/commits/d1f6d0e317fadef90a2036f3a6ab04ac735912e9

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -284,4 +286,110 @@ public static function processBlockForm(array $element, FormStateInterface $form
...
+    switch ($sort_by) {
+      // Creates a join to select the number of comments for a given entity
+      // in a recent timeframe and use that for sorting.
+      case 'most_commented':
+        if ($entity_type_id === 'group') {
+          $query->leftJoin('post__field_recipient_group', 'pfrg', "base_table.${entity_id_key} = pfrg.field_recipient_group_target_id");
+          $query->leftJoin('comment_field_data', 'cfd', "(base_table.${entity_id_key} = cfd.entity_id AND cfd.entity_type=:entity_type) OR (pfrg.entity_id = cfd.entity_id AND cfd.entity_type='post')", ['entity_type' => $entity_type_id]);
+        }
+        // Otherwise only check direct votes.
+        elseif ($entity_type_id === 'node') {
```
Kingdutch: We're not doing anything node specific here so we should just make this `else` so that this also works with other entities (e.g. posts)

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -284,4 +286,110 @@ public static function processBlockForm(array $element, FormStateInterface $form
...
+      // Creates a join to select the number of likes for a given entity in a
+      // recent timeframe and use that for sorting.
+      case 'most_liked':
+        // For groups also check likes on posts in groups. This does not (yet) take into account likes on comments on posts or likes on other group content entities.
+        if ($entity_type_id === 'group') {
+          $query->leftJoin('post__field_recipient_group', 'pfrg', "base_table.${entity_id_key} = pfrg.field_recipient_group_target_id");
+          $query->leftJoin('votingapi_vote', 'vv', "(base_table.${entity_id_key} = vv.entity_id AND vv.entity_type=:entity_type) OR (pfrg.entity_id = vv.entity_id AND vv.entity_type='post')", ['entity_type' => $entity_type_id]);
+        }
+        // Otherwise only check direct votes.
+        elseif ($entity_type_id === 'node') {
```
Kingdutch: We're not doing anything node specific here so we should just make this else. This also ensures that the expression after the if/else doesn't break when this is used with e.g. posts ;-)

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -284,4 +286,110 @@ public static function processBlockForm(array $element, FormStateInterface $form
...
+      // Creates a join that pulls in all related entities, taking the highest
+      // update time for all related entities as last interaction time and using
+      // that as sort value.
+      case 'last_interacted':
+        if ($entity_type_id === 'group') {
+          $query->leftJoin('group_content_field_data', 'gfd',"base_table.${entity_id_key} = gfd.gid");
+          $query->leftjoin('post__field_recipient_group', 'pst', "base_table.${entity_id_key} = pst.field_recipient_group_target_id");
+          $query->leftjoin('post_field_data', 'pfd', 'pst.entity_id = pfd.id');
+          $query->leftjoin('comment_field_data', 'cfd', '(gfd.entity_id = cfd.entity_id) OR (pfd.id = cfd.entity_id)');
+          $query->leftJoin('votingapi_vote', 'vv', '(gfd.entity_id = vv.entity_id) OR (pfd.id = vv.entity_id) OR (cfd.cid = vv.entity_id)');
```
Kingdutch: The join conditions should include entity type checks because IDs may overlap. So if that's not included then a vote for a group could count as a vote for a post for example.

See https://github.com/goalgorilla/open_social/pull/1759/files#diff-07da4bebe69fec9d2e6697ade1478977R336

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -284,4 +286,110 @@ public static function processBlockForm(array $element, FormStateInterface $form
...
+
+          $query->groupBy("base_table.${entity_id_key}");
+          $query->orderBy('newest_timestamp', 'DESC');
+        }
+        elseif ($entity_type_id === 'node') {
+          $query->leftJoin('node_field_data', 'nfd', "base_table.${entity_id_key} = nfd.nid");
+          // Comment entity.
+          $query->leftjoin('comment_field_data', 'cfd', 'nfd.nid = cfd.entity_id');
+          // Like node or comment related to node.
+          $query->leftjoin('votingapi_vote', 'vv', '(nfd.nid = vv.entity_id) OR (cfd.cid = vv.entity_id)');
```
Kingdutch: The join conditions should include entity type checks because IDs may overlap. So if that's not included then a vote for a group could count as a vote for a post for example.

See https://github.com/goalgorilla/open_social/pull/1759/files#diff-07da4bebe69fec9d2e6697ade1478977R336

modules/social_features/social_content_block/src/ContentBuilder.php
rollki: Fixed.

modules/social_features/social_content_block/src/ContentBuilder.php
rollki: Fixed.

modules/social_features/social_content_block/src/ContentBuilder.php
rollki: Fixed

modules/social_features/social_content_block/src/ContentBuilder.php
rollki: Fixed

rollki:
@Kingdutch 
I added sorting support by most commented and most liked for the group and fix all your comments.
Please check it out )

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -312,14 +312,15 @@ protected function sortBy(SelectInterface $query, EntityTypeInterface $entity_ty
...
+          $query->leftJoin('comment_field_data', 'cfd', "(base_table.${entity_id_key} = cfd.entity_id AND cfd.entity_type=:entity_type) OR (pfrg.entity_id = cfd.entity_id AND cfd.entity_type='post') OR (gfd.entity_id = cfd.entity_id AND cfd.entity_type='node')", ['entity_type' => $entity_type_id]);
         }
         // Otherwise only check direct votes.
-        elseif ($entity_type_id === 'node') {
+        else {
           $query->leftJoin('comment_field_data', 'cfd', "base_table.${entity_id_key} = cfd.entity_id AND cfd.entity_type=:entity_type", ['entity_type' => $entity_type_id]);
         }
 
-        $query->addExpression('COUNT(cfd.cid)', 'comment_count');
+        $query->addExpression('COUNT(distinct cfd.cid)', 'comment_count');
```
Kingdutch: @rollki Could you add why the `distinct` is needed here? I believe in the way we select the comments they should already be distinct, otherwise there's a different error.

`distinct` can cost quite a bit of performance so I'd prefer to avoid its use if we can.

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -333,13 +334,14 @@ protected function sortBy(SelectInterface $query, EntityTypeInterface $entity_ty
...
+          $query->leftJoin('group_content_field_data', 'gfd', "base_table.${entity_id_key} = gfd.gid");
+          $query->leftJoin('votingapi_vote', 'vv', "(base_table.${entity_id_key} = vv.entity_id AND vv.entity_type=:entity_type) OR (pfrg.entity_id = vv.entity_id AND vv.entity_type = 'post') OR (gfd.entity_id = vv.entity_id AND vv.entity_type = 'node')", ['entity_type' => $entity_type_id]);
         }
         // Otherwise only check direct votes.
-        elseif ($entity_type_id === 'node') {
+        else {
           $query->leftJoin('votingapi_vote', 'vv', "base_table.${entity_id_key} = vv.entity_id AND vv.entity_type=:entity_type", ['entity_type' => $entity_type_id]);
         }
-        $query->addExpression('COUNT(vv.id)', 'vote_count');
+        $query->addExpression('COUNT(distinct vv.id)', 'vote_count');
```
Kingdutch: Same question here :)

Kingdutch:
Tests are green but test 1 is failing conding standards.

```
FILE: ...ules/social_features/social_content_block/src/ContentBuilder.php

----------------------------------------------------------------------

FOUND 1 ERROR AND 1 WARNING AFFECTING 2 LINES

----------------------------------------------------------------------

 334 | WARNING | [ ] Line exceeds 80 characters; contains 165

     |         |     characters

 359 | ERROR   | [x] Expected one space after the comma, 0 found

----------------------------------------------------------------------

PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY

----------------------------------------------------------------------
```

modules/social_features/social_content_block/src/ContentBuilder.php
rollki: @Kingdutch 
We can not COUNT without `distinct` groups entities, because we have joined to `group_content_field_data` and `post__field_recipient_group` and this multiplies the sample. 
I could not solve this problem in a simple way, maybe with a subquery it is possible, but there will be more code, it will be unreadable and even slower.

But we can add `if` condition and don't do this for other entities.

rollki:
@Kingdutch 
Coding standards were fixed.

Kingdutch:
Backported to 8.x-8.x in b09e3ec0a

PR #1760, opened 2020-03-13
Issue #3119681 by chmez: The body of the custom content block is absent on the Layout Builder page

ghost:
## Problem
When the **Custom content list block** custom block is added to the section of the **Layout Builder** module then only the block title is visible on the overview page

## Solution
Use entity extra field for the block body

## Issue tracker
- https://www.drupal.org/project/social/issues/3119681
- https://getopensocial.atlassian.net/browse/YANG-2544

## How to test
- [ ] Enable the following modules: **Social Content Block**, **Layout Builder**
- [ ] Create the **Test Block** custom block of the **Custom content list block** type
- [ ] Create the **Test Content Type** content type with the field of the **Layout Builder** module
- [ ] Create the **Test Node** of the **Test Content Type** content type and add the **Test Block** custom block to the section of the **Layout Builder** module
- [ ] When you save the node then you should see topics in the body of the **Test Block** custom block

## Release notes
*Probably not applicable because it fixes a bug that has never been in a released version.*

> Improve the content block for using on the pages based on the **Layout Builder** module sections

### DIFF
--- a/modules/custom/social_auth_extra/social_auth_extra.module
+++ b/modules/custom/social_auth_extra/social_auth_extra.module
@@ -202,7 +202,7 @@ function social_auth_extra_form_user_register_form_submit($form, FormStateInterf
 
             // Redirect to profile.
             $form_state->setRedirect(
-              'entity.profile.type.user_profile_form', [
+              'profile.user_page.single', [
                 'user' => $account->id(),
                 'profile_type' => 'profile',
                 [],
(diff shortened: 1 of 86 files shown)


Kingdutch (approved):

Kingdutch:
Cherry picked to 8.x-8.x as fe3a9106d

PR #1764, opened 2020-03-16
Fix compatibility with Mink 1.8.0

Kingdutch:
Mink 1.8.0 no longer automatically starts a session when `getSession` is
called. This breaks some `beforeScenario` steps that rely on the
sessions (for example, testing JS support or resizing the window).

This commit simply checks whether the session is started before calling
any methods on the session, starting it if needed. This is only done in
`beforeScenario` calls that actually use the session.

For `FeatureContext` the beforeScenario calls are merged because I'm
unsure whether we can rely on call order.

Change in Mink: https://github.com/minkphp/Mink/pull/705

*Should be backported to all supported branches*

## Issue tracker
No issue, issue in testing.

## How to test
- [ ] Travis should be green again

## Release notes
N/A

## Change Record
N/A


Kingdutch:
Cherry-picked on 8.x-8.x to c1221a34f and on 8.x-7.x to 766a74639

PR #1768, opened 2020-03-18
Issue #3120580 by Kingdutch: Label of hero image in Sky flavour is vi…

Kingdutch:
…sible

<h2>Problem</h2>

The templates added for the Sky theme were improved in that they more closely render what is configured in the Drupal display layer. However, because this was previously not the case, some configurations are set to display field labels where this shouldn't happen.

<h2>Solution</h2>
Configure the hero view mode for basic pages so that it doesn't show the label for the image field.

## Issue tracker
https://www.drupal.org/project/social/issues/3120580

## How to test
- [ ] Create and view a basic page in the sky flavour

## Release notes
The label of the image field in the hero when using the Sky theme is now properly hidden.

### DIFF
--- a/modules/social_features/social_page/config/install/core.entity_view_display.node.page.hero.yml
+++ b/modules/social_features/social_page/config/install/core.entity_view_display.node.page.hero.yml
@@ -29,7 +29,7 @@ content:
   field_page_image:
     type: image
     weight: 0
-    label: above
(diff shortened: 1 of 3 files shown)


Kingdutch:
Cherry-picked to 599972f24 on 8.x-8.x

PR #1769, opened 2020-03-18
Issue #3115796 by Kingdutch, ronaldtebrake: Create update hook after …

Kingdutch:
…replacing SocialLanguageMetadataBubblingUrlGenerator by Entity::toUrl implementations

[#3098046] In this issue we replaced the SocialLanguageMetadataBubblingUrlGenerator by proper Entity::toUrl implementations.
one of the issues tackled in there was to alter the social autocomplete search view. This was never covered in an update hook.
https://github.com/goalgorilla/open_social/pull/1626/files#diff-9e6738ee851907d213dcf26d545c3eeb

This resulted in weird autocomplete URL's where there was still a prefix for groups / users so url's for search result lead to a page not found.

Ensure the update hook is written so people have the correct changes in their configuration.

## Issue tracker
https://www.drupal.org/project/social/issues/3115796

## How to test
- [ ] Set site language to be non-english and ensure strings like "user" and "group" are  translated
- [ ] Confirm that Search Autocomplete fails (requires install from before 8.0)
- [ ] Run updates
- [ ] Confirm that Search Autocomplete works

## Release notes
The search suggestions now work properly for non-English sites after updating. This was previously only fixed for new installations and not for existing platforms.

### DIFF
--- a/modules/social_features/social_search/modules/social_search_autocomplete/config/install/views.view.search_all_autocomplete.yml
+++ b/modules/social_features/social_search/modules/social_search_autocomplete/config/install/views.view.search_all_autocomplete.yml
@@ -11,8 +11,6 @@ dependencies:
     - search_api
     - serialization
     - user
-_core:
-  default_config_hash: OqzYowGv0kur_1XWmN81Ha5sf6JuKniFTDMmEyUsUi0
 id: search_all_autocomplete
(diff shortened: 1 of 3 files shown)


ronaldtebrake:
Also in 8.2 :)

PR #1770, opened 2020-03-18
Add update and install hook to remove follow content button from topic small teassers

Kingdutch:
<h2>Problem</h2>
The follow content button is added after modules have been installed. 
It's also the result of an optional module so it's possible that the 
button doesn't exist. These factors don't allow the view display 
configuration for the field to be set in configuration that is shipped 
with our modules. 

<h2>Solution</h2>
To work around this an modules_installed hook is implemented that makes 
the change for us. An update hook is provided to fix platforms that 
installed a version that had the new templates to render before this fix 
was present.

## Issue tracker
https://www.drupal.org/project/social/issues/3120644

## How to test
- [ ] Install a platform and login to view the default stream page
- [ ] Before: with follow content buttons in recent topic block
- [ ] After: no follow content buttons in recent topic block

## Release notes
A new style for topic teasers caused the "follow content" button to show up in the teaser erroneously. This has been resolved.

### DIFF
--- a/modules/social_features/social_follow_content/config/update/social_follow_content_update_8001.yml
+++ b/modules/social_features/social_follow_content/config/update/social_follow_content_update_8001.yml
@@ -1,0 +1,11 @@
+core.entity_view_display.node.topic.small_teaser:
+  expected_config:
+    content:
+      flag_follow_content: { }
+  update_actions:
+    add:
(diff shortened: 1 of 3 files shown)


ronaldtebrake:
Also cherry picked :) nice work!

PR #1766, opened 2020-03-16
Issue #3120072 by Kingdutch: Move featured items to a separate module

Kingdutch:
<h2>Problem</h2>
The featured items paragraphs should ideally be re-used by the Dashbaord
module. 

<h2>Solution</h2>
Move the functionality for featured items out into a
separate module so that the Dashboard module can use what is needed and
nothing should functionally change for the landing page module.

## Issue tracker
https://www.drupal.org/project/social/issues/3120072

## How to test
- [ ] Featured items on landing pages should still work as before

## Release notes
The featured items block for landing pages is now in a separate module so that it may be reused in other modules.

PR #1779, opened 2020-03-26
Issue #3122686 by chmez: Move featured content to a separate module

ghost:
## Problem
The featured content paragraphs should ideally be re-used by the Dashboard module

## Solution
Move the functionality for featured content out into a separate module so that the Dashboard module can use what is needed and nothing should functionally change for the landing page module

## Issue tracker
- https://www.drupal.org/project/social/issues/3122686
- https://getopensocial.atlassian.net/browse/YANG-2577

## How to test
- [ ] Featured content on landing pages should still work as before

## Release notes
The featured content block for landing pages is now in a separate module so that it may be reused in other modules.


modules/social_features/social_landing_page/templates/group--featured.html.twig
```diff
@@ -95,7 +95,9 @@
     </span>
   {% endif %}
 
-  <a href="{{ url }}" class="card__link" title="{{ label }}">
-    {% trans %}Read more{% endtrans %}
-  </a>
+  {% if not hide_card_link %}
```
ribel, line 98: @OH17L Have you tested if this condition will work correctly without Dashboard module? I mean if it will be always displayed if there is no variable set?

OH17L: @ribel, yes, I tested it and it works correctly

ribel (approved):
I have checked the functionality and the code and all looks good to me.

ribel:
@Kingdutch Could you review this one, since you made similar changes for Featured Items?
If needed - I can provide a link to the environment, where you can test this.

modules/social_features/social_landing_page/social_landing_page.install
```diff
@@ -82,8 +83,16 @@ function _social_landing_page_get_permissions($role) {
...
+  _social_landing_page_install_module('social_featured_items');
+}
 
+/**
+ * Install a separate module of the moved section.
+ *
+ * @param string $module
+ *   The module name.
+ */
+function _social_landing_page_install_module($module) {
```
Kingdutch, line 95: I don't think we should create a separate function here but should instead opt for code duplication.

A separate function indicates that this is something that can be done more often and encourages re-using the solution. Duplicating the code shows that it was a solution used for a very specific case (which this is).

Duplicating the code also makes cleaning up old update hooks easier in the future.

modules/social_features/social_landing_page/social_landing_page.libraries.yml
```diff
@@ -10,11 +10,6 @@ section.introduction:
     theme:
       css/social_landing_page.section.introduction.css: {}
 
-section.featured:
-  css:
-    theme:
-      css/social_landing_page.section.featured.css: {}
-
```
Kingdutch, line 17: Removing this library is a backwards-incompatible change and may break other modules or themes overriding/extending the library. Ideally we mark the library as deprecated but keep using it for now.

modules/social_features/social_landing_page/social_landing_page.libraries.yml
```diff
@@ -30,11 +25,6 @@ section.block:
     theme:
       css/social_landing_page.section.block.css: {}
 
-featured:
-  css:
-    theme:
-      css/social_landing_page.featured.css: {}
-
```
Kingdutch, line 37: Removing this library is a backwards-incompatible change and may break other modules or themes overriding/extending the library. Ideally we mark the library as deprecated but keep using it for now.

modules/social_features/social_landing_page/templates/group--featured.html.twig
```diff
@@ -95,7 +95,9 @@
...
+    <a href="{{ url }}" class="card__link" title="{{ label }}">
+      {% trans %}Read more{% endtrans %}
+    </a>
```
Kingdutch, line 101: Is there a reason we don't just generate this link in a preprocess hook and render it as `{{ read_more }}` ? Then if a preprocess hook wants to change it or hide it that can be done without introducing a special logic variable.

modules/social_features/social_landing_page/templates/node--featured.html.twig
```diff
@@ -156,11 +156,13 @@
...
-            <span class="visually-hidden">{% trans %}about {{ label }}{% endtrans %} </span>
-          </a>
-        </div>
+        {% if not hide_card_link %}
+          <div class="card__link">
+            <a href="{{ url }}" rel="bookmark">{{ 'Read more'|t }}
+              <span class="visually-hidden">{% trans %}about {{ label }}{% endtrans %} </span>
+            </a>
+          </div>
+        {% endif %}
```
Kingdutch, line 165: Is there a reason we don't just generate this link in a preprocess hook and render it as {{ read_more }} ? Then if a preprocess hook wants to change it or hide it that can be done without introducing a special logic variable.

modules/social_features/social_landing_page/templates/profile--featured.html.twig
```diff
@@ -39,9 +39,11 @@
...
-      <span class="visually-hidden">{% trans %}about {{ label }}{% endtrans %} </span>
-    </a>
-  </div>
+  {% if not hide_card_link %}
+    <div class="card__link">
+      <a href="{{ profile_stream_url }}" rel="bookmark">{{ 'Read more'|t }}
+        <span class="visually-hidden">{% trans %}about {{ label }}{% endtrans %} </span>
+      </a>
+    </div>
+  {% endif %}
```
Kingdutch, line 48: Is there a reason we don't just generate this link in a preprocess hook and render it as {{ read_more }} ? Then if a preprocess hook wants to change it or hide it that can be done without introducing a special logic variable.

themes/socialblue/templates/node/event/node--event--featured--sky.html.twig
```diff
@@ -77,11 +77,13 @@
...
-        <span class="visually-hidden">{% trans %}about {{ label }}{% endtrans %} </span>
-      </a>
-    </div>
+    {% if not hide_card_link %}
+      <div class="card__link">
+        <a href="{{ url }}" rel="bookmark">{{ 'Read more'|t }}
+          <span class="visually-hidden">{% trans %}about {{ label }}{% endtrans %} </span>
+        </a>
+      </div>
+    {% endif %}
```
Kingdutch, line 86: Is there a reason we don't just generate this link in a preprocess hook and render it as {{ read_more }} ? Then if a preprocess hook wants to change it or hide it that can be done without introducing a special logic variable.

themes/socialblue/templates/node/node--featured--sky.html.twig
```diff
@@ -73,11 +73,13 @@
...
-        <span class="visually-hidden">{% trans %}about {{ label }}{% endtrans %} </span>
-      </a>
-    </div>
+    {% if not hide_card_link %}
+      <div class="card__link">
+        <a href="{{ url }}" rel="bookmark">{{ 'Read more'|t }}
+          <span class="visually-hidden">{% trans %}about {{ label }}{% endtrans %} </span>
+        </a>
+      </div>
+    {% endif %}
```
Kingdutch, line 82: Is there a reason we don't just generate this link in a preprocess hook and render it as {{ read_more }} ? Then if a preprocess hook wants to change it or hide it that can be done without introducing a special logic variable.

Kingdutch (changes requested):

ribel:
@OH17L could you take a look to this tomorrow morning?

modules/social_features/social_landing_page/social_landing_page.install
OH17L: Done

modules/social_features/social_landing_page/social_landing_page.libraries.yml
OH17L: I did the same as you did [here](https://github.com/goalgorilla/open_social/pull/1766/files#diff-37d3a37e02d24dff66f6a231753c2c3d)

modules/social_features/social_landing_page/social_landing_page.libraries.yml
OH17L: I did the same as you did [here](https://github.com/goalgorilla/open_social/pull/1766/files#diff-37d3a37e02d24dff66f6a231753c2c3d)

modules/social_features/social_landing_page/templates/group--featured.html.twig
OH17L: This is done to make it easier to add new classes, wrappers, etc. by front-end developers

modules/social_features/social_landing_page/templates/node--featured.html.twig
OH17L: This is done to make it easier to add new classes, wrappers, etc. by front-end developers

modules/social_features/social_landing_page/templates/profile--featured.html.twig
OH17L: This is done to make it easier to add new classes, wrappers, etc. by front-end developers

themes/socialblue/templates/node/event/node--event--featured--sky.html.twig
OH17L: This is done to make it easier to add new classes, wrappers, etc. by front-end developers

themes/socialblue/templates/node/node--featured--sky.html.twig
OH17L: This is done to make it easier to add new classes, wrappers, etc. by front-end developers

modules/social_features/social_landing_page/social_landing_page.libraries.yml
ribel: Moving this one is fine as for me. As Oleksandr said it is the same as we did with featured items before.
But maybe we need to create Change records for both: https://www.drupal.org/list-changes/social

modules/social_features/social_landing_page/social_landing_page.libraries.yml
ribel: Moving this one still need some work, as this change indeed breaks `node--featured.html.twig`, where we have a string `{{ attach_library('social_landing_page/featured') }}`.
So I think if we want to make `social_featured_content` fully decoupled we also need to move this library and all 
`--featured.html.twig` templates from `social_landing_page`, but not sure if to the `social_featured_content`, or just to the `socialbase` theme - to be consistent with other entity templates.
@Kingdutch what do you think?

modules/social_features/social_landing_page/templates/group--featured.html.twig
ribel: I think it was just done like this before, and within this story, we didn't refactor templates themselves.
Oleksandr just adds a condition, which allows hiding read more link, which is relevant for the Dashboard.

modules/social_features/social_landing_page/templates/node--featured.html.twig
ribel: I think it was just done like this before, and within this story, we didn't refactor templates themselves.
Oleksandr just adds a condition, which allows hiding read more link, which is relevant for the Dashboard.

modules/social_features/social_landing_page/templates/profile--featured.html.twig
ribel: I think it was just done like this before, and within this story, we didn't refactor templates themselves.
Oleksandr just adds a condition, which allows hiding read more link, which is relevant for the Dashboard.

themes/socialblue/templates/node/event/node--event--featured--sky.html.twig
ribel: I think it was just done like this before, and within this story, we didn't refactor templates themselves.
Oleksandr just adds a condition, which allows hiding read more link, which is relevant for the Dashboard.

themes/socialblue/templates/node/node--featured--sky.html.twig
ribel: I think it was just done like this before, and within this story, we didn't refactor templates themselves.
Oleksandr just adds a condition, which allows hiding read more link, which is relevant for the Dashboard.

modules/social_features/social_landing_page/social_landing_page.libraries.yml
Kingdutch: The templates are for the `featured` view mode so I think it depends:

1) Should this template only be used on landing pages, then we should probably renamed them so they only apply on landing pages. This requires us to update the `attach_library` call.
2) (more likely) If these templates are for any place where this view mode is used (which the name `node--featured` suggests, then it makes sense to move the template files to the module that provides those view modes.

My underlying philosophy for 2 and not moving to `socialbase` is that I think we have quite a few templates with crucial functionality in `socialbase` that should actually have been put in modules. The idea is that those modules should work even if the Open Social themes are not used. (That is not currently true)

modules/social_features/social_landing_page/templates/group--featured.html.twig
Kingdutch: Okay : ) Maybe we can check in the future if we can improve this a little bit. I think from a `link` class a themer should be able to add classes if needed. I have the feeling that we're adding quite a few one-off flags like this in the template and would prefer to keep that logic out of the templates themselves and higher up in the theme layer.

Lets leave it as is for now :)

modules/social_features/social_landing_page/social_landing_page.libraries.yml
ribel: Since with this PR, we want to use Featured view mode not only on Landing Pages, but also on a Dashboard, the answer is 2 ;)
Currently, both templates and configs for featured view mode are still in `social_landing_page`, but if I understand correctly - we need to move them to this new `social_featured_content` - I will do that now.

Kingdutch: Ah okay, yeah I think the view mode needs to move too. Otherwise the `social_featured_content` module would have a dependency on `social_landing_page` because the featured-item paragraphs render in that view mode.

ribel: I have moved all featured configs and templates in 18f7bbf and 79e213a

ribel (approved):
I have moved all featured configs and templates to the `social_featured_content` module.
Also applied it as a patch on ECI Demo and all works good there, with modules that extend this featured view mode.

ribel:
Behat tests pass :white_check_mark:
Merged and cherry-picked to the 8.x-8.x 🍒

PR #1780, opened 2020-03-26
Issue #3120797: Access permission checks for event invite

bramtenhove:
## Problem
The event invite feature needs access checks.

## Solution
Added access checks for all the blocks and routes.

## How to test
- [ ] Create an event outside of any group
- [ ] Go to /admin/config/opensocial/event-invite
- [ ] Disable the global permission
- [ ] Go to the manage enrollment tab of the event, see that you do not have a block to invite people
- [ ] Go directly to the invite pages: /node/*/invite/user, /node/*/invite/email and /node/*/all-enrollment-invites. Observe that you do not have access
- [ ] Enable the global permission
- [ ] Observe for the above that you do have access
- [ ] Play around with the group type settings, they should not influence the access
- [ ] Place the event in a group
- [ ] Set the group type setting to off for the group type, results in access denied for those pages
- [ ] Set the group type setting to on for the group type, results in correct pages
- [ ] Play around more :)

Also test in combination with event organizers.


modules/social_features/social_event/modules/social_event_invite/config/install/views.view.event_manage_enrollment_invites.yml
```diff
@@ -28,7 +28,7 @@ display:
     position: 0
     display_options:
       access:
-        type: event_owner_or_organizer
+        type: none
```
bramtenhove: This is now handled by the `RouteSubscriber` and `Access` check together with other routes and blocks.

modules/social_features/social_event/modules/social_event_invite/src/Routing/RouteSubscriber.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    if ($route = $collection->get('view.event_manage_enrollment_invites.page_manage_enrollment_invites')) {
+      $requirements = $route->getRequirements();
+      $requirements['_custom_access'] = 'social_event_invite.access::eventFeatureAccess';
+      $route->setRequirements($requirements);
+    }
+
+    // @todo: Needs the correct view route for the user invites overview.
+    if ($route = $collection->get('view...')) {
```
bramtenhove: The correct route needs to be added in the main branch when the user invites overview is ready.

modules/social_features/social_event/modules/social_event_invite/src/SocialEventInviteAccessHelper.php
```diff
@@ -0,0 +1,159 @@
...
+    $enabled_global = $config->get('invite_enroll');
+
+    // If it's globally disabled, we don't want to show the block.
+    if (!$enabled_global) {
+      return AccessResult::forbidden();
+    }
+
+    // Get the group of this node.
+    $node = $this->routeMatch->getParameter('node');
+    if ($node instanceof NodeInterface) {
```
bramtenhove: The `$node` in this case is sometimes a full node object, sometimes just the ID.

modules/social_features/social_event/modules/social_event_invite/config/install/message.template.invite_event_enrollment.yml
```diff
@@ -4,6 +4,9 @@ status: true
 dependencies:
   module:
     - activity_logger
+  enforced:
```
bramtenhove: Now the module cleanly uninstalls.

modules/social_features/social_event/modules/social_event_invite/config/install/social_event_invite.settings.yml
```diff
@@ -0,0 +1,10 @@
+invite_enroll: true
+invite_subject: '[user:name] has invited you to join the event [node:title] on [site:name]'
```
ronaldtebrake: should this be display name instead? This will ensure correct fallbacks for example when users are hiding their name

$tests['[user:name]'] = $account->getAccountName();
    $tests['[user:account-name]'] = $account->getAccountName();
    $tests['[user:display-name]'] = $account->getDisplayName();

PR #1783, opened 2020-03-29
Issue #3123324 by SV: As a CM I want to filter items relevant to a specific tag on the stream block

volodymyr-sydor-lemberg:
## Problem
- Should be possible to add activities blocks that shows items filtered base on specific tags on the landing/dashboard pages

## Solution
- Create additional views display plugin with dependent taxonomy filter for activity views block  

## Issue tracker
- https://getopensocial.atlassian.net/browse/YANG-2580
- https://getopensocial.atlassian.net/browse/YANG-3463
- https://www.drupal.org/project/social/issues/3123324

## How to test
- [ ]  Click edit layout on any dashboard node
- [ ]  Scroll to any section and click to "Add block"
- [ ]  After click - dashboard panel should be displayed at the right side
- [ ]  Select at the bottom an activity stream views block (without post)
- [ ]  On the adding block form should be displayed optional dependent dropdown to filter items in the block 
   - first dropdown: select any Taxonomy Vocabulary
   - second multiple dropdown: select any Taxonomy terms (which should belong to selected 
vocabulary from 1-st dropdown)
  -  or we skip those two selections then items shouldn't be filtered
- [ ]  After click save/update button, should be displayed activity stream block which is filtered by selected terms

## Screenshots
Edit layout for Test dashboard   Global Dev Hub

Edit layout for Test dashboard   Global Dev Hub (1)

Test dashboard   Global Dev Hub
Screenshot from 2020-12-27 21-02-12

## Release notes
Add optional module Social Activity Filter, which allows filter content in the "Activity stream (without post)" block by specific tags.


ribel:
@bramtenhove, @ronaldtebrake, @Kingdutch 
Is this feature interesting for the Distro? If yes - to what release 8.3 or 9.0?
Or should we move it to the enterprise module?

modules/social_features/social_activity/modules/social_activity_filter/src/Plugin/views/display/FilterBlock.php
```diff
@@ -0,0 +1,338 @@
...
+      switch ($type) {
+        case 'vocabulary':
+          $form['override']['vocabulary'] = [
+            '#type' => 'select',
+            '#title' => $this->t('Vocabulary'),
+            '#options' => $this->getVocabularyOptionsList(),
+            '#default_value' => $block_configuration['vocabulary'],
+            '#empty_option' => t('None'),
+            '#required' => TRUE,
+            '#ajax' => [
```
ribel: @volodymyr-sydor-lemberg need to check if it works in all cases:

- Blocks library
- Dashboard
- Landing Page

modules/social_features/social_activity/modules/social_activity_filter/src/Plugin/views/filter/ActivityFilterTags.php
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Filters out activity items by the taxonomy tags.
+   */
+  public function query() {
+    $tags = $this->view->filter_tags;
+    $taxonomy_field = $this->view->filter_vocabulary;
+    $taxonomy_table = "node__{$taxonomy_field}";
+
+    // @todo: replace it!
```
ribel: @volodymyr-sydor-lemberg As you noted yourself this need to be improved, with using DI.

modules/social_features/social_activity/modules/social_activity_filter/src/Form/FilterSettingsForm.php
```diff
@@ -0,0 +1,149 @@
...
+   * @return array
+   *   Mapped array: vid => taxonomy_field.
+   */
+  public function getReferencedTaxonomyFields(array $vocabulary_list) {
+
+    $field_names = [];
+    foreach ($vocabulary_list as $vocabulary) {
+
+      foreach (NodeType::loadMultiple() as $content_type => $type) {
+        $field_definitions = \Drupal::service('entity_field.manager')
```
ribel: @volodymyr-sydor-lemberg Please use Dependency Injection in this and other classes.

modules/social_features/social_activity/modules/social_activity_filter/src/Form/FilterSettingsForm.php
```diff
@@ -0,0 +1,149 @@
...
+
+    $vocabulariesList = [];
+
+    /** @var \Drupal\taxonomy\Entity\Vocabulary */
+    foreach (Vocabulary::loadMultiple() as $vid => $vocabulary) {
+      $referencedField = isset($taxonomyFields[$vid]) ? $taxonomyFields[$vid] : $this->t('none');
+      $vocabulariesList[$vid] = $vocabulary->get('name') . " (field: $referencedField)";
+    }
+
+    $form['social_activity_filter']['vocabulary'] = [
```
ribel: @volodymyr-sydor-lemberg Please add a description which explains that field will be added automatically.
What will happen if both content types use the same taxonomy but with different fields?

modules/social_features/social_activity/modules/social_activity_filter/src/Plugin/views/filter/ActivityFilterTags.php
```diff
@@ -0,0 +1,74 @@
...
+            'value' => 'node',
+          ],
+        ],
+      ];
+      $join = Views::pluginManager('join')
+        ->createInstance('standard', $configuration);
+      $this->query->addRelationship($taxonomy_field, $join, $taxonomy_table);
+
+      $and_wrapper = db_and();
+      $and_wrapper->condition("{$taxonomy_field}.{$taxonomy_field}_target_id", $tags, 'IN');
```
ribel: @volodymyr-sydor-lemberg Does it work only for Nodes, or also for Posts in activity stream if they have taxonomy field?

modules/social_features/social_activity/modules/social_activity_filter/src/SocialActivityFilterOverride.php
```diff
@@ -0,0 +1,108 @@
...
+class SocialActivityFilterOverride implements ConfigFactoryOverrideInterface {
+
+  /**
+   * Load overrides.
+   */
+  public function loadOverrides($names) {
+    $overrides = [];
+
+    // Override activity_stream views.
+    $config_name = 'views.view.activity_stream';
```
ribel: @volodymyr-sydor-lemberg As discussed we need to support it for both: Home and Community streams.

modules/social_features/social_activity/modules/social_activity_filter/src/Plugin/views/filter/ActivityFilterTags.php
volodymyr-sydor-lemberg: @ribel Yes, it works only for Nodes, for Posts we should extend all related functionality

modules/social_features/social_activity/modules/social_activity_filter/src/Form/FilterSettingsForm.php
volodymyr-sydor-lemberg: @ribel I've added description for those cases

modules/social_features/social_activity/modules/social_activity_filter/src/Plugin/views/display/FilterBlock.php
volodymyr-sydor-lemberg: @ribel For blocks and dashboard it works good, only I added an additional condition to skip it for landing page

volodymyr-sydor-lemberg: done

modules/social_features/social_activity/modules/social_activity_filter/src/Plugin/views/filter/ActivityFilterTags.php
volodymyr-sydor-lemberg: done

modules/social_features/social_activity/modules/social_activity_filter/src/Form/FilterSettingsForm.php
volodymyr-sydor-lemberg: done

modules/social_features/social_activity/modules/social_activity_filter/src/Form/FilterSettingsForm.php
volodymyr-sydor-lemberg: done

modules/social_features/social_activity/modules/social_activity_filter/src/Plugin/views/filter/ActivityFilterTags.php
volodymyr-sydor-lemberg: done

modules/social_features/social_activity/modules/social_activity_filter/src/SocialActivityFilterOverride.php
volodymyr-sydor-lemberg: done

modules/social_features/social_activity/modules/social_activity_filter/src/Form/FilterSettingsForm.php
```diff
@@ -0,0 +1,174 @@
...
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\Core\Form\ConfigFormBase;
+use Drupal\Core\Form\FormStateInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Class FilterSettingsForm.
+ *
+ * @package Drupal\unpd_cop\Form
```
ribel, line 17: @volodymyr-sydor-lemberg please remove all references to UNPD or COP or from this PR.

modules/social_features/social_activity/modules/social_activity_filter/src/Plugin/views/filter/ActivityFilterTags.php
```diff
@@ -0,0 +1,169 @@
...
+      // Apply filter by tags to commented entity activity.
+      $join = Views::pluginManager('join')
+        ->createInstance('standard', $configuration);
+      $this->query->addRelationship('commented_nodes', $join, $comment_table);
+      $and_comment_wrapper->condition("commented_nodes.{$taxonomy_field}_target_id", $tags, 'IN');
+
+      $or->condition($and_comment_wrapper);
+
+      // Filter Posts by selected tags.
+      $post_cop_tags_table = 'post__field_cop_tags';
```
ribel, line 128: @volodymyr-sydor-lemberg Please move this to UNDP project.

volodymyr-sydor-lemberg: done

modules/social_features/social_activity/modules/social_activity_filter/src/Form/FilterSettingsForm.php
volodymyr-sydor-lemberg: done

modules/social_features/social_activity/social_activity.install
```diff
@@ -167,3 +172,90 @@ function social_activity_update_8801() {
...
+}
+
+/**
+ * Create "Featured" view mode/display for post.
+ *
+ * @todo: should be moved into social_post module.
+ */
+function social_activity_update_8803() {
+  // Create a new post featured entity view mode.
+  EntityViewMode::create([
```
ribel, line 209: @volodymyr-sydor-lemberg Please move this to social_post module.

modules/social_features/social_comment/social_comment.install
```diff
@@ -120,3 +120,17 @@ function social_comment_update_8002() {
...
+
+/**
+ * Set trimmed format for comment body in activity comment view mode.
+ */
+function social_comment_update_8802() {
+  $config = \Drupal::service('config.factory')
+    ->getEditable('core.entity_view_display.comment.comment.activity_comment');
+
+  if (!empty($config)) {
+    $config->set('content.field_comment_body.settings', ['trim_length' => 300]);
```
ribel, line 132: @volodymyr-sydor-lemberg Why you didn't use update helper for this change?

modules/social_features/social_activity/social_activity.install
volodymyr-sydor-lemberg: done

modules/social_features/social_comment/social_comment.install
volodymyr-sydor-lemberg: it just requires more source code
done

volodymyr-sydor-lemberg:
merged branch 8.x-9.x back and solved conflicts.

modules/social_features/social_activity/social_activity.install
```diff
@@ -199,44 +200,19 @@ function social_activity_update_8802() {
...
 function social_activity_update_8803() {
-  // Create a new post featured entity view mode.
-  EntityViewMode::create([
-    'targetEntityType' => 'post',
-    'id' => 'post.featured',
-    'status' => TRUE,
-    'label' => t('Featured'),
-  ])->save();
+  $config_file = 'views.view.activity_stream';
+  $config = drupal_get_path('module', 'social_activity') . '/config/optional/' . $config_file . '.yml';
```
ronaldtebrake, line 207: We need to make sure we use a static file in update hooks. 
The config/optional file might change before the update hook is run, using a static file in a separate folder.

e.g. config/static/views.view.activity_stream_8803.yml could already be an option. 
Or just use the update helper, i think in this case the update helper should definitely be able to do it.

volodymyr-sydor-lemberg: done

ronaldtebrake:
Hi @ribel @volodymyr-sydor-lemberg 

Thanks for this amazing feature :) i think it will be great if we can finish of some last details!

It seems this also works for landing pages, but I don't see it mentioned in the issue itself. Jira ticket is closed though.
Could you make sure Drupal and github issues reflect the full scope of this story? It's hard for people to understand what to test and what to expect. For example only an SM can access the settings page and configure what vocabularies are available while it looks like a CM can also add tags to the landing page / dashboard (?) It's confusing.

Some improvements from testing it on landing pages:
1. Where does a SM/CM add their settings, i can't find any menu link. (I can find it through looking in the code though) 
2. When going to a landing without configuring anything these fields don't add anything. Should we hide them?
Screenshot 2020-07-29 at 20 17 44
3. We need to spend some love and attention to the config form.
Screenshot 2020-07-29 at 20 17 57
This for me is confusing, not sure how a SM or CM should be able to understand how to configure this. We need to explain to them what they can do here. this (field: none) doesn't add anything nor the description below. Can we take a look at this from a UX point of view?
4. After I added a tag as filter, but this tag doesn't have any content referenced to it I would expect to not see any items in the activity stream. However I see all items anyhow. 
5. Doesn't it make sense to combine this with social_tagging? If social_tagging isn't enabled we are missing out on these tagging vocabulary which in my eyes is a great preset. So we might need to check on install if we can enable social_tagging perhaps.

For now I will need to move this to 10.0.

volodymyr-sydor-lemberg:
Hi @ronaldtebrake , it's about that config form, we could try to explain and improve UX but it's actually more for admin because If we allow SM or CM to add any of taxonomy vocabulary we should cover any cases to prevent issues. 

Probably we don't need it and we can implement simpler solution. For example we could just provide hook_activity_filter_option and use it in other modules if we need to add some vocabulary to allowed filter list. What do you think?

modules/social_features/social_activity/modules/social_activity_filter/social_activity_filter.info.yml
```diff
@@ -0,0 +1,6 @@
+name: 'Social Activity Filter'
+description: 'Provides filtered activity stream for Open Social.'
+type: module
+core: 8.x
+package: Social
```
ribel, line 5: @volodymyr-sydor-lemberg I think it's better to set package `Social (experimental)`

ribel (changes requested):
@volodymyr-sydor-lemberg Please fix my comments, and also:

1. Don't forget to add an installer_options.yml file so this optional module can be installed during the Drupal installation process.
2. On settings page we need to differentiate blocks with and without post form, because now we have two blocks with title `Community activities block` and `Homepage stream block`
3. Update the description of PR and Drupal.org issue with updated settings form other updates.

modules/social_features/social_activity/social_activity.install
```diff
@@ -198,49 +198,6 @@ function social_activity_update_8802() {
   EntityViewDisplay::create($display)->save();
 }
 
-/**
- * Create "Featured" view mode/display for post.
```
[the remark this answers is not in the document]
volodymyr-sydor-lemberg: done

modules/social_features/social_activity/modules/social_activity_filter/social_activity_filter.info.yml
volodymyr-sydor-lemberg: done

ribel (approved):
I have tested functionality and all works well now.

modules/social_features/social_activity/modules/social_activity_filter/config/update/social_activity_filter_update_8002.yml
```diff
@@ -0,0 +1,402 @@
+views.view.activity_stream:
+  expected_config:
```
ronaldtebrake, line 2: Why is this necessary?

ronaldtebrake: I don't think we want to remove metadata for this right?

modules/social_features/social_activity/modules/social_activity_filter/social_activity_filter.install
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ * Implements hook_install().
+ *
+ * Perform actions related to the installation of social_activity_filter module.
+ */
+function social_activity_filter_install() {
+  user_role_grant_permissions('sitemanager', ['administer social_activity_filter']);
+  social_activity_filter_update_8001();
```
ronaldtebrake, line 25: Please don't invoke update hooks in a install.

Update hooks are there to update existing installations, if you want to do something on install please use the `config/install` files, or create a separate method which you can invoke both from a hook_update_n and the hook_install.

ronaldtebrake (changes requested):
Hi guys, thanks for the work!
Some feedback from my side

1. I get a warning after installing for the activity stream, not sure if you've checked
```
The following extensions will be enabled: social_activity_filter
Do you really want to continue? (y/n): y
[warning] Unable to update configuration for views.view.activity_stream.
[info] Configuration views.view.community_activities has been successfully updated.
```

Most likely this is due to the issue with social_activity_filter_update_8002 in code where it tries to update cache metadata while to me that seems odd.

2. Based on 1 I can't get it to work on landing pages so i checked it out on eci_demo instead where I encounter an issue with adding multiple streams.

https://www.loom.com/share/47ff37877b9a4650b2f188b4fd28a29a

You see when trying to select something in the second section / block, it updates the first one instead.

Other than that it seems to work well on eci demo :) 👍

modules/social_features/social_activity/modules/social_activity_filter/config/update/social_activity_filter_update_8002.yml
volodymyr-sydor-lemberg: This part was automatically generated by update_helper but I will try to clean up and check it

ronaldtebrake: I understand, unfortunately we can't always trust this :( it takes in to account your current active config. So also optional modules that were enabled/disabled for example.

modules/social_features/social_activity/modules/social_activity_filter/social_activity_filter.install
volodymyr-sydor-lemberg: ok, I will create a separate method for that

volodymyr-sydor-lemberg:
> 1. Based on 1 I can't get it to work on landing pages so i checked it out on eci_demo instead where I encounter an issue with adding multiple streams.
> 
> https://www.loom.com/share/47ff37877b9a4650b2f188b4fd28a29a
> 
> You see when trying to select something in the second section / block, it updates the first one instead.

HI @ronaldtebrake It's a known issue that could be tricky to fix. Can we merge this part and create a follow-up ticket for that?

ronaldtebrake:
> > 1. Based on 1 I can't get it to work on landing pages so i checked it out on eci_demo instead where I encounter an issue with adding multiple streams.
> > 
> > https://www.loom.com/share/47ff37877b9a4650b2f188b4fd28a29a
> > You see when trying to select something in the second section / block, it updates the first one instead.
> 
> HI @ronaldtebrake It's a known issue that could be tricky to fix. Can we merge this part and create a follow-up ticket for that?

Sure, we have a work around if we save the first section, edit the landing page, and add a second stream it does work.

ronaldtebrake:
Hi @volodymyr-sydor-lemberg thanks for the updates, that seems to work way better!
Can you add a known issue for the work around? I'll merge it with that in mind.

PR #1791, opened 2020-04-06
Issue #3125249 by ribel: Fix complementary region moving page content to the side

ribel:
## Problem
The socialblue_local_actions is rendered lazily, this can cause the simple `if` check to be false. This is not a problem in the normal layout because the content is limited to 67% width and this region just fills the remaining 1/3rd of the page.

However, in Sky, this is problematic because the region is moved to the left which causes the other content to be shifted to the right even if there appears to be no actual content.

## Solution
Hide socialblue_local_actions on those pages where we don't have any blocks in the sidebar.

## Issue tracker
- https://www.drupal.org/project/social/issues/3125249

## How to test
- [ ] Run hook update
- [ ] Go to /user/*/edit page
- [ ] You should not see layout issue like on the screenshot below

## Screenshots
image

## Release notes
Fix complementary region moving page content to the side

### DIFF
--- a/modules/social_features/social_core/config/update/social_core_update_8809.yml
+++ b/modules/social_features/social_core/config/update/social_core_update_8809.yml
@@ -1,0 +1,10 @@
+block.block.socialblue_local_actions:
+  expected_config:
+    visibility:
+      request_path:
+        pages: "/admin/reports/*\r\n/group/*/membership\r\n/node/*/all-enrollments"
+  update_actions:
+    change:
+      visibility:
+        request_path:
+          pages: "/admin/reports/*\r\n/group/*/membership\r\n/group/*/join\r\n/group/*/leave\r\n/group/*/delete\r\n/node/*/all-enrollments\r\n/node/*/delete\r\n/user/*/edit"
--- a/modules/social_features/social_core/config/update/social_core_update_8809.yml
+++ b/modules/social_features/social_core/config/update/social_core_update_8809.yml
@@ -7,4 +7,4 @@ block.block.socialblue_local_actions:
     change:
       visibility:
         request_path:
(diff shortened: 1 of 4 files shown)


ribel:
@Kingdutch, @ronaldtebrake I reverted commit, which also trims Profile Statistic block In Sky theme:
image

I have fixed issue by updating configs of `socialblue_local_actions` block. 
I added all pages where I found this issue. And I didn't remove this block from all pages, because it is Drupal Core block, which can be used by other developers in other places, which we can't predict.

modules/social_features/social_core/config/update/social_core_update_8809.yml
```diff
@@ -0,0 +1,10 @@
+block.block.socialblue_local_actions:
+  expected_config:
+    visibility:
+      request_path:
+        pages: "/admin/reports/*\r\n/group/*/membership\r\n/node/*/all-enrollments"
```
Kingdutch, line 5: I'm wondering whether we should just remove this line altogether in case someone has already manually adjusted the block for some pages. But I'm not sure.

ribel: That would be easier for us, but for BC with other developers, I think it is safer just add more pages here.

Kingdutch: Well, I mostly mean that in case someone has already done something like set the pages to 

```
"/admin/reports/*\r\n/group/*/membership\r\n/node/*/all-enrollments\r\n/node/*/delete"
```
or even
```
"/admin/reports/*\n/group/*/membership\n/node/*/all-enrollments"
```

Then the update hook as-is will fail.

ribel:
Merged and cherry-picked to 8.x-8.x

PR #1793, opened 2020-04-06
Issue #3125293 by SV: Create "Medium Teaser" view mode for profile.

volodymyr-sydor-lemberg:
## Problem
Currently, we already have a profile--profile--medium_teaser.html.twig template in the socialbase theme but a view mode doesn't exist and we are not able to use that template/display in the views or etc.

## Solution
Create a "Medium Teaser" view mode for the profile.

## Issue tracker
- https://getopensocial.atlassian.net/browse/YANG-2482
- https://www.drupal.org/project/social/issues/3125293

## Screenshots
Screenshot from 2020-04-10 12-45-16 (1)

PR #1795, opened 2020-04-07
Issue #3125680 by navneet0693: Removed title property of field_profile_image.

navneet0693:
## Problem
Currently, when a user doesn't have a profile picture and if you hover over the user's picture, you will see "Default profile image" text. 
This is because browsers are picking up the 'title' property of a user's profile which set <code></code> tag in https://github.com/goalgorilla/open_social/blob/8.x-9.x/themes/socialbase/templates/profile/profile--compact.html.twig#L26

## Solution
Remove title's value in https://github.com/goalgorilla/open_social/blob/8.x-9.x/modules/social_features/social_profile/config/install/field.field.profile.profile.field_profile_image.yml#L34

## Issue tracker
https://www.drupal.org/project/social/issues/3125680

## How to test
- [ ] Run drush updb
- [ ] Create a user without any profile picture
- [ ] Login with new user and Enroll in an event
- [ ] Hover over the profile picture of the user where enrollees are shown in an event.
- [ ] You should be able to see username of the user.

## Screenshots
**Before**: 
image

**After**:
image

## Release notes
Profile Image field will not have a default title by the installation.

## Change Record
The default image title property in the profile image field (field.field.profile.profile.field_profile_image) is removed.

## Translations
N.A

### DIFF
--- a/modules/social_features/social_profile/config/install/field.field.profile.profile.field_profile_image.yml
+++ b/modules/social_features/social_profile/config/install/field.field.profile.profile.field_profile_image.yml
@@ -6,8 +6,6 @@ dependencies:
     - profile.type.profile
   module:
     - image
-_core:
-  default_config_hash: tSyQuKXSwEWCqPytdKi5gnT04E2SCEUfF6-3AUegANk
 id: profile.profile.field_profile_image
 field_name: field_profile_image
 entity_type: profile
@@ -31,7 +29,7 @@ settings:
   default_image:
     uuid: 4eb1d927-28f4-402a-8c87-017e637f434a
     alt: 'Default profile image'
-    title: 'Default profile image'
+    title: ''
     width: 200
     height: 200
   handler: 'default:file'
--- a/modules/social_features/social_profile/config/update/social_profile_update_8008.yml
+++ b/modules/social_features/social_profile/config/update/social_profile_update_8008.yml
@@ -1,0 +1,7 @@
+field.field.profile.profile.field_profile_image:
+  expected_config: {  }
+  update_actions:
+    change:
+      settings:
+        default_image:
+          title: ''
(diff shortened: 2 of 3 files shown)


navneet0693:
> This looks like it's the first update hook on the Social Profile module that uses the Update Helper. This means that the hook must get a dependency on a specific social_core hook (I believe 8805 enables the Update Helper).
> 
> Other than that it looks good. This would not change the title if someone had it already changed from "Default profile image". That is fine :)
> 
> @navneet0693 Could you test if this update applies properly for a Dutch site where "Default profile image" is translated?

@Kingdutch I checked with one such site, however, yes I found that configuration changes are not reflecting event after the hook_update ran. Following is how field config looks like even after running update: 

`  default_image:
    uuid: 4eb1d927-28f4-402a-8c87-017e637f434a
    alt: 'Image de profil par défaut'
    title: 'Image de profil par défaut'
    width: 200
    height: 200
  handler: 'default:file'
  handler_settings: {  }
field_type: image`

navneet0693:
As a workaround suggested by mtodor from Thunder in Drupal Slack, we can remove "expected configuration" part from the udpate yml as we are completely removing the title, so no matter what is the translations on production our update will be still working.

https://github.com/goalgorilla/open_social/pull/1795/files#diff-f3663e62cefd108be90834f465235d8a

navneet0693:
> @navneet0693 I think we should move the module update out of this PR for a few reasons.
> 
> 1. It's not needed for the code changes in this PR to work.
> 2. It makes administration (and review/testing) easier because I believe module updates are listed separately.
> 3. The .make file will also need to be updated with the composer.json
> 
> The code changes for the update hook itself look good (although I have not run it myself).
> 
> The update to Update Helper contains two bugs that Open Social has contributed to so we can probably merge that pretty easily in a small issue + separate PR.

I will do it right away! Thanks @Kingdutch :D

ribel:
This PR was not cherry-picked to 8.x-8.x, which cause conflict after merge of PR #1793 
I have resolved this conflict and cherry-picked remaining code in 42b2735a3039564b1e6ec41bcd082efff1d8d411

PR #1799, opened 2020-04-14
Issue #3127555 by Kingdutch, ronaldtebrake: Like buttons have gone mi…

Kingdutch:
…ssing from posts and comments

<h2>Problem</h2>

After the release of 8.0 the way configuration is managed has been changed. The like and dislike module takes care of adding the like field to entities so these fields shouldn't be in the exported configuration.

However, the like_and_dislike module adds these field as hidden by default: https://git.drupalcode.org/project/like_and_dislike/-/blob/8.x-1.x/like_and_dislike.module#L164

Priority major because this is an oft and very visible feature.

<h2>Solution</h2>

It's probably difficult to make the like_and_dislike module change their way so we can implement an alter hook to make the default visibility for this field type visible. We can assume that if the field shouldn't be visible, like and dislike is probably not enabled for an entity/bundle.

## Issue tracker
https://www.drupal.org/project/social/issues/3127555

## How to test
- [ ] Open the site on the branch
- [ ] Refresh the cache (cache-rebuild)
- [ ] See the like buttons

## Release notes
Due to the configuration change the like button had gone missing. We'd like to report that it is now present again.


modules/social_features/social_like/social_like.install
```diff
@@ -18,6 +18,28 @@ function social_like_install() {
   _social_like_set_defaults();
 }
 
+/**
+ * Implements hook_entity_extra_field_info_alter().
```
bramtenhove, line 22: This should be in the `.module` file instead of the `.install` file.

Kingdutch: Damn busted, good catch 😂

ronaldtebrake:
Unfortuately this only works when we update the field_group module or patch https://www.drupal.org/project/field_group/issues/2951175 

Updating could be worth it, will solve multiple issues I believe

modules/social_features/social_like/social_like.module
```diff
@@ -65,6 +65,28 @@ function social_like_invalidate_cache(EntityInterface $entity) {
...
+ */
+function social_like_entity_extra_field_info_alter(&$info) {
+  // By default the like_and_dislike module adds the like count as hidden field.
+  // However, in Open Social it should be visible in all places. This hook
+  // changes the default. See like_and_dislike_entity_extra_field_info.
+  $entity_type_manager = \Drupal::entityTypeManager();
+  foreach (\Drupal::config('like_and_dislike.settings')->get('enabled_types') as $entity_type_id => $bundles) {
+    // The entity type has no bundles. Add display component to the default one.
+    if (!$entity_type_manager->getDefinition($entity_type_id)->hasKey('bundle')) {
+      $info[$entity_type_id][$entity_type_id]['display']['like_and_dislike']['visible'] = TRUE;
```
ronaldtebrake, line 79: Seems like weight needs to be added as it's requested (see failing test)

Kingdutch:
New commit should fix the issue, waiting for Travis to confirm.

ronaldtebrake:
also cherry picked to 8.x

PR #1805, opened 2020-04-17
Issue #3099608 by robertragas: Add patch to fix time() callback filling up logs for the votingapi module

robertragas:
## Problem
Logs filling up with the following warning so many times that it can create logs up to 400mb per day.

Warning: time() expects exactly 0 parameters, 2 given in /var/www/drupal/web/core/lib/Drupal/Core/Field/BaseFieldDefinition.php on line 469

## Solution
Apply the patch from the votingagpi module
https://www.drupal.org/project/votingapi/issues/3048085

## Issue tracker
https://www.drupal.org/project/social/issues/3099608

## How to test
- [ ] Just browse through a platform and notice the logs filling up with these warning. 

## Screenshots

## Release notes
We applied a patch for the votingapi module that will prevent the logs being filled up with warnings.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -98,7 +98,8 @@
                 "Make sure select all gets fired for socialbase theme": "https://www.drupal.org/files/issues/2019-09-19/3042494-trigger-vbo-action-on-jquery-checkbox-change-5.patch"
             },
             "drupal/votingapi" : {
-                "votingapi_views_data_alter uses incorrect empty check for missing entity error handling": "https://www.drupal.org/files/issues/2020-01-31/votingapi-3110353-2.patch"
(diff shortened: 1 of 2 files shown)


ribel:
@robertragas, @ronaldtebrake, I also noticed this issue on a couple of sites with OS 7.5, and we use the same patch in those projecs. So I think it would be good to include this in a distro.

robertragas:
Cool, if since you approved it, I will merge it, as this is already merged into the votingapi module, so no need for Ronald to also look at it.

robertragas:
Merged and cherry picked to 8 and 7. For 7 I needed to manually apply it, since cherry pick gave conflicts.

PR #1808, opened 2020-04-21
Issue #3128618 by frankgraave: refactor if statement to separate func…

frankgraave:
## Problem
The module `social_profile_privacy` comes with a permission:

```
social profile privacy view hidden fields:
  title: 'View all hidden profile fields'
  description: 'Access to view all hidden profile fields'
```

This permission should work as an 'bypass' to view fields on a profile that are hidden. However, this permission does not seem to work correctly with the email on a profile since that is not a field, but is merely an added value to the $variables array in the social_profile preprocess based on some checks.

## Solution
Refactor the _if statement_ in `social_profile.module` that checks if the email "field" should be rendered or not by putting the logic inside a separate function for better readability. The module social_profile_privacy comes with a permission to bypass access to hidden fields. Added this permissions to the check if the email should be rendered or not.

## Issue tracker
https://www.drupal.org/project/social/issues/3128618

## How to test
- [ ] Make sure you have a user with the permission 'social profile privacy view hidden fields' and log in with that user.
- [ ] Log in as another user and set your profile email to hidden in the profile settings.
- [ ] As the user with the **correct permission**, view the other user his profile and see that you can view his email address even though he/she preferred it hidden.
- [ ] Check the user profile with another user that **does not have the permission** and see that you are **not able** to see the email address.

## Release notes
Viewing emails on profiles that were preferred hidden by a user are now visible to the roles with the correct permissions.


modules/social_features/social_profile/social_profile.module
```diff
@@ -398,6 +391,37 @@ function social_profile_preprocess_profile(array &$variables) {
...
+ * @param \Drupal\Core\Session\AccountProxyInterface $current_user
+ *   The current user.
+ *
+ * @return bool
+ *   Return TRUE if current user is allowed to see the email, FALSE otherwise.
+ */
+function social_profile_check_email_field_access(Profile $profile, AccountProxyInterface $current_user) {
+  // Prepare some variables for checking on the email field.
+  $global_show_email = \Drupal::config('social_profile.settings')->get('social_profile_show_email');
+  if (isset($profile->field_profile_show_email)) {
```
navneet0693, line 408: $profile_show_email = $profile->field_profile_show_email->value ?? NULL;

modules/social_features/social_profile/social_profile.module
```diff
@@ -398,6 +391,37 @@ function social_profile_preprocess_profile(array &$variables) {
...
+    $profile_show_email = $profile->field_profile_show_email->value;
+  }
+
+  // If the current user has this permission, return true straight away.
+  if ($current_user->hasPermission('social profile privacy view hidden fields')) {
```
navneet0693, line 413: How about moving this statement above all?

modules/social_features/social_profile/social_profile.module
```diff
@@ -398,6 +391,37 @@ function social_profile_preprocess_profile(array &$variables) {
...
+    return TRUE;
+  }
+  // If the current user doesn't have the above permission, perform additional
+  // checks to see if the current user is still able to view this email field.
+  if ($global_show_email || (!$global_show_email && !empty($profile_show_email)) || $current_user->hasPermission('view profile email')) {
```
navneet0693, line 418: I think the hasPermission conditions can be combined.

modules/social_features/social_profile/social_profile.module
frankgraave: Good one!

modules/social_features/social_profile/social_profile.module
frankgraave: Done!

modules/social_features/social_profile/social_profile.module
frankgraave: Correct, done!

navneet0693:
Cherry-picked on 8.8.x - 47643dad2e5a2d73e3a9f739619096d51650e82d

PR #1812, opened 2020-04-24
Issue #3111945 by frankgraave: fix dependency of role delegation modu…

frankgraave:
…le in social_user.info.yml and add update hook to enable the role delegation module

## Problem
There was a wrong dependency declaration in social_user.info.yml.

## Solution
Fixed the dependency and also added an update hook to enable the module for existing platforms.

## Issue tracker
https://www.drupal.org/project/social/issues/3111945

## How to test
- [ ] See that it's not breaking anymore
- [ ] Run the drush updb and see that the module is enabled on an existing site

### DIFF
--- a/modules/social_features/social_core/social_core.install
+++ b/modules/social_features/social_core/social_core.install
@@ -1208,3 +1208,10 @@ function social_core_update_8809() {
   // Output logged messages to related channel of update execution.
   return $updateHelper->logger()->output();
 }
+
+/**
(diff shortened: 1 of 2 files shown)


navneet0693:
Cherry-picked on 8.8.x : 798a253050f845dd345114c56ec1e62d2a00c868

Cherry-picked on 8.7.x : 61c67f1ff0a8025116d089aacf24af7f5497bdd6

PR #1815, opened 2020-04-26
Issue #3089990 - When a group has no members, the Manage Members page errors

ronaldtebrake:
## Problem
When a group has no members, the Manage Members page errors with the following error message:

```
Error: Call to a member function getArguments() on null in Drupal\social_group\Plugin\views\field\SocialGroupViewsBulkOperationsBulkForm->viewsForm() (line 120 of profiles/contrib/social/modules/social_features/social_group/src/Plugin/views/field/SocialGroupViewsBulkOperationsBulkForm.php).
```

## Solution

Skip processing the SocialGroupViewsBulkOperationsBulkForm when there are no results in the group_manage_members view.

## Issue tracker
- https://www.drupal.org/project/social/issues/3089990
- Rebased from: #1577

## How to test
- [ ] Create an OpenSocial group
- [ ] Go to the "Manage Members" tab
- [ ] Remove all members
- [ ] Error page should display as described above


ronaldtebrake:
Also in 7.x and 8.x

PR #1817, opened 2020-04-28
Issue #3131506 by robertragas: add the post profile view to the check…

robertragas:
… in the postform for removing the duplicate visibility

## Problem
The post visibility when going to another user stream shows community visibility twice.

## Solution
Add the post profile condition to the postForm where the duplicate visibility gets unset.

## Issue tracker
https://www.drupal.org/project/social/issues/3131506

## How to test
- [x] Go to the stream another user and see the community visibility is showing twice
- [x] Check out to this branch and see it only shows once now

## Release notes
We fixed a problem where the post visibility would show "community" twice when posting to the stream of another user.


frankgraave:
This has been backported to 8.x

PR #1823, opened 2020-05-01
Issue 3132823 by AlanHDev: Update social features field group config

AlanHatch:
## Problem
Field group 3.0 requires that region is specified in the config for the group. There is a post-update hook provided by the Field Group contrib module that adds the required config to existing form display config on the site. 

Having updated to Open Social that includes Field Group 3.0, enabling a module such as social_landing_page or another module that provides install config for entity_form_display containing field group settings causes issues. The install config does not contain the required 'region: content' and, as the field group update has already run, the config is incomplete.

## Solution
The resolution is to add region: content to any config provided by open social modules that contain entity_form_display config with field group settings.

If an affected module has already been installed, this won't fix it. In that case, you'd need to ensure that the config is updated manually and then re-import.

## Issue tracker
https://www.drupal.org/project/social/issues/3132823

## How to test
- [ ] Enable social_landing_page on a site where it has not been previously enabled
- [ ] Attempt to create a new landing page
- [ ] Observe a notice: Undefined property: stdClass::$region in field_group_form_process()
- [ ] Enable Field UI
- [ ] Visit Manage form display for the landing page content type
- [ ] Observe the field groups and field order is broken in the UI and a notice is present
- [ ] Roll back the database and repeat tests having applied the changes
- [ ] No notices present
- [ ] Field drag drop UI is as expected

### DIFF
--- a/modules/social_features/social_book/config/features_removal/core.entity_form_display.node.book.default.yml
+++ b/modules/social_features/social_book/config/features_removal/core.entity_form_display.node.book.default.yml
@@ -30,6 +30,7 @@ third_party_settings:
         required_fields: true
         id: visibility
         classes: 'card'
+      region: content
     group_description:
       children:
         - body
@@ -42,6 +43,7 @@ third_party_settings:
         classes: ''
         id: ''
         required_fields: true
+      region: content
     group_title:
       children:
         - title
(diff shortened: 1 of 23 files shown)


navneet0693:
This is quite old. Please open a new PR if required.

PR #1824, opened 2020-05-04
Feature/tb 3755 update linkedin api

robertragas:
## Problem
Linkedin API v1 is deprecated so we are not able to use the SSO anymore for this provider.

## Solution
Since the Happyr does not got updated to support the V2, move to another third party library called https://github.com/zoonman/linkedin-api-php-client

## Issue tracker
https://www.drupal.org/project/social/issues/3133350

## How to test
- [ ] Enable the social_sso and social_auth_linked module and clear the cache
- [ ] Go to Linkedin Developers on https://www.linkedin.com/developers and create an app.
- [ ] In the auth tab, the default permissions are correct, but you need to add the following **callback urls**
http://social.local/social-api/link/linkedin/callback
http://social.local/user/register/linkedin/callback
http://social.local/user/login/linkedin/callback
- [ ] Copy the clientid and secret and fill it on your social installation settings for Linkedin on http://social.local/admin/config/social-api/social-auth/linkedin and check to activate the module
- [ ] Logout and try to sign up using the linkedin link, it should prefill the email adress and username in step 1.
- [ ] In step 2 you should also see a prefilled first name, last name and profile picture.
- [ ] Also try logging in with linkedin
- [ ] Also try unlinking and linking the provider in the account settings page.

Make sure you run a Proxy locally so you have a dedicated url you can use in your callback urls.

## Release notes
Authorization with the social network LinkedIn for Open Social was broken due to deprecation of the V1 API.
We have changed the library to support the V2 API, which means you can use the LinkedIn provider again to log in to Open Social.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -165,7 +165,7 @@
         "php-http/curl-client": "^1.7",
         "guzzlehttp/psr7": "^1.3",
         "php-http/message": "^1.7",
-        "happyr/linkedin-api-client": "^1.0",
+        "zoonman/linkedin-api-php-client": "dev-master#d06531c48d5efc8aaf7dc074a5320f5ba3f906c9",
         "abraham/twitteroauth": "^0.7.4",
         "swiftmailer/swiftmailer" : "v5.4.12",
         "bower-asset/waves": "0.7.6",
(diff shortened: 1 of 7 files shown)


sjoerdvandervis:
Cherry-picked to 8.x-8.x and 8.x-7.x

PR #1826, opened 2020-05-05
Issue #3133763 by rolki: Add an extra condition before invalidating cache tags

rollki:
## Problem
After creating a new group type and adding visibility field to this group then I use _social_flexible_group_edit_submit callback to change visibility options for all content related to this group, but I faced with the error:
`TypeError: Argument 1 passed to _social_group_cache_tags() must implement interface Drupal\group\Entity\GroupInterface, null given, called in /var/www/html/profiles/contrib/social/modules/social_features/social_group/social_group.module on line 769 in _social_group_cache_tags() (line 1197 of /var/www/html/profiles/contrib/social/modules/social_features/social_group/social_group.module`

## Solution
Add an extra condition before invalidating cache tags.

## Issue tracker
https://www.drupal.org/project/social/issues/3133763

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -780,10 +780,12 @@ function _social_group_type_edit_submit($form, FormStateInterface $form_state) {
     }
   }
 
-  // Make sure we clear cache tags accordingly.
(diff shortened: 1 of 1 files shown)


ribel:
Merged and cherry-picked to the 8.x-8.x

PR #1827, opened 2020-05-05
Issue #3133780 by mdeny: Fix "Group members" post visibility for non group members

denysyukm:
## Problem
Users have access to posts (with "Group members" visibility) by using direct URL: "/post/{post_id}" when they shouldn't have

## Solution
Adjust access controller for the Post entity

## Issue tracker
https://www.drupal.org/project/social/issues/3133780
https://getopensocial.atlassian.net/browse/YANG-2889

## How to test
- [ ] create a flexible group
- [ ] leave a post with "Group members" visibility option
- [ ] copy direct post URL and give it to user2 who is not a flexible group member
- [ ] as a user2 open page with post URL
- [ ] user2 should not have access to the post 

## Release notes
Fix issue with accessing to the posts with "Group members" visibility option


ribel:
@ronaldtebrake, @bramtenhove, I want to bring your attention to this prio. I think it was a serious issue for a long time.

modules/social_features/social_post/src/PostAccessControlHandler.php
```diff
@@ -116,8 +116,8 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
                     $group_roles = $group_role_storage->loadByUserAndGroup($account, $group);
                     /** @var \Drupal\group\Entity\GroupRoleInterface $group_role */
                     foreach ($group_roles as $group_role) {
-                      if ($group_role->isOutsider() && $group_role->hasPermission($permission)) {
-                        return AccessResult::allowed()->cachePerUser()->addCacheableDependency($entity);
+                      if ($group_role->isOutsider()) {
```
ronaldtebrake, line 119: Will admins / SM still be able to check it out?

ribel: I've checked and no - CM/SM, which are not group members don't have access to Group only posts after that.
So this probably needs to be fixed in a different way, @denysyukm.

denysyukm:
@ronaldtebrake , @ribel , check my last commit. I added possibility for outsiders with manager roles see posts with 'Group members' visibility.

ribel:
@denysyukm Please fix coding standards.

ribel:
@denysyukm Please fix coding standards and try to use permissions instead of roles in code.

modules/social_features/social_post/src/PostAccessControlHandler.php
denysyukm: So I did some investigation and found that if we revoke "access posts in group" permission for outsiders in flexible_group then everything will work properly. And no needed to do changes in this PostAccessControllHandler.

denysyukm:
So I did some investigation and found that if we revoke "access posts in group" permission for outsiders in flexible_group then everything will work properly. And it won't need to do any changes in the PostAccessControllHandler.

ribel:
🍒 picked to 8.x and 9.x

PR #1828, opened 2020-05-06
Issue #3134100 by navneet0693: Add Group block on user profile's group doesn't takes "create group" permissions in context

navneet0693:
## Problem
"Add Group" block on /user/{uid}/groups on the current user project doesn't take the context of permissions which enables the user to add groups in the system. If the user has permission to create only one group, the + icon in the header takes directly to that group but Add block doesn't, it will take you to /group/add.

## Solution
Use the same logic to create the link as used here: https://github.com/goalgorilla/open_social/blob/8.x-9.x/modules/social_features/social_group/social_group.module#L1745

## Issue tracker
https://www.drupal.org/project/social/issues/3134100

## How to test
- [ ] Login as an authenticated user with permission to create 1 group.
- [ ] Go to "My groups" on your profile page
- [ ] Click on "Add block"
- [ ] You should be directed to /group/add/{group_type}
- [ ] Add more permissions to create another group
- [ ] Check "Add block" button URL
- [ ] You should be redirected to /group/add

## Screenshots
N.A

## Release notes
The Add Group block on the user profile page will now have a computed URL as per permission to create a group of the current user. It will have either of the two URLs
1. If user have permission to create only 1 group, URL will be /group/add/{group_type}
2. If the user has permissions to create more than 1 group, URL will be /group/add

## Change Record
The SocialGroupHelperService service will have a new function called getGroupsToAddUrl()  to return the group creation URL for a given account.

## Translations
N.A

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -729,9 +729,16 @@ function social_group_form_alter(&$form, FormStateInterface $form_state, $form_i
       array_unshift($form['actions']['submit']['#submit'], '_social_group_delete_group');
     }
   }
-
   if (in_array($form_id, ['group_flexible_group_edit_form', 'group_flexible_group_add_form'])) {
     $join_method_default_value = 'added';
+    // Ensure we have a better descriptive label.
+    if (array_key_exists('added', $form['field_group_allowed_join_method']['widget']['#options'])) {
+      $form['field_group_allowed_join_method']['widget']['#options']['added'] = t('Be added only - users can only be added by group managers');
+    }
+    if (array_key_exists('direct', $form['field_group_allowed_join_method']['widget']['#options'])) {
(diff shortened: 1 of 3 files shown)


robertragas:
Working correctly, as discussed I reverted the code that shouldn't be in this branch.
I will merge it, but @navneet0693 could you still fill in the Release notes in this PR?

navneet0693:
> Working correctly, as discussed I reverted the code that shouldn't be in this branch.
> I will merge it once travis is green, but @navneet0693 could you still fill in the Release notes in this PR?

I have added release notes and change the record. Thanks :-)

robertragas:
Sorry, after merging I noticed that cache invalidate in the beginning of social_group.module. Is this needed? Don't see where it gets the variabele from because it's outside of the foreach.

robertragas:
Merged and cherry picked

PR #1830, opened 2020-05-06
Issue TB-4358 by frankgraave: refactor the event enrollment options a…

frankgraave:
## Problem
The checkbox if people were allowed to enroll in an event was becoming obsolete due to all the different enrollment options.

## Solution
Refactored the options to be more simply and straight forwards, which will be easier to understand.

## Issue tracker
-

## How to test
- [ ] Log in as any LU and create a new event
- [ ] See that you no longer see the checkbox "Allow users to enroll in event"
- [ ] See that you now have three radio's with:
- [ ] 1. Open to enrol - users can enrol for this event without approval
- [ ] 2. Request to enrol - users can ‘request to enrol’ for this event which event organiser approve/decline
- [ ] 3. Invite-only - users can only enrol for this event if they are added/invited by event organiser
- [ ] _(Optionally: although this is only UI/UX change and it should not have impact in any way for the current logic of enrollment, you could test the enroll / invite / request which should all work as is designed)_

## Screenshots
-

## Release notes
-

## Change Record
-

## Translations
-


robertragas:
Merging, but will be doing some changes in the main branch of event invite still to everything adds up.

PR #1832, opened 2020-05-11
TB-4368 - Group content visibility options

frankgraave:
# Note:
**Don't merge yet.**

## Problem
The only way to manage the content visibility options for flexible groups was through permissions. Also, when there is only one option available there was a redundant checkbox.

## Solution
Added a settings in the social group settings page where the SM can select which content visibility options are available for when someone creates/edits a flexible group. Now when there is only one option, it shows a clear message to the user.

## Issue tracker
https://getopensocial.atlassian.net/browse/TB-4368

## How to test
- [ ] Log in as SM
- [ ] Go to /admin/config/opensocial/social-group
- [ ] See that you have new options for selecting content visibility options
- [ ] Go to /group/add/flexible_group
- [ ] See you have all three options available
- [ ] Go back to /admin/config/opensocial/social-group and deselect **two** options so only one remains.
- [ ] Go to /group/add/flexible_group again and see the message
- [ ] _(Optional: test the other messages as well)_

## Screenshots
CleanShot 2020-05-11 at 16 41 40@2x

## Release notes
As a site manager, you're now able to select which content visibility options should be available for the users. When you only have one option selected, it will show a message to the user who is creating a new flexible group about the "default" visibility option.


robertragas:
A few remarks
- Think we have to be very clear that these settings are for flexible groups and not open/closed/public groups.
- What happens if someone has a public group and the sitemanager will change the visibility. We should make it clear also that it does not alter all the old content to the new visibility settings.

- Last but not least, but think this is a follow-up as it already seems to be broken in the Distro for flexible groups, it does not respect the admin/config/opensocial/visibility settings where you can disable public visibility in general. 

And I think we should think in general about declaring visibilities, because we have it now in multiple places which is prone to error once someone wants to make a change in the visibility naming or add a new one. But I guess this can be done as a follow-up.

robertragas:
Also I think social_group_flexible_group is optional when coming from 8.x so these settings are then particular confusing.

robertragas:
I changed 1 more thing to make sure it respect the "disable public visibility" settings.
For now we will use it as a patch for 1 specific client. To get it further in the Distro it needs more testing with optional modules such as courses to see it does not conflict in any way.

Also it currently does not respect hook_social_group_default_visibility_alter which would need to get fixed. And while we are at it, this hook also is broken for open groups (already is) but can be fixed in the same PR I guess.

robertragas:
Also 1 additional comment

Last but not least, we haven’t really thought about it when the settings get changed midway. For example When a flexible group has 2 visibilities, **group** and **community**.  And suddenly they change to **group** only. They cannot change it anymore on the group edit page because it only shows 1 visibility. Luckely the visibility of the content is also controlled by this, so they are not able to make extra **community** content, but still would be nice to think about it. Also with the content itself, because that will keep it’s visibility, but don’t think we want to mess with this too much as it will be very heavy if content of every group suddenly gets changed.

Kingdutch:
@frankgraave The top of your PR said "Don't merge this yet" so I've converted this to a draft PR :) Once you're ready, you can convert this to a regular PR and it can be reviewed and merged 🎉

navneet0693:
Closed in favor of: https://github.com/goalgorilla/open_social/pull/2139

PR #1833, opened 2020-05-11
TB-4373 - Event visibility options

frankgraave:
## Note:
**Do not merge yet!**

## Problem
As a site manager, there was no easy way to set allowed/available visibility options for creating new events and to decide if events outside of groups are allowed.

## Solution
Added a settings in the event settings page where the SM can select which content visibility options are available for when someone creates/edits an event. When there is only one option, it shows a clear message to the user. Selecting a group will always still respect the group visibility options. So in case of a flexible group, its visibility options will be presented (unless they are also limited).

## Issue tracker
https://getopensocial.atlassian.net/browse/TB-4373

## How to test
- [ ] Log in as SM
- [ ] Go to /admin/config/opensocial/event
- [ ] See that you have new options for selecting content visibility options
- [ ] Go to /node/add/event
- [ ] See you have all three options available
- [ ] Go back to /admin/config/opensocial/event and deselect **two** options so only one remains.
- [ ] Go to /node/add/event again and see the message
- [ ] _(Optional: test the other messages as well)_

## Screenshots
CleanShot 2020-05-15 at 14 07 50@2x

## Release notes
As a site manager, you're now able to select which content visibility options should be available for the users. When you only have one option selected, it will show a message to the user who is creating a new event about the "default" visibility option.


navneet0693:
Closed in favor of : https://github.com/goalgorilla/open_social/pull/2139

PR #1835, opened 2020-05-13
Issue #3136189 by rolki: Changing visibility of content items for group doesn't work correctly with social_role_visibility because of hard-coded visibility options

rollki:
## Problem
After installing the social_role_visibility module, adding a new role as visibility option, then go to flexible group and for example choose a role as visibility option and uncheck all other options, and this case visibility options for group content should be changed automatically to role visibility, but it's doesn't work because of hard-coded options in the distro.

## Solution
Add role visibility support to the distro.

## Issue tracker
https://www.drupal.org/project/social/issues/3136189

## How to test
Create a new role and choose this role as visibility option, then go to some flexible group and check the role as visibility option and uncheck all other options, and this case visibility options for group content should be changed automatically to role visibility.


ribel:
🍒cherry-picked to 8.x-8.x

PR #1836, opened 2020-05-14
Issue #3136472 by Kingdutch, ronaldtebrake: Group comment join pop-over breaks with event organisers

Kingdutch:
<h2>Problem</h2>
The group pop-over very much looks like a teaser but applying the actual 
styling from a teaser will break the pop-over. This is normally not a 
problem because the module doesn't load the teaser styling. However, 
when combined with Event Organisers this causes issues as that feature 
does load the teaser styling.

<h2>Solution</h2>
The solution is to change the component itself instead of trying to 
remove the teaser stylesheet as other future features may also need that 
styling themselves.

The teaser also no longer displays behind the secondary navigation.

## Issue tracker
https://www.drupal.org/project/social/issues/3136472

## How to test
- [ ] Using the Sky flavour
- [ ] Create a group that anyone can join with public content
- [ ] Create an event in the created group
- [ ] Add an event organiser to the event
- [ ] View the event as a user who is not a member of the group does not have permission to ignore comment restrictions (i.e. not admin)
- [ ] Compare the difference before and after this fix

## Screenshots
**Before**
Screenshot 2020-05-14 at 12 46 19

**After**
Screenshot 2020-05-14 at 12 41 20

## Release notes
The group preview for content where a user needs to become a group member in order to comment now displays properly when the group organiser feature is used at the same time. The teaser also no longer displays behind the secondary navigation.


robertragas:
Approved and will merge it.

robertragas:
Verified cherry pick to 8.x

PR #1837, opened 2020-05-14
Issue #3136493 by chmez: Absent post visibility label in the custom content list block

ghost:
## Problem
When the **Custom content list block** custom block show posts with the same visibility then only the first post has visibility label

## Solution
Remember visibility settings only per visibility type

## Issue tracker
- https://www.drupal.org/project/social/issues/3136493
- https://getopensocial.atlassian.net/browse/ECI-1084

## How to test
- [ ] Create a new plugin instance based on the **Content Block** plugin type for posts and add filter by group
- [ ] Enable the **Social Content Block** module and log in as **SM+**
- [ ] Create the **Group 1** closed group and add a few posts to this group
- [ ] When you create a new custom block of the **Custom content list block** type based on the created post plugin instance with the **Group 1** group in the group filter then you should see visibility label for each post

## Release notes
Show visibility label for each post in the custom content list block

### DIFF
--- a/modules/social_features/social_post/post.page.inc
+++ b/modules/social_features/social_post/post.page.inc
@@ -58,8 +58,9 @@ function template_preprocess_post(array &$variables) {
   }
 
   // Visibility icon and label for template use.
-  $variables['visibility_icon'] = social_post_get_visibility_details($post->get('field_visibility')->value, 'icon');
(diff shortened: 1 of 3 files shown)


ribel (approved):
All looks good to me.

ribel:
🍒cherry-picked to 8.x-8.x

PR #1841, opened 2020-05-18
Issue #3137445 by mdeny: As a SM I want to specify which parts of the address are shown in the profile forms

denysyukm:
## Problem
As a SM I want to specify which parts of the address are shown in the profile forms

## Solution
Implement possibility to enable/disable address's profile fields: country, city, address

## Issue tracker
https://www.drupal.org/project/social/issues/3137445
https://getopensocial.atlassian.net/browse/YANG-2479

## How to test
- [x] Go to Social profile fields settings form (/admin/config/opensocial/profile-fields)
- [x] Disable city and address settings options
- [x] Go to profile edit form
- [x] The city and address fields shouldn't be displayed
- [x] Go back to the Social profile fields settings form
- [x] Click Flush profile data button
- [x] Run removing data
- [x] The city and address fields values should be removed from all user profiles

## Release notes
Implement possibility to enable/disable address's profile fields: country, city, address


ronaldtebrake:
@ribel //cc @bramtenhove lets test this for the userexport as well

ribel:
@ronaldtebrake  I've checked that it still takes into account export settings from `admin/config/opensocial/export`, but if for example City and Address fields was unchecked from `admin/config/opensocial/profile-fields`, they still keeps selected in `admin/config/opensocial/export`, and so appears in export - so I think this needs to be fixed?
I've checked that with other fields when you uncheck it in Fields settings, it then also doesn't appear in Export.
// cc @denysyukm @bramtenhove

ribel:
@ronaldtebrake Issue with Export is fixed now. Do you want to test and review it?
Or can I merge it?

ronaldtebrake:
Hi @ribel @denysyukm thanks for fixing the export.

I've got two questions
1. 
Screenshot 2020-07-29 at 19 29 00

I dont really understand the Country checkbox. I do understand the country field is special, and we can't disable it without the rest because thats how the field widget works. But I think it's confusing if we render a checkbox that always stays disabled. 
We can better explain this as part of the Address field and remove the checkbox for country.

2. 
Screenshot 2020-07-29 at 19 35 26
I'm missing the postal code as a setting in the admin ui. 

Can we still fix this?

ronaldtebrake:
@ribel @denysyukm 
For this one I will merge it and we can fix it during bugfixing days :) will add these as known issues.

PR #1843, opened 2020-05-18
Issue #3137603 by chmez, agami4: Custom content list block section on landing pages

OH17L:
## Problem
As SM I want homepage Blocks automatically pulling content from other nesting pages

## Solution
- Should be possible to add a section where CM can choose content list
  - blocks should be kept the way they are
  - create new section just for content list
- Items should be displayed according to current view mode
- Should be possible to select only existing content lists

## Issue tracker
- https://www.drupal.org/project/social/issues/3137603
- https://getopensocial.atlassian.net/browse/YANG-2705

## How to test
- [x] Enable the **Social Content Block on Landing pages** module and log in as **SM**
- [x] Navigate to the **/block/add/custom_content_list** page and fill the **Topics list** in the **Block description** field
- [x] Fill **Topics** in the **Subtitle** field and press **Save**
- [x] Navigate to the **/node/add/landing_page** page and fill the **Test landing page** in the **Title** field
- [x] When you press **Add Section** then you should see the **Add Custom content list block** button
- [x] When you press the **Add Custom content list block** then you should see the **Block** field
- [x] When you start typing **Topics** in the **Block** field then you should see a drop-down menu with one item - **Topics list**
- [x] Select the **Topics list** item and press **Save**
- [x] You should see the **Custom content list block** custom block with the **Topics** title

## Release notes
Create the **Custom content list block** section for landing page for showing a custom block of the **Custom content list block** type. It's an optional module you can enable.


bramtenhove:
Great job so far @OH17L, @agami4 and @ribel! What will be needed in order for this to be merged?

ribel:
@bramtenhove @ronaldtebrake It is already completed from ECI side and we already using it as a patch on a project.
But since this is kind of new feature, it requires input from distro team

PR #1847, opened 2020-05-20
Updated labels of enrollment methods on event & group add form.

navneet0693:
## Problem
Change labels of Enrollment options for events and join methods for group.

## Solution
Change labels of Enrollment options of events to the following:

```
Open to enroll - users can enroll for this event without approval

Request to enroll - users can ‘request to enroll’ for this event which event organisers approve/decline

Invite-only - users can only enroll for this event if they are added/invited by event organisers

```

Change join method labels of group to the following:

```
Open to join - users can join this group without approval

Request to join - users can ‘request to join’ this group which group managers approve/decline

Invite-only - users can only join this group if they are added/invited by group managers

```

##Translations
• Source strings are added to the translations.php file.


modules/social_features/social_group/social_group.module
```diff
@@ -733,10 +733,10 @@ function social_group_form_alter(&$form, FormStateInterface $form_state, $form_i
     $join_method_default_value = 'added';
     // Ensure we have a better descriptive label.
     if (array_key_exists('added', $form['field_group_allowed_join_method']['widget']['#options'])) {
-      $form['field_group_allowed_join_method']['widget']['#options']['added'] = t('Be added only - users can only be added by group managers');
     }
```
robertragas, line 736: @navneet0693 what is this?  the if statement has nothing in it anymore.

modules/social_features/social_event/social_event.install
```diff
@@ -1502,3 +1502,17 @@ function social_event_update_8807() {
...
+
+/**
+ * Update the enrollment method text of an event on event add form.
+ */
+function social_event_update_8808() {
+  /** @var \Drupal\update_helper\Updater $updateHelper */
+  $updateHelper = \Drupal::service('update_helper.updater');
+
+  // Execute configuration update definitions with logging of success.
+  $updateHelper->executeUpdate('social_event', 'social_event_update_8808');
```
robertragas, line 1514: Did you test this or am I doing something wrong? I get the following message
[warning] Unable to update configuration for field.storage.node.field_enroll_method.

robertragas:
So this one got some code changes in the end by me to change the way to load in the configuration changes. Using the structure of the update helper to create a static file that will never change, but using the loading of the yml file, because update helper seems to have trouble converting array keys to php.

Comment about this can also be seen in the update hook file.

Also showed Navneet these changes and he approved.

Kingdutch:
Hey @robertragas ,

I see that in your last commit you introduced `modules/social_features/social_event/config/update/social_event_update_8809.yml`. This file is named and located as if it's an Configuration Update Definition file. However, the contents is no longer structured that way. This could lead to issues in the future or confuse developers who find the file before the hook.

To avoid this confusion I think we should move the file somewhere else. Maybe to `config/update_<date>/<config_name>.yml` for example. That way you can see it's just intended as a static version of the configuration file and not suitable for tools that are using the Configuration Update Definition format. (Hopefully in the future there will be tools besides the Update Helper module that can use this format).

Cheers :D

PR #1848, opened 2020-05-20
Issue #2995501: Group filters show values that a user does not need access to

ronaldtebrake:
## Problem
On the all groups page if a user has access to two different types of groups then the user can filter on all group types. Even if he/she does not have access to a certain group type.

## Solution
We need to ensure we check the current user roles, and see if any of those roles have acces to the 'view group' permission of a certain group type if that its the case we can show the filter option.

The way it's done is the following.

1. If AN can see the groups of a group type - We show the filter
2. If group type has a outsider role and that role can see the groups of a group type  - We show the filter
3. If group type has a member role and that role can see the groups of a group type  - We show the filter (This is necessary so we don't need to invalidate caches or calculate which group a user belongs to every time he joins / leaves a group. Once a member can see it and a user can potentially become a member it's good enough for us to show the filter option)
4. For all the group type advanced outsider roles matching any of the user roles, and that role can see the groups of a group type we show the filter.
For example the Administrator, CM, SM roles can have view group permissions while the rest doesn't. We show the filter option as well.

## Issue tracker
https://www.drupal.org/project/social/issues/2995501

## How to test
- [x] See that for AN users on the Group search and the All Groups overview the filter block only shows the group types Anonymous users have "view group" permissions for. You can play around with this by going to for example
`/admin/group/types/manage/closed_group/permissions`
Enable the General " View group " permission for anonymous to see the filter option becomes available. 
- [x] See for secret groups nothing breaks so access checks are still in place
- [x] See that for SM / CM you can change the advanced outsider permissions and they work
- [x] See that if there is only one group type you can filter, the block is removed. One option doesn't make sense.

## Release notes
On the search groups and all groups page we now ensure you can only filter on group types of groups you are able to see.

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -1895,14 +1895,14 @@ function social_group_can_view_groups_of_type(string $type, AccountInterface $ac
   // Authenticated is a locked Role,
   // So first we check outsiders also easier on the performance :)
   $outsider = $group_type->getOutsiderRole();
-  if (NULL !== $outsider && $outsider->hasPermission('view group'))  {
+  if (NULL !== $outsider && $outsider->hasPermission('view group')) {
     return TRUE;
   }
 
   // If members can see it, we also need to show group types because
   // we are not calculation memberships.
   $member = $group_type->getMemberRole();
-  if (NULL !== $member && $member->hasPermission('view group'))  {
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
Also cherry-picked to 8.x and 7.x

PR #1851, opened 2020-05-22
#3119327: Make the confirmation join text more meaningful for users accepting an invite to a group

bramtenhove:
## Problem
Screenshot 2020-05-22 at 13 39 26
The above text does not refer to the user accepting an invite.

## Solution
By altering the text when someone accepts an invite it is clearer for the user what is about to happen.

Screenshot 2020-05-22 at 13 35 59

## Issue tracker
https://www.drupal.org/project/social/issues/3119327

## How to test
- [ ] Enable the `social_group_invite`
- [ ] Create a group where users can join directly
- [ ] Join the group as a normal user, observe that you see the original text
- [ ] Invite a different user as the group manager
- [ ] Observe that when the invitee accepts to join the group the text is more meaningful

## Screenshots
N/a

## Release notes
N/a, should be part of the Group Invite feature.

## Translations
N/a

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -611,9 +611,17 @@ function social_group_form_alter(&$form, FormStateInterface $form_state, $form_i
 
   // Set some helpful text on the group join form now it's there.
   if (in_array($form_id, $join_forms)) {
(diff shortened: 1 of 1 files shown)


bramtenhove:
The group invite feature needs tests written. This should be part of that test.

robertragas:
Awesomesauce, works!

PR #1852, opened 2020-05-25
TB-4416 - Add directly page access

sjoerdvandervis:
## Problem
A user with sufficient access has no access to add people to a group directly; he / she will be refused access because of access checks.

## Solution
When a user has sufficient rights to a group and wants to add people directly to the group, we will now grant this person access.

## How to test
*For example*
- [x] Login as a GM and go to your group
- [x] Go to `group/{{GROUP ID}}/content/add/group_membership`
- [x] See that you are able to access this page and that you can add people to your group.

## Release notes
*A group manager is now able to add people to his/her flexible group directly.*


ronaldtebrake:
Hi @navneet0693 @sjoerdvandervis 
Make sure to add this to the correct milestone.
Also ensure we have drupal.org tickets, so we are open on what we are working on. 
Bugs can be backported and I assume this one needs a backport because adding directly to a group was already in 8.x.

If you are unsure about this feel free to PM me or ask in the slack channels.

PR #1862, opened 2020-05-27
by robertragas: add defensive check for mentions in social tokens

robertragas:
## Problem
When we send out notifications, it can be that in cases a mention is made to a person. These tokens reside in social_features/social_mentions/social_mentions.tokens.inc

The problem is that when these are run through the cron it can be that the mentioned user is already deleted. In this case the cron will crash and stop. 

## Solution
Add defensive checks when converting mention tokens.

## Issue tracker
https://www.drupal.org/project/social/issues/3150253

## How to test
- [ ] Disable cron, and mention someone in a post.
- [ ] After the activity creator has made the notification to send out with the email worker, delete the user you mentioned.
- [ ] Run the cron and see it crashes
- [ ] Try again with this patch and see it will continue.

## Release notes
We have added defensive checks to the cron when sending out notifications with mentions in it.

### DIFF
--- a/modules/social_features/social_mentions/social_mentions.tokens.inc
+++ b/modules/social_features/social_mentions/social_mentions.tokens.inc
@@ -7,6 +7,7 @@
 
 use Drupal\Component\Utility\Unicode;
 use Drupal\Core\Render\BubbleableMetadata;
+use Drupal\mentions\Entity\Mentions;
 use Drupal\user\Entity\User;
 use Drupal\Core\Render\Markup;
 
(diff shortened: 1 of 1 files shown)


robertragas:
@navneet0693 could you review this? Is the issue we worked on, which is a part of crons breaking.

ronaldtebrake:
Hi @navneet0693 @robertragas 
I have decided to re-open and merge this. Otherwise this fix won't land because the PR #1873 isn't ready yet. 
Once you merge the base branch back in it should be all fine :)

Also if you decide to close a PR, which is fine. Please ensure the linked PR also contains this d.org ticket so we know what to close and what ended up in what release.

ronaldtebrake:
Also cherry-picked to 8.x

PR #1865, opened 2020-05-28
Issue #3143472 by SV: Join flexible group doesn't work

volodymyr-sydor-lemberg:
## Problem
Currently, normal members not able to join flexible group with enabled "Join directly" option.

## Solution
Use 'Allowed join methods' instead of static visibility settings to provide the ability to join a flex group.
## Issue tracker
- https://getopensocial.atlassian.net/browse/YANG-2924
- https://www.drupal.org/project/social/issues/3143472

## How to test
- [ ] Create a Flexible group with “Group members“ visibility and "Join directly" method
- [ ] Log in as a normal member
- [ ] Try to Join the group
- [ ] Page will be reloaded but without any action


ribel:
@bramtenhove @ronaldtebrake We had an issue when we had custom visibility for Role - users were not able to Join groups, because in distro we had an assumption that we will have only Public, community and Group members role.
I think check for `social_group_flexible_group_community_enabled` is not needed since `social_group_flexible_group_can_join_directly` should be enough.

Any concerns about it?

ronaldtebrake:
I have just tested this and the concerns I had were all still working. The access handler ensures the topics / events overview are indeed still not accessible for this scenario. So I'm happy for this to be merged!

ronaldtebrake:
Next time don't forget to fill in as much of the PR template as possible by the way :)

volodymyr-sydor-lemberg:
> We might want to split this up and use your code only for the group join method route.
> Right now the groups and events overview return an access denied for non group members when there is only group members visibility. Also the group stream still shows activities when tried to access directly.
> 
> screen
> 
> So for the `entity.group.join` route we use the
> 
> ```
> elseif (!$group->getMember($user) && in_array($routeMatch, $routes, FALSE)	      // "Join directly" in this group.
>         && !social_group_flexible_group_community_enabled($group)	      
> ```
> 
> For the other routes we use the
> 
> ```
>         && !social_group_flexible_group_public_enabled($group)) {	        !$group->getMember($user) &&
>         in_array($routeMatch, $routes, FALSE) &&
>         !social_group_flexible_group_can_join_directly($group)
>       ) {```
> ```

done

ribel:
Hi @ronaldtebrake, we made changes that you asked. Could you review it again?

ribel:
🍒 picked to 8.x and 9.x

PR #1866, opened 2020-06-01
Issue #3144386 by agami4: Horizontal scroll always visible on the landing page

agami4:
## Problem
Horizontal scroll always visible on the landing page

## Solution
Remove styles(negative margin for all components on the landing page) and remove general 'container' class

## Issue tracker
https://www.drupal.org/project/social/issues/3144386
https://getopensocial.atlassian.net/browse/YANG-3047

## How to test
*For example*
- [x] Go to/create a landing page
- [x] Create a new component
- [x] Check if you have horizontal scroll

## Screenshots
before:
YANG-3047

after:
tratata

## Release notes
The landing page does not have horizontal scroll


robertragas:
Sorry, just happen to run across this issue and was wondering. We have a client that also has these horizontal scrolling on mobile on stream. Will this also be fixed with this solution or is this specifically for landing pages?

agami4:
Hey @robertragas, This solution works only for the landing pages, because we have horizontal scroll only those pages. Can you attach a screenshot on this issue or send me a private msg, what the page has a horizontal scroll?

modules/social_features/social_landing_page/templates/page--node--landing-page.html.twig
```diff
@@ -27,6 +27,12 @@
  */
 #}
 
+{%
+  set classes = [
+  'row',
```
ribel, line 32: @agami4 maybe it makes sense to add those classes together with `$variables['attributes']->addClass('container'`) in `social_landing_page.module` ?
Also if we will not remove those classes: `with-title-region', 'layout--with-complementary` it will not break any section of Landing Page?

modules/social_features/social_landing_page/css/social_landing_page.section.introduction.scss
```diff
@@ -1,12 +1,7 @@
 .introduction {
   padding: 10px 0;
   background-color: #fff;
-  @media(min-width: 1200px) {
-    .paragraph--section & {
-      margin-left: calc(-100vw / 2 + #{1200px} / 2);
```
ribel, line 6: @agami4 can we simplify this by just using: `calc(-50vw + #{600px});` ?

agami4: @ribel, I've removed this part of the styles because this case add horizontal scroll

modules/social_features/social_landing_page/templates/page--node--landing-page.html.twig
agami4: @ribel,
The `row` class adds to the current landing page.
The `container` class adds to the edit landing page

agami4: And I've checked all paragraphs to the landing page and all looks good

ribel:
@agami4 I see it is fixed for Landing pages, but would be good to fix it also for Topics, Events and other content types.
See an example here: https://x-bitbucket-pr-79-mkq4mvq-vncdokyquw6mc.eu.platform.sh/topic/faqs-about-open-social

agami4:
@ribel, According to your last comment -> If I remember correctly it was fixed in another PR. I've checked Topic and everything is okay

ribel (approved):

ribel:
🍒 picked to 8.x and 9.x

PR #1869, opened 2020-06-03
Issue #3143273 - Add filter with group join methods for flexible groups

ronaldtebrake:
## Problem
Whenever a user selects flexible groups from the Group type filter, we want the user to be able to filter on join methods. Especially now with the request to join it makes a lot of sense. In the future all platforms will migrate to flexible groups only. So we can remove some of the additional alter logic that we need to take in to account now.

## Solution
Add a filter for the allowed join method for flexible groups. On the group overview and search page.

## Issue tracker
https://www.drupal.org/node/3143273

## How to test
- [x] Ensure you checkout this branch
- [x] Add 2 flexible groups one with Join directly one with Be added
- [x] See that on the explore groups and search groups you can now filter on join method once flexible group is selected as group types
- [x] Try to enable social group request
- [x] Add a new flexible group with this join method
- [x] See that you can also filter on this new join method

## Screenshots
https://www.loom.com/share/8242b0f475da47ffb2a589fd32000b6f

## Release notes
We now have the group join method available as filter on the explore group and search group pages whenever you filter on flexible groups.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,6 +75,9 @@
             "drupal/r4032login": {
               "r4032login should perform access check for /user/login as anonymous user": "https://www.drupal.org/files/issues/2018-11-01/3010747-3-perform-access-check-as-an-user.patch"
             },
+            "drupal/search_api": {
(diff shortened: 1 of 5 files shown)


robertragas:
Approved, green, merging.

PR #1870, opened 2020-06-03
Issue #3145315 Add filter to the explore events page for enrollment method

ronaldtebrake:
## Problem
As a User it would be helpfull to filter events on the events overview based on their enrollment method.

## Solution
Adds a filter by default to the community or explore events overview.

## Issue tracker
https://www.drupal.org/node/3145315

## How to test
- [x] Enable the social event invite module
- [x] Create several events (at least 6) one for every enrollment method (Open / Request / Invite)
- [x] And one in the future and one in the past
- [x] Go to Explore -> Events
- [x] See that you can now filter on the enrollment method
- [x] See that the upcoming / passed events filter also still works

## Screenshots
https://www.loom.com/share/8d8698956e0d4475a0484830228b2f1a

## Release notes
Adds a new filter for the enrollment method on the explore events overview. So users can quickly identify which events they can easily enroll for.

### DIFF
--- a/modules/social_features/social_event/config/install/views.view.upcoming_events.yml
+++ b/modules/social_features/social_event/config/install/views.view.upcoming_events.yml
@@ -12,8 +12,6 @@ dependencies:
     - options
     - social_event
     - user
-_core:
-  default_config_hash: jICmCvyBLhkwjQbswDAncFY0kUOuDGtZmMJf26vPvxg
 id: upcoming_events
 label: '(Upcoming) Community events'
 module: views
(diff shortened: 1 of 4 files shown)


robertragas:
Works like it should, I just wonder about one thing. In the form edit itself it will hide enroll method invite only when the submodule is not enabled. In the frontend it shows regardless. Could be that people wonder what this option is as they don't see it when creating events.

@ronaldtebrake What do you think?

ronaldtebrake:
@robertragas good point.

So the way the field is created can be optimized and rather than hiding it everywhere the fix for this would be to ensure the field gets created within the module by creating your own allowed values. So if that is blocking us, I wouldnt want to hide it with form alters but do it good instead. 
However that is a bit too big for this scope, so I can live with it and i will create a drupal.org ticket so people are aware of this issue. 

Just for some future reference: 

```/**
 * Allowed join method values.
 */
function social_group_request_allowed_join_method_values(FieldStorageConfig $definition, ContentEntityInterface $entity = NULL, $cacheable) {
  $allowed_values = $definition->getSetting('allowed_values');
  $allowed_values['request'] = t('Request to join - user requests to join need to be approved by group managers');
  return $allowed_values;
} ```

PR #1872, opened 2020-06-05
fix multiple issues with event invite

robertragas:
## Problem
We encountered multiple problems with the event invite feature
1. When you are the event owner or organizer you should always be able to enroll via the hero, even when it's invite only
2. When you create an event with "request to join" and have multiple people joined, you can convert it to "invite only" but then all enrollments will be lost that got accepted.
3. When you are invited for an event and accepted, you are not able to un-enroll anymore as the button is gone.
4. When I want to un-enroll to an invite only invite, I need to press the button twice before it actually un-enrolls me.
5. When changing an event from "request to join" to "invite only" the enrollments block on the detail pages is gone.

## Solution
1. Create exceptions for enrolling when you are the event owner or organizer
2. Add extra check in checking values for enrollment status so it does not delete the wrong enrollments.
3. Make sure it checks on enrollments before unsetting the button for invite only
4. When un-enrolling for invite only add an additional check to get the current enrollment status and change it accordingly.
5. In the blockaccess add an additional check to see if the invite only is enabled and make an exception for that.

## Issue tracker
https://www.drupal.org/project/social/issues/3149795

## How to test
1.
- [x] Create an event with enroll method "invite only"
- [x] See that you can still enroll as owner
- [x] Create an event with enroll method "invite only" as assign another person as organizer
- [x] Login as this organizer and see you can still enroll

2. 
- [x] Create an event with enroll method "request to join"
- [x] Enroll as owner, but also request with other people and accept some, other leave them pending.
- [x] Convert the event to "invite only". See that everyone who was enrolled is still enrolled.

3.
- [x] Create an event with enroll method "invite only"
- [x] Invite someone and accept as that person
- [x] See you are still able to un-enroll after accepting

4. 
- [x] Create an event with enroll method "invite only"
- [x] Invite someone and accept as that person
- [x] Un-enroll for the event and see it only takes 1 click.

5.
- [x] Create an event with enroll method "invite only"
- [x] Invite someone and accept as that person
- [x] Go to the detail page of the event and you will still see the enrollment block on the right

## Release notes
We have solved multiple issues concerning the event invite feature, such as 
- properly converting invite statuses when changing the event enroll method
- Enrollment access controls to the event

### DIFF
--- a/modules/social_features/social_event/modules/social_event_invite/src/Form/EventInviteEnrollActionForm.php
+++ b/modules/social_features/social_event/modules/social_event_invite/src/Form/EventInviteEnrollActionForm.php
@@ -30,7 +30,6 @@ class EventInviteEnrollActionForm extends EnrollActionForm {
     $nid = $this->routeMatch->getRawParameter('node');
     $current_user = $this->currentUser;
     $uid = $current_user->id();
-    $isNodeOwner = ($node->getOwnerId() === $uid);
 
     if (!$current_user->isAnonymous()) {
       $conditions = [
@@ -43,8 +42,7 @@ class EventInviteEnrollActionForm extends EnrollActionForm {
       // Unless you are the node owner or organizer.
       if (empty($enrollments)) {
         if ((int) $node->field_enroll_method->value === EventEnrollmentInterface::ENROLL_METHOD_INVITE
-          && !$isNodeOwner
-          && social_event_manager_or_organizer() === FALSE) {
+          && social_event_owner_or_organizer() === FALSE) {
           return [];
         }
       }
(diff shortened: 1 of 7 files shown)


robertragas:
Made a small change in using social_event_owner_or_organizer so it also checks correctly if this module has been enabled. Waiting for travis now as I expect tests to go green now.

robertragas:
I made one more additional change in the group-create-open.feature, in the end the steps are
# DS-722 As an outsider I am not allowed to enrol to an event in group

The thing is, that the event is created by the user itself, and a user should be always allowed to enroll in his/her own event. This is what my checks add with social_event_owner_or_organizer in the EnrollActionForm.

So I changed the behat test by adding a third person called "outsider" who has no connection whatsoever with the event, and this one should not be able to enroll.

PR #1873, opened 2020-06-09
Issue #3069371 by robertragas: reintroduce fix with mention tokens in email notifications

robertragas:
## Problem
When mentioning people it will convert there id to the displayname. The message templates also send out email notifications with a small preview also containing these mentions.

These tokens are not getting converted correctly. The problem already exists for a while, and a workaround was made for a client, but now more clients are reporting the same problem.

## Solution
Alter the way the tokens are getting generated, by applying the mention filter with the settings that do not require an access check. This way the cron will also be able to convert the tokens.

## Issue tracker
https://www.drupal.org/project/social/issues/3069371

## How to test
- [ ] Trigger the message template that contain previews for email notifications such as:
* create_comment_reply
* create_post_profile
You need to make sure you also mention the person in the comment or in the post on the profile with the @mention function.
- [ ] Run the cron through terminal and see the Id does not get converted
- [ ] Check out to this branch and try again

## Release notes
We have solved a problem where email notifications containing mentions would not get converted to the correct display name but instead showed an id.


robertragas:
@Kingdutch also wanted to mention you as you helped with the initial patch. Reason for re-opening this is that we are seeing this problem more and more, and it's becoming quite major for the clients. Patching each client and maintaining is a lot of overhead we cannot really afford.

Even the patch is not ideal, I think we really need a solution on the short term for this, also considering that a lot people outside of our client base will face this problem.

So input from Shipsters is appreciated how you guys look at it.

modules/custom/mentions/src/Plugin/Filter/MentionsFilter.php
```diff
@@ -227,7 +227,7 @@ public function filterMentions($text) {
           '#render_link' => $output_settings['renderlink'],
           '#render_value' => $output['value'],
         ];
-        $mentions = $this->renderer->render($build);
+        $mentions = $this->renderer->renderPlain($build);
```
ronaldtebrake, line 230: maybe add a quick comment as to why we need renderPlain here.

modules/custom/mentions/src/Plugin/Mentions/Entity.php
```diff
@@ -40,17 +43,43 @@ public static function create(ContainerInterface $container, array $configuratio
...
+      // Get the profile.
+      $profile = $this->entityManager->getStorage($mention['target']['entity_type'])
+        ->load($mention['target']['entity_id']);
+
+      // Get the output value according to the config settings.
+      $config = $this->config->get('mentions.settings');
+      switch ($config->get('suggestions_format')) {
+        case SOCIAL_PROFILE_SUGGESTIONS_FULL_NAME:
+        case SOCIAL_PROFILE_SUGGESTIONS_ALL:
+          $output['value'] = $profile->getOwner()->getDisplayName();
```
ronaldtebrake, line 70: Doesn't this already fall back to the account name? 

``` $name = $this
    ->getAccountName() ?: \Drupal::config('user.settings')
    ->get('anonymous');
  \Drupal::moduleHandler()
    ->alter('user_format_name', $name, $this);
  return $name;``` 

Or did we alter it empty somewehere? 🤔

File: modules/custom/mentions/src/Plugin/Mentions/Entity.php [hunk 3]
```diff
@@ -66,8 +95,11 @@ public function outputCallback($mention, $settings) {
   public function targetCallback($value, $settings) {
     $entity_type = $settings['entity_type'];
     $input_value = $settings['value'];
-    $query = $this->entityQueryService->get($entity_type);
-    $result = $query->condition($input_value, $value)->execute();
+    $query = $this->entityQueryService
+      ->get($entity_type)
+      ->condition($input_value, $value)
+      ->accessCheck(FALSE);
```
ronaldtebrake, line 101: this should probably not be here

navneet0693: @ronaldtebrake any suggestions on this?

robertragas: I have been looking at this morning. I think we cannot get past by this accesCheck false.

The targetCallback is being used on quite some places though, not just the cron. So to make it any safer I can imagine we set a variable or such in the the activity_send_email_cron so when the functions gets run from the cron it will disable the access check only during the cron.

@ronaldtebrake @navneet0693 what do you think? As anonymous users are not allowed to view profiles I don't think we can remove this check in total.

Kingdutch: Hey @robertragas and @ronaldtebrake ,

I see some movement in this PR. I'm also working on the `social_profile` module which as one of the changes will give anonymous users access to profile information (which will then be checked on a per-field level depending on configuration).

So this may already solve this problem without removing access checks?

navneet0693:
Closing this, as it is quite old now. Please re-open or create a new one if required.

PR #1874, opened 2020-06-10
Issue #3150451 by rolki: Add the ability to hide/show flexible group members tab and newest members block for outsiders

rollki:
## Problem
Outsiders shouldn't be able to see members of a flexible group and newest members block

## Solution
- Make it configurable on group edit page
- Block with newest members and members tab should be shown / hidden depending on settings

## Issue tracker
* https://www.drupal.org/project/social/issues/3150451
* https://getopensocial.atlassian.net/browse/YANG-3022

### DIFF
--- a/modules/social_features/social_group/modules/social_group_flexible_group/config/update/social_group_flexible_group_update_8802.yml
+++ b/modules/social_features/social_group/modules/social_group_flexible_group/config/update/social_group_flexible_group_update_8802.yml
@@ -1,0 +1,66 @@
+core.entity_form_display.group.flexible_group.default:
+  expected_config:
+    content:
+      path:
+        weight: 3
+    third_party_settings:
+      field_group:
+        group_content:
+          weight: 1
+        group_location:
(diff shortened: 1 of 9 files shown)


rollki:
@ribel 
Text for labels was confirmed by Xinyu in this ticket
https://getopensocial.atlassian.net/browse/YANG-3022

ronaldtebrake:
Hi @ribel @rollki i have added some remarks in the YANG ticket because I'm not so sure about this solution.

ronaldtebrake:
Also please when reviewing and opening up a PR make sure the entire PR template is filled in. Because the info in there could have answered my questions already :)

ronaldtebrake:
We are probably better of making sure we can migrate everything to Flexible groups which also means we need to make sure Secret Group types are supported. This should cover this scenario and saves us from adding more options in the Group add / edit screen (which is already difficult).

We will park this for now but can be used as a patch still.

PR #1875, opened 2020-06-10
Issue #3150640 by Kingdutch: Content tagging filters on search and ov…

Kingdutch:
…erview don't update when terms or settings change

<h2>Problem</h2>
Deleting a content tag from the vocabulary will not change the filters that are available on the overview and search pages. A cache clear is required for this to be resolved.

The suspicion is that this is caused by the SocialTaggingOverride which is responsible for the fields added but provides an empty CacheableMetadata object causing it to be cached endlessly.

The view on a node seems to go well by accident because a term is deleted from a node when the term itself is removed. This invalidates the cache for the node and the SocialTagsBlock has cache tags for a specific node. The same seems to be the case for a group at quick glance.

<h2>Proposed solution</h2>
Ensure that for the configs that the class overrides the correct cache contexts and tags are applied so that the configuration override invalidates when the taxonomy terms change.

## Issue tracker
https://www.drupal.org/project/social/issues/3150640

## How to test
- [ ] Using version 8.9.0 of Open Social with the Social Tagging module enabled
- [ ] As a sitemanager
- [ ] Create a content tagging hierarchy
- [ ] Create some content with the content tags
- [ ] Remove terms from the tagging vocabulary
- [ ] Observe that on the explore and search pages the removed terms still seem to be present
- [ ] Apply this fix
- [ ] Observe that when repeating the steps the removed terms are also immediately cleaned up on the explore and search pages.

## Release notes
Tags for content tagging are now properly removed from filters after being deleted instead of lingering around like an unwelcome guest. 

## Change record
No developer facing code affected, implementation details fixed only.

## Translations
No user facing texts affected

### DIFF
--- a/modules/social_features/social_tagging/src/SocialTaggingOverrides.php
+++ b/modules/social_features/social_tagging/src/SocialTaggingOverrides.php
@@ -19,11 +19,18 @@ class SocialTaggingOverrides implements ConfigFactoryOverrideInterface {
   protected $tagService;
 
   /**
-   * {@inheritdoc}
+   * Whether this config override should apply to the provided configurations.
+   *
+   * Used to check what to return in loadOverrides as well as in the metadata
+   * method.
+   *
+   * @param array $names
+   *   The names of the configs for which overrides are being loaded.
+   *
+   * @return bool
+   *   Whether we override those configs.
    */
-  public function loadOverrides($names) {
-    $overrides = [];
-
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
Changed milestone to 7.6, looks like it was already in there. Seems to work fine though. :)

ronaldtebrake:
also in 7.x and 8.x

PR #1876, opened 2020-06-12
Feature/3151112

rollki:
## Problem
*Describe the problem you're trying to solve, this should motivate why the changes you're proposing are needed.*

## Solution
*Describe the solution you've created, elaborate on any technical choices you've made. Why is this the right solution and is a different solution not the right one?*

## Issue tracker

## How to test
*For example*
- [ ] As a sitemanager

## Screenshots

## Release notes

## Change Record

## Translations

### DIFF
--- a/modules/custom/social_views_infinite_scroll/social_views_infinite_scroll.info.yml
+++ b/modules/custom/social_views_infinite_scroll/social_views_infinite_scroll.info.yml
@@ -1,0 +1,7 @@
+name: Social Views Infinite Scroll
+description: 'Social Views Infinite Scroll'
+type: module
+core: 8.x
+package: Social
+dependencies:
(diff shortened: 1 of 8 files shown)


Kingdutch:
Hey @rollki ,

I'm assuming you've opened this PR to run Travis tests. These can also be run locally, but if that's not possible, please consider opening a Pull Request as a Draft PR instead of a normal PR. Especially when nothing in the Pull Request template is filled in.

Opening a PR as a Draft PR makes it clearer that the PR can be safely ignored by maintainers and causes less noise than a normal PR with the "needs work", "don't merge" labels.

Thanks!

rollki:
Hi @Kingdutch 
Sure, thanks for the explanation.

PR #1877, opened 2020-06-12
Issue 3151179 by frankgraave - Fix flexible group with only public content visibility

frankgraave:
## Problem
When creating a flexible group with only public content visibility, the group page does not show certain links such as the stream to the LU any longer. They are also not accessible by logged in users.

## Solution
Inside the `FlexibleGroupContentAccessCheck.php` class, there is the following if statement :

```
    // It's a non member but Community isn't enabled.
    // No access for you only for the about page.
    if ($account->isAuthenticated() && !social_group_flexible_group_community_enabled($group)
      && $route_match->getRouteName() !== 'view.group_information.page_group_about'
      && $route_match->getRouteName() !== 'entity.group.canonical'
      && $route_match->getRouteName() !== 'view.group_members.page_group_members') {
      return AccessResult::forbidden()->addCacheableDependency($group);
    }
```

This if statement should include `&& social_group_flexible_group_public_enabled($group)` to ensure the correct access results. We also should include this check in the tests, which I have done with the first commit. To show that it's actually breaking for LU and not for AN.

## Issue tracker
https://www.drupal.org/project/social/issues/3151179

## How to test
- [ ] Do not check out to this branch yet
- [ ] Create a flexible group with only public content visibility
- [ ] Visit the group as AN user
- [ ] See that you can see the Stream, About, Events, Topics and Members tabs
- [ ] Log in as LU and visit the group
- [ ] See that you only have About and Members
- [ ] Check out to this branch, repeat the above checks and see it works now

## Screenshots
N.A.

## Release notes
Bug fixed where flexible groups with public visibility only restricted authenticated users to view the group properly.

## Change Record
N.A.

## Translations
N.A.

### DIFF
--- a/modules/social_features/social_group/modules/social_group_flexible_group/src/Access/FlexibleGroupContentAccessCheck.php
+++ b/modules/social_features/social_group/modules/social_group_flexible_group/src/Access/FlexibleGroupContentAccessCheck.php
@@ -73,6 +73,7 @@ class FlexibleGroupContentAccessCheck implements AccessInterface {
     // It's a non member but Community isn't enabled.
     // No access for you only for the about page.
     if ($account->isAuthenticated() && !social_group_flexible_group_community_enabled($group)
+      && social_group_flexible_group_public_enabled($group)
       && $route_match->getRouteName() !== 'view.group_information.page_group_about'
       && $route_match->getRouteName() !== 'entity.group.canonical'
       && $route_match->getRouteName() !== 'view.group_members.page_group_members') {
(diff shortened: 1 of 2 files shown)


frankgraave:
On commit 03518dd you can see that the added test goes [wrong](https://travis-ci.org/github/goalgorilla/open_social/jobs/697602698#L2131).

On commit ca1dc16 you can see that the added test now goes [right](https://travis-ci.org/github/goalgorilla/open_social/jobs/697623181#L2132)!

robertragas:
Merged and backported to 8.4

PR #1879, opened 2020-06-15
by robertragas: replace select2 from theme for contrib module by remo…

robertragas:
…ving the js, but keeping the custom styling

## Problem
For the social_tagging we currently use the select2 library in our theme to apply this widget. As this widget is quite nice and we would to apply it to different fields, we should replace it with the contrib module so we can easily apply this widget.

## Solution
Remove the custom JS from the theme and implement the select2 contrib module. Keep the custom styling as we do have our custom theming that needs to be applied still.

## Issue tracker
https://www.drupal.org/node/3151727

## How to test
- [ ] Enable the social_tagging
- [ ] Configure the social_tagging and create a topic
- [ ] See the widget is still the same and working
- [ ] Create a new select field and choose the select2 widget
- [ ] See this also works out of the box

## Release notes
We have replaced our custom select2 widget by the contrib module. This makes it easier to apply this widget on more places.


robertragas:
Merged in 9.x, backport will happen after 8.4 has been released

robertragas:
8.4 is released, so I backported this to 8.5

PR #1883, opened 2020-06-18
Feature/3152934 invite preview

volodymyr-sydor-lemberg:
## Problem
There is no possibility to preview of invite email for Event/Group

## Solution
After the recipients are selected - "Show an example of the message that will be sent to the invited anonymous user.

## Issue tracker
- https://getopensocial.atlassian.net/browse/YANG-3171
- https://www.drupal.org/project/social/issues/3152934

## How to test
- [ ] Using version 8.x-9.x of Open Social with the social_group_invite/social_event_invite modules enabled
- [ ] As a group manager
- [ ] Try to invite a user to Event and preview of sent mail should be shown on the page (/node/[nid]/invite/email)
- [ ] Go to any group 
- [ ] Try to invite a user and preview of sent mail should be shown on the page (/group/[gid]/invite-members)

## Release
You are now able to preview emails before sending them out as part of the event and group invitation feature. You will see this preview on the invite pages itself where you would fill in the email addresses. Note that these previews are not exact visual representations of how they will look in every email client.


robertragas:
I approved it. It's a nice addition but I do think we need a follow-up for this. But more in general for Open Social. (we will picking up the follow-up during refinement of team B)

We need to see how we can separate the logic that we put into buildForm functions, otherwise in the end the code will become unreadable, and also if we could put it in a service or something else, we could reuse it in a better way because I can see we eventually want this preview everywhere where a bulk email gets send out and we want to see the preview first.

PR #1891, opened 2020-06-30
Issue #3155587 by robertragas: alter the config override for paths to…

robertragas:
… a generic block access

## Problem
Currently there are some modules that use a config override to alter the visibility path of blocks. This is getting the original state of the block visibility and adding some new paths to it.

Unfortunately this will only work to a max of 1 config override, because adding a second one will overwrite the first one, as we are getting the original config.

Also since loading configs in a config override is not best practice, I am proposing to make a more generic way in our block access where we can deny access to a block if the paths are matching.

## Solution
Create a general hook where people can add their block_ids and paths to be hidden and use a invokeall to load those changes in a hook block access and restrict access based on the paths.

## Issue tracker
https://www.drupal.org/node/3155587

## How to test
- [ ] Check that the changes from the current \Drupal\social_event\EventEnrollmentConfigOverride are still being picked up
- [ ] Try adding your own implementation of the new hook_social_core_block_visibility_path and try to hide a block on a certain page and see it works. 
Hint: Try hiding the block that are also being changes in social_event_social_core_block_visibility_path so you can also test the arrays are getting merged correctly when multiple instances are calling the same hook on the same block, but with different paths.

## Release notes
We have added a more flexible way of hiding blocks on certain paths in code without the need to use a config override. See the hook_social_core_block_visibility_path for more information.


modules/social_features/social_core/social_core.module
```diff
@@ -522,6 +524,35 @@ function social_core_filter_process_format($element) {
...
+ * Hide blocks on extendable visibility path.
+ *
+ * Implements hook_block_access().
+ */
+function social_core_block_access(Block $block, $operation, AccountInterface $account) {
+  $blocks = &drupal_static(__FUNCTION__);
+  $blocks = [];
+
+  // Allow other modules to change the block visibility.
+  \Drupal::moduleHandler()->alter('social_core_block_visibility_path', $blocks);
```
Kingdutch, line 537: I like the optimisation of not running the hook more than once, but to be effective we should check if we actually need to recompute the blocks.

```suggestion
if (empty($blocks)) {
  \Drupal::moduleHandler()->alter('social_core_block_visibility_path', []);
 }
```

This also shows that you're not really altering anything but only collecting arrays. With this in mind it may be better to use https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21ModuleHandler.php/function/ModuleHandler%3A%3AinvokeAll/8.2.x

Note that that method already merges the arrays for you, so you may need to move to a slightly different array structure to ensure merging happens correctly. 
```
'block_id' => [
  'path' => TRUE,
]
```

The benefit of this is that Drupal takes care of merging the array for you and you can remove the loops from each of the hook implementations.

If you want to alter what another module does then you'd have to implement an alter hook separately. The added benefit of this is that page definitions can be altered mostly without changing the module weight (as they happen in separate hooks).

Moving the entire creation and caching of `$blocks` into a separate function can make it easier to reason about as well :)

modules/social_features/social_core/social_core.module
```diff
@@ -522,6 +524,35 @@ function social_core_filter_process_format($element) {
...
+
+  if ($operation === 'view' && array_key_exists($block->getPluginId(), $blocks)) {
```
Kingdutch, line 539: Instead of `array_key_exists` we can use `isset`. It's [slightly faster](https://stackoverflow.com/a/9522522/576060) and provides [a bit more robustness](https://stackoverflow.com/a/3210982/576060) about incorrectly specified values.

```suggestion
  if ($operation === 'view' && isset($blocks[$block->getPluginId()])) {
```

modules/social_features/social_core/social_core.module
```diff
@@ -522,6 +524,35 @@ function social_core_filter_process_format($element) {
...
+    // Get the current path.
+    $current_path = \Drupal::service('path.current')->getPath();
+
+    // Go through the request paths and see if they match.
+    foreach ($blocks[$block->getPluginId()] as $page) {
+      if (\Drupal::service('path.matcher')
+          ->matchPath($current_path, $page) === TRUE) {
+        return AccessResult::forbidden();
+      }
+    }
```
Kingdutch, line 549: Cool! Nice catch on making sure you match wildcard paths!

modules/social_features/social_core/social_core.module
robertragas: Moved to the invokeall.

modules/social_features/social_core/social_core.module
```diff
@@ -522,6 +524,50 @@ function social_core_filter_process_format($element) {
...
+
+    // Set the cache expire to 1 week.
+    $date = new DateTime(date('Y-m-d'));
+    $date->modify('+7 day');
+
+    $cache->set('social_core.block_visibility_cache', $blocks, $date->getTimestamp(), ['social_core.block_visibility_cache']);
+  }
+  else {
+    $blocks = $cache->get('social_core.block_visibility_cache');
+    $blocks = $blocks->data;
```
Kingdutch, line 551:
```suggestion
    $blocks = $cache->get('social_core.block_visibility_cache')->data;
```

Kingdutch (approved):
Looks good to me :D 

We can merge two lines into one to avoid a temporary variable but that's mostly to make it easier to read :D

PR #1898, opened 2020-07-07
#3155032: Fix a caching issue with the group join button

bramtenhove:
## Problem
There's a caching issue with the join CTA buttons for groups when the Sky theme is enabled.

When a user joins the group, the next user will see that he already joined the group because it's already cached.

## Solution
Apply the same fix as #1890 to `8.x-8.x`. This is a custom backport of that of that issue.

## Issue tracker
https://www.drupal.org/project/social/issues/3155032

## How to test
- [x] Create a group
- [x] As user x join the group
- [x] Log in as a user y, and go to the group
- [x] Observe that the button does not say that you joined the group, but that you can join the group

## Screenshots
N/a

## Release notes
The caching for the group join button, in combination with the Sky theme, was incorrect. The caching contexts are now applied correctly.

## Change Record
N/a

## Translations
N/a

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -335,6 +335,9 @@ function social_group_preprocess_group(array &$variables) {
 
   // Prepare variables for statistic block.
   if ($variables['view_mode'] === 'statistic') {
+    $variables['#cache']['contexts'][] = 'group';
+    $variables['#cache']['contexts'][] = 'user';
+
     if ($group->getGroupType()->hasContentPlugin('group_node:event')) {
       $variables['group_events'] = $group_statistics->getGroupNodeCount($group, 'event');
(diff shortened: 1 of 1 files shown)


navneet0693:
The solution works great. But this job is failing somehow: https://travis-ci.org/github/goalgorilla/open_social/jobs/705756354

navneet0693:
As @sjoerdvandervis found out:

It fails because of this: https://github.com/Roave/SecurityAdvisories/blob/master/composer.json#L59
Open Social 8.x-8.x is using 7.x as base to check whether the update path still works there. However, OS 8.x-7.x has core set to ~8.7.14, which conflicts as you would see above (it conflicts when the Drupal version specified in composer is >8, which would throw a conflict if Drupal core < 8.8.8 is being installed


And it is a backport, so it is safe to merge.

PR #1903, opened 2020-07-13
Issue #3158812 by tBKoT: Disable access to the blocked users pages

dmytriikaiun:
## Problem
Disable access to blocked users' pages through the direct link

## Solution
Add custom permission to allow view blocked users' pages and add it to routes

## Issue tracker
- https://www.drupal.org/project/social/issues/3158812
- https://getopensocial.atlassian.net/browse/YANG-3225

## How to test
- [x] Log in as LU (not as SM)
- [x] Go to the `user/{user_id}/information` page
- [x] You should get the `403` page

### DIFF
--- a/modules/social_features/social_user/social_user.install
+++ b/modules/social_features/social_user/social_user.install
@@ -195,3 +195,13 @@ function social_user_update_8809() {
   // The administrator should be able to assign all roles.
   user_role_grant_permissions('administrator', ['assign all roles']);
 }
+
+/**
+ * Update permissions for the sitemanager role.
+ */
+function social_user_update_8810() {
+  // Grant the sitemanagers new permissions.
+  user_role_grant_permissions('sitemanager', [
+    'view blocked user',
+  ]);
+}
--- a/modules/social_features/social_user/social_user.install
+++ b/modules/social_features/social_user/social_user.install
@@ -175,6 +175,7 @@ function _social_user_get_permissions($role) {
     'administer social_user settings',
     'assign contentmanager role',
     'assign sitemanager role',
+    'view blocked user',
   ]);
 
   if (isset($permissions[$role])) {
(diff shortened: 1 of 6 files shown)


ribel:
@dmytriikaiun there are multiple failing behat tests on a step `When I am on /user` - please check and fix.

modules/social_features/social_user/social_user.install
```diff
@@ -195,3 +195,13 @@ function social_user_update_8809() {
...
   user_role_grant_permissions('administrator', ['assign all roles']);
 }
+
+/**
+ * Update permissions for the sitemanager role.
+ */
+function social_user_update_8810() {
+  // Grant the sitemanagers new permissions.
+  user_role_grant_permissions('sitemanager', [
+    'view blocked user',
```
ribel, line 205: @dmytriikaiun please also make sure that new permission will be added on hook_install, so add it to the `_social_user_get_permissions()`.

modules/social_features/social_user/src/Routing/RouteSubscriber.php
```diff
@@ -45,6 +45,21 @@ protected function alterRoutes(RouteCollection $collection) {
       $route->setOption('_admin_route', FALSE);
     }
 
+    // Routes for which needs to disable access if the user is blocked.
+    $disable_access_for = [
+      'view.groups.page_user_groups',
```
ribel, line 50: @dmytriikaiun LU still have access to this route, because it has different _custom_access in `modules/social_features/social_group/src/Routing/RouteSubscriber.php`, so please check if we can fix it there.

modules/social_features/social_user/src/Routing/RouteSubscriber.php
```diff
@@ -45,6 +45,21 @@ protected function alterRoutes(RouteCollection $collection) {
...
+      'view.user_information.user_information',
+      'view.user_discussion.page_user_discussions',
```
ribel, line 52: @dmytriikaiun Since discussions are an external module, and this route doesn't exist in the clean distro, so we usually recommended to remove such lines of code from here and fix it in the related module(s).
But I think this is not so critical so please create follow up to fix it in other modules (Discussions, Challenges, Courses)

ribel:
🍒 Cherry-picked to 8.x-8.x

PR #1904, opened 2020-07-15
Issue #3159259: Book pages need to be available in the search all and search autocomplete

ronaldtebrake:
## Problem
When enabling social book, when you search for a book you will see the result in the Content tab but not in the All tab although that also shows all other pieces of content.
The search autocomplete through the magnifying glass also doesn't show this in that case.

## Solution
Ensure we also index and show book as search results for the All index. This will make sure it works for both.
Also ensure we trigger a re-index and save so the new config is added.

## Issue tracker
https://www.drupal.org/project/social/issues/3159259
https://getopensocial.atlassian.net/browse/OS-303

## How to test
- [x] Login as LU and create a book page
- [x] Try to find it in the search all tab or using the autocomplete, it's not there
- [ ] See that it is there on the Content tab
- [ ] Checkout this branch, run updates
- [ ] The expected result is the same book page is now available in search all and search autocomplete

## Screenshots
Screen Shot 2020-07-15 at 11 08 31
Screen Shot 2020-07-15 at 11 08 21

## Release notes
Book pages can now also be found using the search autocomplete using the magnifying glass, or on the search all tab of the search itself.

### DIFF
--- a/modules/social_features/social_book/social_book.install
+++ b/modules/social_features/social_book/social_book.install
@@ -5,6 +5,8 @@
  * Install, update and uninstall functions for the social_book module.
  */
 
+use Drupal\Core\Entity\EntityStorageException;
+use Drupal\search_api\Entity\Index;
 use Drupal\user\Entity\Role;
 use Symfony\Component\Yaml\Yaml;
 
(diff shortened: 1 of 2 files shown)


ronaldtebrake:
I have set the milestone to 9.0, it can be a pretty big change for sites with a lot of book pages. Feel free to use the patch if needed on 8.x.

ronaldtebrake:
🍒  picked to 8.x

PR #1908, opened 2020-07-17
Issue #3159810 by SV: Replace static classes with attributes variable for the body field of node

volodymyr-sydor-lemberg:
## Problem
*Using static classes for the body field of node prevents adding dynamic attributes by other modules (like quickedit) which could cause issues.*

## Solution
*Use the "attributes" variable.*

## Issue tracker

- https://www.drupal.org/project/social/issues/3159810
- https://getopensocial.atlassian.net/browse/YANG-3361


ribel:
🍒 Cherry-picked to 8.x-8.x

PR #1910, opened 2020-07-17
Issue #3159954 by tBKoT: Check if activity related entity exists

dmytriikaiun:
## Problem
Email notifications contain unconverted tokens

## Solution
Check if related entity of activity exist otherwise delete activity and exit from the queue worker process

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-3130
https://www.drupal.org/project/social/issues/3159954


modules/custom/activity_send/modules/activity_send_email/src/Plugin/QueueWorker/ActivitySendEmailWorker.php
```diff
@@ -57,6 +67,15 @@ public static function create(ContainerInterface $container, array $configuratio
   public function processItem($data) {
     // First make sure it's an actual Activity entity.
     if (!empty($data['entity_id']) && $activity = Activity::load($data['entity_id'])) {
+      // Check if activity related entity exist.
+      if (!$activity->getRelatedEntity()) {
```
ribel, line 71: @dmytriikaiun There is already a function for it, so I think it is better to use it. See `\Drupal\activity_creator\ActivityNotifications::deleteNotificationsbyIds`

Kingdutch: Agreed with @ribel but please note there's a bug in that function that I discovered today, I have not been able to report it on Drupal.org yet.

The fix is relatively simple but also requires an update hook to delete any possibly orphaned notifications.

dmytriikaiun: Hi @Kingdutch 
Could you describe the bug in that function?

ribel:
🍒 picked to 8.x and 9.x

PR #1912, opened 2020-07-20
Issue #3152906: Ensure likes are rendered correctly for posts and comments and fix regression on topic view mode

ronaldtebrake:
## Problem
The like widget is misplaced in a couple of view modes and on several entities.

For example:
Screen Shot 2020-07-20 at 15 48 53
Screen Shot 2020-07-20 at 15 48 12

This is due to changes in:
https://github.com/goalgorilla/open_social/commit/0626294fedde062cbc60bab76cad83c4dce7b9a9#diff-7a427f5ae42721d382705d79148f7117

Where we added the social_like_entity_extra_field_info_alter to make sure the likes popped up again, they were removed after features removal from Open Social. However that's when we noticed during tests the were issues and notices for the weight not being added in the hook
so we updated the weight in there:
https://github.com/goalgorilla/open_social/commit/b08ae44bec6e7663e8817375f16d6c36795d8ece#diff-7a427f5ae42721d382705d79148f7117

The above resulted in a scenario where we weren't able to change the weight correctly based on view modes and it's configuration in the manage display of the entities.

## Solution
We removed all of it again basically reverting back to the original issue, where likes were not rendering on posts and comments anymore.
We didn't notice back than this was due to the configuration not being changed correctly after features removal. 
So that is what we did now. We made sure using the update helper to update the configuration for the view modes.

## Issue tracker
https://www.drupal.org/project/social/issues/3152906
https://getopensocial.atlassian.net/browse/OS-217

## How to test
- [ ] Check a fresh install of Open Social and see likes are not rendered in the correct spot for comments on posts, but also for other scenario's like the featured topics in the sidebar block.
- [ ] Checkout this branch, run the updates
- [ ] See that likes are now working for:
- [ ] Posts
- [ ] Posts including a photo (post_photo)
- [ ] Comments on Posts
- [ ] Comments on Posts including a photo (post_photo)
- [ ] Not rendered anymore on any of the entities in the sidebar blocks
- [ ] Still rendered correctly on Nodes and comments on nodes
- [ ] Not rendered in the Emails or any other notification centre related items 

## Screenshots
Screen Shot 2020-07-20 at 15 30 34
Screen Shot 2020-07-20 at 15 30 18
Screen Shot 2020-07-20 at 15 30 51
Screen Shot 2020-07-20 at 15 42 37
Screen Shot 2020-07-20 at 15 31 19

## Release notes
We have made sure likes are now rendered again for posts and comments, also we have hidden them for the topics in the sidebar blocks to make it consistent again. You should be able to change this however using the view modes.


modules/social_features/social_comment/config/update/social_comment_update_8901.yml
```diff
@@ -0,0 +1,15 @@
...
+    delete:
+      hidden:
+        like_and_dislike: true
+    add:
+      content:
+        like_and_dislike:
+          region: content
+          settings: {  }
+          third_party_settings: {  }
+          weight: 2
```
Kingdutch, line 15: This config update looks good :D

modules/social_features/social_post/config/update/social_post_update_8901.yml
```diff
@@ -0,0 +1,57 @@
+core.entity_view_display.comment.post_comment.default:
+  expected_config:
+    content:
+      field_comment_body:
+        type: basic_string
```
Kingdutch, line 5: I think we should adjust this config update. We write this update to change the like_and_dislike field. However, if someone has changed the configuration for `field_comment_body` then our change for like_and_dislike also wouldn't be applied. This means the automatically generated expectation doesn't match what we're trying to do and in what scenarios we want it to happen.

Removing the expectation and the change for the field_comment_body field is probably the way to go. If we do want to update the body field we should create a separate update definition.

This applies to multiple configurations in this file.

modules/social_features/social_post/config/update/social_post_update_8901.yml
```diff
@@ -0,0 +1,57 @@
...
+core.entity_view_display.post.post.activity_comment:
+  expected_config: {  }
+  update_actions:
+    add:
+      content:
+        like_and_dislike:
+          region: content
+          settings: {  }
+          third_party_settings: {  }
+          weight: 2
```
Kingdutch, line 42: This is an example of what we probably want to adjust the other configurations to match: update the `like_and_dislike` regardless of existing configuration. :)

modules/social_features/social_topic/social_topic.install
```diff
@@ -195,3 +196,17 @@ function social_topic_update_8801() {
   // Output logged messages to related channel of update execution.
   return $updateHelper->logger()->output();
 }
+
+function social_topic_update_8901() {
```
navneet0693, line 200: No description, noticed while running the updates.

ronaldtebrake:
Also 🍒 in 8.x-8.x

PR #1914, opened 2020-07-22
Issue #3160720 Add expertise and interests as full text indexed fields on the search_user index

ronaldtebrake:
## Problem
When searching for users it might come in handy when their Expertise and Interests are also available as options to search for in the search bar. Right now sometimes they get indexed as part of the Profile rendered output. But this isn't done consistently.

## Solution
According to https://www.drupal.org/project/search_api/issues/2923870#comment-12361946
It's best to add the Terms as fulltext indexed fields. Right now we only have them as filter options (being their term ids)
So in here we need to update our search api configuration to also add the Expertise and Interests term names.

## Issue tracker
https://www.drupal.org/node/3160720

## How to test
- [ ] Login to Open Social and add some expertise and interests to users
- [ ] See that sometimes they are indexed, as part of the "Rendered output" of a users profile, but this isn't consistently (see screenshot below)
- [ ] As a LU try to find a user by its expertise and interests
- [ ] See that filtering works, but using the expertise or interest as full text doesn't
- [ ] Checkout this branch and see that you can now find users by searching for the expertise or interest taxonomy term name

## Screenshots
screenshotkaas

What you see here is the `search_api_db_social_users_text` table. 
This table contains all of the indexed words with their score for a certain user. 
I have added "Kaas" as interest and expertise to certain users. What happened is that for 2 users, the word Kaas was indexed as part of the Rendered Profile entity so they would show up if I were to search for "kaas".

After that I have added these fields to the user index 
Screen Shot 2020-07-22 at 11 38 24

And reindexing showed more results based on the new fields!
Screen Shot 2020-07-22 at 12 00 01

And now all the users can be found using the full text search.

## Release notes
We ensure expertise and interests are now consistently indexed for users so LU can find a user based on their expertise and interests.


ronaldtebrake:

PR #1821, opened 2020-05-01
Issue #3134118 by agami4: Update styles for the newest members block and profile small teasers

agami4:
## Solution
Should update newest member block

## Issue tracker
https://www.drupal.org/project/social/issues/3134118
https://getopensocial.atlassian.net/browse/YANG-2794

## How to test
- [ ] Go to the 'Group' page and you can see 'Newest member' block in the sidebar

## Screenshots
old:
old

new:
new

## Release notes
The newest member block has a new design

PR #1916, opened 2020-07-27
Issue #3134118 - Update hook for newest members block

ronaldtebrake:
## Problem
#1821 
Didnt add the correct update hook, however this was encountered after merging.

## Solution
Create a new update hook to trigger it correctly.

## Issue tracker

## How to test
*For example*
- [ ] See that the update hook passes and the screenshots in #1821 are now matching the newest members block

## Release notes
The newest members block is now also updated on existing platforms with the organization and function name.


ribel:
🍒 picked to 8.x-8.x

PR #1922, opened 2020-07-30
Issue #3162344 by robertragas: add translationcontext when retrieving…

robertragas:
… labels from the social_tagging taxonomy

## Problem
Open Social offers the possibility to insert content tags as taxonomy, these are then shown altered/shown through code in the social_tagging module.
This module does not take translations into account when getting the labels. As a result it will always show the source language aka english when your site is multilangual.

## Solution
Change the way the labels are getting retrieved in the social_tagging module to take translation context into account.

## Issue tracker
https://www.drupal.org/project/social/issues/3162344

## How to test
- [ ] Enable the Social language modules
- [ ] Enable the Social tagging module
- [ ] Make sure you go to the Open Social -> Tag settings and save the form once
- [ ] Add a second language
- [ ] Make sure you edit the content tags taxonomy category and make it translatable
- [ ] Create some taxonomy terms in this category and translate them
- [ ] Create a topic when switched to the secondary language. See that the content tags are found in the original language
- [ ] Save the content and see the sidebar block also has the original language content tags block. Same goes when trying to use the search content function and want to apply the filter.

## Release notes
We have fixed an issue where content tags taxonomy terms would show in source language when you would have multiple languages. Now it will show the correct translation according to the active language.

### DIFF
--- a/modules/social_features/social_tagging/src/SocialTaggingService.php
+++ b/modules/social_features/social_tagging/src/SocialTaggingService.php
@@ -177,7 +177,7 @@ class SocialTaggingService {
   /**
    * Returns a multilevel tree.
    *
-   * @param array $terms
+   * @param array $term_ids
    *   An array of items that are selected.
    * @param string $entity_type
    *   The entity type these tags are for.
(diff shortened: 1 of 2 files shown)


modules/social_features/social_tagging/src/SocialTaggingService.php
```diff
@@ -122,12 +133,29 @@ public function getChildren($category) {
...
+  /**
+   * Returns the translated label of the taxonomy.
+   *
+   * @param int $tid
+   *   The term id.
+   *
+   * @return string
+   *   The translated label.
+   */
+  public function getTranslatedLabel($tid) {
```
frankgraave, line 151: This method might use some more defensive check(s). 

Where's its trying to get the taxonomy term `$taxonomy_term = $this->termStorage->load($tid);` I think we might want to check if the `$tid` is actually having some value.

robertragas: Hey @frankgraave, I have been thinking about it to make it more defensive, but I don't think it's needed. You could always ask Shipsters for their second opinion.

Reason is, I call the getTranslatedLabel on 4 occasions which are only in this file itself:
1.    line 115 and 135
foreach ($this->termStorage->loadTree('social_tagging', 0, 1) as $category) {
$options[$category->tid] = $category->name;
}

- It load the terms tree, if there would be no taxonomy terms it would not even enter the foreach loop.

2. line 205 and 224
      $current_term = $this->termStorage->load($term['target_id']);
      // Must be a valid Term.
      if (!$current_term instanceof TermInterface) {
        continue;
      }
      // Get current terms parents.
      $parents = $this->termStorage->loadParents($current_term->id());
      $parent = reset($parents);
      $category = $this->getTranslatedLabel($parent->id());

      $tree[$parent->id()]['tags'][$current_term->id()] = [
        'url' => $url->toString(),
        'name' => $this->getTranslatedLabel($current_term->id()),
      ];

- If the current_term would be empty, it would return right away not even entering my function. 
- For the parent, this buildHierarchy function is only being called when the tagging split is being set in the .module file, so it always has a tree, thus parent.

modules/social_features/social_tagging/src/SocialTaggingService.php
```diff
@@ -26,20 +27,30 @@ class SocialTaggingService {
...
+   *   Injection of the languageManager.
    *
    * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
    * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
-  public function __construct(EntityTypeManagerInterface $entityTypeManager, ConfigFactoryInterface $configFactory) {
+  public function __construct(EntityTypeManagerInterface $entityTypeManager, ConfigFactoryInterface $configFactory, LanguageManagerInterface $language_manager) {
     $this->termStorage = $entityTypeManager->getStorage('taxonomy_term');
     $this->configFactory = $configFactory;
+    $this->languageManager = $language_manager;
```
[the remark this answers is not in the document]
robertragas: Reason why I call it like this, is because language_manager is the service name and this is being done throughout Drupal standards.
see
https://api.drupal.org/api/drupal/core!core.services.yml/service/language_manager/8.2.x
or for 1 of the examples under it
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21Sql%21SqlContentEntityStorage.php/8.2.x

or for the config factory
https://api.drupal.org/api/drupal/core%21modules%21rest%21src%21LinkManager%21TypeLinkManager.php/8.2.x

so I think coding standards wise, mine is actually the way it should be.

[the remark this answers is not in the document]
robertragas: So we should refactor the other two :octocat:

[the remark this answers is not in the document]
frankgraave: You are correct! Good one!

modules/social_features/social_tagging/src/SocialTaggingService.php
frankgraave: Fair enough. Then I can agree it is not needed I guess!

navneet0693:
This PR results in PHP memory outage while doing any operations via Drush, include drush cr.

modules/social_features/social_tagging/src/SocialTaggingService.php
```diff
@@ -174,7 +202,7 @@ public function buildHierarchy(array $terms, $entity_type) {
       // Get current terms parents.
       $parents = $this->termStorage->loadParents($current_term->id());
       $parent = reset($parents);
-      $category = $parent->getName();
+      $category = $this->getTranslatedLabel($parent->id());
```
navneet0693, line 205: While the function getTranslatedLabel receives tid and then it loads the term inside that function, however, so far, in the instances where this function is called, terms have been already loaded, so can we improve this function to just accepting taxonomy terms as an entity and then returning the correct language version?

robertragas:
Closing this one in favor of https://github.com/goalgorilla/open_social/pull/1950

PR #1923, opened 2020-07-30
Issue #3089852 by robertragas: featured item link markup to make the whole button clickable

robertragas:
## Problem
When you create a featured item on a landing page, you can define an url with a "read more". Only the text within this button is clickable while the whole button should work.

## Solution
Move the classes around so the button itself is styled on the link and not the surrounding div.

## Issue tracker
https://www.drupal.org/project/social/issues/3089852

## How to test
- [x]  Enable the social_landing_page module or run drush updb.
- [x]  Create a landing page
- [x]  Add a section item
- [x]  Add a featured item
- [x]  Add URL and a text
- [x]  See that the whole button is clickable now.

## Release notes
When creating a featured item for a landing page, only the link text was clickable. This has been fixed and now the complete button can be clicked.

### DIFF
--- a/modules/social_features/social_core/src/Plugin/Field/FieldFormatter/LinkClassFormatter.php
+++ b/modules/social_features/social_core/src/Plugin/Field/FieldFormatter/LinkClassFormatter.php
@@ -1,0 +1,79 @@
+<?php
+
+namespace Drupal\social_core\Plugin\Field\FieldFormatter;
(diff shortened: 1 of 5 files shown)


navneet0693:
🍒 on 8.x-8.x on 7c3debcbae8097414a398fe4494543c77a12ee06

PR #1932, opened 2020-08-05
Issue #3162597 by tBKoT: Change sorting and filtering for events. Add 'Ongoing' label

dmytriikaiun:
## Problem
Change filter and sort process for events

## Solution
Create custom filters and sorting plugins for views.

## Issue tracker
https://www.drupal.org/project/social/issues/3162597

## How to test
- [x] Go to the `/community-events` page
- [x] There must be the ongoing events first
- [x] All ongoing events should be marked as `Ongoing` (see screenshot)

## Screenshots
image


modules/social_features/social_event/social_event.install
```diff
@@ -1448,7 +1448,8 @@ function social_event_update_8805(&$sandbox) {
 
   switch ($parts[0] . '.' . $parts[1]) {
     case 'field.storage':
-      $entity_type = \Drupal::service('config.manager')->getEntityTypeIdByName($name);
```
ribel, line 1451: @dmytriikaiun looks like this change is not relevant.

modules/social_features/social_event/social_event.module
```diff
@@ -1022,3 +1053,31 @@ function _social_event_get_enrollment_status(NodeInterface $event) {
...
+
+/**
+ * Change views filters order.
+ *
+ * @param string $view_config
+ *   The views config name.
+ * @param string $display
+ *   The views display id.
+ */
+function _social_event_views_update($view_config, $display) {
```
ribel, line 1065: @dmytriikaiun Since this function is used only in `social_event.install` file I think it is better to move it there.

modules/social_features/social_event/src/Plugin/views/sort/EventDate.php
```diff
@@ -0,0 +1,24 @@
+<?php
```
ribel, line 1: @dmytriikaiun I think instead of writing new plugin, we can just change the logic of existing `event_passed_upcoming_sort` - since it will be not used.

@ronaldtebrake, @bramtenhove, or do you think we need to leave it and mark ass deprecated?

modules/social_features/social_event/social_event.module
dmytriikaiun: @ribel It's also used in the `social_group.install`

modules/social_features/social_event/src/Plugin/views/sort/EventDate.php
ronaldtebrake: @ribel 
I think this is actually a nice solution :) if people still use it in different views which we don't provide this will help them out and they can change it themselves. Marking the other as deprecated in favor of this new one is a good solution!

modules/social_features/social_event/social_event.module
ribel: @dmytriikaiun Ok, then it's fine.

modules/social_features/social_event/src/Plugin/views/sort/EventDate.php
ribel: @dmytriikaiun Good, could you then just `@deprecated` annotation to `event_passed_upcoming_sort` plugin?

dmytriikaiun: @ribel Yes, I could

ribel:
Hi @ronaldtebrake and @bramtenhove, we finished work on this feature. Let us know if you are fine to have this landed in 9.2.

ronaldtebrake:
Sounds good to me @ribel

ribel:
🍒 picked to 9.x

PR #1935, opened 2020-08-09
DS-7231 - The request membership button does not work

ronaldtebrake:
## Problem
I create a flexible group with request membership option selected. Logging in as another user, when I click on the request to join button, nothing happens. The rendering in sky worked fine, however in normal it didnt. 
What happens is when enabling social_group_request, the button should be invalidated. Before it goes to group/group_id/join while after enabling “Request to join” as join method, it should go to group/group_id/request-membership

## Solution
Ensure the group statistics from sky and the group hero have matching twig files. 
We have also added a new cache context that will cache the hero based on the join method selected for better cacheability. So the caching and rendering has improved.

## Issue tracker
https://getopensocial.atlassian.net/browse/DS-7321

## How to test
*Enable social_group_request*
- [x] As LU login add a flexible group with request to join as join method
- [x] Login as a different user and see the join button says request to join and leads to opening a modal form
- [x] Logout and as AN see that it still works as expected
- [x] No regression should be caused on other group types so try public/closed/open

## Screenshots
Screenshots of the issue
Screenshot 2020-08-05 at 12 20 55
Screenshot 2020-08-05 at 12 21 00

Screenshots after this issue has been fixed:
Screenshot 2020-08-09 at 14 31 57

## Release notes
*probably not needed because this issue only occurred with the new request to join method*
The hero should now show the correct button text and lead you to the correct steps to join a group based on the selected join method.


bramtenhove:
I noticed the button became smaller:
Screenshot 2020-08-09 at 15 57 38

Fixed in 571eaf4.

modules/social_features/social_group/social_group.module
```diff
@@ -320,6 +333,15 @@ function social_group_preprocess_group(array &$variables) {
 
   // Add group edit url for management.
   if ($group instanceof Group) {
+    // Ensure all groups get the group.type (differentiating labels in teasers
```
navneet0693, line 336: closing bracket 😅

modules/social_features/social_group/social_group.module
```diff
@@ -320,6 +333,15 @@ function social_group_preprocess_group(array &$variables) {
...
+    // hero's and full nodes.
+    // But also if the allowed join method is there, this should be added.
+    $variables['#cache']['contexts'][] = 'group.type';
```
navneet0693, line 339: Groups will have cache context "group" which will have vary per group, which is a larger aspect than group.type
Please correct me if I am wrong here.

ronaldtebrake: I don't think the group context is added to the hero render array. And it should not be necessary perse to cache this per group, but you are right that it's a larger aspect than group.type and possible solves our problems.

If you think of the data that is shown in the hero block, it's either based on the group type (closed / public have different labels) or the join methods (making sure the buttons render correctly as well as the popover) or the members (which is group content) 

so you could just vary it per group, or we could try to move this to a "smart" caching strategy where we can enhance from sharing parts of this for more than that. 

Although it also seems for the statisctics block for Sky we cache it per group and user, so maybe we should just make it consistent now. 🤔

ronaldtebrake:
https://github.com/goalgorilla/open_social/pull/1945
Closed in favor of #1945 

This is due to the twig file changes, which resulted in some more regressions. In the end we figured these werent necessary. However these changed might still be useful in the future. 
So I will close this and remove the branch but for history purposes we can still see it. :) 

Also the cache context is moved because that is still a lot of added value!

PR #1936, opened 2020-08-09
DS-7306: Event Invite button to Enroll should be removed when Invite is selected for AN

ronaldtebrake:
## Problem
When invite is selected as enrollment option, for AN the "Enroll" button is still rendered.

## Solution
Similar to Group Invite we just remove the enrollment form. However since the Form itself is overridding we will unset the element just to make sure it doesn't render. 
We have also added cache context to ensure it doesn't get cached incorrectly, however we can improve this by making a cache context which varies the cache in the future. Based on the enrollment method.

Since the setting to allow for AN enrolments is only added for direct enrolling to the Event there is no need to make it compatible yet.

## Issue tracker
None needed, its bugfixing from the Event Invite also doesn't need a backport.

## How to test
*Enable social_event_invite*
- [ ] Login as LU and create a public event with Invite only
- [ ] As a AN visit the event and see there is no Enroll button anymore.
- [ ] Login as LU and see the button is still gone, make sure however if you are invited you get to see the button and can accept / decline
- [ ] Make sure for AN when enroll to event option is selected it works

## Release notes
None needed, bugfixing day and the scope is in the Event Invite story.


bramtenhove:
If possible we should follow-up with a story to make the button show for AN and LU, but disabled. This way the user does know why he can't join out of the blue.

PR #1938, opened 2020-08-10
Issue #3164307 by tBKoT: Add the Gin Toolbar module as dependency

dmytriikaiun:
## Problem
The Gin Toolbar fails to be installed

## Solution
Add the Gin toolbar module as a dependencie

## Issue tracker
https://www.drupal.org/project/social/issues/3164307


ronaldtebrake:
Unfortunately the failing test was related to gin_toolbar.
Before the test would return true on this step: 
`And I should see the link "behatsitemanager"` because it was rendered in the `admin_toolbar` however with GIN this link is gone, and apparantly it now fails.

Screenshot 2020-08-12 at 14 29 45

Although the link is still there if you test it out locally 
Screenshot 2020-08-12 at 14 12 42

So I have now updated the test to `And I should see "behatsitemanager" in the "Main content" region` 
so it actually checks in the region, although it doesnt verify if its a link, i think the main thing we want to check is to see if we can find the author.

ronaldtebrake:
Thank you so much @ribel @dmytriikaiun works like a charm!

PR #1942, opened 2020-08-11
Issue #3164579 by SV: Add Explore activity stream block on the dashboard page

volodymyr-sydor-lemberg:
## Problem
Add Explore activity stream block on the dashboard page

## Solution
In the [related PR](https://github.com/goalgorilla/open_social/pull/1783) we already solved it but all those updates will be merged a bit later. So, proposal is to move a part of updates separately from other not related changes.

## Issue tracker
- https://www.drupal.org/project/social/issues/3164579
- https://getopensocial.atlassian.net/browse/YANG-3473

## How to test
- [ ] Using latest version of Open Social
- [ ] As a sitemanager
- [ ] Try to create a dashboard page
- [ ] And add Activity Stream: Explore Stream Block to the section.
- [ ] When saving I expect that items (like events, discussion) has the same layout as in other stream block(featured)

## Release notes
Added a new Explore Stream Block


ronaldtebrake:
Hi @ribel @volodymyr-sydor-lemberg am I correct this feature actually just creates a block for the exlpore stream instead of fixing UI issues?
Can you please update the PR to reflect what it's actually trying to solve?

ronaldtebrake:
Also in 9.x 🍒

PR #1944, opened 2020-08-11
Issue #3137445 by mdeny: Make sure postalcode can also be hidden

denysyukm:
## Problem
Make sure postalcode can also be hidden

## Solution
Implement possibility to enable/disable postalcode address's profile field 

## Issue tracker
https://www.drupal.org/project/social/issues/3137445
https://getopensocial.atlassian.net/browse/YANG-3511

## How to test
- [ ] Go to Social profile fields settings form (/admin/config/opensocial/profile-fields)
- [ ] Disable postalcode settings option
- [ ] Go to profile edit form
- [ ] The postalcode field shouldn't be displayed
- [ ] Go back to the Social profile fields settings form
- [ ] Click Flush profile data button
- [ ] Run removing data
- [ ] The postalcode field values should be removed from all user profiles

## Release notes
Implement possibility to enable/disable Postal code address's profile field.

### DIFF
--- a/modules/social_features/social_profile/modules/social_profile_fields/config/install/social_profile_fields.settings.yml
+++ b/modules/social_features/social_profile/modules/social_profile_fields/config/install/social_profile_fields.settings.yml
@@ -15,4 +15,5 @@ profile_profile_field_profile_self_introduction: true
 profile_profile_field_profile_show_email: null
 profile_address_field_city: true
 profile_address_field_address: true
+profile_address_field_postalcode: true
 nickname_unique_validation: false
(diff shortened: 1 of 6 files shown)


ronaldtebrake:
Hi @denysyukm probably the reason the tests are failing is because we need to add the default setting to our
`social_profile_fields/config/install/social_profile_fields.settings.yml` and make it `profile_address_field_postalcode: true` this is because otherwise when this module is enabled it immediately starts stripping away the fields that are not configured in the settings on install.

PR #1946, opened 2020-08-12
Issue #3164733 by SV: Allow modules to override default filter format of elements

volodymyr-sydor-lemberg:
## Problem
Allow module to override default filter format because some content type could require specific one which could cause an issue with existing implementation:

     $full_html = FilterFormat::load('full_html');
     $permission_name = $full_html->getPermissionName();
     $account = \Drupal::currentUser();
     if ($element['#type'] == 'text_format' && $element['#format'] == 'full_html' && !$account->hasPermission($permission_name)) {
     $element['#format'] = 'basic_html';
     $element['value']['#format'] = 'basic_html';
     $element['format']['format']['#default_value'] = 'basic_html';
     $element['format']['format']['#value'] = 'basic_html';
     $element['value']['#disabled'] = FALSE;
     $element['format']['format']['#access'] = FALSE;
     $element['format']['#access'] = TRUE;
     $key = array_search('filter_form_access_denied', $element['value']['#pre_render']);

      if (isset($element['value']['#pre_render'][$key])) {
        unset($element['value']['#pre_render'][$key]);
      }
    }
    elseif (
      $element['#type'] === 'text_format' &&
      $account->hasPermission($permission_name) &&
      in_array($element_id, $full_html_field_ids)
    ) {
      $element['#format'] = 'full_html';
      $element['format']['format']['#access'] = FALSE;
      $element['format']['format']['#value'] = 'full_html';
      $element['format']['help']['#access'] = FALSE;
      $element['format']['format']['#options'] = ['full_html' => 'Full HTML'];
    }

## Solution
Allow module to override default filter format in the social_core_filter_process_format()

## Issue tracker
- https://www.drupal.org/project/social/issues/3164733
- https://getopensocial.atlassian.net/browse/YANG-3404

## How to test
- [ ] Using latest version of Open Social with the example module enabled
- [ ] As a administrator
- [ ] Select that filter format as default for some field of content type
- [ ] Set permission to allow LU/SM/CM role to use that filter format
- [ ] Create new "Specific HTML" filter format
- [ ] As a sitemanager create new public node with configured new filter format
- [ ] Login as co-author (LU user which have access to edit node)
- [ ] Try to edit and save changes
- [ ] Expected result: text area field still have the same specific format
- [ ] Actual result: it was overridden with basic_html (or full_html if user have access to that format).

## Release notes
Allow module to override default filter format and prevent issue with editing node with specific filter format

## Change Record
Provide a method to alter the default filter format for a content type.


modules/social_features/social_core/social_core.module
```diff
@@ -512,10 +512,23 @@ function social_core_filter_process_format($element) {
...
-  if ($element['#type'] == 'text_format' && $element['#format'] == 'full_html' && !$account->hasPermission($permission_name)) {
+  // Default filter format for text area fields.
+  $social_filter_format = 'full_html';
+
+  // Allow module to override default filter format because some content type
+  // could require specific one.
+  Drupal::moduleHandler()->alter('social_filter_format_default', $social_filter_format);
+
+  /* @var \Drupal\filter\Entity\FilterFormat $social_filter_format */
+  $social_filter_format = FilterFormat::load($social_filter_format);
```
ribel, line 523: I think we need to have an additional check here if filter format was loaded, to prevent fatal errors if wrong value was provided from hook.

ribel (changes requested):
In general, all looks good, but I think we need to add one more check here.
Also, I was not able to test it for LU on test env, but I think this might be an issue related to the 
 module implementation.
Also, this needs to be checked wit Discussion module.

modules/social_features/social_core/social_core.module
volodymyr-sydor-lemberg: Added checks bellow

ronaldtebrake:
@ribel I'm happy for this to be added :) thanks for the work @volodymyr-sydor-lemberg as well :)

ribel:
🍒 picked to 9.x

PR #1950, opened 2020-08-13
Issue #3162344, #3165009 by robertragas, navneet0693: Taxonomy terms should appear in site's current language.

navneet0693:
_Original PR: https://github.com/goalgorilla/open_social/pull/1922_

## Problem
Open Social offers the possibility to insert content tags as taxonomy, these are then shown altered/shown through code in the social_tagging module.

This module does not take translations into account when getting the labels. As a result, it will always show the source language aka English when your site is multilingual.

## Solution

1. Include the Drupal core patch from https://www.drupal.org/project/drupal/issues/3123561#comment-13786738 because we use loadTree in https://github.com/goalgorilla/open_social/blob/8.x-9.x/modules/social_features/social_tagging/src/SocialTaggingService.php
2. Make changes to https://github.com/goalgorilla/open_social/blob/8.x-9.x/modules/social_features/social_tagging/src/SocialTaggingService.php#L160 got use `getTranslation()` function of TermStorage. 

## Issue tracker
https://www.drupal.org/project/social/issues/3165009
https://www.drupal.org/project/social/issues/3162344

## How to test
- [ ] Enable the Social language modules
- [ ] Enable the Social tagging module
- [ ] Make sure you go to the Open Social -> Tag settings and save the form once
- [ ] Add a second language
- [ ] Make sure you edit the content tags taxonomy category and make it translatable
- [ ] Create some taxonomy terms in this category and translate them
- [ ] Create a topic when switched to the secondary language. See that the content tags are found in the right language
- [ ] Save the content and see the sidebar block also has the right language content tags block.
- [ ] Trying to use the search functions. Tags should appear in the right language.

## Screenshots
N.A

## Release notes
We have fixed an issue where content tags taxonomy terms would show in source language when you would have multiple languages. Now it will show the correct translation according to the active language.

## Change Record
N.A

## Translations
N.A


robertragas (approved):

robertragas:
Currently on hold, because we needed to set the correct language for certain taxonomy, otherwise this fix can crash WSOD the platform. Until a valid solution is found for this, we cannot release this to clients where we don't have direct control over and might experience the same.

It could be a change record, but I think a better solution should be possible here.

robertragas (approved):

navneet0693:
🍒 picked to 8.x-8.x and 8.x-9.x

PR #1954, opened 2020-08-16
Issue #3165476 by tBKoT: Add lazy_builder for custom content block

dmytriikaiun:
## Problem
Enable `#lazy_builder` for activity stream blocks and custom content block

## Solution
1. Add service for the `#lazy_builder` to render views.
2. Preprocess views block to override render array for using the `#lazy_builder`
3. Override the `ContentBuilder` to use the `#lazy_builder`

## Issue tracker
https://www.drupal.org/project/social/issues/3165476

## How to test
- [ ] Enable the `big_pipe` module.

**Activity stream**
- [ ] Add any stream block to dashboard
- [ ] First, you should see the content without a lazy builder
- [ ] Wait till JS load stream block

**Content block**
- [ ] Add the `Custom content list` block to landing page
- [ ] First, you should see the content without a lazy builder
- [ ] Wait till JS load content list block

## Screenshots
N/A

## Release notes
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_activity/social_activity.module
+++ b/modules/social_features/social_activity/social_activity.module
@@ -240,21 +240,34 @@ function social_activity_views_post_build(ViewExecutable $view) {
  * Implements hook_preprocess_HOOK().
  */
 function social_activity_preprocess_block(&$variables) {
+  $route_match = \Drupal::routeMatch();
   $allowed_routes = [
     'entity.node.canonical',
     'layout_builder.overrides.node.view',
   ];
-  if (($node = \Drupal::routeMatch()->getParameter('node')) &&
-    $node instanceof NodeInterface) {
-    if (in_array(\Drupal::routeMatch()->getRouteName(), $allowed_routes) &&
-      ($node->bundle() === 'dashboard' ||
-        $node->bundle() === 'landing_page') &&
+  $allowed_bundles = [
+    'dashboard',
+    'landing_page',
+  ];
+  $allowed_content_names = [
+    'activity_stream',
+    'community_activities',
+    'activity_stream_profile',
+    'activity_stream_group',
+  ];
+
+  if (($node = $route_match->getParameter('node')) && $node instanceof NodeInterface) {
+    // Use lazy builder for stream blocks only if:
+    // 1. It on the node or landing builder page.
+    // 2. Node is a Dashboard or a Landing page
+    // 3. This block is from views.
+    // 4. View one of the stream views.
+    if (
+      in_array($route_match->getRouteName(), $allowed_routes) &&
+      in_array($node->bundle(), $allowed_bundles) &&
       $variables['base_plugin_id'] === 'views_block' &&
       isset($variables['content']['#name']) &&
-      ($variables['content']['#name'] === 'activity_stream' ||
-        $variables['content']['#name'] === 'community_activities' ||
-        $variables['content']['#name'] === 'activity_stream_profile' ||
-        $variables['content']['#name'] === 'activity_stream_group')
+      in_array($variables['content']['#name'], $allowed_content_names)
     ) {
       if ($node->bundle() === 'dashboard') {
         $variables['attributes']['class'][] = 'block-dashboard-activity-stream card';
--- a/modules/social_features/social_activity/social_activity.module
+++ b/modules/social_features/social_activity/social_activity.module
@@ -259,6 +259,8 @@ function social_activity_preprocess_block(&$variables) {
       if ($node->bundle() === 'dashboard') {
         $variables['attributes']['class'][] = 'block-dashboard-activity-stream card';
       }
+      /** @var \Drupal\views\ViewExecutable $view */
+      $view = $variables['content']['#view'];
       $variables['content'] = [
         '#create_placeholder' => TRUE,
         '#lazy_builder' => [
@@ -267,6 +269,7 @@ function social_activity_preprocess_block(&$variables) {
             $variables['content']['#name'],
             $variables['content']['#display_id'],
             $node->bundle(),
+            $view->getItemsPerPage(),
           ],
         ],
       ];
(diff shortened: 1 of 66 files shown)


modules/social_features/social_activity/social_activity.module
```diff
@@ -234,3 +235,44 @@ function social_activity_views_post_build(ViewExecutable $view) {
     }
   }
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function social_activity_preprocess_block(&$variables) {
```
rollki, line 242: I think it would be better to stay consistent with our current code style about new lines and etc. And also do not duplicate of `$variables['content']['#name']`, `$node->bundle()`.
Here's my proposal:
```
function social_activity_preprocess_block(&$variables) {
  $route_match = \Drupal::routeMatch();
  $allowed_routes = [
    'entity.node.canonical',
    'layout_builder.overrides.node.view',
  ];
  $allowed_bundles = [
    'dashboard',
    'landing_page',
  ];
  $allowed_content_names = [
    'activity_stream',
    'community_activities',
    'activity_stream_profile',
    'activity_stream_group',
  ];

  if (($node = $route_match->getParameter('node')) && $node instanceof NodeInterface) {
    if (
      in_array($route_match->getRouteName(), $allowed_routes) &&
      in_array($node->bundle(), $allowed_bundles) &&
      $variables['base_plugin_id'] === 'views_block' &&
      isset($variables['content']['#name']) &&
      in_array($variables['content']['#name'], $allowed_content_names)
    ) {
      if ($node->bundle() === 'dashboard') {
        $variables['attributes']['class'][] = 'block-dashboard-activity-stream card';
      }
      /** @var \Drupal\views\ViewExecutable $view */
      $view = $variables['content']['#view'];
      $variables['content'] = [
        '#create_placeholder' => TRUE,
        '#lazy_builder' => [
          'social_activity.lazy_builder:viewsLazyBuild',
          [
            $variables['content']['#name'],
            $variables['content']['#display_id'],
            $node->bundle(),
            $view->getItemsPerPage(),
          ],
        ],
      ];
    }
  }
}
```

Also, I think it would be good to describe conditions a bit.

rollki: Also, after the changes, the code will become more readable.

modules/social_features/social_activity/src/SocialActivityLazyBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+   *   Node bundle.
+   * @param int $item_per_page
+   *   Items to display.
+   *
+   * @return array|null
+   *   Render array.
+   */
+  public function viewsLazyBuild($view_id, $display_id, $node_type, $item_per_page) {
+    // Get view.
+    $view_entity = $this->entityTypeManager
```
rollki, line 59: I think similar things can be write in one line
$view_entity = $this->entityTypeManager->getStorage('view')->load($view_id);

modules/social_features/social_content_block/src/ContentBuilder.php
```diff
@@ -84,16 +84,22 @@ public function __construct(
...
    * @return array
    *   Returns the entities found.
    *
    * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginException
    * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
-  protected function getEntities(BlockContentInterface $block_content) {
+  public function getEntities($block_id, $entity_bundle) {
+    $block_content = $this->entityTypeManager->getStorage('block_content')
```
rollki, line 100: The same: I think similar things can be written in one line

rollki (changes requested):
The changes look good, but I think we need to do some improvements to make the code more readable.
We can discuss these.

rollki (approved):
The changes look good to me.

ribel:
@dmytriikaiun please update PR Description according to the template.

modules/social_features/social_group/modules/social_group_request/social_group_request.module
```diff
@@ -187,8 +187,10 @@ function social_group_request_preprocess_group(&$variables) {
     return;
   }
 
+  $variables['closed_group'] = TRUE;
```
ribel, line 190: @dmytriikaiun after switching to 10.x branch I see this not relevant change - please fix

ribel (approved):

modules/social_features/social_group/modules/social_group_request/social_group_request.module
```diff
@@ -187,7 +187,6 @@ function social_group_request_preprocess_group(&$variables) {
     return;
   }
 
-  $variables['closed_group'] = TRUE;
```
ronaldtebrake, line 190: This change probably shouldnt be reverted in here?

modules/social_features/social_activity/social_activity.module
```diff
@@ -234,3 +235,57 @@ function social_activity_views_post_build(ViewExecutable $view) {
...
```
ronaldtebrake, line 242: Just a question out of curiosity, since we have views in place. Doesn't views allow us to use the lazy building already by default? It seems we are adding a hell of a lot of alteration just for this, while i assume this core behavior to work and I don't see any explanation in code why its necessary to use this preprocess. 
Also looking at the `viewsLazyBuild` the only thing we do there are things views should take care of. 

Did you guys check that scenario if core isn't able to do this easier for us?

ronaldtebrake (changes requested):

modules/social_features/social_activity/social_activity.module
dmytriikaiun: Hi, @ronaldtebrake we investigated it with a team before and unfortunately Drupal doesn't have a solution for that.

ronaldtebrake: Hi @dmytriikaiun alright, thanks for the explanation. While im thinking about it, I think this is actually very valuable to add to core in this case. So I can agree that this is necessary for now and we might be able to port this to core inside the rendering of views in a follow-up. Let's make core better with your awesome work :D !
So for me it's good to merge! :)

dmytriikaiun:
@ronaldtebrake could you check again?

ronaldtebrake (approved):

ribel:
🍒 picked to 9.x

PR #1956, opened 2020-08-19
Fix backport of improvements on statistics block tags and caching to 8.x

ribel:
## Problem
After update to latest 8.x-8.x, there is an error with Group Statistic block:
`Error: Class 'Drupal\social_group\Plugin\Block\Cache' not found in Drupal\social_group\Plugin\Block\GroupStatisticBlock->getCacheTags() (line 121 of profiles/contrib/social/modules/social_features/social_group/src/Plugin/Block/GroupStatisticBlock.php).`

## Solution
Make sure this file is the same as in 9.x and 10.x
See: https://github.com/goalgorilla/open_social/commit/fd1289a2b518c19625bcbeca3e4dc99bfa75fc61

## How to test
*For example*
- [ ] Using version 8.x-8.x of Open Social with the sky theme enabled
- [ ] As a user go to the group page
- [ ] You should not see fatal errors, and see a correct statistic block

## Screenshots

### DIFF
--- a/modules/social_features/social_group/src/Plugin/Block/GroupStatisticBlock.php
+++ b/modules/social_features/social_group/src/Plugin/Block/GroupStatisticBlock.php
@@ -4,6 +4,7 @@ namespace Drupal\social_group\Plugin\Block;
 
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Cache\Cache;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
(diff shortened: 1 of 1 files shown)


ribel:
@ronaldtebrake Looks like some fixes were not backported correctly to 8.x, and now we have issue described above. 
I was trying to fix it directly from 8.x-8.x, but now see some behat tests failing. So not sure how to deal with that. Maybe you can help?

ronaldtebrake:
Hi @ribel,

These tests are all written with updated labels for 9.x + so thats understandable why it fails. 
Looking at the changes I don't mind this being backported, shouldn't be a big deal and if this fixes it I'm happy if this gets a backport!

PR #1965, opened 2020-08-31
Issue #3162557 by tBKoT, Kingdutch: Update delete notification condition

dmytriikaiun:
## Problem
There is an error in \Drupal\activity_creator\ActivityNotifications::deleteNotificationsByIds. On line 260 of ActivityNotifications.php the line ->condition('aid', $activity_ids) is used, where $activity_ids is an array. However, the default comparison operator for condition is = which causes issues in the query, resulting in unremoved activities.

## Solution
- Ensure the correct operator is used in the condition, in this case, add "IN" as the third argument. Similar to what's done on line 196 of the same file
- Create an update hook that cleans up any notifications for whom the user or associated entity no longer exists.

## Issue tracker
https://www.drupal.org/project/social/issues/3162557

### DIFF
--- a/modules/custom/activity_creator/activity_creator.install
+++ b/modules/custom/activity_creator/activity_creator.install
@@ -5,6 +5,7 @@
  * Installation code for the activity_creator module.
  */
 
+use Drupal\activity_creator\ActivityInterface;
 use Drupal\Core\Database\Database;
 
 /**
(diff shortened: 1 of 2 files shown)


ribel (approved):
Looks good to me.

ribel:
Hey @ronaldtebrake, 
We had issues when users received notification about entities that were already deleted.
This PR should prevent this. I think @Kingdutch suggested solution, which @dmytriikaiun implemented.
Could you check if we can include this in upcoming 8.5 release?

ronaldtebrake:
Hi @ribel, nice work looks good to me also in 8.5! :)

ribel:
🍒 picked to 8.x and 9.x

PR #1966, opened 2020-08-31
Issue #3168287 by robertragas: replace addmember query with mention v…

robertragas:
…ersion to keep consistent

## Problem
When trying to find members on their full name to add them to the group or events it will find no results.

## Solution
The query of finding members does not take the full name into account, while the mention query does. To keep code simple and clean, the adding members query should use the same one as the mentioning. So integrate that query, so we can remove double and obsolete code.

## Issue tracker
https://www.drupal.org/project/social/issues/3168287

## How to test
- [x] Add a new member and name him "John Doe" (first and lastname)
- [x] Create a group or event and try to add members
- [x] Search for "John" you will get results
- [x] Search for "Doe" you will get results
- [x] Search for "Jonh Doe", no results
- [x] Check out to the branch and check again

## Release notes
When adding members in groups or events we fixed a problem where a member could not be found when searching for the full name. This was already possible for mentions, but not adding members. This has been equalized.

### DIFF
--- a/modules/social_features/social_profile/src/Plugin/EntityReferenceSelection/UserSelection.php
+++ b/modules/social_features/social_profile/src/Plugin/EntityReferenceSelection/UserSelection.php
@@ -32,12 +32,44 @@ class UserSelection extends UserSelectionBase {
   /**
    * {@inheritdoc}
    */
-  protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS') {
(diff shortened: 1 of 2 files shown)


sjoerdvandervis:
This works great! If we manage to get the tests green, it is ready to be merged.

navneet0693 (approved):

PR #1967, opened 2020-08-31
Fix logic errors due to invalid tokens in cache_contexts

vredeling:
## Problem
The social_tagging module sets a Cache Context that contains an invalid token: "config". This should be a cache tag instead. It leads to countless errors in the logs saying:

> `assert(): assert(\Drupal::service('cache_contexts_manager')->assertValidTokens($cache_contexts)) failed Cache.php:33`

## Solution
Call addCacheTags() instead of addCacheContexts()


navneet0693 (approved):

navneet0693:
@vredeling Thanks for spotting this and raising a PR. Would you be able to merge back 10.0.x branch in your branch so that the tests can pass?

vredeling:
@navneet0693 I've created a new PR, because my own 10.0.x has diverged too much from goalgorilla/10.0.x
Please review #2080 instead.

PR #1304, opened 2019-04-05
Issue #3046160 by ribel: As a LU I want to see more information about the item in email notifications I get

ribel:
## Problem
Currently, in email notifications, we just send a short message that someone commented on your comment, or replied to you.
We want to add more context to these emails.
UPD:
Currently, in email notifications, we just send a short message that someone created a new post.
We want to add more context to these emails.

## Solution
Improve email notifications for new comments and replies like in the example below.
UPD: 
Improve email notifications for new posts on profile like in the example below. 
Improve email notifications for new posts, topics or events in a joined group.
Please notice that we display the full text only if the message is shorter than 280 characters, otherwise we display the first 280 characters of the text with an ellipsis at the end.

Create few comment:
email-comment

Create new post in another user's profile or in a group:
Create post

Create new topic or event in a group: 
Create topic

## Issue tracker
- https://www.drupal.org/project/social/issues/3046160
- https://jira.goalgorilla.com/browse/ECI-871
- https://jira.goalgorilla.com/browse/ECI-875

## How to test
- [ ] Checkout to this branch and revert features

Test comments email notifications
- [ ] Create few comments and replies on nodes, posts
- [ ] See email notification in a new style with context information

Test posts email notifications
- [ ] Create new post in another user's profile 
- [ ] Create new post in a group 
- [ ] Create new topic or event in a group 
- [ ] See email notification in a new style with context information

## Release notes
Improved email notifications for new posts, comments and replies so user know if he want to go back to the site or not.

PR #1968, opened 2020-09-02
YANG-3478: Allow to use social activity token in other content types, like Discussion

ribel:
## Problem
The e-mail notification for a reply to a comment on a discussion is not working properly. It doesn't contain a snippet of the text/comment.

## Solution
Remove hardcoded limitation only for topics and events. We don't need it here.

## Issue tracker

- https://getopensocial.atlassian.net/browse/YANG-3478
- https://github.com/goalgorilla/open_social/pull/1304

## How to test
- [ ]  Create few comments and replies on discussions
- [ ] Email notification should contain a snippet of the text/comment, the same as Topics or Events

### DIFF
--- a/modules/social_features/social_activity/social_activity.tokens.inc
+++ b/modules/social_features/social_activity/social_activity.tokens.inc
@@ -8,6 +8,7 @@
 use Drupal\Core\Render\BubbleableMetadata;
 use Drupal\Core\Url;
 use Drupal\group\Entity\GroupContentInterface;
+use Drupal\node\NodeInterface;
 
 /**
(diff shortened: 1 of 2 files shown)


ribel:
Hey, @ronaldtebrake, we are using this as a patch at the moment, could you check if we can merge it?

ronaldtebrake:
Hi @ribel, seems like a very good addition, and agreed its nice to have it backported as well!

ronaldtebrake (approved):

ribel:
🍒 picked to 8.x and 9.x

PR #1970, opened 2020-09-03
Issue #3169027 - Add to calendar button

dmytriikaiun:
## Problem
As a LU I want to be able to add an event from OS to my calendar

## Solution
- Create an extra field with calendar dropdown
- Create new plugin type for calendar links

## Issue tracker
https://www.drupal.org/project/social/issues/3169027
https://getopensocial.atlassian.net/browse/YANG-1902

## How to test
- [ ] Go to any event page as LU
- [ ] You should see the Add to Calendar dropdown
- [ ] Every link in the dropdown should create an event in your calendar

### Sky theme
image
### Default theme
image
### Config page
image

## Release notes
SM now can enable/disable the `Add to calendar` button and decide which calendars users can use (see screenshot).

## Change Record
N/A

## Translations
N/A


ribel (approved):
Code looks good to me.

@bramtenhove, @ronaldtebrake, @robertragas 
I know that we usually don't want to add parts related to non-distro extensions to distro code, but I think this is the easiest way to support **Add to calendar** button. Or you have any other ideas about implementation?

ronaldtebrake:
Hi @ribel, I agree this is the easiest way, but this also shows that the template isn't using rendering of fields correctly.
I really want us to continue to work with the field ui, so people can manage the fields through the UI themselves. So adding modules and fields becomes way easier, for everybody. For the community, for extensions, literally everybody. 

If we do it correctly we dont need any update in the distro if you want to add a field, hide it's label etc. 
We have already started working on that a bit with the field labels before but if we continu like this we are just adding legacy. 

Of course that means more work, but we will benefit from this in the long run. 
I'll discuss internally how to proceed with this, perhaps hardening / refactoring could benefit from this!

bramtenhove:
@ribel, it would be interesting to know what the effort is of the changes @ronaldtebrake is proposing.

His solution is more scalable for the future, also in the case we'll move away to a different front-end theme for example. Or even with the difference of Sky vs. Socialblue, etc.

ribel:
@dmytriikaiun could you check comments above and think about some different implementation including some code refactoring maybe to make it more scalable, but also backwards compatible?

dmytriikaiun:
Hi @ronaldtebrake @bramtenhove! One of the easiest ways is to replace the template in the module, but we need to maintain it every time something changes in the distro. Another way is to create custom variables on preprocessing, but it still needs modifications in the distro templates. Maybe you have suggestions on how to do it?

ribel:
@ronaldtebrake @dmytriikaiun As we discussed we need to update this PR and add this new optional add to calendar module to the distro. So I am moving it to milestone 9.3.

ribel (approved):
Looks good to me

ribel:
@dmytriikaiun just please fix failing coding standards

ronaldtebrake:
Thanks for your great work, all of you!
I'm not sure how you are on time, the only thing I would suggest to still see improved here is:
1. Enable this module by default.
2. Add a settings form for SM, where an SM can enable the individual plugins. I can imagine for a lot of our SaaS customers they don't necesarilly want all of these calendar plugins, maybe one or 2 would suffice. Also I can imagine there being a couple of more calendar integrations we would like to make in the future, you can imagine having 10 items in the list not really being to user friendly. 
3. By default have none of the plugins enabled.
4. We can add a link to the Dashboard for SM to enable it and add it to our manuals.

Shouldnt take to much time, but curious if you see this still fit or if we need to consider doing it separately.
@ribel @dmytriikaiun

dmytriikaiun:
@ronaldtebrake check the last changes, please

ronaldtebrake:
Thank you @dmytriikaiun will have a look!

ronaldtebrake:
Great work @dmytriikaiun i'll merge it and backport it to 9.3 :) ! Travis is failing due to another issue int he main branch which i will try and fix later.

ronaldtebrake:
Cherry-picked to 9.x 🍒

PR #1977, opened 2020-09-09
Issue #3169982 by chmez: Make the redirect subscriber service of the flexible group module more expandable

ghost:
## Problem
When need to extend the redirect subscriber service of the **Social Flexible Group** module in the additional module then must be copy-pasted the code of class of that service.

## Solution
Move code for creating a redirect to a separate method.

## Issue tracker
- https://www.drupal.org/project/social/issues/3169982
- https://getopensocial.atlassian.net/browse/YANG-3641

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -1244,9 +1244,12 @@ function _social_group_get_member_profile(GroupContent $group_content) {
 }
 
 /**
- * Get current Group entity from the route.
+ * Gets current Group entity from the route.
  *
- * @return \Drupal\group\Entity\GroupInterface
+ * @param \Drupal\node\NodeInterface|null $node
+ *   (optional) The node object or NULL.
+ *
+ * @return \Drupal\group\Entity\GroupInterface|null
  *   Returns the group object.
  */
 function _social_group_get_current_group($node = NULL) {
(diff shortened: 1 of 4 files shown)


ribel (approved):

modules/social_features/social_group/social_group.module
```diff
@@ -1244,9 +1244,12 @@ function _social_group_get_member_profile(GroupContent $group_content) {
 }
 
 /**
- * Get current Group entity from the route.
```
rollki, line 1247: I think the previous state was better "Get" instead of "Gets"

modules/social_features/social_group/social_group.module
```diff
@@ -1277,7 +1280,7 @@ function _social_group_get_current_group($node = NULL) {
   }
   else {
     $node = is_object($node) ? $node : \Drupal::routeMatch()->getParameter('node');
-    if (is_object($node) && !is_null($node)) {
+    if (is_object($node)) {
```
rollki, line 1283: I think it's better to add condition `if($node instanceof Nodeinterface)` instead of `is_object `

rollki (commented):
All look good, but please check it only one comment from me.

rollki (approved):

ribel:
🍒 picked to 9.x

PR #1978, opened 2020-09-10
Issue #3162557: Increase number of activities per one batch operation

ribel:
## Problem
In PR #1956 we introduced batch update of all activity entities in `activity_creator_update_8802()`.
During update to latest 8.x-8.x on one of project with around 30000 activities, I received memory outage during this batch, both locally and on Platform.sh.

## Solution
I propose to increase the number of items from 10 to at least 100 to have this batch finished.
When I tried this locally all was good, but on Platform.sh it was not enough even when I increased to 200, but I am also not sure to what value it's good to increase.

## Issue tracker

- https://github.com/goalgorilla/open_social/pull/1965

## How to test
- [ ] Using latest version 8.x-8.x of Open Social
- [ ] Run update hook
- [ ] You should have activity_creator_update_8802 run without error

## Screenshots
N.A

## Release notes
N.A

## Change Record
N.A

## Translations
N.A

### DIFF
--- a/modules/custom/activity_creator/activity_creator.install
+++ b/modules/custom/activity_creator/activity_creator.install
@@ -7,6 +7,7 @@
 
 use Drupal\activity_creator\ActivityInterface;
 use Drupal\Core\Database\Database;
+use Drupal\Core\Site\Settings;
 
 /**
  * Implements hook_schema().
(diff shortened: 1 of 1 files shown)


ribel:
@dmytriikaiun, I am wondering if this `activity_creator_update_8802()` is actually needed?
Can we just ignore all old activities, and just have fixed `IN` in `ActivityNotifications.php` for future cases?

ronaldtebrake:
Hi @ribel 
I can imagine the struggle, what I would propose is to use settings for this.

`use Drupal\Core\Site\Settings;
$activities_per_batch = Settings::get('activity_update_batch_size', 50);`

This allows us and everybody to change this, also gives us the easy access to override this using platformsh variables on a per environment base.

ronaldtebrake:
Cherry picked to 8.x and 9.x

PR #1981, opened 2020-09-14
Issue #3170795 by rolki: Hide entities from custom content list block if user do not have access

rollki:
## Problem
Users that don’t have access to entities can see these entities by custom content list block. Behavior should be the same as in drupal core (if do not access then hide the entity include teasers)

## Solution
Add access check condition

## Issue tracker
https://www.drupal.org/project/social/issues/3170795

## How to test
- Create a custom content list block /block/add/custom_content_list
- Select some entity type to which for example authenticated user does not have access
- Go to the page with this block under authenticated user
- You can see entities

## Screenshots
image


ribel (approved):
Looks good to me

PR #1988, opened 2020-09-21
Issue #3171563 by navneet0693: Moved logic from activity_creator_update_8802 to hook_post_update_name and refactored.

navneet0693:
## Problem
I started working on the issue https://www.drupal.org/project/social/issues/3171563 and I created a solution here: https://github.com/goalgorilla/open_social/pull/1986. But even this solution was leading to memory outage no matter if the batch size is 10 or 10000. So, I decided to take a different approach to solve the issue.

## Solution
1. Use database queries instead of Entity APIs to solve the problem reported here: https://www.drupal.org/project/social/issues/3162557
2. Split the updates into two updates.
3. First update hook should delete non-existent Activity IDs from **activitiy_notification_status** table.
4. Second update hook should delete the Activity entities which doesn't have any value in "field_activity_entity".

## Issue tracker
https://www.drupal.org/project/social/issues/3171563

## How to test
- [ ] Using version <8.5 of Open Social.
- [ ] Upgrade to 8.5.
- [ ] Check for no memory outage.

## Screenshots
N.A

## Release notes
In a past release we made a update hook to delete old activities that had no more related entities. This update hook was not working properly for bigger websites, so we changed the logic to make it less performance heavy.

## Change Record
N.A

## Translations
N.A

### DIFF
--- a/modules/custom/activity_creator/activity_creator.install
+++ b/modules/custom/activity_creator/activity_creator.install
@@ -5,9 +5,7 @@
  * Installation code for the activity_creator module.
  */
 
-use Drupal\activity_creator\ActivityInterface;
 use Drupal\Core\Database\Database;
-use Drupal\Core\Site\Settings;
 
 /**
  * Implements hook_schema().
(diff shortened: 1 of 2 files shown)


robertragas:
@navneet0693 
1 question about the change, not sure if i'm misreading it, but in the old situation when the activity does not exist, it would delete the activity_notification_status, like you have in your first update hook

In the second part it would delete the whole activity is the related_entity does not exist, in the old one it would again also delete the activity_notification_status, but in yours it only deletes the activity. Is this not needed anymore?

navneet0693:
> @navneet0693
> 1 question about the change, not sure if i'm misreading it, but in the old situation when the activity does not exist, it would delete the activity_notification_status, like you have in your first update hook
> 
> In the second part it would delete the whole activity is the related_entity does not exist, in the old one it would again also delete the activity_notification_status, but in yours it only deletes the activity. Is this not needed anymore?

Thanks for pointing out Robert. I have fixed it last commit :-)

robertragas (approved):

robertragas:
You made the last change and it goes a lot faster. Still some timeouts on Pantheon, but that's due to their plan being to small. 
Will merge it.

PR #1993, opened 2020-09-22
Issue #3171792 by agami4: Update styles for the show/hide search popup

agami4:
## Problem
The form with ID search-content-form is used to show the search overlay. However, even when this overlay is not visible the form still captures focus.

## Solution
Adjust the `search-take-over` class to use `display:none` and add `display: initial` to the `.mode-search .search-take-over ` selector. The search icon itself should be keyboard and screenreader accessible to make the search overlay visible.

## Issue tracker
https://www.drupal.org/project/social/issues/3171792

## How to test
*For example*
- [ ] Open a landing page with the stream.
- [ ] Tab through the page.

## Screenshots
search-form-popup

## Release notes
The search form is no longer in the tab order when it's not visible. No visual changes.

### DIFF
--- a/themes/socialbase/assets/css/navbar.css
+++ b/themes/socialbase/assets/css/navbar.css
@@ -346,6 +346,7 @@
     display: none;
   }
   .search-take-over {
+    display: none;
     height: 50px;
     position: fixed;
     top: 20%;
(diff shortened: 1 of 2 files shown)


Kingdutch (approved):

Kingdutch:
Cherry-pick to 8.x and 9.x

PR #1995, opened 2020-09-22
Issue #3172477 by chmez, agami4, rolki: As a AN I want to have a Albums so I can view all media items in one place

OH17L:
## Problem
As a AN I want to have a Albums so I can view all media items in one place.

## Solution
- It should be possible to place images in albums
  - Only the creator of the Album can upload images to the album
- AN can open the images and it will show in a lightbox
  - should be possible to click through all images of the album
  - the post on the right should change appropriately
  - should be possible to see and add comments there
  - when adding a comment it should not reload
- Pictures can be removed by the creator of the album
  - Creator can remove the picture from the post
  - Creator can delete the whole post
  - Creator can remove the image from the album
  - CM+ can also delete those images (as in deleting a whole post )
  - If user should delete the last image the album stays
- The Album should have a default image if no image is uploaded
- Pictures can be placed in albums via
  - A post on a stream
  - The album itself
    - If placed via the album directly it should be still a post (with description and the possibility to comment)
    - An activity item is created for that
- Albums should contain
  - Title
  - Visibility
  - Creator
  - Creation date
  - Referenced images from posts
  - number of images (on the teaser and detail page
  - add image button
  - navigational element to go back to the Album overview
- There should be an overview page for all Albums of one user
  - should be a separate tab on a profile
  - each album should have a teaser
- The album should be a new content type
- SM can enable and disable the tab
- should be a new module: social_album

## Issue tracker
- https://www.drupal.org/project/social/issues/3172477
- https://getopensocial.atlassian.net/browse/YANG-3686

## How to test
- [ ] Install the "Social Album" module.
- [ ] When you log in as "Test User" and open the user menu then you should see the "My albums" item.
- [ ] When you click on that menu item then you should be redirected to the "Albums" tab of the user profile page.
- [ ] You should see the "Create new album" button but when you navigate to the same tab on the profile page of another user then you shouldn't see this button.
- [ ] When you open the home page or the user stream page and add at least one image to the post creation form then you should see the "Add image(s) to an album" field which appears in this form.
- [ ] When you select the "Create new album" item then you should see the "Album name" field.
- [ ] When you fill "My first album" in the "Album name" and add some text to the post text field and press "Post" then you should see "Images posted in album My first album" in the teaser of the created post.
- [ ] When you open the "My albums" page from the user menu then you should see a tile teaser for the "My first album" album.
- [ ] You should see the first post image in the teaser.
- [ ] When you click on the image or album name or the "View album" link then you should be redirected to the album node overview page.
- [ ] You should see the "Test User's Albums" link over the page title in the hero section.
- [ ] You should see the number of images and the "Add images" button between the hero section and the content section.
- [ ] You should see all images from the created post in the content section.
- [ ] You should see the votes widget in the main content section and you should see the comment section under the content section.
- [ ] When you click "Test User's Albums" then you should be redirected to the "Albums" page.
- [ ] When you log in as SM and navigate to the "Configuration > Open Social Settings" page from the administrative menu then you should see the "Album settings" link.
- [ ] When you click "Album settings" then you should be on the settings page and you should see the "Active" checkbox.
- [ ] Uncheck the "Active" checkbox and press "Save configuration".
- [ ] When you back to the home page and open the user menu then you shouldn't see the "My albums" item.

## Release notes
Implement functionality for creating user photo albums from the post creation form (community stream or user stream) or you can do it from a new tab on the user profile page.


modules/social_features/social_album/social_album.libraries.yml
```diff
@@ -0,0 +1,21 @@
...
+  css:
+    theme:
+      assets/css/social-album--post.css: {}
+
+social_album_block:
+  css:
+    theme:
+      assets/css/social-album--block.css: {}
+
+image:
```
rollki, line 16: Maybe it's better to say `social_album_image` to be consistent.

modules/social_features/social_album/social_album.module
```diff
@@ -0,0 +1,351 @@
...
+  }
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function social_album_preprocess_block__social_post(&$variables) {
+  if (
+    \Drupal::routeMatch()->getRouteName() === 'social_album.post' &&
+    ($content = \Drupal::service('social_post.helper')->buildCurrentUserImage())
```
rollki, line 65: Maybe it's better to say `$user_image` or something instead of `$content`.

modules/social_features/social_album/social_album.module
```diff
@@ -0,0 +1,351 @@
...
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function social_album_preprocess_views_view(&$variables) {
+  /** @var \Drupal\views\ViewExecutable $view */
+  $view = $variables['view'];
+
+  if ($view->id() === 'albums') {
+    $library = [
```
rollki, line 99: $libraries instead of $library

modules/social_features/social_album/social_album.module
```diff
@@ -0,0 +1,351 @@
...
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function social_album_preprocess_social_album_post(&$variables) {
+  $items = [];
+
+  foreach ($variables['items'] as $item) {
+    if (!isset($items[$pid = $item['pid']])) {
```
rollki, line 117: Not sure if it is a good idea to create a new variable in the array brakets, also `$pid = $item['pid']`, and are you sure that the `$pid` is always not NULL?

modules/social_features/social_album/social_album.module
```diff
@@ -0,0 +1,351 @@
...
+  }
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_create_access().
+ */
+function social_album_node_create_access(AccountInterface $account, array $context, $entity_bundle) {
+  if (
+    $entity_bundle === 'album' &&
+    \Drupal::routeMatch()->getRouteName() === 'view.albums.page_albums_overview' &&
```
rollki, line 148: In this case, I think we can use a variable (`$route_match`) for `\Drupal::routeMatch()` because it calls twice.

modules/social_features/social_album/social_album.module
```diff
@@ -0,0 +1,351 @@
...
+    if ($first = &drupal_static(__FUNCTION__, TRUE)) {
+      $first = FALSE;
+    }
+    else {
+      $form['actions']['submit']['#attributes']['class'][] = 'hide';
+    }
+  }
+
+  if (!(
+    !\Drupal::moduleHandler()->moduleExists('social_ajax_comments') ||
```
rollki, line 215: In this case, I think we can use variables (`$route_match` and `$module_handler` because it calls twice.

modules/social_features/social_album/src/Form/SocialAlbumPostForm.php
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * The visibility options mapping.
+   *
+   * The associative array where keys are node options and values are the
+   * corresponding post options.
+   *
+   * @see field.storage.node.field_content_visibility.yml
+   * @see field.storage.post.field_visibility.yml
+   */
+  const VISIBILITY_MAPPING = [
```
rollki, line 25: Why you can not get this option from the config itself? Because, in this case, it seems, we don't support Social role visibility module.

modules/social_features/social_album/src/Plugin/Field/FieldWidget/SocialAlbumOptionsSelectWidget.php
```diff
@@ -0,0 +1,107 @@
...
+ * )
+ */
+class SocialAlbumOptionsSelectWidget extends OptionsSelectWidget {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getOptions(FieldableEntityInterface $entity) {
+    $options = parent::getOptions($entity);
+    $option = $options['_none'];
```
rollki, line 32: I think it is an unclear name, probably it's better to use $none_option

modules/social_features/social_album/src/Plugin/Field/FieldWidget/SocialAlbumOptionsSelectWidget.php
```diff
@@ -0,0 +1,107 @@
...
+  public static function validateElement(array $element, FormStateInterface $form_state) {
+    $field = $element['#parents'][0];
+    $has_images = $form_state->hasValue(['field_post_image', 0, 'fids', 0]);
+
+    if (
+      $element['#value'] === '_add' &&
+      ($title = $form_state->getValue([$field, 'title']))
+    ) {
+      if ($form_state->getTriggeringElement()['#name'] === 'op' && $has_images) {
+        $node = \Drupal::entityTypeManager()->getStorage('node')->create([
```
rollki, line 83: Please use dependency injection here.

modules/social_features/social_album/src/Plugin/views/field/SocialAlbumEntityOperations.php
```diff
@@ -0,0 +1,77 @@
...
+   * @param int $post_id
+   *   The post entity ID.
+   * @param int $file_id
+   *   The file entity ID.
+   *
+   * @return array
+   *   A renderable array representing the post links.
+   */
+  public static function renderLinks($node_id, $post_id, $file_id) {
+    $entity = \Drupal::entityTypeManager()->getStorage('post')->load($post_id);
```
rollki, line 56: We can call `\Drupal::entityTypeManager()` only once here, so it's better to create a variable I think.

rollki (approved):
All changes look very good, but please note my comments, these not critical but it will be good to take a look.

PR #1997, opened 2020-09-26
Access denied on group manage members

slowflyer:
## Problem
Just figured out, that I can not access the group manage members tab, after one of the last upgrades. fe. /group/34/membership
Even with user 1 I receive a "You are not authorized to access this page."

## Solution
This seems to be a "multilayer issue" on most installs I got rid of the behaviour by rebuilding node access permission.
But on one install I still got "You are not authorized to access this page." after rebuilding the permission.

See issue https://www.drupal.org/project/social/issues/3173334

## Issue tracker
https://www.drupal.org/project/social/issues/3173334

## How to test
*For example*
- [ ] As a sitemanager

## Screenshots

## Release notes

## Change Record

## Translations


navneet0693 (changes requested):
The value of $group variable will not be a group object, if Line 50 is TRUE

navneet0693:
Also, this is quite old. Please open a new PR if required.

PR #2003, opened 2020-10-02
Issue #3164294: Make sure we render the group_type field on group edit as part of a fieldset

ronaldtebrake:
## Problem
After the field_group update we noticed in 9.x the group_type field on editing a group is rendered outside of a fieldset. While on adding a group it is rendered correctly. This is due to the fact the FormBuilder renders radios within a container, instead of just a plain radio form element.

## Solution
We are now mimicking this behavior by also providing our radio element as container.

## Issue tracker
https://www.drupal.org/project/social/issues/3164294

## How to test
- [ ] As a SM edit an open group and make it a closed one
- [ ] Make sure the Ajax command shows the message it's tricky to change group types and content is updated
- [ ] Make sure all group content, e.g. memberships, nodes are all updated in a batch process as well when you change group type
- [ ] Make sure the permission to change a group type still is taken in to account
- [ ] Make sure it looks good.

## Release notes
The group type field is now part of the Settings field group again.


navneet0693 (approved):

navneet0693:
Cherry picked on 4b744c0db2bb7a95201c55a3c2f44157fce3621c 🍒

PR #2004, opened 2020-10-02
Issue #2947667 by ok_lyndsey, Kingdutch: Style skip to content so it …

Kingdutch:
…is easily visible on each page

## Problem
There is a skip to content link for people using a keyboard (or screen reader) to skip the menu structure (which can have lots of selectable elements). However, this element doesn't become visible when it receives focus.

## Solution
The style is borrowed from Drupal.org. For the colors we've chosen to
use the navigation active link color as it's usually different from the
navbar color itself. Additionally it has a matching text color that
should be configured by platform users to something accessible. This has
the highest chance of resulting in a skip link that is styled to the
platform styles while being accessible.

## Issue tracker
https://www.drupal.org/project/social/issues/2947667

## How to test
*For example*
- [ ] In a browser that supports tab navigation to links
- [ ] Open the homepage
- [ ] Use the tab button to select the first link on the page
- [ ] This should be the skip link that is now visible on top of the navigation bar.

## Screenshots
The default styling on desktop
Screenshot 2020-10-02 at 15 24 09

The default styling on mobile
Screenshot 2020-10-02 at 15 24 23

The link follows the color configuration of the socialblue theme.
Screenshot 2020-10-02 at 15 22 39

## Release notes
The link that is available to skip the navigation for users navigating the page with a keyboard will now show up when it receives focus.

## Change Record
Not needed

## Translations
None affected

### DIFF
--- a/themes/socialbase/assets/css/base.css
+++ b/themes/socialbase/assets/css/base.css
@@ -145,6 +145,23 @@ body {
   -webkit-font-smoothing: antialiased;
 }
 
+#skip-link {
+  text-align: center;
+  position: absolute;
+  left: 0;
+  right: 0;
+  z-index: 15000;
+}
+
+#skip-link a {
+  background: #222222;
+  box-shadow: 0 0 4px rgba(0, 0, 0, 0.14), 0 4px 8px rgba(0, 0, 0, 0.28);
+  color: #f3f3f3;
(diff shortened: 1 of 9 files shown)


Kingdutch:
Cherry-pick to 8.x and 9.x

PR #2008, opened 2020-10-05
Notifications for unpublished content

navneet0693:
## Problem
Users received notification about unpublished content posted in the group.

## Solution
Add conditions to check for unpublished content before sending notifications. 

## Issue tracker
https://www.drupal.org/project/social/issues/3052900

## How to test
- [ ] Login as user A
- [ ] Create a group and invite User B to it.
- [ ] Create both unpublished and published content in the group.
- [ ] Run cron
- [ ] Login as User B and check for notifications
- [ ] User B should not receive notification about any unpublished content.

## Screenshots
N.A

## Release notes
Users of the group will not receive any notification about unpublished content created in a group.

## Change Record
Message templates `moved_content_between_groups, create_topic_group, create_topic_community, create_event_community, create_event_group, and create_content_in_joined_group` have new activity conditions to exclude notifications for unpublished content.

## Translations
N.A


modules/social_features/social_activity/config/install/message.template.create_content_in_joined_group.yml
```diff
@@ -21,7 +21,7 @@ third_party_settings:
       email: email
     activity_create_direct: 0
     activity_aggregate: 0
-    activity_entity_condition: ''
+    activity_entity_condition: node_published
```
ronaldtebrake, line 24: Hi @navneet0693, don't we need an update hook here?

navneet0693: @ronaldtebrake yes, I will add it. 🙈

modules/custom/activity_basics/src/Plugin/ActivityEntityCondition/NodePublishedActivityEntityCondition.php
```diff
@@ -0,0 +1,28 @@
...
+ *  label = @Translation("Entity Published"),
+ *  entities = {"node" = {}}
+ * )
+ */
+class NodePublishedActivityEntityCondition extends ActivityEntityConditionBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isValidEntityCondition($entity) {
```
ronaldtebrake, line 21: Did you check what will happen with entities like Post? You are adding it now to `message.template.create_content_in_joined_group.yml` which is also triggered for posts, did you check if this doesn't apply there? 
Returning FALSE might not cut it.

navneet0693:
https://github.com/goalgorilla/open_social/pull/1586

PR #2009, opened 2020-10-05
Issue #3174889 by tBKoT: Bug with only one selected calendar in settings

dmytriikaiun:
## Problem
For the `Add to calendar` feature we have a configuration page. But if we select only one calendar for it the button 'Add to Calendar' on the event page doesn't work.

## Solution
Check the number of links before the render button. If the count of links is 1 don't use the 'no-split' attribute.

## Issue tracker
https://www.drupal.org/project/social/issues/3174889

## How to test
*For example*
- [ ] Go to Configuration -> Open Social Settings -> Add to calendar settings
- [ ] Enable the `Add to calendar` feature
- [ ] Select only one calendar for use.
- [ ] Go to any event page.
- [ ] The `Add to Calendar` button does not respond. 

## Screenshots
N/A

## Release notes
N/A

## Change Record
N/A

## Translations
N/A


ronaldtebrake:
Also 🍒  to 9.x

PR #2012, opened 2020-10-06
Issue #3171502 by agami4: Make post remove image button not visible for the accessibility tree

agami4:
## Problem
When adding an image to a post, the image will show up and be overlaid with a button containing a cross that allows the removal of the image. This button correctly has a text for screenreaders but it does not mark the SVG cross image as being decorative.

## Solution
Remove focus styles and add `aria-hidden="true"` for the `post remove image` button.

## Issue tracker
- https://www.drupal.org/project/social/issues/3171502
- https://www.drupal.org/project/social/issues/3134488

## How to test
*For example*
- [ ] Login to an Open Social community
- [ ] On the main stream page add an image to a new post
- [ ] You should now see the image preview in the post form with `post remove image` button

## Screenshots
post-remove-image-button

## Release notes
The `post remove image` button has not focus styles and has `aria-hidden="true"`

## Change Record
-

## Translations
-

### DIFF
--- a/themes/socialbase/templates/file/image-widget.html.twig
+++ b/themes/socialbase/templates/file/image-widget.html.twig
@@ -43,11 +43,9 @@
     <div class="hidden">{{ data.remove_button }}</div>
     <button type="button" id="post-photo-remove" class="btn--post-remove-image">
       <svg class="btn-icon">
+        <title>{% trans %}Remove image{% endtrans %}</title>
         <use id="btnicon" xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#icon-close"></use>
       </svg>
-      <span class="sr-only">
-        {% trans %}Remove image{% endtrans %}
-      </span>
     </button>
 
   {% else %}
--- a/themes/socialbase/templates/file/image-widget.html.twig
+++ b/themes/socialbase/templates/file/image-widget.html.twig
@@ -13,14 +13,11 @@
  */
 #}
 
-{% if data.upload['#id'] starts with "edit-field-post-image-0-upload" %}
-  {% set in_post = true %}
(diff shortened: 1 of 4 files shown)


themes/socialbase/templates/file/image-widget.html.twig
```diff
@@ -41,7 +41,7 @@
   {% if in_post %}
     {# Render a "fake" remove button that triggers the actual (hidden) remove button. #}
     <div class="hidden">{{ data.remove_button }}</div>
-    <button type="button" id="post-photo-remove" class="btn--post-remove-image">
+    <button type="button" id="post-photo-remove" class="btn--post-remove-image" aria-hidden="true">
```
Kingdutch, line 44: @agami4 have you tested this with a screenreader?

This change makes it no longer possible to remove an image using a screenreader. We should not hide the button entirely but only the SVG icon.

We can also move the `.sr-only` text as `<title>` into the SVG element. This would accomplish the same result I think.

Kingdutch (changes requested):

themes/socialbase/templates/file/image-widget.html.twig
agami4: @Kingdutch, I've removed the `aria-hidden` attribute and moved help text to the title on the SVG

Kingdutch (approved):
I've tested the accessibility improvement and added a fix for https://www.drupal.org/project/social/issues/3134488 which was rediscovered while testing that.

Kingdutch:
Cherry picked to 9.x in cb9d86c72 6250a4fad 

Cherry picked to 8.x in c6bb0df11 759bcd11f

PR #2013, opened 2020-10-06
Issue #3171785 by agami4: Make help text search button not visible for the accessibility tree

agami4:
## Problem
The search icon in the menu bar contains a help text. However, this help text isn't actually visible for screen readers.

## Solution
Replace <span class="invisible">Search</span> with <span class="sr-only"Search</span> and add `aria-hidden="true"`

## Issue tracker
https://www.drupal.org/project/social/issues/3171785

## How to test
*For example*
- [ ] Open inspector element and check new help text search button

## Screenshots
search-button-help-text

## Release notes
The help text search button not visible for the accessibility tree.

## Change Record
-

## Translations
-

### DIFF
--- a/themes/socialbase/templates/block/block--system-menu-block--main.html.twig
+++ b/themes/socialbase/templates/block/block--system-menu-block--main.html.twig
@@ -49,8 +49,8 @@
   {% if content.links.search_block is defined %}
     <div class="navbar__open-search-control">
       <a href="{{ '/search/all' }}" class="navbar__open-search-block" rel="search" title="{{ 'Click to open search box'|t }}">
-        <span class = "sr-only">{{ 'Search'|t }}</span>
         <svg class="navbar-nav__icon navbar-nav__icon--search">
+          <title>{{ 'Search'|t }}</title>
           <use xlink:href="#icon-search"></use>
         </svg>
       </a>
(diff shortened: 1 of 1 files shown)


themes/socialbase/templates/block/block--system-menu-block--main.html.twig
```diff
@@ -49,7 +49,7 @@
   {% if content.links.search_block is defined %}
     <div class="navbar__open-search-control">
       <a href="{{ '/search/all' }}" class="navbar__open-search-block" rel="search" title="{{ 'Click to open search box'|t }}">
-        <span class="invisible">{{ 'Search'|t }}</span>
+        <span class = "sr-only" aria-hidden="true">{{ 'Search'|t }}</span>
```
Kingdutch, line 52: @agami4 Why did you add the `aria-hidden="true"`? This way it's still not visible to screenreaders.

Kingdutch (changes requested):

themes/socialbase/templates/block/block--system-menu-block--main.html.twig
agami4: @Kingdutch, I've removed the `aria-hidden` attribute and checked via 'Voice Over'

Kingdutch (approved):

PR #2016, opened 2020-10-06
Issue #3171814 by agami4: Replace id with class and update js behavior

agami4:
## Problem
The ID `btnicon` is used in image-widget.html.twig as well as dropdown.html.twig. This usage causes it to show up multiple times on the same page. Having an ID duplicated on a page is invalid.
This ID seems to be used in the visibility-settings.js file.

## Solution
Replace id with class and update js behavior

## Issue tracker
https://www.drupal.org/project/social/issues/3171814

## How to test
*For example*
- [ ] Open web inspector and check current buttons

## Screenshots
-

## Release notes
Replaced `btnicon` id with class and updated js behavior

## Change Record
-

## Translations
-

### DIFF
--- a/themes/socialbase/templates/file/image-widget.html.twig
+++ b/themes/socialbase/templates/file/image-widget.html.twig
@@ -41,7 +41,7 @@
     <button type="button" id="post-photo-remove" class="btn--post-remove-image">
       <svg class="btn-icon">
         <title>{% trans %}Remove image{% endtrans %}</title>
-        <use id="btnicon" xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#icon-close"></use>
+        <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#icon-close"></use>
       </svg>
     </button>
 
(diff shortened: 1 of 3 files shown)


themes/socialbase/templates/file/image-widget.html.twig
```diff
@@ -43,7 +43,7 @@
     <div class="hidden">{{ data.remove_button }}</div>
     <button type="button" id="post-photo-remove" class="btn--post-remove-image">
       <svg class="btn-icon">
-        <use id="btnicon" xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#icon-close"></use>
+        <use class="btnicon" xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#icon-close"></use>
```
Kingdutch, line 46: Do we think this class is even needed?

Kingdutch (changes requested):
@agami4 When I tested this, the icon no longer changed. Please let me know if you need help with the JavaScript

themes/socialbase/templates/file/image-widget.html.twig
agami4: @Kingdutch, what do you mean `the icon no longer changed`? I think this class no needed in this case because this class does not use for `post photo remove` button

Kingdutch: I've checked in the code base and we only use the class for this specific button so we don't need it on other buttons (it has no styling and the JavaScript is scoped to this dropdown).

I've updated the code accordingly :)

Kingdutch (approved):

PR #2018, opened 2020-10-06
Issue #2896916 by rolki: Fix issue with #lazy_builder and ajax comments for nodes

rollki:
## Problems
- The social_ajax_comments module doesn't work with #lazy_builder
- We display only the first 50 comments in topics and events

## Solution
Remove pager since we don't have a pager for comments and we display only 50 first items. After the fix, we will display all comments and it's not a problem because the list of comments will be rendered by #lazy_builder.

## Issue tracker
https://www.drupal.org/project/ajax_comments/issues/2896916#comment-13845503 

## How to test
- Enable BigPipe and social_ajax_comments modules
- Go to any topic or event and try to leave a comment
- The commenting will be failing because of a pager (see screenshot bellow) 

## Screenshots
Снимок экрана 2020-10-02 в 15 53 17


ronaldtebrake:
Hi guys, i'm actually working on fixing social_ajax_comments on more than all occations, once i've opened up a PR will link it and made sure this work is also in there. (including the patch with updates in ajax_comments)

ribel (commented):
@rollki I see tests is failing here, but as Ronald said he is also working on those issues, so we can wait for his input.

ronaldtebrake:
https://github.com/goalgorilla/open_social/pull/2033
See the above for the finished PR

PR #2033, opened 2020-10-09
Issue #3108608 - Stability of Ajax comments

ronaldtebrake:
## Problem
We have tried to solve several issues which were encountered when using Ajax comments.

1. Ajax comments breaks landing pages and dashboards, when using the Views block, content in de activity stream is rendered using a wrong display mode.
2. Ajax comments and the lazy builder do not work well together, see: 2018
3. Replying on a reply, or replying on a comment do not render correctly. Messages are in the wrong place.
4. Error with "top" undefined when clicking on the reply button

## Solution
Make sure we stabilize social_ajax_comments. Update the ajax_comments patch and help it being committed.

## Issue tracker
All the below should be fixed:
https://www.drupal.org/project/social/issues/3108608
https://www.drupal.org/project/social/issues/3167492

## How to test
- [ ] Enable social_landing_page
- [ ] Enable social_ajax_comment
- [ ] As LU+ we should be able to comment on a post, topic or node without a page refresh.

## Release notes
Ajax comments are now stable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -101,7 +101,7 @@
                 "Make sure lazy doesnt do anything on cron": "https://www.drupal.org/files/issues/2019-07-30/3071331-lazy-cron-empty-path-2.patch"
             },
             "drupal/ajax_comments": {
-                "Fix display mode issue": "https://www.drupal.org/files/issues/2020-10-07/ajax_comments-ajax_not_working_when_using_non_default_view_mode-2896916-34.patch"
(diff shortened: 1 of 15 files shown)


ronaldtebrake:
With thanks to @slowflyer - 
We need to fix `$entity->getEntityType()->getLabel()` on several places, in the patch as well as in our code.

bramtenhove (approved):
After the last set of commits it works quite well!

PR #2044, opened 2020-10-20
Issue #3177932 by Kingdutch, ronaldtebrake: Allow modules to add cust…

Kingdutch:
…om CSS to improved_theme_settings

Some modules may require different selectors than are supported in the improved_theme_settings module. They may need this to hook into theme settings with their own styling. It's impractical to support every possible module in the improve_theme_settings module itself. 

Add an info hook that is called when improved_theme_settings generates its CSS. This allows modules to query the current theme settings and generate their own selector. One such use case is an internal Open Social extension that is adopting utility based CSS.

Follow ups to this ticket could be created to move some of the very specific queries currently included in improved_theme_settings to their own modules, this cleans up the CSS when those modules are not enabled.

## Issue tracker
https://www.drupal.org/project/social/issues/3177932

## How to test

- [ ] Implement a custom hook in a separate module
- [ ] Enable that module
- [ ] See that the styles are added

## Screenshots
N.a.

## Release notes
Modules can now inject their own CSS that depends on theme settings using `hook_improved_theme_settings_add`.

## Change Record
N.a.

### DIFF
--- a/modules/custom/improved_theme_settings/improved_theme_settings.api.php
+++ b/modules/custom/improved_theme_settings/improved_theme_settings.api.php
@@ -1,0 +1,42 @@
+<?php
+
+/**
+ * @file
+ * Hooks provided by the improved_theme_settings module.
+ */
+
+/**
+ * @addtogroup hooks
+ * @{
+ */
+
+/**
+ * Allows modules to add additional CSS based on theme settings.
+ *
+ * @param string $theme
+ *   The name of the them that's being rendered.
+ *
+ * @return string
+ *   Returns a string that will be added as CSS.
+ *
+ * @ingroup improved_theme_settings
+ */
+function hook_improved_theme_settings_add(string $theme) {
+  $style_to_add = '';
+
+  $card_radius = improved_theme_settings_get_setting('card_radius', $theme);
+
+  if ($card_radius >= 0) {
+    $style_to_add .= '
+      .my-custom-selector {
+        border-radius: ' . $card_radius . 'px;
+      }
+    ';
+  }
+
+  return $style_to_add;
+}
+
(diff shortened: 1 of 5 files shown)


Kingdutch:
@ronaldtebrake Would like your input on:

- Whether you think this is a good idea
- Whether you agree with the chosen hook name or think we should name it differently

I've left this as Draft PR because I need the styling in `tokens.(s)css` for an extension and may still need to make changes there. That styling can not live outside of the socialblue theme but otherwise does not impact code in the distro. Ping me on Slack if you have questions :)

ronaldtebrake:
Hi @Kingdutch 
Very happy with this approach :) and the chosen name.
Would you be able to investigate if we need some validation? Since we will use Markup::create, I'm not entirely sure what happens if people use this hook with wrong data, which we are opening up with this possibility.
As long as we are aware I'm also happy to hear if it's a no :) other than that 👍

Kingdutch:
I'm not sure we can validate here since we're requesting freeform CSS from developers. Quick testing shows that we do the proper input validation necessary in the theme settings. So as long as that remains correct there is no validation needed (that attack vector doesn't change).

Of course Developers will need to make sure they don't introduce sources of insecure user input, which is a similar concern in other places. I'll update the hook API documentation to indicate that the string returned from the hook is not sanitized before output so that developers of hook implementations should be careful of user input.

(In any case I think sanitizing is not an option because it may break valid CSS, so this will always require developer discretion)

Kingdutch:
Given the approval I'm merging this when tests are green. It was originally paused in case we want to add more tokens already, but we can easily do that in another PR and merging this unblocks some other things :)

PR #2047, opened 2020-10-28
Issue #3179128 by Kingdutch: Teasers contain inaccessible image links

Kingdutch:
On the default stream or landing pages with the stream and blocks, the various blocks contain some related issues.

<ol>
<li>The SVG icons in the teasers (when no image is present) contain a title element with the content type. However in the context the functionality of the SVG icons are decorative so they should have the attribute <code>aria-hidden="true"</code></li>
<li>The teasers that have an image contain no alt text so these are in essence empty links</li>
</ol>

On landing pages the images in the teasers have been hidden from accessibility software using <code>aria-hidden="true"</code>. However, the contained link elements are still visible when navigating the page by keyboard. (See https://web.dev/aria-hidden-focus/ )

<h4>Original description from Dutch</h4>
<blockquote>in the right column under the heading "New topics" there are svg images that are decorative. However, they are formatted as informational images. These are the SVG images entitled 'Topic'. To make it decorative, the title element of these svg's can be removed, adding the attribute aria-hidden = true is another option. For some topics the icon is also a link to the topic. If an image works as a link, the alternative text should describe the link's purpose. Now the alternative text is empty. Since the title also refers to the topic, you can also choose to keep only that link and not to make a link of the image (or to combine these links).</blockquote>

<blockquote>contains messages from different people. The persons have a profile picture. This is a link to the profile page. These images do not have alternative text. The person's name is also a link to the profile page. These links could therefore be combined.</blockquote>

With regards to a landing page with content as featured items.

<blockquote>These images act as links and therefore need alternative texts that give these links a link purpose. Because these images are decorative and these links seem unnecessary, it is recommended to combine the link around the image with the next link (the title of the post). That way, the blank text alternative is okay, the clickable area is large, and the link has clear link text. There are also icons in the top left corner of the image. The icon and image are now hidden with aria-hidden. However, these elements still receive keyboard focus and the link can be followed, they are interactive elements. Auxiliary software cannot determine the name and role of these elements. Now there are four links to the same page on every post. These links can therefore be better combined into one link or links can be removed. </blockquote>

<ul>
<li>On an Open Social installation go to <code>/stream</code></li>
<li>Alternatively create a landing page with the desired blocks</li>
<li>Also look at the "Upcoming Birthdays" block (only available on landing pages).</li>
</ul>

The media in teasers is hidden with <code>aria-hidden="true"</code> since the images are decorative and useful to users using touch but are confusing to keyboard users or users of screenreaders. 

Additionally, JavaScript is added that fixes the issue where links are still focusable when contained in a part of the DOM that's hidden from the accessibility tree. This should fix the problem with keyboard focus disappearing in a part of the page hidden for screenreaders.

## Issue tracker
https://www.drupal.org/project/social/issues/3179128

## How to test
Navigate using the keyboard and a screenreader. Validate according to our internal a11y audit document and test the examples linked to this ticket.

## Screenshots
N.a.

## Release notes
An issue has been fixed where some decorative images that were previously hidden for screenreaders still had links that could be focused but could not be seen. Additionally, some previously missed images in teasers have been hidden from screenreaders as well. These changes should make it easier to navigate pages in Open Social using the keyboard and/or assistive technologies.

## Change Record
N.a.

## Translations
N.a.


ronaldtebrake:
Also in 9.x 🍒

PR #2052, opened 2020-11-03
Issue #3180468 by tBKoT: Change redirection for Social Login from profile to the front page

dmytriikaiun:
## Problem
Currently, after login with SSO user is redirected to the profile page, instead of the home page.

## Solution
Change redirect implementation in the 'userLoginCallback'

## Issue tracker
https://www.drupal.org/project/social/issues/3180468
https://getopensocial.atlassian.net/browse/YANG-4062

## How to test
- [x] Login to the system using a social network
- [x] You should be redirected to the front page.

## Screenshots
N/A

## Release notes
Now after login through the `Social Accounts` user will be redirected to the front page. 

## Change Record
N/A

## Translations
N/A


ribel (approved):

ronaldtebrake:
Good job! :) 
I think this makes sense,
Ideally we check after SSO if their Profile is complete to maybe delegate them that way first, but that didn't work before either so the front page makes more sense to me.

ribel:
🍒 picked to 9.x

PR #2053, opened 2020-11-04
Issue #3180663 by tBKoT: Improve caching for Invite group members block

dmytriikaiun:
## Problem
Sometimes SM can no longer add members to a group.

## Solution
Improve caching for Invite block by covering case when $this->routeMatch->getParameter('group') returns group id but not the group object.

## Issue tracker
https://www.drupal.org/project/social/issues/3180663
https://getopensocial.atlassian.net/browse/YANG-4042

## How to test
N/A

## Screenshots
N/A

## Release notes
Improve caching for Invite block by covering case when $this->routeMatch->getParameter('group') returns group id but not the group object.

## Change Record
N/A

## Translations
N/A


ribel (approved):

modules/social_features/social_group/modules/social_group_invite/src/Plugin/Block/SocialInviteLocalActionsBlock.php
```diff
@@ -154,4 +153,11 @@ public function build() {
     return $build;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheMaxAge() {
```
navneet0693, line 159: This might not permanently help as this block is rendered here https://github.com/goalgorilla/open_social/blob/10.0.x/modules/social_features/social_group/modules/social_group_invite/social_group_invite.module#L122

ronaldtebrake:
Are you sure the max age works correctly?

https://www.drupal.org/project/drupal/issues/2828136
https://www.drupal.org/node/2352009 

It seems to me there are still quite the issues with it. 
And this blocks in general seems cacheable to me..

dmytriikaiun:
@ronaldtebrake I was not able to reproduce this issue on local env so that's why I disable caching.
Also, I found that cache tags not created on some group pages because `$this->routeMatch->getParameter('group')` returns group id but not the group object.

ribel (approved):

ronaldtebrake:
Hi @dmytriikaiun, //cc @ribel 

Thanks for your explanation. I can imagine it's difficult.
But it seems you have updated the latest code to now fix that. Thanks for that!

Could you please update the PR and title with the new information since caching is not disabled anymore?

The updated code looks way better and mergable to me 👍

ribel:
Thanks for review @ronaldtebrake!
I have updated title and description for Dmytri since he is on vacation. 
Will merge it as well ;)

ribel:
🍒 picked to 9.x

PR #2056, opened 2020-11-09
Issue #3178408 - part 1: Add new slick version of dropdown field for usage on other places than post.

navneet0693:
## Problem
When adding a field of type Dropdown, it is always renderd in the same way as Posts.
This doesn't work on other places, due to the nature of the dropdown button / list and placement. Also we want it to look different.

## Solution
Update the dropdown with a new setting, this slick dropdown version will render it as shown in the design below..

Screenshot 2020-11-09 at 10 46 18

## Issue tracker
http://drupal.org/node/3178408

## How to test
- [ ] Enable field_ui
- [ ] See there is now a new setting in the Dropdown field called is slick.
- [ ] Create a new field and see this option works as expected.
- [ ] Make sure the post visibility field still works on all streams
- [ ] Make sure you can add multiple fields, so the JS accepts this

## Release notes
Implementation is done in a separate story and we will add release notes accordingly there.


ronaldtebrake:
Closed, design is updated. Can re-open if we ever need this element.

PR #2057, opened 2020-11-09
TB-4516 - Determine where user goes after registration.

sjoerdvandervis:
## Problem
* When a new user signs up for an open social platform, we let him / her fill in the account details and profile details. Afterwards, they are sent towards their own profile. This does not help onboarding them; it would be nice if we could send this user to a configurable page.

## Solution
Added an argument to use when the user first time logs in. If so, we check on saving the already created profile (it is not new, unfortunately) where we send the user. When this is configured correctly, we send the user to the configured page

## Issue tracker
https://www.drupal.org/project/social/issues/3181460

## How to test
- [ ] As admin, enable the alternative_frontpage module and configure their paths at `/admin/config/alternative_frontpage`
- [ ] Register as a new user
- [ ] Use your one-time login link
- [ ] Complete your account and your profile and see you get sent to the path you configured.

## Release notes
We made it possible to configure where a user goes after registering with the platform, using the alternative frontpage module.

### DIFF
--- a/modules/custom/alternative_frontpage/src/Form/AlternativeFrontpageSettings.php
+++ b/modules/custom/alternative_frontpage/src/Form/AlternativeFrontpageSettings.php
@@ -82,6 +82,14 @@ class AlternativeFrontpageSettings extends ConfigFormBase {
       '#size' => 64,
       '#default_value' => $config->get('frontpage_for_authenticated_user'),
     ];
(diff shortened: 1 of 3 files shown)


navneet0693 (approved):

navneet0693:
The changes related to this PR will not land in product for now. We will work on this in future.

PR #2058, opened 2020-11-09
Travis self-updates to v2 of composer while we don't support it (yet).

ronaldtebrake:
## Problem
Travis runs

```

composer self-update
Upgrading to version 2.0.6 (stable channel).
   
Use composer self-update --rollback to return to version 2.0.4
```

While we don't support it yet, breaking our tests.

## Solution
**Solution 1. Rollback to v1 untill we support it.** 
Solution 2. In a separate PR is to remove hirak/prestissimo and see if v2 actually does work in tests. hirak/prestissimo isn't necessary anymore for composer v2 since performance is much improved (also they announced it themselves) 
Solution 2 is tested in #2059 

So lets see which one is green!

## Issue tracker
None

## How to test
See if tests succeed.


navneet0693 (approved):

navneet0693:
The failure of tests is very much unrelated to changes in this PR, in fact, this PR fixes the Travis issues.

navneet0693:
🍒 picked to 8.x and 9.x

PR #2059, opened 2020-11-09
Explore Composer 2 support in Open Social CI (Travis)

ronaldtebrake:
## Problem
Travis runs

```

composer self-update
Upgrading to version 2.0.6 (stable channel).
   
Use composer self-update --rollback to return to version 2.0.4
```

While we don't support it yet, breaking our tests.

## Solution
Solution 1. Rollback to v1 untill we support it and is tested in #2058 
**Solution 2. In a separate PR is to remove hirak/prestissimo and see if v2 actually does work in tests. hirak/prestissimo isn't necessary anymore for composer v2 since performance is much improved (also they announced it themselves)**

So lets see which one is green!

## Issue tracker
None

## How to test
See if tests succeed.


ronaldtebrake:
Can reopen once we have composer v2 support.

Kingdutch:
goalgorilla/drupal_social adds repositories for modules that are also found in the Drupal.org packagist. This is no longer allowed in Composer 2. [These repositories](https://github.com/goalgorilla/drupal_social/blob/master/composer.json#L79) need to be removed and the reference to the modules should use the `drupal/*` format rather than the `goalgorilla/*` format.

Can retest this after those changes and https://github.com/goalgorilla/drupal_social/pull/355

ronaldtebrake:
Interesting for this PR:
https://blog.bitexpert.de/blog/fast-faster-composer2/

And as per comment above 
https://github.com/goalgorilla/drupal_social/pull/356

ronaldtebrake:
Duplicate of #2101

PR #2062, opened 2020-11-10
by robertragas: override the metadata info block to remove the creati…

robertragas:
…on date for events in the hero as we only need to event date itself

## Problem
Users are confused about what is the event date.

## Solution
Remove creation date from the Event detail page. This can be easily done because we already extend for events and can remove it there.

## Issue tracker
https://www.drupal.org/project/social/issues/3181695

## How to test
- [ ] Create an event
- [ ] Go the the event detail page
- [ ] The hero contains the creation date of the event next to the author
- [ ] Check out to this branch and see it's gone now

## Screenshots
Before
Screen Shot 2020-11-10 at 3 35 43 PM

After
Screen Shot 2020-11-10 at 3 35 58 PM

## Release notes
We've removed the system creation date of the event on the full display in the SKY theme. This is because we noticed that some users found it confusing with the startdate of the event itself.


jochemvn (approved):
1 question:
Why is the entire block in a `{% if display_submitted %}` block? Is that needed? For the rest all is ok.

jochemvn (approved):

robertragas:
> 1 question:
> Why is the entire block in a `{% if display_submitted %}` block? Is that needed? For the rest all is ok.

Saw your remark a bit too late, think you might be right that the display_submitted might not be needed, will check it out tomorrow.

robertragas:
Did some testing and you are correct yes, so created a follow-up on https://github.com/goalgorilla/open_social/pull/2109 which I merged.

PR #2063, opened 2020-11-10
by robertragas: add a white color for the badge-accent to increase re…

robertragas:
…adability

## Problem
In the new SKY theme we have an accent color which is configurable, but the text color of the accent is not. In some cases this can conflict visually making it unreadable. 

## Solution
Make the accent text color also configurable so you can avoid getting this contrast conflict.

## Issue tracker

## How to test
- [ ] Enable social_private_message
- [ ] Send a Private message to someone
- [ ] See the counter is black
- [ ] Check out to this branch
- [ ] See the counter is now white

## Screenshots
Before
Screen Shot 2020-11-10 at 6 36 17 PM

After
Screen Shot 2020-11-10 at 6 36 24 PM

## Release notes
We have added the new color "Accent color text" to the configuration of the SKY theme. This way you can make sure the Accent color fits together with the text color which used to be static.


robertragas:
To be honest. This came out of UAT of a client and was agreed to change the color. But now that I see this I think we should reconsider as the contrast of black on orange is a lot better.

Screen Shot 2020-11-10 at 6 37 31 PM
Screen Shot 2020-11-10 at 6 37 21 PM

jochemvn (changes requested):
You've now set the number to fixed white. This introduces problems for the default Open Social badge, since that's now an unreadable white on yellow... Also other platform might get into trouble. See attached screenshots.

Screenshot 2020-11-24 at 10 06 33
Screenshot 2020-11-24 at 10 08 29

robertragas:
> You've now set the number to fixed white. This introduces problems for the default Open Social badge, since that's now an unreadable white on yellow... Also other platform might get into trouble. See attached screenshots.
> 
> Screenshot 2020-11-24 at 10 06 33
> Screenshot 2020-11-24 at 10 08 29

I made it a bit more flexible by adding a new color to the theme, where you can change the accent color text. This way everyone can be happy, by default I keep it "4d4d4d" like it was before.

jochemvn (approved):

PR #2064, opened 2020-11-10
by robertragas: add pagination a white background instead of transpar…

robertragas:
…ent for better accessibility

## Problem
The pagination for the next and previous buttons are transparent in the new SKY theme. This makes it harder to read, also to be consistent with the rest it should be white.

## Solution
Add a white background color to the next previous navigation. Needs to be so specific otherwise the navigation counter for example "1/2" will also be white and that should stay transparent.

## Issue tracker
https://www.drupal.org/project/social/issues/3181780

## How to test
- [x] Enable the social course module
- [x] Create a course with a section
- [ ] Create 3 parts in it
- [ ] Navigate through the course where you have a next and previous button
- [ ] See the button is transparant
- [ ] Now check out this fix and see it became white and better to read

## Screenshots
Before
Screen Shot 2020-11-10 at 7 56 58 PM

After
Screen Shot 2020-11-10 at 7 56 46 PM

## Release notes
The navigation pager buttons have been more accessible by replacing the transparant background with a white background.


jochemvn (approved):
Colors can be hard coded in this part of opensocial, since they're not configurable.

PR #2067, opened 2020-11-11
Issue #3181951 by chmez: As a CM+ I want to have exclusive right to create groups

ghost:
## Problem
As a CM+ I want to have exclusive right to create groups.

## Solution
- SM should be able enable/disable this functionality on group settings page
- if the setting is disabled, LU should not be able to create a group:
  - the option “create group should not appear for LU in
    - user menu
    - my groups page
- CM should still be able to create challenges, courses and groups

## Issue tracker
- https://www.drupal.org/project/social/issues/3181951
- https://getopensocial.atlassian.net/browse/YANG-4106

## Screenhots
**Before**
image

**After**
image

## How to test
- [ ] Log in as SM.
- [ ] When you navigate to the "Group settings" page from the "Configuration > Open Social Settings" admin menu then you should see the "Allow regular users to create new groups" checkbox and this checkbox should be checked.
- [ ] Uncheck the "Allow regular users to create new groups" checkbox and press "Save configuration".
- [ ] Log in as CM+.
- [ ] You still should see the "New Group" item in the sub-menu of the "Create New Content" top menu item.
- [ ] When you open the "My groups" page from the user menu then you still should see the "Add a group" button in the sidebar.
- [ ] Log in as a normal member.
- [ ] When you go to the home page then you shouldn't see the "New Group" item in the sub-menu of the "Create New Content" top menu item.
- [ ] When you open the "My groups" page from the user menu then you shouldn't see the "Add a group" button in the sidebar.

## Release notes
Add a new field to the group settings page for prohibiting access to pages for creating groups by logged-in users without manager roles.


modules/social_features/social_group/config/install/social_group.settings.yml
```diff
@@ -1 +1,2 @@
+create: true
```
ribel, line 1: @OH17L Please rename this property to something like `allow_group_create` to be consistent with existing setting, because `create` is too general.

ribel (changes requested):

modules/social_features/social_group/config/install/social_group.settings.yml
OH17L: Done

rollki (approved):
All changes look good to me.

ribel (approved):
I have tested new functionality, also in the SaaS version, and it looks good to me.

ronaldtebrake:
Hi @ribel @OH17L 

Thanks, this is a lovely improvement :) 
Would be great to have this accompanied by a behat test next time, should be fairly easy but this has quite the impact if we break it.

Don't want 10.0.0 to be blocked by that but would be great if you can add a follow-up for it.

ribel (approved):

ribel:
@ronaldtebrake We have created follow up for adding behat tests here: https://getopensocial.atlassian.net/browse/YANG-4546

PR #2075, opened 2020-11-17
Issue #3183029 by chmez: As a SM I want to be able to limit profile visibility

OH17L:
## Problem
A user can choose to make their profile private (accessible by the site administrator only), but it should at a minimum display name and institution.

## Solution
- use social_profile_privacy module as a basis
- as a SM I should be still able to see profile information even when the information is private
- improve the copy and structure of SM settings page
- make sure that user should be able to hide separately:
  - phone number
  - birthday
- include selection for hiding groups on the cards itself

## Issue tracker
- https://www.drupal.org/project/social/issues/3183029
- https://getopensocial.atlassian.net/browse/YANG-4105

## Screenshots
**SM Settings**
image

**LU Settings**
image

## How to test
- [ ] Enable the "Social Profile Privacy" module.
- [ ] When you log in as SM and navigate to the profile settings page ("/admin/config/people/social-profile") then you should see the "Profile fields visibility" section instead of the following fields: "Privacy settings on display of user form", "Privacy settings on display of profile form". Also, you shouldn't see the following sections: "Field groups", "Disclaimer".
- [ ] You should see a table and the "Disclaimer" field in the "Profile fields visibility" section.
- [ ] You should see all fields from the profile edit page in the first column of the table.
- [ ] You should see the following radio buttons in the second and next columns of the table: "Always show for everyone", "Show, but can be hidden by each user", "Hide for others".
- [ ] You should see that the "Always show for everyone" radio button is checked by default for all fields except the following fields: "Address", "Phone number".
- [ ] You should see that the "Show, but can be hidden by each user" radio button is checked by default for the following fields: "Address", "Phone number".
- [ ] You should see that the "Disclaimer" fields contain a value that consists of the fields of the "Disclaimer" removed sections.
- [ ] When you navigate to the user edit page then you should see toggle switcher fields for all fields from the profile edit page instead of the "Show "Phone number and location" on my profile" toggle switcher field in the "Privacy settings" section.
- [ ] When all radio buttons in the field row on the profile settings page are disabled then the field with the same label in the "Privacy settings" section of the user edit page should be disabled as well.
- [ ] When the "Always show for everyone" radio button in the field row on the profile settings page is checked then the field with the same label in the "Privacy settings" section of the user edit page should be disabled and checked.
- [ ] When the "Show, but can be hidden by each user" radio button in the field row on the profile settings page is checked then the field with the same label in the "Privacy settings" section of the user edit page should be enabled and checked.
- [ ] When the "Hide for others" radio button in the field row on the profile settings page is checked then the field with the same label in the "Privacy settings" section of the user edit page should be disabled and unchecked.
- [ ] When a field in the "Privacy settings" section of the user edit page is unchecked then another user can not see this field on the "Information" profile tab of the current user.
- [ ] Enable the "Social Profile Fields" module.
- [ ] Navigate to the "/admin/config/opensocial/profile-fields" page and uncheck the "Expertise" field and press "Save configuration".
- [ ] When you back to the profile settings page then you should see that all radio buttons in the row of the "Expertise" field are disabled and the "Hide for others" radio button is checked.

## Release notes
A user can manage the visibility of each profile field instead of managing the visibility of groups of fields.


modules/social_features/social_profile/modules/social_profile_privacy/social_profile_privacy.install
```diff
@@ -63,3 +65,178 @@ function _social_profile_privacy_get_permissions($role) {
 
   return $permissions[$role];
 }
+
+/**
+ * Define visibility options of separate fields per user.
+ */
+function social_profile_privacy_update_8001(&$sandbox) {
```
rollki, line 72: Do we also need these changes on `hook_isntall`?

modules/social_features/social_profile/modules/social_profile_privacy/social_profile_privacy.install
```diff
@@ -63,3 +65,178 @@ function _social_profile_privacy_get_permissions($role) {
...
+      ->load('profile.profile.default');
+
+    foreach ($display->getThirdPartySettings('field_group') as $fields_group => $settings) {
+      $states = [];
+
+      foreach ($settings['children'] as $field) {
+        $states[$field] = SocialProfilePrivacyHelperInterface::HIDE;
+      }
+
+      $sandbox['fields'][$fields_group] = serialize($states);
```
rollki, line 85: It's better to use "Serialization API" instead.
https://www.drupal.org/docs/8/api/serialization-api/serialization-api-overview

rollki (commented):
Generally, look very good, please check my 2 comments.

modules/social_features/social_profile/modules/social_profile_privacy/social_profile_privacy.install
OH17L: No, we don't. This update converts an old configuration for a new one but when a module will be enabled on a project then a new configuration is used.

modules/social_features/social_profile/modules/social_profile_privacy/social_profile_privacy.install
OH17L: Done.

ribel (approved):
Code and functional changes look good to me.
Hook update applies without errors on ECI demo and new customer site.

ronaldtebrake:
Hi @chmez //cc @ribel 

Thank you so much for this, it works like a charm mostly, a couple of remarks:

1) So in the old way of working, a user on it's edit page was only able to toggle settings they were actually able to toggle.
So in the below case, only that fieldset was configured as being able to hide for others.

Screenshot 2020-12-28 at 17 02 31

Versus

Screenshot 2020-12-28 at 16 58 51

All these settings are very confusing, most of them by default are not even able to be changed and also they don't add any informational purpose since users know what fields are available.

While the design shows it a bit more informational
Screenshot 2020-12-28 at 17 11 53


2) It seems the update hook to update the individual settings for a user didn't work well.

Before i have hidden the below
Screenshot 2020-12-28 at 17 14 37

After update it did print out it updated it for the correct UID.
```
Profile privacy data has been processed for the following users: 5
```

However when I login as that user and go to my edit settings I see this:

Screenshot 2020-12-28 at 17 15 51
and also the phone number shows up on the profile so it seems that failed
Screenshot 2020-12-28 at 17 52 05

PR #2076, opened 2020-11-18
Issue #3178408 - As a LU, I want to configure groups more easily - Add group visibility field 

ronaldtebrake:
## Problem
We plan to provide only flexible groups in future. However to do this, we need to make Flexible groups easier to configure.
Currently, the configuration interface for flexible groups is
1) overwhelmed with all the options and it is
2) missing the option to configure the group visibility.

## Solution
This PR focusses on part 2 - missing the option to configure the group visibility. 
It adds a new field to flexible groups, this field will be the base of all what is to come. 
- It will provide default values for the content visibility and join methods
- It will provide the access for where a certain group is visible
- It will help explaining what certain configurations of the flexible groups mean.

Out of scope for this PR will be dealt with separately:
- Access checks based on the input
- State/Ajax updates of the content visibility fields
- Theming 

## Issue tracker
https://www.drupal.org/node/3178408

## How to test
- [x] Test install / update
- [x] As a LU - See that you can add a flexible group and there is now a new Group visibility field
- [x] See there is an explanation of what the options actually mean.
- [x] When saving I expect the field to keep its value
- [x] On edit the value should still be there.

## Screenshots
Screenshot 2020-11-18 at 15 49 18
Screenshot 2020-11-18 at 15 49 22

## Release notes - Only included once all the work is done.
Flexible groups used to be very overwhelming and confusing. The configuration of visibility and join methods weren't transparent enough on what it meant for a Group, it's members and content. To overcome this unclarity we gave a new look to the UX of flexible groups. Allow you to specify the visibility of the group itself, not just it's content, and provide insights to what the different settings mean.


navneet0693 (approved):

ronaldtebrake:
Closed in favor of https://github.com/goalgorilla/open_social/pull/2077

PR #2079, opened 2020-11-20
Issue #3183708 by SV: Groups do not show up in the discussions for administrator

volodymyr-sydor-lemberg:
## Problem
Groups do not show up in the widget of discussions for the site administrator if those discussions were created by other users. So, seems that option is only available for authors with SM/CM roles

```
$account = $entity->getOwner();
.............
.............
// Remove groups the user does not have create access to.
if (!$account->hasPermission('manage all groups')) {
   $options = $this->removeGroupsWithoutCreateAccess($options, $account, $entity);
}
```
Screenshot from 2020-11-20 16-16-59

## Solution
Use a current user instead of the author of the node to prevent an access issue for admin users.

## Issue tracker
- https://www.drupal.org/project/social/issues/3183708
- https://getopensocial.atlassian.net/browse/YANG-4268

## How to test

- [ ] Using version 10.0.x of Open Social with the Social Group and Social Discussion modules enabled
- [ ] As LU create a new discussion node
- [ ] Open that discussion as admin and I expect the group filed includes a list of available groups

## Screenshots
 Extend the message of group selector widget, from:
Screenshot from 2020-12-14 13-10-24
to the following:
Screenshot from 2020-12-14 18-45-39

## Release notes
 Resolved an access group widget issue for admin users, to be able to assign some created discussion to a group


ribel (approved):

ronaldtebrake:
Hi guys,

This needs some changes.
Imagine as user I create a topic, in a open group. If a Sitemanager decides to edit this and select a group the author is not a member of. This might result in the content not being editable anymore by the author and thus they arent in control of their own content. 
I can see why it adds benefits to have this enabled for a SM. But the way it's written right now it's also the case for co-authors.

So potentially we need to special case this for users with certain permissions.

So we can say the following:

```
$account = $this->userManager->load($this->currentUser->id());
// If the user can administer content and groups, we allow them to override this.
if (!$account->hasPermission('administer content') && !$account->hasPermission('manage all groups')) {
$account = $entity->getOwner();
}
``` 

Screenshot 2020-11-24 at 10 02 34

Also it would be great to extend this message, to ensure SM understands that when they change it, it might impact not only the visibility settings but also the original authors flow.

ribel:
@ronaldtebrake Thanks for the feedback! 
We will add this special check for permission as you proposed.

volodymyr-sydor-lemberg:
Hi @ronaldtebrake, I've added a check for permissions and adjusted the message of the group selector widget. 
Could you take a look at it, please

modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
```diff
@@ -262,7 +262,7 @@ public static function validateGroupSelection(array $form, FormStateInterface $f
 
       $ajax_response->addCommand(new InvokeCommand('#edit-field-content-visibility-' . $visibility, 'change'));
     }
-    $text = t('Changing the group may have impact on the <strong>visibility settings</strong>.');
+    $text = t('Changing the group may have impact on the <strong>visibility settings</strong> and <strong>access</strong> for <strong>author/co-authors</strong> of the current page.');
```
ronaldtebrake, line 265: Would be great if we can remove the specific "of the current page" as this will be shown on events, nodes etc.

Just

`t('Changing the group may have impact on the <strong>visibility settings</strong> and may cause <strong>author/co-authors</strong> to lose access.)`

ronaldtebrake:
Hi @volodymyr-sydor-lemberg, thanks for the updates, almost there, just a small update to the text will help showing this in all the different contexts. After that it's perfect for me, thank you!

volodymyr-sydor-lemberg:
thank you @ronaldtebrake!  I've just changed it

modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
volodymyr-sydor-lemberg: done)

ribel:
🍒 picked to 9.x

PR #2081, opened 2020-11-23
Issue #3184207 by tBKoT: Update morris.js library from version 0.5 to 0.6

dmytriikaiun:
## Problem
6 years was passed since the last release of the morris library so I suggest using another version on this library. Also, it needs to provide a new horizontal bar chart for KPI (screenshot).

## Solution
Use the new library https://www.npmjs.com/package/morris.js06

## Issue tracker
https://www.drupal.org/project/social/issues/3184207
https://getopensocial.atlassian.net/browse/YANG-3983

## How to test
N/A

## Screenshots
image

## Release notes
A new version of Morris library.

## Change Record
The process of using morris is not changed. The new library support all features from old, just add something new.

## Translations
N/A

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -187,7 +187,7 @@
         "bower-asset/waves": "0.7.6",
         "bower-asset/timepicker": "~1.11.14",
         "bower-asset/tablesaw": "~3.1.0",
-        "bower-asset/morris.js": "0.5.1",
+        "npm-asset/morris.js06": "^0.6.6",
         "bower-asset/raphael": "v2.2.8",
         "bower-asset/bootstrap": "v3.4.1",
         "bower-asset/select2": "~4.0.5",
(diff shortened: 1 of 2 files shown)


robertragas:
Just to note, this worked for Nacho. He tried to get OS running locally, had problems with Morris and said to try out this PR.

I don't know if the kpi analytics are still functionally working, didn't test that.

dmytriikaiun:
> Just to note, this worked for Nacho. He tried to get OS running locally, had problems with Morris and said to try out this PR.
> 
> I don't know if the kpi analytics are still functionally working, didn't test that.

@robertragas It's working. This is the same Morris library but with some new features. I've used it to implement horizontal bars.

ribel (approved):

ronaldtebrake:
I'm happy to get it in in that case seeing all the input from everybody :)

ribel:
@ronaldtebrake great! 
Just wondering if it is fine to include this in 9.5 milestone, or should it be different?

ronaldtebrake:
Good question @ribel 
In this case considering the amount of features come with this increase in versioning i'd say 9.5 is the correct milestone.

ribel:
🍒 picked to 9.x

PR #2083, opened 2020-11-24
Issue #3184502 by sjoerdvandervis: Between date on people overview not working when selecting one day for both fields

sjoerdvandervis:
## Problem
When filtering users on the admin user overview, if you select the Between operator and the same date twice, you will get no results.

## Solution
Apply the patch in https://www.drupal.org/node/2842409

## Issue tracker
https://www.drupal.org/project/social/issues/3184502

## How to test
- [ ] Go to /admin/people and select the Between operator for dates;
- [ ] Choose the same date twice. Users registered on the specified days show up.

## Release notes
We made a small improvement in the way users can be searched in the site manager's user overview. When using the Between operator and selecting two days, you will always see the users that registered on those days.


ronaldtebrake:
Screenshot 2020-12-02 at 15 12 02
Screenshot 2020-12-02 at 15 12 13

Can't figure out why yet, but it doesn't seem to work well for the last login Between date.

Also probably the same issue occurs in our searches, where we use Search API between filters, that might need the same consistent fix. Although I can understand we need a separate story for that, but would be great if we can add that once we figure out the best solution.

Marking it as needs work for now.

ronaldtebrake:
This is used as patch on GPI and works there, we are working on the search api between filters in a separate story as that would contain a little more work. Approved, waiting for final tests results after fixing merge conflict.

ronaldtebrake (approved):

navneet0693:
Closed in favor of https://github.com/goalgorilla/open_social/pull/2314

PR #2086, opened 2020-11-30
Create drupal-cs.yml

ronaldtebrake:
Create a github action for running drupal coding standards on every PR

## Problem
We have some issues with travis running drupal coding standards, these issues result in behat tests not running due to the fact we stop travis tests on errors.

## Solution
Use github actions instead, this doesn't block travis, makes it faster as it can run in paralel as well.

## How to test
Ensure tests pass. 

## Release notes
Not needed. Scripts still remain in tact.


ronaldtebrake:
Closed in favor of a new one with better commits.

ronaldtebrake:
#2101 fixes this :)

PR #2087, opened 2020-11-30
Issue #3185627 by tBKoT: Add the settings form for email notification frequency

dmytriikaiun:
## Problem
Users receive a lot of emails notification as we have the `immediately` frequency as default for all emails.

## Solution
Add a settings page where the admin can edit the default frequency for every message template.

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-4245
https://www.drupal.org/project/social/issues/3185627

## Screenshots
image

## Release notes
Now site admin can specify the default frequency for every email notification instead of hardcoded `immediately`. It will take effect only for users that never saved/updated their own settings.

## Change Record
Do not use `immediately` frequency as default for email notifications. Get settings from config `social_swiftmail.settings` -> `template_frequencies`

## How to test
N/A

## Translations
N/A


modules/social_features/social_swiftmail/social_swiftmail.routing.yml
```diff
@@ -2,6 +2,6 @@ social_swiftmail.settings:
   path: '/admin/config/opensocial/swiftmail'
   defaults:
     _form: 'Drupal\social_swiftmail\Form\SocialSwiftmailSettingsForm'
-    _title: 'Social Swiftmail settings'
+    _title: 'Email notification settings'
```
ribel, line 5: @dmytriikaiun In the menu list, it still says `Email settings`. I think we should also change there to `Email notification settings` to be consistent.

ribel (approved):
Code refactoring and new functionality look really good.
Good job @dmytriikaiun!

ribel:
Hi, @ronaldtebrake, could you take a look at this new feature? Here we bring the ability to set default frequency to different message templates. 
We keep backwards-compatibility, so we can include it to 9.5 milestone, but want to confirm it with you.

ronaldtebrake:
Hi @ribel @dmytriikaiun 
Well done, from a BC perspective the code looks good to me too :) if we want this in 9.5 we do need to explain this a bit better though (in the release notes but also in the UI).

Example:
If an SM now decides to alter the default frequency on an existing site, and a user has never saved/updated their frequency, the default will be used because there are no settings to load from the database. Which is what the SM expects, but maybe the user doesn't. Because without them knowing, their settings have changed. 

So for existing sites this could cause some unexpected behavior for LU's. 

That's probably functional wise a bit BC breaking if we can't inform the user about this.

dmytriikaiun:
@ronaldtebrake please check release notes. Is it more clear? Also, about UI changes, we need to discuss with @xinyumaGG what text and where we can add for clarification to the admin.

ribel:
I have asked Xinyu to take a look at the UI texts in a ticket.

ribel:
🍒 picked to 9.x

PR #2089, opened 2020-12-02
TB-5244 by jochemvn: improve JS, so profile tag hover doesn't break

jochemvn:
## Problem
*In some cases, when passing  the mouse over the organisation nub, it is not displaying the organisation name*

## Solution
*Improve the JS so it doesn't break anymore*

## Issue tracker
https://getopensocial.atlassian.net/browse/TB-5244

## How to test
*For example*
- [ ] Using SAAS with the `social_profile_organization_tag` module enabled
- [ ] As a site manager
- [ ] Handout a couple of org tags to a couple of users
- [ ] On certain pages where a lot of tags are shown, the JS doesn't work.
- [ ] Checkout this branch, clear the cache and reload the page... notice now it does work.

## Release notes
An issue was fixed, where in some rare cases, the organisation of a user wasn't shown when hovering over the badge behind the users name.


robertragas (approved):

robertragas:
Changes look good, travis was red though and seemed unrelated so I restarted the test and need to wait for that.

robertragas:
Merging the main branch and restarted a few times didn't help so I guess we need to run the behat manually for 1 test to see if we have the same issue locally.

Also a bit weird as the organizational tag is an optional module which is not enabled by default.

jochemvn:
🍒 -picked to 8.x-8.x and 8.x-9.x

PR #2091, opened 2020-12-02
Reporting (ajax) comments leads to white screen 

jochemvn:
## Problem
Critical bug in AJAX Comments + Report. When:
a) posting a comment
b) directly report that comment for abuse
c) system crashes

## Solution
The solution is to create a different destination url on the report form, when the comment was inserted in the DOM through an ajax request.

## Issue tracker
https://getopensocial.atlassian.net/browse/TB-5356
https://www.drupal.org/project/social/issues/3186203

## How to test
- [ ] Enabled both `social_ajax_comments` and `social_content_report`
- [ ] Create a comment somewhere
- [ ] Immediately report is as abusive
- [ ] Notice you end up on a page with a textarea (with callback information)
- [ ] Try again on this branch and notice you're being redirected to the parent entity (the comment was made on)

## Release notes
An issue was fixed, where one would report one's own comment (made with ajax), a redirect would be made to a white screen with nothing more than a textarea.

### DIFF
--- a/modules/social_features/social_content_report/social_content_report.module
+++ b/modules/social_features/social_content_report/social_content_report.module
@@ -14,6 +14,12 @@ use Drupal\social_post\Entity\PostInterface;
  * Implements hook_form_alter().
  */
 function social_content_report_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+
+  // No need to do anything here, if it's not a flagging related form.
+  if (strpos($form_id, 'flagging_') === FALSE) {
+    return;
+  }
+
   // Get all 'report_' flags.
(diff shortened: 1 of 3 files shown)


bramtenhove (approved):
Nice find!

bramtenhove:
@jochemvn, I set the milestone for you at `8.8`.

bramtenhove:
@jochemvn, one thing you still need to do is to create a Drupal.org ticket and link it in the PR.

If you could that do, that would be appreciated. Thanks!

robertragas:
Looks good, I guess in the end it would be nice if something like this can also be processed in ajax without a reload of the page.

robertragas (approved):

robertragas:
merged and cherry-picked

PR #2092, opened 2020-12-02
Issue #3185840 by rolki: All topic in the all-topics view have the same number of comment

rollki:
## Problem
The number of comments shown in the counter on a discussion preview is not the actual number of comments in a discussion

## Solution
The number the same for each teaser because we use static cache to get the comments to count number. Generally, it's the proper way to get this amount, but in this case, it's too much. Also, I removed the loading array of comments object for performance reason. (see bellow)
And I think in the feature it's better to don't use static cache for this case, since this kind of cache work for page query, so maybe it's better to use cache tags.
Снимок экрана 2020-12-02 в 17 57 55

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-4294 
https://www.drupal.org/project/social/issues/3185840 

## Screenshots
image


robertragas:
@rollki so this PR is still in progress as you disabled the cache but want a better solution to optimize the way how we count the comments?

ronaldtebrake:
FYI - https://www.drupal.org/project/social/issues/3185840 seems to be related to this issue.
A good idea here would be to use the entity id as identifier so it only statically caches this per entity and doesn't have to count the comments more than once on a single request.

OH17L (approved):

ribel:
@rollki Please answer on comments above and create a related issue on drupal.org. I think it would be good to get that in tomorrow release.

rollki:
@robertragas 
So, actually, this ticket was in-progress but now it is done. The problem was in loading all entities (see below) and in the first commit I disabled cache but didn't remove the loading array of objects, but now loading entities overrides by the query into database.

image

rollki:
@ronaldtebrake 
Indeed I think this is exactly the same issue.
And yes I tried to use entity id as an identifier for static cache, and also use MemoryCache instead of drupal_static, but it didn't work correctly (
Therefore, so far I have implemented this way, it will not create performance problems, but we can create a separate ticket to implement a static cache by an entity identifier.

rollki:
@ribel 
I think we can reuse the existing one right? https://www.drupal.org/project/social/issues/3185840

ribel:
@rollki yes we can, then please change a title of PR according to our standards.

ribel (approved):

ribel:
@ronaldtebrake Could you check it again and let know if we can merge it?
Also since we change function arguments - maybe it's better to add it just to 9.5 milestone?
But I have checked that this function is not used in any of our enterprise modules.

ronaldtebrake:
Hi @ribel @rollki 

I think I need some more time to see how we should fix the cache before we merge it unfortunately.
Can confirm this PR does fix the bug, but since we stopped serving this from cache I need to test the performance impact, especially on overviews / search / landing pages or dashboards. 

While we should be able to cache it and only invalidate it once a comment is placed, also @rollki is right in trying something different than static cache when I think about it, you want this cached across requests.

rollki:
@ronaldtebrake 
Mmm, indeed we can try to use cache tags or some other things, but also it seems the following code should work correctly, and it still will be static cache, but per entity.
```
function _socialbase_node_get_comment_count(NodeInterface $node) {
  $comment_count = &drupal_static(__FUNCTION__ . $node->id());

  // Calculate the comment_count for this page request.
  if (is_null($comment_count)) {
```

rollki:
@ronaldtebrake 
Please check this commit: https://github.com/goalgorilla/open_social/pull/2092/commits/4694650161af01c145ef7e0dd9903593386169fc 

It is about things that I described above :)

ronaldtebrake:
@rollki that is correct, however if you look at it from a caching perspective.. you'd probably dont have a lot of pages where this entity gets counted more than once. So static cache might make less sense versus something that lives across requests.
But for now the new commit seems to indeed has a better caching mechanism in mind than nothing :D thanks!

ribel:
🍒 picked to 9.x

PR #2098, opened 2020-12-09
Event - Update enrollment method, show inline and add popover with explanation

navneet0693:
## Problem
After the new changes in the flexible group visibility design, we want to make join methods consistent. 

## Solution
1. Update labels
2. Remove explanatory label descriptions
3. Add new descriptive text under labels
4. Add popover and re-use logic from Flexible groups - Join method element
5. Make sure you add the popover render array element 
6. Make sure you add a hook to allow extensions to add visibility options

## Issue tracker
https://www.drupal.org/project/social/issues/3186003

## How to test
- [ ] Checkout to this branch, run drush updb
- [ ] Create an event
- [ ] Check for enroll methods labels & text
- [ ] Check for popover in enroll methods

## Screenshots
image

## Release notes
N.A

## Change Record
N.A

## Translations
N.A


modules/social_features/social_event/social_event.install
```diff
@@ -1621,3 +1621,36 @@ function social_event_update_8904() {
...
+  $updateHelper->executeUpdate('social_event', 'social_event_update_8906');
+
+  // Output logged messages to related channel of update execution.
+  return $updateHelper->logger()->output();
+}
+
+/**
+ * Update the enrollment method text of an event on event add form.
+ *
+ * Load in a config file from an specific update hook that will never change.
```
ronaldtebrake, line 1642: This only goes for static files, in this case you generated a nice update helper one. :D

modules/social_features/social_event/social_event.install
```diff
@@ -1621,3 +1621,36 @@ function social_event_update_8904() {
...
+ * Update helper did not update all fields correctly.
+ */
+function social_event_update_8907() {
+  $config_file = drupal_get_path('module', 'social_event') . '/config/update/social_event_update_8907.yml';
```
ronaldtebrake, line 1646: These statically yml files are better of in a separate folder. eg

`/config/static/social_event_update_8907.yml`

ronaldtebrake: This so we can distinguish those from the update helper.

The downside though is now if anybody has added custom storage options, they are removed.
Maybe you are better of just writing your own update hook that loops over the labels similar to 

https://github.com/goalgorilla/open_social/pull/2096/files#diff-76261ef8588b9c0e238662e1597bf7c1dc0d0db483c3e21fd3fb897e8feb6976R1642

ronaldtebrake (changes requested):

modules/social_features/social_event/social_event.module
```diff
@@ -1126,3 +1128,129 @@ function _social_event_views_update($view_config, $display) {
...
+  $description = '';
+
+  // We need it to be specified otherwise we can't build the markup.
+  if (empty($key)) {
+    return $description;
+  }
+
+  // Add explanatory descriptive text.
+  switch ($key) {
+    case '1':
```
ronaldtebrake, line 1148: would be cool to make a `const` out of this, so it's understandable for everybody what these keys mean.

PR #2099, opened 2020-12-09
Issue #3188292 by rolki: Add featured view mode template for the landing page

rollki:
## Problem
Dashboard and landing page don't look fine as featured item in the landing page or some other place with featured view mode.

## Solution
* Create a featured template for the landing page
* Add a default image for this view mode

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-4029 
https://www.drupal.org/project/social/issues/3188292 

## How to test
- [ ] Create a new landing page with futured items
- [ ] Select other landing pages as items
- [ ] You should see featured items with images

## Screenshots
7dbc6890-20eb-404e-b9a2-b9c546c0cbaf

image


themes/socialbase/src/Plugin/Preprocess/Node.php
```diff
@@ -209,6 +209,15 @@ protected function preprocessElement(Element $element, Variables $variables) {
       unset($variables['content']['links']['#lazy_builder']);
     }
 
+    // A landing page has a different way of determining this.
+    if ($node->getType() === 'landing_page') {
```
ribel, line 213: @rollki It looks like we don't have any other cases with no_image here, so maybe it needs to be added to social_landing_page_preprocess_node() or if it doesn't work - then to social_core_preprocess_node().
Could you check and fix that?

rollki: Moved, everything works fine after changes.

ribel (approved):

ribel:
🍒 picked only to 9.x, was not able to 8.x because of conflict

PR #2101, opened 2020-12-09
Developer task - Use github actions for coding standards check on a PR.

ronaldtebrake:
## Problem
Travis gave us quite some issues with coding standards, we decided travis is not the way forward as it's also not fully free for open source software anymore.

## Solution
We took the liberty to move towards github actions instead. While doing so we also added the following:
- (Potentially Composer v2 support, not verified yet but it seems it runs on v2)
- phpcs.xml where we configure the coding standards 
- Update drupal/coder so we benefit from the latest coding standards
- Unfortunately this gave a ton of errors which we cant fix that easily, so we decided to exclude some in our phpcs.xml

## Issue tracker
None - Only internal usage as coding standards applied before.

## How to test
See the new github actions are green!

## Release notes
None - Internal only.

### DIFF
--- a/.github/workflows/drupalCodingStandards.yml
+++ b/.github/workflows/drupalCodingStandards.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      # Install composer depdencies so we have drupal/coder which adds our codesniffer, the bin/phpcs file
+      # Install composer dependencies so we have drupal/coder which adds our codesniffer, the bin/phpcs file
       # and dealerdirect/phpcodesniffer-composer-installer which changes the installer_path configuring all sniffs.
       - name: Install dependencies
         run: composer install --dev --prefer-dist --no-progress --no-suggest
--- a/.github/workflows/drupalCodingStandards.yml
+++ b/.github/workflows/drupalCodingStandards.yml
@@ -1,0 +1,19 @@
+name: Drupal coding standards
+
+on: [pull_request]
+
+jobs:
+  phpcs:
+    name: PHPCS
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      # Install composer depdencies so we have drupal/coder which adds our codesniffer, the bin/phpcs file
+      # and dealerdirect/phpcodesniffer-composer-installer which changes the installer_path configuring all sniffs.
+      - name: Install dependencies
+        run: composer install --dev --prefer-dist --no-progress --no-suggest
+
+      - name: PHPCS check
+        uses: chekalsky/phpcs-action@v1
+        with:
+          phpcs_bin_path: './vendor/bin/phpcs'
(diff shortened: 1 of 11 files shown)


phpcs.xml
```diff
@@ -0,0 +1,39 @@
...
+    <config name="ignore_warnings_on_exit" value="0"/>
+    <config name="default_standard" value="Drupal"/>
+
+    <arg name="extensions" value="php,module,inc,install,test,profile,theme" />
+
+    <exclude-pattern>*/node_modules/*</exclude-pattern>
+    <exclude-pattern>*.css</exclude-pattern>
+    <exclude-pattern>*/assets/*</exclude-pattern>
+    <exclude-pattern>*/components/*</exclude-pattern>
+    <exclude-pattern>*social.info.yml</exclude-pattern>
```
bramtenhove, line 20:
```suggestion
    <exclude-pattern>social.info.yml</exclude-pattern>
```

Shouldn't this be the specific file instead of a wildcard in front of it?

phpcs.xml
```diff
@@ -0,0 +1,39 @@
...
+    <exclude-pattern>*.md</exclude-pattern>
+    <exclude-pattern>PULL_REQUEST_TEMPLATE\.md</exclude-pattern>
+
+    <config name="drupal_core_version" value="8" />
```
bramtenhove, line 24: We'd have to update this to `9` when we get to that right?

.github/workflows/drupalCodingStandards.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  phpcs:
+    name: PHPCS
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      # Install composer depdencies so we have drupal/coder which adds our codesniffer, the bin/phpcs file
+      # and dealerdirect/phpcodesniffer-composer-installer which changes the installer_path configuring all sniffs.
+      - name: Install dependencies
+        run: composer install --dev --prefer-dist --no-progress --no-suggest
```
bramtenhove, line 14: Do we install all composer dependencies here? Or just Open Social?

phpcs.xml
ronaldtebrake: Yep, but that is in a separate PR which we havent reviewed yet!

.github/workflows/drupalCodingStandards.yml
ronaldtebrake: its running in the root, so it's adding all from the distributions composer.json including the dev. (There is no such a thing as just installing dev without the normal unfortunately)

phpcs.xml
ronaldtebrake: Not sure, got this from our check-coding-standard scripts, we can revise those later same as the other excluded patterns.

.github/workflows/drupalCodingStandards.yml
bramtenhove: Check, ok!

PR #2102, opened 2020-12-10
Issue #3186003 by navneet0693: DS-7431 - Updated fieldsets with new labels and re-arranged fields in CT Topic

navneet0693:
## Problem
Rename the fieldset labels and move the field to new fieldsets.

## Solution
1. Rename Content → Basic details
2. Rename Group & Visibility → Access permissions
3. Rename Attachments → Additional details
4. Add Settings
5. Move description field to Basic details
6. Remove Description fieldset
7. Add the following fields to the new “Settings” fieldset: Comment setting, URL Alias, Author information, Published status

## Issue tracker
https://www.drupal.org/node/3186003

## How to test
- [x] Check out this branch, do an update of your DB
- [x] As a LU create a topic
- [x] See that the labels are updated
- [x] Go to Structure -> Content type -> Topic -> Manage form display and check for new "Settings" fieldset.
- [ ] Check that the fields are re-structured.

## Screenshots
image

**Topic - Before**
image

**Topic - After**
image

## Release notes
Will be dealt with in the parent issue.


ronaldtebrake (changes requested):
Hi @navneet0693,

Great work so far, you're off to a good start! I like how the settings are tackled :D 
Couple of remarks to be done.

1. When I enable social tagging, it's not in the correct fieldset yet.
Screenshot 2020-12-14 at 17 47 14

2. The order of fields should be as design, see first screenshot vs second
Type - title - image - description

Screenshot 2020-12-14 at 17 48 00

Screenshot 2020-12-14 at 17 48 00

3. Add a new file should become attachments 
Screenshot 2020-12-14 at 17 49 30

4. And all the fields within the Settings tab should be as per design, so some alteration and updates of weights are necessary there. You can for example just leave out the revision information etc. People can move it back through the UI themselves :D so by the end when I do an update of the database it should look (without the CSS updates) exactly like this 

Screenshot 2020-12-14 at 17 51 19

modules/social_features/social_topic/social_topic.install
```diff
@@ -247,3 +247,40 @@ function social_topic_update_8903() {
...
+
+  // Output logged messages to related channel of update execution.
+  return $updateHelper->logger()->output();
+}
+
+/**
+ * Remove group_topic_description fieldgroup from topic content type.
+ */
+function social_topic_update_8906() {
+  if ($group = field_group_load_field_group('group_topic_description', 'node', 'topic', 'form', 'default')) {
```
ronaldtebrake, line 283: Would be great if we can add the check we discussed, to see if there are any fields attached to this fieldset. 
Eventhough we move the description out, it could be others have added more fields to this.
We don't want to break it for them, we rather give them the option to move that field through the UI to the new fieldsets if they wish to do so.

PR #2106, opened 2020-12-11
Issue #3188242 by rolki: Add filter to display only ongoing events and fix issue with end event date

rollki:
## Problem
An event that is ongoing does not show on the custom content list once it has started 

## Solution
* Add ongoing filter to the `EventContentBlock.php` plugin
* Fix issue with end date of event. Because now it is always working with the `node__field_event_date` table but not with `node__field_event_date_end`

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-4133 
https://www.drupal.org/project/social/issues/3188242 

## How to test
- [x] Create a new dashboard and then add a custom content list block there or create a new custom content list block and then place this block somewhere
- [x] In filters you should see the "Ongoing" option
- [x] Shoose this option
- [x] And you should see only ongoing events

## Screenshots
Снимок экрана 2020-12-11 в 13 45 16
Снимок экрана 2020-12-11 в 13 46 16


modules/social_features/social_content_block/modules/social_event_content_block/src/Plugin/ContentBlock/EventContentBlock.php
```diff
@@ -74,6 +77,15 @@ public function query(SelectInterface $query, array $fields) {
               $range['end'] = new \DateTime('last day of next month 23:59');
               break;
 
+            case 'ongoing':
+              $today = new \DateTime();
```
ribel, line 81: @rollki since we don't use `$today` variable in other places, but just new `\DateTime()`, I think we need to be consistent here as well or replace it to `$today` everywhere.

ribel:
@rollki please create a ticket on drupal.org and check my comment after review.

modules/social_features/social_content_block/modules/social_event_content_block/src/Plugin/ContentBlock/EventContentBlock.php
rollki: Fixed. Variable was removed.

rollki:
@ribel 
Variable was removed, the ticket was created and pasted to the description.

ribel:
@rollki Thanks! Please update PR with new issue number title as well ;)

ribel:
🍒 picked to 8.x and 9.x

PR #2111, opened 2020-12-15
Issue #3188391 by mdeny: Change Text on Group Join Confirmation Page

denysyukm:
## Problem
Text on Group Join Confirmation Page does not meet the requirements.

## Solution
The text should describe what will happen for the user once they join.

## Issue tracker
https://www.drupal.org/project/social/issues/3188391

## How to test
- [x] Text on Group Join Confirmation Page should be: "By submitting this form you will become a member of the group. As a member of the group you will begin to receive notifications about changes and interactions within the group. Your profile will also be included in group membership."


ribel (approved):
Changes look good to me.
Also, the text was proposed by Zach and confirmed by Xinyu and Stefano.

@ronaldtebrake I just wondering since other distro users might actually have some fields on group join - then maybe we need to include this change just to 10.0 release.
Or do you you think we need to backport it?

ronaldtebrake:
Hi @ribel 

Good question! I think we can backport it in this case, its a better text even if there are fields on this confirmation page.
Don't forget the translations.php in that case :D

ribel:
@denysyukm please add text of old string to the `translations.php` file.

denysyukm:
@ribel , I updated translations.php file.

ribel:
🍒 picked to 9.x

PR #2113, opened 2020-12-17
Issue #3189357 by rolki: Add the ability to invite users via email

rollki:
## Problem
Previously, we were able to copy/paste an email address and it would easily find the user. Now, email seems to not work, but when I delete and only use username it seems to work fine

## Solution
Add the ability to search users via email in the group invitation form

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-4150 
https://www.drupal.org/project/social/issues/3189357 

## How to test
- [ ] Go to some group manage members page
- [ ] Click "Add directly"
- [ ] Try to add a user via email
- [ ] Save the form
- [ ] The user that was added via email should be as a group member

## Screenshots
image
image


ribel (approved):

ribel:
@ronaldtebrake I see you worked on Group Invite module, so wondering if there are any objections against merging this PR?
I think there was the ability to add a user by email before, but now it's missing.
Also, we removed the description, to not confuse and reflect current state just by title.

ronaldtebrake:
Hi @ribel and @rollki 

Thanks for this, this seems to work well and it's a nice and consistent solution.

1) I have thought about it a bit and I think this should be restricted when social profile privacy is enabled only to those who can see a profile field even if its hidden. 

Yes, people can still be found in search by e-mail, even when it's hidden. 
But now we provide an easy way to guess this using autocomplete.

typing `drupal_social+c` nicely shows chris hall, if you dont know the rest you can just try typing a letter and see if you get a resultset back. 

2) Most likely it now also works on multiple places, since you updated all the autocompletes using:

```
      '#selection_handler' => 'social',
      '#target_type' => 'user',
```

With that in mind it would be great to update the consistent label / descriptions there too. I have found at least the below:

Screenshot 2020-12-28 at 16 04 54

Afterwards can be merged 👍

rollki:
@ronaldtebrake 
So I checked all social #selection_handler with the following result:
* InviteEmailBaseForm class - without any changes since the description is correct, we are able to paste multiple entries and title also fine.
* SocialBulkGroupInvitation class - without any changes since the description is correct, we are able to paste multiple entries and title also fine.
* SocialEventManagersAddEnrolleeForm class - changed the title and removed description since it doesn't valid.
* EnrollInviteUserForm class - changed the title since we are able to find by email also.

All cases were checked on ECI Demo.

ribel:
@rollki What about the first question? 
I think we can add a simple check if the current user has access to administer people - if yes, then we will also add a condition to search by email, what do you say @ronaldtebrake?

rollki:
It seems only SM has the `administer users` permission, so we can use this permission, but if CM also should have an access to find users via email then we should use another permission.

rollki:
I found another permission `view profile email` so we can try to use this one.
cc @ronaldtebrake, @ribel

ribel:
I think that should be enough, and so ready to merge now.
Any concerns @ronaldtebrake?

ribel:
🍒 picked to 9.x

PR #2127, opened 2020-12-23
Issue #3189637 by rolki: Add comments pagination to the LazyRenderer social service

rollki:
## Problem
There is no way to see pagination for comments in topic or discussion or all other content types

## Solution
Add pagination to comments build in the SocialCommentLazyRenderer service

## Issue tracker
- https://www.drupal.org/project/social/issues/3189637
- https://www.drupal.org/project/drupal/issues/3189538
- https://getopensocial.atlassian.net/browse/YANG-4174

## How to test
- [ ] Go to the comments edit field page (E.G. for topic type) and set the limit for view in 5 items e.g.
- [ ] Go to some topic and add at least 6 comments
- [ ] You should see only 5 comments and there is no way to see the last one

## Screenshots
After implementation:
image


dmytriikaiun (approved):

dmytriikaiun:
@rollki looks good to me

ribel (approved):

ribel:
Hi, @ronaldtebrake,
We were able to fix the issue with comments pager.
For this, we also were needed to create and add a patch for drupal/core.
Also, we found that pager doesn't work correctly with AJAX comments.
We already created follow up ticket for that, but not sure if we will able to fix it this sprint.
Do you think we can merge this bugfix without AJAX fix, or both needs to be fixed in one PR?

ronaldtebrake:
Hi @ribel,

It's okay to merge it separately so people not using ajax comments don't have to wait :) as long as the PR / Drupal.org ticket reflect that people are aware its a known issue.

ribel:
@ronaldtebrake I have created a separate issue on drupal.org: https://www.drupal.org/project/social/issues/3190205

modules/social_features/social_comment/social_comment.module
```diff
@@ -360,3 +360,24 @@ function social_comment_field_info_alter(&$info) {
...
+/**
+ * Implements hook_preprocess_pager().
+ */
+function social_comment_preprocess_pager(&$variables) {
+  $items = &$variables['items'];
+  $pattern = '/%2C.$/';
+
+  // Change href for all pager items except "pages"
+  // (E.G. first, previous, next, last, etc.).
+  foreach ($items ?? [] as $key => $item) {
```
ribel, line 373: Need to add check if $items are not empty because sometimes it throws errors.

rollki: In this case, items can not be empty, because they exist or not, so it's can be null or array.

rollki: Fixed.

ribel:
I was not able to cherry-pick it to `9.x` due to conflict in `composer.json` file. Also don't resolve it manually, since it still may cause conflicts in future. I think it's fine to have this code just in `10.0.x` since it will be released soon.

PR #2128, opened 2020-12-23
Developer task - Update the custom step.

ronaldtebrake:
## Problem
Whenever you embed a link in an iFrame sometimes it strips out the query fragments/attributes resulting in the src not completely matching what you put in.

That could result in adding a link like:
`https://www.youtube.com/embed/lh3xOx8eutQ"

But the link inside the iFrame contains

`src="https://www.youtube.com/embed/lh3xOx8eutQ?feature=oembed"`

## Solution
Try to not exactly match the source, but see if the source contains.


ronaldtebrake:
The test passed at least once.

PR #2141, opened 2021-01-05
correct permission

slowflyer:
## Problem
*Describe the problem you're trying to solve, this should motivate why the changes you're proposing are needed.*

The route to the admin pages of Social Font (/admin/config/fonts) uses the general permission 'access administration pages' instead of the more granular 'administer font entities' permission.

## Solution
*Describe the solution you've created, elaborate on any technical choices you've made. Why is this the right solution and is a different solution not the right one?*

Replace 'access administration pages' by 'administer font entities'

## Issue tracker

https://www.drupal.org/project/social/issues/3191095

## How to test
*For example*
- [ ] As a sitemanager

## Screenshots

## Release notes

## Change Record

## Translations

### DIFF
--- a/modules/custom/social_font/social_font.routing.yml
+++ b/modules/custom/social_font/social_font.routing.yml
@@ -3,4 +3,4 @@ social_font.fonts:
   defaults:
     _title: 'Fonts'
   requirements:
-    _permission: 'access administration pages'
+    _permission: 'administer font entities'


navneet0693 (approved):

navneet0693:
Thank you @slowflyer for this PR :-)

PR #2153, opened 2021-01-13
Issue #3192493 by sjoerdvandervis: Canonical URL for flexible group points to stream

sjoerdvandervis:
## Problem
The new "See group information” command (introduced with the Skytheme) does not point anywhere, and ends up misleading users. It should point to the /about section of the group.

## Solution
Change the canonical URL for the group, so that this link points to the about page.

## Issue tracker
https://www.drupal.org/project/social/issues/3192493

## How to test
*For example*
- [x] Create a flexible group
- [x] On the group stream, click the link "See group information" in the statistics block
- [x] See that you end up on the /about page of that specific group.

## Release notes
We changed the link of the Flexible group's statistics block to point to the about page of the group, instead of the stream.


navneet0693 (approved):
Works like charm even with Social Group Default Router, Social Path Manager and Path Auto modules 👍

navneet0693:
Cherry-picked to 8.x-9.x

PR #2154, opened 2021-01-13
by jochemvn: fix notice on almost every page load

jochemvn:
## Problem
There's a notice on almost every page load: `Notice: Trying to access array offset on value of type null in Drupal\social_group\SocialGroupHelperService->getGroupFromEntity() (line 69 of profiles/contrib/social/modules/social_features/social_group/src/SocialGroupHelperService.php).`

## Solution
Check if the protected cache is an array itself, before trying to access its value.

## How to test
- [ ] Turn on notices and clear caches
- [ ] Reload the homepage
- [ ] See you get the notice described
- [ ] Checkout this branch and clear caches again
- [ ] See that the notice is now gone.


navneet0693 (approved):
**Before**
image

**After**
image

PR #2095, opened 2020-12-08
#3187257 by sjoerdvandervis: show thumbnails on image uploads on topic / events / pages

sjoerdvandervis:
## Problem
When a platform allows uploading images as attachments with for instance topics, these files are not shown as thumbnails, so a user has no clue when images have been uploaded what the content of the image is.

## Solution
We move the already existing FieldFormatter in `social_comment_upload` to `social_core` so that it is always available. Furthermore, we update the FieldFormatter wherever we use the field_files as field on content types and we've added the possibility to add images as attachments to Events, Topics & Pages.

## Issue tracker
https://www.drupal.org/project/social/issues/3187257

## How to test
- [x] See that you can attach images to a topic, event or page;
- [x] Create a new topic and add different attachments to it. Include an image file
- [x] Visit the newly created topic and see that for the images, the thumbnail is shown.

## Screenshots
Screenshot 2020-12-08 at 15 14 20

## Release notes
We've included the option to add images to events, topics and pages as attachments. When uploading images as attachments to events / topics / pages, they will now be shown as thumbnails instead of a regular file icon.

PR #2156, opened 2021-01-13
Issue #3192510: Name for files in the topics.

Ressinel:
## Problem
The file names functionality for the topics is currently a bit limited. If the name of my file is a bit long, it is not possible for the user to read the full name. It would be good if there was more space for showing the file name, or if by hovering the cursor on top of the file, the user could see the full name of the file.

## Solution
- We can make the title visible in 3 lines instead of 2, so more text can be displayed
- Add hover text that shows the full title of the document

## Issue tracker
- https://www.drupal.org/project/social/issues/3192510
- https://getopensocial.atlassian.net/browse/YANG-3796

## How to test
In the example from the picture below, I have three files whose names start with "Interactive roundtable on...". Although they have 3 different names, the users have no way to know which is which.

## Screenshots
image

### DIFF
--- a/themes/socialbase/assets/css/file.css
+++ b/themes/socialbase/assets/css/file.css
@@ -87,7 +87,7 @@
 .card-file__link {
   display: block;
   position: relative;
-  height: 100px;
+  height: 145px;
   padding: 12px;
   color: inherit;
   text-decoration: none;
@@ -97,7 +97,7 @@
   overflow: hidden;
   position: relative;
   line-height: 1.35em;
-  max-height: 2.7em;
+  max-height: 8em;
   padding-right: 1em;
   display: block;
   word-break: break-word;
--- a/themes/socialbase/templates/file/file-link--card.html.twig
+++ b/themes/socialbase/templates/file/file-link--card.html.twig
@@ -17,7 +17,7 @@
 {% spaceless %}
 
   <li class="card-file">
-    <a class="card-file__link" href="{{ link.url.uri }}" title="{{ 'Open or download file'|t }}" target="_blank">
+    <a class="card-file__link" href="{{ link.url.uri }}" title="{{ link.text|replace({'_': ' '}) }}" target="_blank">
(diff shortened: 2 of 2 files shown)


ribel (approved):
Changes look good to me.

ronaldtebrake:
Hi guys,

Thanks for your work.

This does require a little bit more work though, as the 100px was chosen for images. Images attached as files are rendered as using the thumbnail image style (100x100).. with this change images and files won't match in size anymore.

Ressinel:
> Hi guys,
> 
> Thanks for your work.
> 
> This does require a little bit more work though, as the 100px was chosen for images. Images attached as files are rendered as using the thumbnail image style (100x100).. with this change images and files won't match in size anymore.
> 
> See f.e. #2095

Hi @ronaldtebrake 

Thanks for your response!

We have tested on OS 10.0.x with current PR changes and have next result: 

topic-files

Can you share any screenshots how it looks on your side?

Thanks!

ronaldtebrake:
Hi @Ressinel @ribel 

So the idea is, that we can show a Thumbnail if it's an image. 
That's why I linked to https://github.com/goalgorilla/open_social/pull/2095 where we have added that feature.
But it seems in 10.0.x there is some regression which removed this again.

I'll ask @navneet0693 to check it out, as this should be File and image in a table as display setting.


Screenshot 2021-01-25 at 10 39 51

navneet0693:
Hi @ronaldtebrake ,

I tried this branch as a fresh install and also upgrading from 9.x to this branch, both of them resulted the same and I ended up with:

image

It is the same on 10.0.x as well.

navneet0693:
We now understand what the difference is so the PR https://github.com/goalgorilla/open_social/pull/2095 is only added in 10.0.x and this PR will land in 9.7, so what we will do is merge this and fix the image thumbnails in a different pull request for 10.0.x

navneet0693 (approved):

navneet0693:
🍒  picked to 8.x-9.x

PR #2157, opened 2021-01-13
DS-7429: Add updates for the field and fieldset labels

agami4:
## Problem
The fields and fieldsets look and feel are the same as the design for all content types.

## Solution
Add updates for the field and fieldset labels.

## Issue tracker
https://getopensocial.atlassian.net/browse/DS-7429

## How to test
*For example*
- [ ] Go to any create page and check how looks labels(should have font-weight: 700).

## Screenshots
new-labels

## Release notes
The field and fieldset labels have font-weight 700 - its part of the content form improvements and dont need specific release notes.


agami4:
@ronaldtebrake, one thing that I did not add its labels on the `Date & time` block on the event create page(see screenshot).
event-page
Should remove or update styles on the `social_event` module. I've added styles only for socialbase/blue themes

ronaldtebrake:
Hi @agami4 thank you for the update, I think it's OK to change this within social_event. 
Not sure what is the best, do you think removing the styles there is a good idea? I'm ok with that!

agami4:
@robertragas, we can remove only the `font-weight` style for the labels, not all styles😊

ronaldtebrake:
@agami4 ah yeah sorry i was indeed talking about just the `font-weight`, ideally it's not part of the module but part of the theme. Because the font-weight will be set on a theme base rather than a module base. 😊

PR #2160, opened 2021-01-14
Issue #3192752 by rolki: Fix issue with incorrect number of comments

rollki:
## Problem
As a site manager, I can see the correct number of comments for a group, but a normal user only sees 1 in the comment counter for a discussion on a group page.

## Solution
When we use the comment_query variable to add conditions in foreach then it works only for the first item of the array, for all other items we get an empty array. But when we use comment_storage and then call getQuery methon and then add conditions to the query then everything works properly.

## Issue tracker
- https://www.drupal.org/project/social/issues/3192752
- https://getopensocial.atlassian.net/browse/YANG-4628

## How to test
- [ ] Create some flexible group with discussion enabled
- [ ] Create some discussions for this group
- [ ] Add a few comments to this discussion
- [ ] Logged in as a normal member
- [ ] Go to the group about page
- [ ] You should see the list of discussions at the bottom of the page with the incorrect number of comments, also the incorrect number on the discussion itself

## Screenshots
**Issue:**
image
image

**After fix:**
image
image


ribel (approved):

ribel:
🍒 picked to 9.x

ribel:
@rollki I've also checked that we have the same line in `\Drupal\activity_creator\ActivityFactory::getAggregationRelatedActivities`
```
$comment_query = $this->entityTypeManager->getStorage('comment')->getQuery();
```

Should we fix it there to the same variant as here?

PR #2126, opened 2020-12-22
Issue #3189520 by chmez: As a user I want to see logo in the footer

OH17L:
## Problem
As a user, I want to see a logo in the footer.

## Solution
Should be possible to replace: logo, text, link.

## Issue tracker
- https://www.drupal.org/project/social/issues/3189520
- https://getopensocial.atlassian.net/browse/ECI-1113

## How to test
- [ ] When you enable the "Social Footer" module then you should see a block with the following text in the footer section of a page: "powered by Open Social".
- [ ] When you click on that text then you should be redirected to [getopensocial.com](https://getopensocial.com/).
- [ ] Log in as SM+ and navigate to the "/admin/structure/block/manage/socialblue_footer_powered" page.
- [ ] You should see the following fields: "Logo", "Text", "Link".
- [ ] The "Logo" field should be empty.
- [ ] The "Text" field should contain the following text: "powered by Open Social".
- [ ] The "URL" sub-field of the "Link" field should contain the following link: "https://getopensocial.com/".
- [ ] The "Title" sub-field of the "Link" field should contain the following text: "Community Engagement Platform Software".
- [ ] Upload an image in the "Logo" field and press "Save block".
- [ ] When you navigate to the front page then you should see uploaded image in the footer section and this image should be clickable.

## Release notes
There is now an easy option to add your organizations logo to the footer next to the footer menu.

PR #2165, opened 2021-01-18
Issue #3189520 by navneet0693: Add default logo to social footer & settings form for Site Manager.

navneet0693:
## Problem
Currently, we have this block installed by default, and users with setting tray module permissions can easily edit this. 
We don’t want users to have that permission as it will add a lot of pencil icons all over the place.
So, we rather focus on getting a SM settings form, and let the block get the data out of that settings form. Also, add a default branding logo in footer.

## Solution
1. Add a settings form for site manager.
2. Add a default logo.

## Issue tracker
https://www.drupal.org/project/social/issues/3189520
https://getopensocial.atlassian.net/browse/DS-7518
https://getopensocial.atlassian.net/browse/DS-7520

## How to test
- [ ] Install version 10.0.x of Open Social
- [ ] Check if the footer block has logo by default after the fresh installation.
- [ ] As a sitemanager, go to Configuration -> Open Settings -> Footer block settings
- [ ] Check if the form contains the default values
- [ ] Make some changes and check if it is reflected in the footer block.

## Screenshots
image

## Release notes
Can be picked up from https://github.com/goalgorilla/open_social/pull/2126


ronaldtebrake:
Thanks @navneet0693 looks good!

PR #2168, opened 2021-01-20
Issue #3177673 by chmez: As outsider you can create / see albums on closed group you're not a member off

ghost:
## Problem
As outsider you can create / see albums on closed group you're not a member off.

## Solution
Re-use access checker implementation from group events/topics page for group albums page.

## Issue tracker
- https://getopensocial.atlassian.net/browse/YANG-4776

## How to test
- [ ] When you open a closed group page and you aren't a member of this group then you shouldn't see the "Albums" tab.
- [ ] When you don't have access to the group albums page then you shouldn't have access to the group album creation page (/group/<group id>/albums/add).

## Release notes
Improve checking access to group albums page and group album creation page.


modules/social_features/social_album/config/install/views.view.albums.yml
```diff
@@ -946,7 +946,8 @@ display:
             enabled: '1'
       access:
         type: social_album_group
-        options: {  }
+        options:
+          group_permission: 'view group_node:topic content'
```
ribel, line 950: @chmez Should we use album instead of topic here?

modules/social_features/social_album/config/update/social_album_update_8901.yml
```diff
@@ -0,0 +1,22 @@
...
+            contexts:
+              - route.group
+              - user.group_permissions
+    change:
+      display:
+        page_group_albums_overview:
+          display_options:
+            access:
+              options:
+                group_permission: 'view group_node:topic content'
```
ribel, line 22: @chmez Should we use album instead of topic here?

ribel (changes requested):
@OH17L I have tested it and checked code changes and all looks good as for me.
Just please change 2 topic things that I commented.

modules/social_features/social_album/config/install/views.view.albums.yml
OH17L: Fixed.

modules/social_features/social_album/config/update/social_album_update_8901.yml
OH17L: Fixed.

ribel (approved):
All looks good for me now.

ronaldtebrake:
This looks way better, thanks guys. Also false negative on the PR, so can merge it either way.

PR #2179, opened 2021-01-25
Issue/3194414 by navneet0693: Removed socialbase & socialblue theme from Distro.

navneet0693:
## Problem
You can read more information [here](https://www.getopensocial.com/blog/open-source/change-coming-decoupled-open-social).
On our road to going decoupled, we want to make sure Open Social becomes available as a Headless product.
This means we have to move our Socialbase & Socialblue themes out of the Distro as they are not a necessary dependency for this product.

However, as we are promising an out of the box featured community solution we want to make this front-end still available for those users who don't want to use our headless back-end.

## Solution
1. Contributed Socialbase & Socialblue to drupal.org
2. Add them as dependencies to Open Social

## Issue tracker
https://www.drupal.org/project/social/issues/3194414

## How to test
- [ ] See that the themes are now live as a contrib project on drupal.org
- [ ] See that the dependency is declared in Open Social's composer.json.
- [ ] See that contrib themes have necessary changes such as including a composer.json with proper dependencies.
- [ ] Do a composer install and check the themes are downloaded.
- [ ] See that themes are installed by default in Open Social after a fresh install.
- [ ] On existing website, it should not break the website after clearing the cache.

## Screenshots
Not required.

## Release notes
We have moved our both socialbase and socialblue themes as contrib themes on drupal.org and have added it as a dependency in Open Social.

https://www.drupal.org/project/socialblue
https://www.drupal.org/project/socialbase

## Change Record
The themes **socialbase** & **socialblue** are now added as dependency in Open Social's composer.json which means after a composer install it will be moved to module/contrib folder.

The directory changes might affect the website or platform where themes are already enabled and may lead to crashing of website. Please clear the cache before doing the update, you can also follow issue queue at https://www.drupal.org/project/drupal/issues/2651144. We have not opted to clear the cache in hook update as it is not a recommended practice to do, please read more here: https://www.drupal.org/node/2960601, also because hook_post_update_N will help us in this case.

Also, it avoid issues during the deployment please include the following statement in your settings.php:

`
// Deployment identifier.
$settings['deployment_identifier'] = \Drupal::VERSION;
`

You can read about it more here: https://git.drupalcode.org/project/drupal/-/blob/9.1.x/sites/default/default.settings.php#L295

## Translations
N.A


navneet0693:
We are going to merge, tests failures are false negative as it failed in files attachments test.

PR #2184, opened 2021-01-25
Issue #3193963 by tBKoT: Fixes for follow taxonomy

dmytriikaiun:
## Problem
 - Updated the design for the `Add tag` section on the landing page.
 - Update info.yml to reflect what the different modules exactly do
 - The "following bar" on the search disappears when sorting 
> After applying a sort or filter to the result set, the following bar disappears while I'm still looking at the other content
 - The `field_tag` is unknown
> After adding a landing page, then enabling social_follow_landing_page, and trying to add a tag to the section I get a
```
a:6:{s:5:"%type";s:24:"InvalidArgumentException";s:8:"@message";s:27:"Field field_tag is unknown.";s:9:"%function";s:58:"Drupal\Core\Entity\ContentEntityBase->getTranslatedField()";s:5:"%file";s:63:"/var/www/html/core/lib/Drupal/Core/Entity/ContentEntityBase.php";s:5:"%line";i:587;s:17:"@backtrace_string";s:6149:"#0 /var/www/html/core/lib/Drupal/Core/Entity/ContentEntityBase.php(568): Drupal\Core\Entity\ContentEntityBase->getTranslatedField('field_tag', 'x-default')
/#1 /var/www/html/profiles/contrib/social/modules/social_features/social_follow_taxonomy/modules/social_follow_landing_page/social_follow_landing_page.module(169):
```
 - Enabling on clean Saas gives errors about message templates
> drush en social_follow_tag
The following extensions will be enabled: social_follow_tag, social_follow_taxonomyDo you really want to continue? (y/n): y
Drupal\Core\Config\PreExistingConfigException: Configuration objects (message.template.create_node_following_tag, message.template.create_node_following_tag_stream, message.template.update_node_following_tag) provided by social_follow_tag already exist in    [error]active configuration in /var/www/html/core/lib/Drupal/Core/Config/PreExistingConfigException.php:65

## Solution
Fix issues

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-4715
https://getopensocial.atlassian.net/browse/YANG-4747
https://getopensocial.atlassian.net/browse/YANG-4751
https://getopensocial.atlassian.net/browse/YANG-4754
https://getopensocial.atlassian.net/browse/YANG-4756

## How to test
N/A

## Screenshots
image

## Release notes
Now user can select only one tag on the landing page is the `Add tag` section.

## Change Record
N/A

## Translations
N/A


modules/social_features/social_follow_taxonomy/social_follow_taxonomy.module
```diff
@@ -467,3 +476,73 @@ function social_follow_tag_activity_send_email_notifications_alter(array &$items
...
+ * @param \Drupal\node\NodeInterface $node
+ *   The node entity.
+ */
+function _social_follow_taxonomy_invalidate_node_tagging_cache(NodeInterface $node) {
+  // Only for nodes that has the 'social_tagging' field.
+  if (!$node->hasField('social_tagging')) {
+    return;
+  }
+
+  if ($node->isNew()) {
```
ribel, line 513: @dmytriikaiun could you explain this code a little bit more?
Wondering if it will not have a performance impact since we are checking om each node change?
Also see `social_follow_tag_entity_update()` - can we combine it somehow, or it's not really related?

ribel (commented):
Please check my comments and make sure CS and test pass.

modules/social_features/social_follow_taxonomy/social_follow_taxonomy.module
dmytriikaiun: @ribel I guess `social_follow_tag entity update` will not enough because it will just cover the case when we edit the existing node but also we should clear cache when adding a new node with tags to increase following items in follow term block and decrease when deleting a node with tags.

modules/social_features/social_follow_taxonomy/modules/social_follow_landing_page/social_follow_landing_page.module
```diff
@@ -88,12 +92,16 @@ function social_follow_landing_page_preprocess_paragraph(&$variables) {
...
-    $subform = &$element['subform']['field_section_paragraph']['widget'][0]['subform'];
-  }
-  elseif (isset($element['subform']['field_tag'])) {
+  if (isset($element['subform']['field_tag'])) {
     $subform = &$element['subform'];
   }
+  elseif (isset($element['subform']['field_section_paragraph']['widget'])) {
+    $child_keys = Element::children($element['subform']['field_section_paragraph']['widget']);
+    $child_key = reset($child_keys);
+    if (isset($element['subform']['field_section_paragraph']['widget'][$child_key]['subform'])){
```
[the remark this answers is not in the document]
dmytriikaiun: @ribel Fixed

navneet0693:
@dmytriikaiun I have following the steps below:

1. Fresh install Open Social 10.0.x
2. Checkout to this branch
3. Enable the social_follow_landing_page module
4. Create a landing page
5. Add section -> Add tag
However, I don't see any thing inside the block, please see screenshot:

image

navneet0693:
Ok so I was missing tags in hierarchal form of category and children. This condition is checked here: https://github.com/goalgorilla/open_social/pull/2184/files#diff-a8c370e5717b59c7b6de2108024386d9d0fb11ea961cd003bbdea556ce647416R110

 
image

modules/social_features/social_follow_taxonomy/social_follow_taxonomy.module
ronaldtebrake: Hi guys, good to see this discussion going on in here!
Did you already investigate if we can use a custom submit for this? This is way less impactful, and either adding, editing, or deleting a tag will be dealt with when passing this along correctly.
Also don't forget, we allow tags on entities (so more than just a node) like a group or a profile even I believe, so going for the node route isn't the best.

dmytriikaiun: @ronaldtebrake for now on the `Follow tag` block counter shows only related nodes and `See related content` redirects to the content search page. To add groups, profiles, etc. we need to change the counter logic, redirect users to the search all page, and add tag filters on it. If as `custom submit` you mean that we need to add it when the form has the `social_tagging` field it will not work on delete.

ronaldtebrake: Hi @dmytriikaiun it's not just about the block isn't it? 🤔 

Screenshot 2021-02-01 at 11 08 55

Don't we want to invalidate caches for social_tagging regardless of this block? 

I'm ok with the block only being available on nodes, but as you can see in the screenshot we already add this as a field on a group, and the popup also works. Which redirects you nicely to the group search, with the social_tagging filter. But only after cache clear somethings happen. Which to me sounds like this isn't really done correctly.

dmytriikaiun: @ronaldtebrake I've added cache tags for the `Group tag` block. Also, fix the counter because even for groups it calculates related nodes.

modules/social_features/social_follow_taxonomy/modules/social_follow_tag/social_follow_tag.install
```diff
@@ -42,3 +43,35 @@ function social_follow_tag_update_8801() {
     }
   }
 }
+
+/**
+ * Update followtag block.
+ */
+function social_follow_tag_update_8802() {
+  $config_path = drupal_get_path('module', 'social_follow_tag') . '/config/optional';
```
ronaldtebrake, line 51: It's not a good idea to load these config files from optional / install (in the below update hook)
rather move a copy to /static which never changes, so you don't have any issues when these yml files change and the update hook might not have run

ronaldtebrake:
The PR seems to solve the YANG tickets, please next time make sure to explain what I need to test in the HTT that helps me review these PR's a lot

PR failed due to rate limit, so will accept for now with that being red. Rest passed.

PR #2189, opened 2021-01-27
YANG-4490: Make the prepareTermOptions as public method

rollki:
## Problem
It is not possible to use the `prepareTermOptions` method from the `SocialTaggingService` somewhere else outside the service.

## Solution
Make the method public.

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-4490


rollki:
Not needed anymore.

PR #2195, opened 2021-01-27
Issue #3172477 by chmez: Add page title when adding post to image to explain users what they are doing

ghost:
## Problem
Album post creation page is using page title block as node page.

## Solution
Use simple title instead.

## Issue tracker
- https://www.drupal.org/project/social/issues/3172477
- https://getopensocial.atlassian.net/browse/YANG-4777

## How to test
- [ ] When you navigate to the page for creating a post for an album then you should see the simple page title.
- [ ] When you redirect to the post creation page after creating a new album then you should see the following message: "Album <ALBUM_TITLE> is successfully created. Now you can add images to this album.".

## Release notes
Use simple title on the album post creation page.


modules/social_features/social_node/src/SocialNodeForm.php
```diff
@@ -0,0 +1,101 @@
...
+      $container->get('social_node.messenger')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function save(array $form, FormStateInterface $form_state) {
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = clone $this->entity;
```
ronaldtebrake, line 80: Why do we need a cloned version of this?

ronaldtebrake:
Accepted this for now as its a good improvement of the UX for adding albums

modules/social_features/social_node/src/SocialNodeForm.php
OH17L: It's needed for a new hook because when a node isn't cloned then node always determining as a not new node in the new hook.

PR #2197, opened 2021-01-28
Issue #3195069 by navneet0693: Link account display name in account header block to user profile page.

navneet0693:
## Problem
After a user has logged in and it clicks on its user profile icon it is presented with "Signed in as" and followed by current user's display name on next line. it would really great if it becomes clickable and links to user profile page.

## Solution
Link account display after 'Signed in as' in AccountHeaderBlock to user profile page.

## Issue tracker
https://www.drupal.org/project/social/issues/3195069

## How to test
- [ ] Login in as LU
- [ ] Click on user profile icon in account header block
- [ ] It should link to user profile page

## Screenshots
image

## Release notes
We have updated the account header block which contains the user menu. Now, when a user will click it's own display name it will be link to it's own profile page.

## Change Record
N.A

## Translations
N.A


ronaldtebrake:
False negative on the testing, tested it locally and it's ok. Will merge.

ronaldtebrake:
also cherry-picked to 9.x 🍒

PR #2200, opened 2021-01-29
Issue #3177673 by chmez: WSOD when user creates an album and assigns it to course or challenge

ghost:
## Problem
WSOD when a user creates an album and assigns it to course or challenge.

## Solution
Exclude group types which don't support current content type in the group field on the content type creation/edit page.

## Issue tracker
- https://getopensocial.atlassian.net/browse/YANG-4806

## How to test
- [ ] Enable the [Social Course](https://www.drupal.org/project/social_course) module and create a course.
- [ ] When you navigate to the user albums page and press "Create new album" then you shouldn't see created course in the "Groups" field.

## Release notes
Exclude group types which don't support current content type in the group field on the content type creation/edit page.

### DIFF
--- a/modules/social_features/social_album/social_album.routing.yml
+++ b/modules/social_features/social_album/social_album.routing.yml
@@ -5,7 +5,7 @@ social_album.add:
     _title_callback: '\Drupal\node\Controller\NodeController::addPageTitle'
     node_type: album
   requirements:
-    _custom_access: '\Drupal\social_album\Controller\SocialAlbumController::checkGroupAccess'
+    _custom_access: '\Drupal\social_album\Controller\SocialAlbumController::checkGroupAlbumAccess'
     _node_add_access: 'node:{node_type}'
     group: \d+
   options:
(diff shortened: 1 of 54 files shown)


ronaldtebrake:
Hi @OH17L when trying to figure out if it would still work it seems the visibility is not updated once selecting the group field, not sure if this PR is causing that regression but it seems related to this. 

Screenshot 2021-02-02 at 17 33 58

It doesn't work when updating or adding content.

ronaldtebrake:
> Hi @OH17L when trying to figure out if it would still work it seems the visibility is not updated once selecting the group field, not sure if this PR is causing that regression but it seems related to this.
> 
> Screenshot 2021-02-02 at 17 33 58
> 
> It doesn't work when updating or adding content.

Nevermind I have figured out we didn't implement that for albums yet, and will fix this part separately.

ronaldtebrake:
Seems like no impact on the Distro but helpfull for other group types, so will accept it as its tested correclty on ECI demo with the extensions :)

PR #2201, opened 2021-01-29
Issue #3195438 by tBKoT: Improves entity validation in the follow tag activity context

dmytriikaiun:
## Problem
inside the `isValidEntity` method of the `FollowTagActivityContext` and `FollowTaxonomyActivityContext` we check for every user who follows a specific tag if he has access to view that entity. That is has a performance issue on saving and update nodes, also on creating activity.

## Solution
Instead of checking access to the node for all users, easier to check if the node has a field with relation to taxonomy. Also, add a alter hook to modify the list of fields that can be checked.

## Issue tracker
https://www.drupal.org/project/social/issues/3195438
https://getopensocial.atlassian.net/browse/YANG-4870

## How to test
N/A

## Screenshots
N/A

## Release notes
N/A

## Change Record
Added a new hook `hook_social_follow_taxonomy_fields_alter` that modifies the list of fields that needs to be checked in the activity context.

## Translations
N/A


modules/social_features/social_follow_taxonomy/modules/social_follow_tag/src/Plugin/ActivityContext/FollowTagActivityContext.php
```diff
@@ -70,51 +79,4 @@ public function getRecipientsWhoFollowTaxonomy(array $related_entity, array $dat
     return $recipients;
   }
 
-  /**
-   * {@inheritdoc}
-   */
-  public function isValidEntity(EntityInterface $entity) {
```
ronaldtebrake, line 76: Can we completely skip this step? 

Don't we need to tell our activity creator that only valid entities are those who can actually be followed?

navneet0693 (approved):

ribel (approved):

ribel:
@dmytriikaiun I see that behat test `@like-post-stream` is failing on step:

```
And I am on "/notifications"                                                                              
Then I should see "Isaac Newton likes your post"
```
From the code, it looks like it might be related to changes made in code.
Could you take a look at this and fix if needed?

cc @navneet0693 @ronaldtebrake Can we wait for a few hours with release 10.0 and try to include this one, as it was important for the performance?

ronaldtebrake:
That seemed to have fix it @dmytriikaiun thanks, the PR failing seems to be false negative, will restart and test locally as well to see if it passes : )

PR #2202, opened 2021-01-31
Issue #3172477 by ronaldtebrake: Fix albums with one image and demo content

ronaldtebrake:
## Problem
1) Posts with an image on existing sites were preprocessed, this resulted in weird looking images especially for posts with only 1 image attached. 
2) We didn't have the possibility to theme different amount of images in a post which is necessary. One image can be rendered full width, while more than one, need to be rendered as square thumbnails.

## Solution
1. Removed the preprocess function, updated the existing image album formatter to update the image styles there when applicable (e.g. only when there are more than 1 images we need to update the image style)
2. Added a new class called single-image to the wrapper, so we can now style this.

## Issue tracker
It's part of the Album feature and fixes these bugfixes:
- https://getopensocial.atlassian.net/browse/YANG-4764
- https://getopensocial.atlassian.net/browse/YANG-4811

## Todo
Theming - Make sure when the wrapper is `<div class="single-image social-post-album--post-gallery">` 
we update the CSS of `.social-post-album--post-gallery .post-with-image` and ensure it's not set to 25% anymore.

## How to test
- [ ] Re-install or Update to see the demo content images are now rendered in the new format
- [ ] See that when you update more than 1 image it still behaves as it used to

## Screenshots
See below, with this update we get the opportunity to change the CSS for only one image.

Before:

Screenshot 2021-01-31 at 14 17 29

After
Screenshot 2021-01-31 at 14 16 21

And when styling is updated (i did this with inspect element)
Screenshot 2021-01-31 at 14 16 46

## Release notes
Not needed.


OH17L (commented):
1. Delete binary files.
2. Fix merge conflicts.

agami4:
@chmez, done

PR #2204, opened 2021-02-01
3195840: Youtube video's are too short in the stream

jochemvn:
## Problem
When embedding youtube movies, the default dimensions for that are 200x113, which is a bit short to actually watch them properly in a stream. The issue is in the embed vendor, on which we rely to embed external media.

## Solution
The problem cannot be fixed very easy from our end, so therefore we've chosen to quickfix this issue, since many users are having issues with this

## Issue tracker
- https://www.drupal.org/project/social/issues/3195840

## How to test
- [ ] Add posts with youtube links
- [ ] Notice the movies are way too short (before screenshot)
- [ ] Switch to this branch
- [ ] Notice it's a lot beter now (after screenshot)
- [ ] Also tryout stuff like twitter links
- [ ] Also notice that on mobile it's not as high as on desktop but still beter watchable

## Screenshots
before:
Screenshot 2021-02-01 at 14 39 31

after:
after

## Release notes
A quick fix to make sure youtube video's are shown less compact in the stream

### DIFF
--- a/themes/socialbase/assets/css/cards.css
+++ b/themes/socialbase/assets/css/cards.css
@@ -69,6 +69,7 @@
 .card__body iframe {
   max-width: 100% !important;
   width: 100% !important;
+  min-height: 300px;
 }
 
 .card__nested-section {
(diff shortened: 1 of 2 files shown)


navneet0693:
This PR fixes the embeds in the stream page but I think it would be really nice if we can extend it to fix the video embeds on content pages.

**Stream page**
image


**Topic**
image

PR #1894, opened 2020-07-03
Issue #3156708 by chmez: Sign in/up with Apple

ghost:
## Problem
Need support sign-in/sign up with Apple in case they have authentification with any other Social accounts.

## Solution
Implement it base on the [Social Auth Apple Sign-In](https://www.drupal.org/project/social_auth_apple) module.

## Issue tracker
- https://www.drupal.org/project/social/issues/3156708
- https://getopensocial.atlassian.net/browse/YANG-3285

## How to test
- [x] Enable the "Social Auth Apple Extra" module and fill required fields on the settings page of the "Social Auth Apple" module.
- [ ] When you open the "Sign up" or "Log in" page as an anonymous user then you should see the "Apple" button in the "Log in with Social Accounts" section.
- [ ] When you press "Apple" then you should be redirected to the "Sign in" page on the "appleid.apple.com" domain.
- [ ] Fill the "Apple ID" and "Password" field and press the button in the 2nd field.
- [ ] Press the "Continue" button on the "Signed in as..." page.
- [ ] You should be redirected to the "Data policy agreement" on your project environment as a logged-in user.
- [ ] When you click "Settings" in the user menu then you should see the "Unlink" button near the "Apple" label in the "Social Log in Connections" section.
- [ ] When you press "Unlink" then you should be redirected to the "Unlink Apple" page.
- [ ] When you press "Unlink" then you should the "Link" button instead of the "Unlink" button near the "Apple" label in the "Social Log in Connections" section.
- [ ] When you press the "Link" button then you should pass steps from 4 to 5 and you should back to the user settings page.
- [ ] You should see the following message: "You are now able to log in with Apple".
- [ ] You should the "Unlink" button instead of the "Link" button near the "Apple" label in the "Social Log in Connections" section.
- [ ] When you log in as administrator and navigate to the "Configuration > Social API settings > User authentication" page from the admin menu then you should see the "Apple" tab.
- [ ] Click "Apple" and uncheck the "Active" checkbox and press "Save configuration".
- [ ] When you log out and open the "Sign up" or "Log in" page then you shouldn't see the "Apple" button.

## Release notes
Add the "Apple" button on the "Sign up" and "Log in" page for authorizing via Apple account.

PR #2205, opened 2021-02-01
Issue #319416 by navneet0693: Removed SSO related modules.

navneet0693:
## Problem
We will remove all of our custom SSO related modules out of the Distro. We have made our custom modules way back in the days when there were no alternatives available yet.

## Why?
We believe there are better alternatives available for you which are better maintained and documented and we rather have Open Social contribute to those modules than to have to maintain our own.

## Solution
1. We will remove `social_auth_extra, social_auth_google, social_auth_facebook, social_auth_linkedin, social_auth_twitter` from the Distro, and will add it as a new unmaintained project.
2. We won't be supporting this module anymore. However we don't mind other maintainers applying.
3. We advice you and any new users to use the listed integrated modules at social_auth

## Issue tracker
https://www.drupal.org/project/social/issues/3194416

## How to test
- [ ] See that the module is now live as a contrib project on drupal.org
- [ ] See that contrib module necessary changes such as including a composer.json with proper dependencies.
- [ ] Do a composer require and check the module is downloaded.
- [ ] See that upgrading from previous branches work if we add the modules to project via composer.

## Screenshots
Not required as it is a code changes which doesn't affect UI.

## Release notes
We have moved the SSO modules (`social_auth_extra, social_auth_google, social_auth_facebook, social_auth_linkedin, social_auth_twitter and social_sso`) as contrib module on drupal.org.

New module page is https://www.drupal.org/project/social_auth_extra

## Change Record
The module **social_auth** shall now be added as dependency in website via composer.json which means after a composer install it will be moved to module/contrib folder.

The directory changes might affect the website or platform where module is already enabled and may lead to crashing of website. Please clear the cache before doing the update, you can also follow issue queue at https://www.drupal.org/project/drupal/issues/2651144. We have not opted to clear the cache in hook update as it is not a recommended practice to do, please read more here: https://www.drupal.org/node/2960601, also because hook_post_update_N will help us in this case.

Also, it avoid issues during the deployment please include the following statement in your settings.php:

`
// Deployment identifier.
$settings['deployment_identifier'] = \Drupal::VERSION;
`

You can read about it more here: https://git.drupalcode.org/project/drupal/-/blob/9.1.x/sites/default/default.settings.php#L295

## Translations
Not required.


ronaldtebrake:
We will create a new PR once #1894 is merged and also contribute this.

PR #2208, opened 2021-02-01
Bugfix/ds 7491 album post visibility

robertragas:
## Problem
When I place a post album in a group album, which is set to only group member visibility as content visibility, it’s still posted as community visibility setting. That shouldn't be possible. 
The post should reflect that it’s only send to group members.

## Solution
Confine albums to the contexts they are in
- Public stream should only show public albums
- Group stream should only show albums related to the specific group (to avoid mixing up secrets groups etc.)

## Issue tracker
10.x not released yet.

## How to test
- [ ] Enable the social_album
- [ ] Create 1 album in the public stream and multiple albums in multiple groups
- [ ] Go to the public stream and see you can only add images to the public album (or a new one)
- [ ] Go the the group stream and see you can only add images to the album that belongs to this specific group.
- [ ] Go to the other group and see it also works this way.

## Release notes
10.x not released yet


robertragas:
Could the person who reviews this double check the 
    if (empty($nids)) {
      $nids = NULL;
    }

I am not sure if this could pose a problem in combination with the condition "NOT IN" and show too much information (could not get this locally though). Reason why I need to run the query is  because a group can have 0 albums in there and then it needs to alter the query in such a way that it gives 0 results (already tried , because if I would not add the where query in the nids are empty, it would show all the groups.

ronaldtebrake:
Unfortunately this doens't seem to work.

Even if you create an album outside of a group, and set it to community, you can still create a post within that album set to public, which is confusing. 

So we probably need to ditch the query alter approach and move towards an ajax solution which updates the visibility field according to the chosen selector.

PR #2210, opened 2021-02-02
Feature #3195382: Ability to add colour setting for Body background colour in the theme settings of social blue theme

swastikpareek:
## Problem
The open social blue theme is quite well integrated with the color module and provides a way to configure brand colors of the site along with some navigation components. 
  There is a need to add a configuration in the color scheme of the OpenSocial's Social Blue theme, to control the page's `<body>` background color, so that along with all the brand colors and navigation section colors, the background color of the page can also be made configurable from the theme settings page of the Social Blue theme.
 This is done to improve the content editor experience and provide the content editor to change the background color of their site

## Solution
The solution is composed of three steps
#### Step 1 - Add a new configuration in the `Color Scheme` section of the Social Blue theme settings page
The `socialblue/color/color.inc` file was updated and a new entry is made in the `colors` array of all the color schemes (incl, default) and `#f3f3f3` is set as the default values. The fields array is also updated to accommodate the field bale info of this new option. 
  This provided a new text-field with the label `Body background` in the color scheme. The theme settings YAML file was also updated to provide initial value (`color_body_bg` is introduced)

#### Step 2 - Make the change in styles in order to make the body bg color configurable by the color module on the FE side
To make the color module do the necessary changes on the body background color styles, the `background-color` property needs to be added in `brand.css` (which is tracked/parsed by color module - see [color.inc](https://github.com/goalgorilla/open_social/blob/8.x-9.x/themes/socialblue/color/color.inc#L95)). So, related styles were added to the `brand.scss` file (and removed from the `01-base/root/_root.scss` file to prevent duplication). The PR contains compiled CSS files.

#### Step 3 - Update the color preview JS to have live preview.
In `socialblue/color/preview.js` JS is updated to live preview the body background color on the previewed HTML as soon as the input value of the `Body background` option changes

### Slight change in the navigation active link
For the open social blue settings page, the body background color is `#f3f3f3`, but then the navigation active link color is also set to `#f3f3f3`. And on changing the body background color, the navigation active link color changes too (on the page but not on the settings page), This happens (I think) is because the color module replaces all the occurrence of `#F3F3F3` including that of navigation active link color in the stylesheet and hence changing the Body background color also changes the active navigation link.
  To fix this we slightly changed the default color of the navigation active link color (to `#f4f4f4`) so that the color module separates the UI components while replacing the color codes.
The navigation active background color is changed to `#f4f4f4` as it doesn’t bring much visual difference on page because visually, `#F3F3F3` almost looks the same as ( as far as Human eyes are concerned) as `#F4F4F4`. 

So, `$navbar-default-link-active-color` variable is updated to `#f4f4f4` in `00-config/variables/_navbar.scss` and all SASS code was compiled again.

## Issue tracker
https://www.drupal.org/project/social/issues/3195382

## How to test
*For example*
- [ ] Login as administrator and reinstall the social blue theme in the Appearance section
- [ ] Check if the color scheme section has a new option with the label `Body background'
- [ ] Change the Body background color to some other value and check if the website background color changes on the web pages

#### Before
Screenshot 2021-02-02 at 12 39 56 PM

#### After
Screenshot 2021-02-02 at 12 44 35 PM

## Release notes
A new option `Body background` is added and navigation link active color changed from `#f3f3f3` to `#f4f4f4`.

## Change Record
* Navigation active link color  has been changed from `#f3f3f3` to `#f4f4f4`

## Translations
N/A


Kingdutch:
Hi @swastikpareek ,

Thank you for the suggestion.

Unfortunately I think this change introduces some potential pitfalls for users who try to change theme settings and this would require more design as well as finer grained control over how Open Social components look. 

We have some places such as landing pages or on content pages in the sky flavour, where we use a different background color. Additionally there's currently no way for users to change the text color on the pages. So changing the background color can cause pages to become inaccessible.

I don't think this is something we can implement at this time. This will have to wait for a more thorough componentization of our theme and design system.

The recommended way of changing this aspect of the theming would be to implement a sub-theme that changes the styling.

If you do require this to be a setting and can not use a sub-theme then an alternative would be to create a separate module. The module can alter the theme settings form and inject dynamically generated CSS, similar to what our [improved_theme_settings](https://github.com/goalgorilla/open_social/tree/main/modules/custom/improved_theme_settings) module already does.

PR #2213, opened 2021-02-03
Feature/3195834: Add theme settings to override box shadow values of cards for open social blue theme

swastikpareek:
## Problem
Currently, the box-shadow applied on cards by the open social cards was controlled by the CSS file coming from the theme repo of the social blue theme.
  The site administrator couldn't update the box-shadow value for the cards from the theme settings page of the open social blue theme as they can for the border-radius value of the cards. So, in order to give more flexibility to the site administrator to directly update the cards box-shadow in order to match it with other brand-specific colors (which they can provide in the color scheme section of theme settings) and make the UI more configurable, the card box-shadow values are exposed in the theme settings page (under Opensocial settings) as a separate new vertical tab.

## Solution
* The idea is to alter the theme settings form of the social blue theme, for providing the configuration for box-shadow.  The `socialblue_form_system_theme_settings_alter` function is updated to accommodate for this setting. This new setting is added under the `open_social_settings` form element with the `#weight` value so that it appears in the correct order on the vertical tabs with the `Card box-shadow` label.
* A new checkbox is added under this section, to control the overriding settings of the box-shadow. A new collapsible form item group is added which contains all the required box-shadow configs (like color, opacity, x,y-offset, blur and spread values) and will be only visible if the checkbox for overriding the box-shadow is enabled ( This done using[ form states API](https://www.lullabot.com/articles/form-api-states)) 
  ```
  '#states' => [
      'invisible' => [
        ':input[name="override_box_shadow"]' => ['checked' => FALSE],
      ],
   ],
   ```
* All the newly added config's default value is stored in the theme settings schema YAML file ( See `socialblue.settings.yml`)
* To keep the code organized in the function, subform items of the related vertical tabs are grouped in the order of their `#weight` attribute. For. e.g. All the code related to fonts vertical tabs are grouped together and so on. 
* The values will be saved in the config table after updating the form. These values are then fetched in the `improved_theme_settings` module. These fetched values are then used the `.module` file and used to create the required box-shadow CSS styles for the `.card` CSS selector ( see `improved_theme_settings.module`), which is then eventually loaded on the HTML page as CSS code.
* As a part of this PR, the `improved_theme_settings_get_setting` function is updated to provide the raw value of the theme setting, as the box-shadow config needed the `color` value (hex code) for the box-shadow. The current `improved_theme_settings_get_setting` has the logic to return the `FALSE` value if the returned value is not an integer and breaking the logic.
  To make the function consistent, `$raw` variable is added, which when set to `TRUE` return raw value.

## Issue tracker
https://www.drupal.org/project/social/issues/3195834 

## How to test
*For example*
- [ ] Login as site manager
- [ ] Go to appearance tab theme settings section and select Social blue theme settings page
- [ ] Under the `Opensocial settings` section, a new vertical tab link should be available by the name of `Cards Box shadow`
- [ ] The section will contain a checkbox saying `Override default box shadow for cards`
- [ ] Enabling it will make a collapsible section appear with all the required settings for initializing box-shadow.
- [ ] Save the form and check the cards on the FE side of the site and see the box-shadow being updated with the supplied value in the theme settings.

#### Before
Screenshot 2021-02-02 at 6 11 15 PM

#### After
after gif

## Release notes
* A new vertical tab menu link is introduced under the `Opensocial settings` section of the Social Blue theme settings page. This can be used to override and control the default box-shadow on the cards used throughout the site. Also, the theme settings schema of the Social Blue theme is updated to store the default values for the box-shadow values.

* The `improved_theme_settings_get_setting` module is updated to provide the raw value of the theme setting if a third param `$raw` was passed. This is initially set to `false`. So, that it is backward compatible.

## Change Record
* Box shadow is been made configurable through the Social Blue theme settings. 
* The `improved_theme_settings_get_setting` function now also supports providing the raw value of the theme setting.

## Translations
N/A


Kingdutch:
Hey @swastikpareek,

Thanks for this suggestion! 

We're still looking at how best to expand the theme settings and specifically also what not to add. In the case of the shadow values that is a theme aspect that determines visual hierarchy and as a result has certain requirements. Additionally there are a few different drop shadows in use so a single setting is not sufficient (and multiple settings are most likely overwhelming).

With that in mind I don't think this change is something that'll improve the look and feel of platforms for our users. Especially since most users using the theme settings are not themselves designers.

The recommended way of changing this aspect of the theming would be to implement a sub-theme that changes the styling.

If you do require this to be a setting and can not use a sub-theme then an alternative would be to create a separate module. The module can alter the theme settings form and inject dynamically generated CSS, similar to what our [improved_theme_settings](https://github.com/goalgorilla/open_social/tree/main/modules/custom/improved_theme_settings) module already does.

PR #2216, opened 2021-02-03
Issue #3197265 by chmez: An error on group page after adding images to the post and deleting the first image

ghost:
## Problem
An error on the group page after adding images to the post and deleting the first image.

## Solution
- Restore a field for selecting album on post creation form after deleting the first image.
- Fix creating a new post for an album from the stream page of a public group when the post has a default visibility option.

## Issue tracker
- https://www.drupal.org/project/social/issues/3197265
- https://getopensocial.atlassian.net/browse/YANG-4860

## How to test
- [ ] Log in as a normal member and create a public group.
- [ ] Navigate to the stream page of the created group and add a few images to the post creation form.
- [ ] When you delete the first image then the album field still should be available.
- [ ] When you submit that form then you shouldn't see any errors.

## Release notes
Improve the process for creating a new album together with a new post.


modules/social_features/social_album/src/Plugin/Field/FieldWidget/SocialAlbumOptionsSelectWidget.php
```diff
@@ -24,6 +24,22 @@
...
+   * The visibility options mapping.
+   *
+   * The associative array where keys are node options and values are the
+   * corresponding post options.
+   *
+   * @see field.storage.node.field_content_visibility.yml
+   * @see field.storage.post.field_visibility.yml
+   */
+  const VISIBILITY_MAPPING = [
+    '0' => 'community',
```
ribel, line 37: @OH17L could you explain why do we have both 0 and 2 for the community here?

ribel:
@chmez Changes looks good to me, but since original issue was already released - could you create a new one for this fix, so we can specify it in next release notes? 
Also, check my question in the code.

modules/social_features/social_album/src/Plugin/Field/FieldWidget/SocialAlbumOptionsSelectWidget.php
OH17L: Because these values were gotten from allowed values [here](https://github.com/goalgorilla/open_social/blob/10.0.x/modules/social_features/social_post/config/install/field.storage.post.field_visibility.yml#L14).

ribel: Ok, then it's clear.

ribel (approved):

ribel:
🍒 picked to 10.0.x

PR #2219, opened 2021-02-04
Feature/3196481 : Improve FE font loading performance by preloading the webfont

swastikpareek:
## Problem
The custom web fonts are loaded in the browser (fetched) after the webpage is ready for the user. This makes the page to show a flash of unstyled text and later renders the custom font with a layout shift. This is not considered a good practice and can be crucial for the end-user experience. 
This needs to be fixed by preloading web fonts and using a smart and compressed form of fonts (woff/woff2)

## Solution
* First of all getting rid of unwanted font file formats for a custom web font as you just need woff and woff2 for most modern browsers. So, in the `social_font` config install file `field.field.font.font.field_fonts.yml` is updated and the `file_extensions` setting is limited to just allow `woff` and `woff2`
* Now the idea is to load fonts in the order `woff2 -> woff`. For this, we had to refactor the logic which renders CSS for setting up the `src` attribute for the font face in the `social_font_fontfiles_render` function.
 Initially, this logic was outputting the CSS property like this
  ```
   src: URL(<URL1>) format(<FORMAT1>);
   src: URL(<URL2>) format(<FORMAT2>);
   src: URL(<UR3>) format(<FORMAT3>);
   ...
   ```
   This was buggy as if there are multiple property declarations with the same name, only the last CSS property will be read by the browser, and in this case, only the last custom web font file uploaded was being loaded by the browser. This was not performance efficient. 
   To fix this, first of all, we need to use a single `src` property and use it in this way
   ```
   src: URL(<URL1>) format(<FORMAT1>),
          URL(<URL2>) format(<FORMAT2>),
          URL(<UR3>) format(<FORMAT3>);
   ```
   Also, we need to order(sort) the font files in an order of `woff2 -> woff` ([see why](https://nooshu.github.io/blog/2021/01/23/the-importance-of-font-face-source-order-when-used-with-preload/#font-load-priority)) and to achieve these things the logic is bit refactored, and all the fonts are added in a sorted array of font files `$fontStringArray` (sorted on the basis of `weight` key using [usort](https://github.com/goalgorilla/open_social/pull/2219/files?file-filters%5B%5D=.module&file-filters%5B%5D=.yml#diff-e25b5f704af400f4a53c255d82e508fe7979245463cf87c700f0b14716289c01R153-R159), before assigning the weight to them using a [simple switch](https://github.com/goalgorilla/open_social/pull/2219/files?file-filters%5B%5D=.module&file-filters%5B%5D=.yml#diff-e25b5f704af400f4a53c255d82e508fe7979245463cf87c700f0b14716289c01R135-R144)) 
* The next important and crucial part is to preload the font files. (for now, only woff2 files are being preloaded as preload and woff2 has the same support matrix and woff2 takes care of almost all the modern web browsers (except IE11 - taken care of by woff)). 
    To preload, the fonts we had to add them to the `<head>` tag of the page, and `hook_preprocess_html` is used for it. This hook is written in the social blue theme file as this logic is specific to the social blue theme.
    The font file id is being fetched by `font_primary` theme config and eventually, the font file URLs are being loaded, this then filtered by formats and only woff2 font files are being preloaded. See [this code] (https://github.com/goalgorilla/open_social/pull/2219/files?file-filters%5B%5D=.theme&file-filters%5B%5D=.yml#diff-d53df3659a7dd84b99048a5b7f97a7803d668c49cb8a30099b1be7bc6b961cf5R110-R119)
* The last and important piece of the jigsaw was to use [font-display](https://css-tricks.com/almanac/properties/f/font-display/) property to set a rule of how the browser will behave while and after downloading the font. 
    We need the browser to swap the fallback text with a custom downloaded web font as soon as possible, so we set the `font-display` property to `swap` (since we are preloading the font this will be so fast that no flash of invisible text (FOIT) or flash of unstyled text will be noticed - even on a slower connections). 
    This is added in `social_font.module` file in `social_font_render` function, just after `src` declaration [here](https://github.com/goalgorilla/open_social/pull/2219/files#diff-e25b5f704af400f4a53c255d82e508fe7979245463cf87c700f0b14716289c01R88)
   

## Issue tracker
https://www.drupal.org/project/social/issues/3196481

## How to test
*For example*
- [ ] Add a custom font to the font list in (Structure -> Fonts -> Font list ) and set that font as the primary font under the open social blue theme settings page.
- [ ] Set the network speed to 3G or slower bandwidth and open any page to see Flash of unstyled text

### Before - At fast 3G network speed
Before

### After
After

## Change Record
* Valid font formats to be uploaded for a font entity are limited to WOFF and woff2 as they cover almost all the major web browsers and their older versions.
* Fonts are preloaded using HTML `<head />` tag and only woff2 are preloaded because `preload` has the same support matrix as `woff2`. See `caniuse.com`
* Initially, under the `social_font` module multiple `src` properties were used and the last property was overriding all the others. This was buggy. Now, only one `src` property is used now. And the order of loading was set to use `woff2 -> woff`. See why [here](https://nooshu.github.io/blog/2021/01/23/the-importance-of-font-face-source-order-when-used-with-preload/#font-load-priority)

## Translations
N/A


navneet0693:
Hey @swastikpareek,

Thanks for your PR. This PR needs rebase with the latest version of 10.3.x branch.

navneet0693:
Although, this is a nice idea, unfortunately, I am going to close this PR. We are on the path of going decoupled and we will have to refactor the work.

PR #2220, opened 2021-02-04
Issue #3197227 by chmez: Issue with Ajax commenting in album on the stream

ghost:
## Problem
An issue with Ajax commenting in an album on the stream.

## Solution
Fix changing comments wrapper identifier on stream after adding a comment in post modal window.

## Issue tracker
- https://www.drupal.org/project/social/issues/3177902
- https://www.drupal.org/project/social/issues/3197227
- https://getopensocial.atlassian.net/browse/YANG-4830

## How to test
- [ ] Enable the "Social Ajax Comments" module.
- [ ] Log in as a normal member and navigate to the user stream page.
- [ ] Create a post with image and open image modal window.
- [ ] Create the first comment.
- [ ] When you create the second comment then it should be displayed in the comments section of the sidebar.

## Release notes
Add the possibility to create more that one comment in the post modal window opened from stream page.

### DIFF
--- a/tests/behat/features/capabilities/post/post-modal-comments.feature
+++ b/tests/behat/features/capabilities/post/post-modal-comments.feature
@@ -1,0 +1,33 @@
+@api @post @stability @perfect @critical @YANG-4830 @database @stability-2 @post-modal-comments
+Feature: Comment on a Post with an Image
+  Benefit: In order to give my opinion on a post with an image
+  Role: As a LU
+  Goal/desire: I want to comment on a post with an image
+
+  Scenario: Successfully create a comment on a post with an image in the modal window
+
+    Given I enable the module "social_ajax_comments"
+    Given users:
+      | name     | status | pass     |
+      | PostUser |      1 | PostUser |
+    And I am logged in as "PostUser"
+
+    When I attach the file "/files/opensocial.jpg" to hidden field "edit-field-post-image-0-upload"
+    And I wait for AJAX to finish
+    And I fill in "Say something about these image(s)" with "This post with a photo."
+    And I press "Post"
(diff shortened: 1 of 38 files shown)


ribel (approved):

ribel:
@chmez Changes looks good to me, but since Original issue was already released - could you create a new one for this fix, so we can specify it in next release notes?

tests/behat/features/capabilities/post/post-modal-comments.feature
```diff
@@ -0,0 +1,33 @@
+@api @post @stability @perfect @critical @YANG-4830 @database @stability-2 @post-modal-comments
+Feature: Comment on a Post with an Image
+  Benefit: In order to give my opinion on a post with an image
+  Role: As a LU
+  Goal/desire: I want to comment on a post with an image
+
+  Scenario: Successfully create a comment on a post with an image in the modal window
+
+    Given I enable the module "social_ajax_comments"
```
ribel, line 9: @OH17L This behat test looks good, but since we made changes in `social_post_album` module in this PR, I think we need to write a scenario when this module is enabled. Could you enable it here or maybe even better write a separate similar scenario, but with social_post_album enabled.

ribel (changes requested):

tests/behat/features/capabilities/post/post-modal-comments.feature
OH17L: As you can see [here](https://github.com/goalgorilla/open_social/blob/10.1.x/modules/social_features/social_core/social_core.install#L1291) this module is enabled by default.

ribel (approved):

ribel:
🍒 picked to 10.0.x

PR #2223, opened 2021-02-05
Issue #3196173 - Fix Expected argument of type "array or Traversable and ArrayAccess" 

ronaldtebrake:
## Problem
Our routing.yml declaration isn't correct and compatible with modules like Plugin.

## Solution
Fix our routing.yml route requirements section instead of misusing the parameters section.

## Issue tracker
https://www.drupal.org/project/social/issues/3196173
https://www.drupal.org/project/social/issues/3169375

## How to test
- [ ] Enable plugin contrib module
- [ ] See that it crashes the site
- [ ] checkout this branch, it now works again after a drush cr
- [ ] See that event invite and enrollments routes, which had the issues, still works.

## Screenshots
None needed. 

## Release notes
An internal fix which resulted in issues when certain modules were used.

### DIFF
--- a/modules/social_features/social_event/modules/social_event_invite/social_event_invite.routing.yml
+++ b/modules/social_features/social_event/modules/social_event_invite/social_event_invite.routing.yml
@@ -55,6 +55,6 @@ social_event_invite.update_enrollment_invite:
         type: entity:user
       event_enrollment:
         type: entity:event_enrollment
-      accept_decline: '[0-1]+'
   requirements:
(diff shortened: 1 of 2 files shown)


Kingdutch (approved):
Looks like a straightforward change. Only test error is in our infamous human text which is unrelated.

ronaldtebrake:
Thanks @Kingdutch for the review :)

ronaldtebrake:
Also in 9.x / 10.0.x

PR #2224, opened 2021-02-08
Issue #3197235 by SV: Display featured profile section for anonymous users

volodymyr-sydor-lemberg:
## Problem
Profile teasers added as "featured content" items on landing pages do not display for non-logged-in users. Page permissions are set to public and it is a bit confusing that added block displays as empty

## Solution
Provides access for AN (users who don't have `view any profile profile` permission) to preview teasers profiles but only for landing & dashboard pages.

## Issue tracker
- https://www.drupal.org/project/social/issues/3197235
- https://getopensocial.atlassian.net/browse/SUP-560

## How to test
*For example*
- [ ] Using the latest version of Open Social with the example module enabled
- [ ] As a site manager
- [ ] Try to create a landing page, add "featured content" section and select any profile entities
- [ ] When I open the created page as anonymous I expect to see added profiles but instead I don't see any new items.

## Screenshots
Create a section with profiels:
Screenshot from 2021-02-08 17-20-47

Landing page for LU
Screenshot-from-2021-02-08-17-22-47-png-1841×888-

Landing page for AN
Screenshot from 2021-02-08 17-18-58

## Release notes
Allow view profile teasers which added as "featured content" items on landing or dashboard pages for anonymous users.

### DIFF
--- a/modules/social_features/social_profile/social_profile.module
+++ b/modules/social_features/social_profile/social_profile.module
@@ -17,6 +17,7 @@ use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\Core\Url;
 use Drupal\file\Entity\File;
 use Drupal\image\Entity\ImageStyle;
+use Drupal\node\Entity\Node;
 use Drupal\profile\Entity\Profile;
 use Drupal\profile\Entity\ProfileType;
 use Drupal\user\Entity\User;
@@ -838,12 +839,31 @@ function social_profile_social_user_account_header_items_alter(array &$menu_item
  */
 function social_profile_profile_access(EntityInterface $entity, $operation, AccountInterface $account) {
 
+  $display_profile_teaser = FALSE;
+  $route_match = \Drupal::service('current_route_match');
+  $allowed_node_types = [
+    'landing_page',
+    'dashboard',
+  ];
+
+  if ($route_match->getRouteName() === 'entity.node.canonical') {
+    $node = $route_match->getParameter('node');
+
+    if (!is_null($node) && !$node instanceof Node) {
+      $node = Node::load($node);
+    }
+    if ($node instanceof Node && in_array($node->getType(), $allowed_node_types)) {
+      $display_profile_teaser = TRUE;
+    }
+  }
+
   // Provides access only to viewing user profile info, like referenced data in
-  // featured items, if the current user has no permissions.
+  // featured items for landing & dashboard pages, if the current user has no
+  // permissions.
   if (
     $operation === 'view' &&
-    !$account->hasPermission('view any profile profile') &&
-    \Drupal::routeMatch()->getRouteName() !== 'entity.profile.canonical'
+    $display_profile_teaser &&
+    !$account->hasPermission('view any profile profile')
   ) {
     return AccessResult::allowed();
   }
--- a/modules/social_features/social_profile/social_profile.module
+++ b/modules/social_features/social_profile/social_profile.module
@@ -5,6 +5,7 @@
  * The Social profile module.
  */
 
+use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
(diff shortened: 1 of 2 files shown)


ribel (approved):

ribel:
Hi, @ronaldtebrake, 
We discussed this issue earlier in the ticket, and now it is fixed with test coverage.
Just wanted to confirm with you if we can merge it, and also since we are not sure if this is a bug fix or new feature not sure to what milestone better to add it?

ronaldtebrake:
Hi @ribel thanks for that :) 
I think in this case it's a feature request as it has quite the implications for our users. So 10.1.0 would be best.

modules/social_features/social_profile/social_profile.module
```diff
@@ -831,3 +832,19 @@ function social_profile_social_user_account_header_items_alter(array &$menu_item
...
+
+/**
+ * Implements hook_ENTITY_TYPE_access().
+ */
+function social_profile_profile_access(EntityInterface $entity, $operation, AccountInterface $account) {
+
+  // Provides access only to viewing user profile info, like referenced data in
+  // featured items, if the current user has no permissions.
+  if (
+    $operation === 'view' &&
```
ronaldtebrake, line 844: @ribel @volodymyr-sydor-lemberg have we checked any other potential impact?

What happens on the search now for example? We're now checking all routes except a users profile itself. 
Can't find if you've tested / reviewed the unhappy scenario's on this PR.

volodymyr-sydor-lemberg: @ronaldtebrake good catch! Yes, we've checked unhappy scenarios for user stream/information/profile pages but on the search/all we can get user items in the result and need to add this route into exceptions. 
 Also, there are pages like:
- all-members
- birthday (social_birthday module)
we can add them too but they are not accessible for anonymous

ronaldtebrake: Thanks for the explanation, great to hear! :D 
Maybe in this case it's safer to check if it's part of a landing page or dashboard?
In that case we can always add to it when necessary or expose it as a setting in the future. 
Because our community might have even more routes which they want/need to add into exceptions.

volodymyr-sydor-lemberg: I'm afraid that in this `hook_entity_access` we don't have any info about a place where exactly it used
About the list of disabled routes, I think we can implement it as a `hook_disable_view_teaser_profile_routes()` which returns a list of exceptions and can be altered by other modules

ronaldtebrake: Sorry, didn't get that completely out of the PR.

What I want to make sure is that in a minor release we are not adding a security risk like this. Not every user has the ability to create custom code. Right now site builders are in no way possible to deny access. Which is my main concern.

So maybe it would be better to create a custom permission or a site setting so people can remove that access? 
What if there is no public content for example? Or if they are not using landing pages / dashboard. 

We don't want them to get an access allowed for this scenario and them not having an option to deny access by a setting or a permission or something you can use the UI for.

Do you see my point? 😊

ronaldtebrake: Hmm no using that permission seems ok 🤔  ok check that seems to be a good answer :D thanks for the explanation not sure why i missed that haha

volodymyr-sydor-lemberg: Hey @ronaldtebrake, I've improved it a bit and added an additional condition to give access AN to preview teasers profiles  only for landing & dashboard pages, please take a look

ronaldtebrake:
That looks better @volodymyr-sydor-lemberg thanks a lot, could you maybe update the PR template to show that this is only there when the correct permission is set & and only for the routes you checked so everybody is aware of the full impact? 

Thanks so much!

volodymyr-sydor-lemberg:
> That looks better @volodymyr-sydor-lemberg thanks a lot, could you maybe update the PR template to show that this is only there when the correct permission is set & and only for the routes you checked so everybody is aware of the full impact?
> 
> Thanks so much!

done

PR #2225, opened 2021-02-08
Issue #3196837 - validation node form loses content visibility when a group is selected

ronaldtebrake:
## Problem
Whenever you try to save a node and you have a group selected in the access permission group section this isn't referenced correctly by the widget. Which means when the form validation kicks in, the widget renders a default state. 
While you have a group selected and the widget should take the groups visibility settings in to account.

## Solution
As the `$form_state` is being rebuilt already and we can check if it contains any values for the `$form['groups']` selection widget, we can base our default state on that value being set.
So whenever validation kicks in, this is taken in to account by our widget and the correct visibility is rendered including the disabled state.

## Issue tracker
https://www.drupal.org/project/social/issues/3196837

## How to test
- [ ] Login as SM
- [ ] Create a topic / event
- [ ] Don't fill in any details, just select a group 
- [ ] See that the visibility is being updated based on the group settings
- [ ] When you press create topic the validation should kick in nicely
- [ ] See that the visibility and the group is still as you've selected
- [ ] See that you can change the settings as you wish and save it after
- [ ] See that it also works on editing a topic / event 

## Screenshots
https://www.loom.com/share/d7ad06159ca942aeb3a4e09baa5c9cd5

## Release notes
When saving content which is able to be placed in a group, if some validation error was triggerd, it could be the content visibility field value was reset which shouldn't happen. Now if there is a validation error we will make sure to base our visibility setting on the group it's placed in.

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -1250,11 +1250,10 @@ function _social_group_node_form_submit(array $form, FormStateInterface $form_st
  * Implements hook_field_widget_form_alter().
  */
 function social_group_field_widget_form_alter(&$element, FormStateInterface $form_state, $context) {
-
   /** @var \Drupal\Core\Field\FieldDefinitionInterface $field_definition */
   $field_definition = $context['items']->getFieldDefinition();
 
-  // Unset the public options on visibility field.
+  // Unset the visibility field options which are not enabled.
   if ($field_definition->getType() == 'entity_access_field') {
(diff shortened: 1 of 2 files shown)


navneet0693 (approved):

sjoerdvandervis:
Cherry-picked to 8.x-9.x.

ronaldtebrake:
Cherry 🍒  picked to 9.x / 10.0.x / 10.1.x as well.

PR #2230, opened 2021-02-11
Issue #3197910 by navneet0693: Updated all the search indexes to have similar configurations.

navneet0693:
## Problem
The search index "Search all" is a combination of fields from "Search content, users and groups". Currently, not all the fields from content, user and group indexes are present in Search all index. This might result in inconsistent search results.

## Solution
Update the "search all" index to contain all the fields from Search content, users and groups index.

## Issue tracker
https://www.drupal.org/project/social/issues/3197910

## How to test
- [ ] Run drush updb 
- [ ] Re-index the search indexes.
- [ ] Perform search.

## Screenshots
N.A

## Release notes
We have updated "Search All" indexes to now contain similar results are the other three search result categories namely "Content, users and groups"

## Change Record
Search indexes are update so that "Search All" contains the combination of fields from Search group, users and content.

## Translations
N.A


ronaldtebrake:
Hi @navneet0693 

It seems you've done more than just updating the social_all indexes, also would be great to capture your changes in a test to ensure we have better coverage of the issues that (might) occur.

Kingdutch:
> We have updated "Search All" indexes to now contain similar results are the other three search result categories namely "Content, users and groups"

The Search All already had results similar to the other indexes right? Is the change mainly that it now relies on the fields rather than the rendered entity? Perhaps that's a better way to describe it?

> The "Search All" search index now uses the fields of contents in addition to the rendered entity for searches. This ensures that results in "Search All" more closely match results from the type specific searches for "Content", "Users", and "Groups".

Additionally I don't think a change record is needed here right? Unless we remove the rendered entity field entirely (which I didn't see happen in the diff). In that case this would be a BC breaking change because custom fields may stop being indexed.

If we do think a CR is needed then we should create [a draft on D.o](https://www.drupal.org/list-changes/social) and post the link here :D 

Somewhat unrelated: We are missing change records for the removal of modules from Open Social. So we may need better guidelines on when a change record is needed :)

PR #2235, opened 2021-02-15
Issue #3198528 by tBKoT: Improve Follow taxonomy block performance

dmytriikaiun:
## Problem
Improve Follow taxonomy block performance and extendability

## Solution
Implement Lazy Load or AJAX for a popup with counters for performance improvements. Count followed users and related content inside the popup rendering instead of block preprocess.

## Issue tracker
https://www.drupal.org/project/social/issues/3198528
https://getopensocial.atlassian.net/browse/YANG-4810

## How to test
N/A

## Screenshots
N/A

## Release notes
Popup for the following taxonomy badges now loads through the lazy builder.
All counts are calculated in popup rendering.

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_follow_taxonomy/modules/social_follow_tag/social_follow_tag.module
+++ b/modules/social_features/social_follow_taxonomy/modules/social_follow_tag/social_follow_tag.module
@@ -106,7 +106,6 @@ function social_follow_tag_preprocess_social_tagging_split(&$variables) {
             'social_follow_tag.lazy_builder:popupLazyBuild',
             [
               $tag_info['url'],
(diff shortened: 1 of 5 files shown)


ribel:
Code changes look good, but when I tested it on env, it looks like lazy-load doesn't work as expected.
It looks like it loads tags later, but together with popup since it appears immediately on click.
Idea was that most user will not even click on tags, so we don't need to load popup with numbers on each page load.

Also please update the solution in a PR so it matches with actual implementation.

ribel:
🍒 picked to 10.0.x

PR #2237, opened 2021-02-16
Bugfix/3198751 event enrollment requests

robertragas:
## Problem
With events you can let users request to join an event. When there are open requests and you update the node, these requests are gone.

## Solution
The social event module has a protection to keep pending enrollments when changed from request to enroll to invite only, but this does not sufficiently check the state. This check also needs to keep the current state in mind, not only the original.

## Issue tracker

## How to test
- [ ] Create an event with request to enroll
- [ ] Enroll to the event as another user
- [ ] Check the event and see someone requested...now edit the node and update the description.
- [ ] See that the request enrollment is gone
- [ ] Check out to this branch and see it stays now

## Release notes
We solved an issue where if you have an event with pending enrollment request, it would delete those requests after updating the event itself with for example the description or title.


robertragas:
Closed in favor of https://github.com/goalgorilla/open_social/pull/2238/files since I had some commit issues here.

PR #2241, opened 2021-02-17
Issue #3198737: Ensure content within a group can be translated

ronaldtebrake:
## Problem
When group content is translated we don't have to run the entire group content update, this is redundant as translations are now on field level and it is not an entire node translation, so access records et all will stay the same.
 
## Solution
For clarity we disabled the group reference field and the group content visibility for now, as they can create unwanted behavior on editing group content with duplicate access records as it seems like a new group is added but that is not true.

## Issue tracker
https://www.drupal.org/project/social/issues/3198737#comment-14001876

## How to test
*Enable content translation for group content and at least 2 languages*
- [x] Create a topic
- [x] Add it to a group
- [x] See that you can save it in the default language
- [x] When you try to translate it in a second language, just changing the title / description it will result in a `Integrity constraint violation: 1062 Duplicate entry '29-15-gnode:topic-en' for key 'PRIMARY': INSERT INTO {node_access}` 
- [x] Checkout this branch to see that this now should pass and the group reference and content visibility can only be changed on the original default language instead. 
- [x] Checkout when you change the group reference and content visibility also the translated group content is now placed in the newly referenced group.

## Screenshots
Screenshot 2021-02-17 at 16 24 25

## Release notes
When translating group content (like topics/events) in a different language it would result in an error, eventhough the translation was added. We have ensured this is now fixed, however you can not change the group reference or visibility when translating.


navneet0693 (approved):

ronaldtebrake:
Cherry-picked to 10.1 / 10.0 / 9.x 🍒

PR #2245, opened 2021-02-19
Issue #3199367 by SV: Allow member to add a table in the description of node

volodymyr-sydor-lemberg:
## Problem
Currently, authorized users can create/edit new nodes but they are not able to use a table widget because they have access only to the `basic_html` format which does not provide that possibility.

## Solution
Adjust `basic_html` format and add table widget

## Issue tracker
- https://www.drupal.org/project/social/issues/3199367
- https://getopensocial.atlassian.net/browse/YANG-4987

## How to test
- [ ] Using the latest version of Open Social with the social_discussion module enabled
- [ ] Login as a member of group where enabled discussion(or any other node type) 
- [ ] When creating a new node I expect to see a table widget in CKEditor bu it doesn't exist.

## Screenshots
Screenshot from 2021-02-19 20-09-50

## Release notes
Extends basic_html format and add table widget. So, authorized users will be able to use it when create/edit nodes

### DIFF
--- a/modules/social_features/social_editor/config/install/editor.editor.basic_html.yml
+++ b/modules/social_features/social_editor/config/install/editor.editor.basic_html.yml
@@ -40,6 +40,10 @@ settings:
           items:
             - Source
             - Maximize
+        -
+          name: Tables
+          items:
+            - Table
   plugins:
     stylescombo:
       styles: ''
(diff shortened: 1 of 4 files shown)


ribel:
Changes look good to me, but since this is a change request, @ronaldtebrake, could you tell if you have any objections?

ribel (approved):

navneet0693 (approved):

ribel:
I am going to merge this PR since only often failing false-positive tests related to private files and groups are failing.

PR #2227, opened 2021-02-10
Issue #3197706 by rolki: Provide a method to alter array of group types used in social group request

rollki:
## Problem
There is no way to add join methods for group types except flexible group.

## Solution
Use the custom hook to add other group types

## Issue tracker
- https://www.drupal.org/node/3197706
- https://getopensocial.atlassian.net/browse/YANG-4740
- https://www.drupal.org/project/social/issues/3199293

## How to test
- [ ] Add these changes https://github.com/goalgorilla/social_course/pull/38 
- [ ] For the course_basic install social_course_basic_request module
- [ ] Got to some basic course edit page
- [ ] You should be able to select join methods for the course

## Screenshots
image

## Release Notes
Provide a method to alter array of group types used in social group request

PR #2246, opened 2021-02-22
Issue #3199893 by rolki: Notifications don't work for the group types that install the "Group request membership" plugin except for flexible group

rollki:
## Problem
See https://getopensocial.atlassian.net/browse/YANG-5001
When we install the "Group membership" plugin for some group type, then we have an issue with notifications since it hardcoded only for the flexible group.

## Solution
Use hook alter to get the list of request group types instead of the only flexible group.

## Issue tracker
- https://getopensocial.atlassian.net/browse/YANG-5001
- https://www.drupal.org/project/social/issues/3199893

## How to test
- [ ] For example, take the course group
- [ ] Go to some course under not a member with "Request to join" option
- [ ] Click "Request to join button"
- [ ] Fill the message in the pop-up and submit
- [ ] The course manager should receive notification about new request


ribel (approved):
I see this is actually a follow-up to changes made in PR #2227
So this actually good to be released together in 10.1.0

PR #2248, opened 2021-02-23
Issue #3200113 by rolki: Allow extending group and content visibilities to other group types

rollki:
## Problem
There's no way to use visibility fields to other group types since this feature is hardcoded only for the flexible group.

## Solution
Allow extending group and content visibilities to other group types.

## Issue tracker
- https://www.drupal.org/project/social/issues/3200113
- https://getopensocial.atlassian.net/browse/YANG-4702

## How to test
- [x] The easiest way to check the difference are:
- [x] Go to some flexible group
- [x] You should see the "Access permissions" label in a group header
- [x] Hover to this label should display a tooltip with some info about group visibilities

So when you add visibility fields to your custom group type, then it will be broken
E.G. social course group type -> https://github.com/goalgorilla/social_course/pull/39 

## Screenshots
Снимок экрана 2021-02-24 в 14 12 35

## Release notes
Allow extending group and content visibilities that used in Flexible groups to other group types.

### DIFF
--- a/modules/social_features/social_group/social_group.api.php
+++ b/modules/social_features/social_group/social_group.api.php
@@ -182,7 +182,7 @@ function hook_social_group_allowed_join_method_description_alter($key, &$descrip
 }
 
 /**
- * Adds extra group types with visibility fields that work as flexible groups.
+ * Provide a method to extend group and content visibility to other group types.
  *
  * @param string $social_group_types
  *   The group type id.
--- a/modules/social_features/social_group/social_group.api.php
+++ b/modules/social_features/social_group/social_group.api.php
@@ -182,8 +182,7 @@ function hook_social_group_allowed_join_method_description_alter($key, &$descrip
 }
 
 /**
- * Adds extra group types with visibility fields that should work the same as
- * flexible groups.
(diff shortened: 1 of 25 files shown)


ribel (changes requested):
@rollki Unfortunately we can't just replace `flexible_rgroup` with `default`, since it will cause errors on custom group types. You can check on Challenge for example.
Since the main goal of these changes is to display information about settings in the hero I think we can create one more hook alter, for example, `hook_social_group_settings_hero_alter()` similar to `hook_social_group_types_alter()`.
It should be a quick win for you.

rollki:
@ribel 
Fixed.

ribel (approved):

modules/social_features/social_group/social_group.api.php
```diff
@@ -181,6 +181,18 @@ function hook_social_group_allowed_join_method_description_alter($key, &$descrip
   }
 }
 
+/**
+ * Adds extra group types with visibility fields that work as flexible groups.
```
ribel, line 185: @rollki Please add a better description, so it will be clear for developers how they can use this hook. Something like:

> Provide a method to extend group and content visibilities to other group types.

PR #2249, opened 2021-03-01
Remove 'select account cancellation method' from default user roles i…

robincee:
…n social_user.install.

## Problem
Users are able to select the cancellation method when they close their account

## Solution
Remove the permission from the installer

## Issue tracker
none yet

## How to test
- install basic setup
- create test account
- login to test account
- cancel the account
With this patch you should not be able to select the method, the site will just inform you (based on the cancellation settings)

## Screenshots
Before:
image

After:
image

## Release notes
Removed the 'select account cancellation method' permission from authenticated user role.

### DIFF
--- a/modules/social_features/social_user/social_user.install
+++ b/modules/social_features/social_user/social_user.install
@@ -161,7 +161,6 @@ function _social_user_get_permissions($role) {
     'access user profiles',
     'cancel account',
     'change own username',
-    'select account cancellation method',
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
On second thought, I believe we actually want the users to be in controle here. 
If that doesn't work for our SaaS instance we need to update the permissions in there.

PR #2250, opened 2021-03-03
DS-7553: As a developer I want to have precise control over how cronjobs and queue items are processed

navneet0693:
## Problem
At this point, all tasks are executed on every cron run. This isn't necessarily an issue as most of these tasks check for themselves if they have been recently executed.
However, there's no control over these tasks other than what they enforce on themselves.

In order for us to be sure queue items are processed quickly, stable, and consistently, priorities and timings need to be configured. This is not possible by default with Drupal Core.
Another issue is that cron runs cannot run in parallel. A cron run can take a long time to process items. depending on what needs to happen. When cron runs cannot run in parallel they can block each other for a certain amount of time (or theoretically indefinitely if it's stale or stuck).

## Solution
We'll need to implement a new cron management system (e.g. Ultimate Cron) that allows us to precisely configure and control timings and priorities of cron tasks.

## Issue tracker
https://getopensocial.atlassian.net/browse/DS-7553

## How to test
*For example*
- [ ] As a sitemanager

## Screenshots

## Release notes

## Change Record

## Translations


ronaldtebrake:
#2260 will be favored

PR #2252, opened 2021-03-08
Issue #3202292 by vnech: Fix hiding of "Tagging" field in the Inline Entity Form

nechai:
## Problem
We have a few modifications for "Tagging" field in the function social_tagging_form_alter(). The issue is that if we render entities with "Inline Entity Form" module widgets this hook doesn't trigger....

## Solution
Add addition hook allows compatibility with the "Inline Entity Form" module widgets

## Issue tracker
- https://www.drupal.org/project/social/issues/3202292
- https://getopensocial.atlassian.net/browse/YANG-5012

## How to test
- [ ] Enable "Social Course" module and log in as SM
- [ ] Create a new "Course" (for example - _/group/add/course_advanced_)
- [ ] Go to the course view page and click on the tab "Sections", then click on the button "Create a section"
- [ ] In the field "Parts" add any entity which bundle is enabled on "Tagging Settings" page - _/admin/config/opensocial/tagging-settings_
- [ ] You should see "Tagging" field even the entity bundle is disabled on "Tagging Settings" page

### DIFF
--- a/modules/social_features/social_tagging/social_tagging.module
+++ b/modules/social_features/social_tagging/social_tagging.module
@@ -137,14 +137,14 @@ function social_tagging_form_alter(&$form, FormStateInterface $form_state, $form
  * @param \Drupal\Core\Entity\EntityInterface $form_entity
  *   Entity object.
  */
-function social_tagging_social_tagging_field_form_alter(&$form, FormStateInterface &$form_state, EntityInterface $form_entity) {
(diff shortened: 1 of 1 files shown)


ribel (changes requested):
@nechai When I add this patch I see that tagging field is also hidden on Add Topics page.

ribel (approved):
I see the checkbox that allows adding tags was not checked, so all works as expected now.

ribel:
🍒 picked to 9.x and 10.0.x

PR #2255, opened 2021-03-10
Issue #3202830 by sjoerdvandervis: Group creator that leaves group still receives welcome message

sjoerdvandervis:
## Problem
When somebody creates a group (whatever type it is), this user automatically becomes the owner and administrator of the group. If later on, this person leaves the group and somebody else is added by a different group manager, the recipients of the welcome message still include the former group owner.

## Solution
Check whether the owner of the group is actually still part of the group, before adding this person to the list of recipients for the welcome message.

## Issue tracker
https://www.drupal.org/project/social/issues/3202830

## How to reproduce
- [ ] Enable the social_group_welcome_message module
- [ ] As a LU, create a group and set a welcome message
- [ ] Add another person to the group and grant this person group management role
- [ ] Now leave the created group
- [ ] Log in as the added user, and add another person to the group
- [ ] See that the welcome message is sent, but also sent to the user that created (and left) the group

## How to test
- [ ] Execute the above steps of reproduction
- [ ] See that the group creator is not one of the recipients anymore, after using this branch.

## Release notes
We made changes to the way we send group's welcome messages to the group managers. Before these changes, all creators of groups would receive the group's welcome messages in a thread, while it could be that the group creator already left the group and should thus not be one of the recipients of the welcome message. We added an extra condition to check whether the creator of the group is actually still a member of this group, before adding this user to the recipient list of the group's welcome message.


modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
```diff
@@ -168,7 +168,14 @@ function social_group_welcome_message_group_content_insert(GroupContentInterface
     }
 
     // Merge all managers of the group to recipients array.
-    $recipients = array_merge($g_admins_users, $g_managers_users, [$group->getOwner()]);
+    $recipients = array_merge($g_admins_users, $g_managers_users);
+
+    // Check what groups the group owner is in, before adding this person
+    // to the list of recipients.
+    $owner_groups = \Drupal::service('social_group.helper_service')->getAllGroupsForUser($group->getOwnerId());
```
navneet0693, line 175: @sjoerdvandervis I do feel that we should remove "Group owner" at all from this equation, we should only include group managers. By default, in Open Social, the group creator is assigned with "group manager" role. These settings are present in each group type, ex: /admin/group/types/manage/closed_group. This will also solve the situation where a group creator leaves.

navneet0693 (changes requested):

modules/social_features/social_group/modules/social_group_welcome_message/social_group_welcome_message.module
sjoerdvandervis: Thanks for this, we indeed don't need this check anymore and we don't want the group owner in the list of recipients, as the group owner already has the group manager role indeed. So we can just remove the group owner from the list of recipients.

navneet0693 (approved):

navneet0693:
🍒  - picked to 10.1.x, 10.0.x, 8.x-9.x!

PR #2256, opened 2021-03-11
Issue #3208583: Optimize Email notifications work flow for direct emails.

navneet0693:
## Problem
When an activity is created it is parsed through the activity system.

Through message templates, we determine where activity should be sent to, e.g. the stream, notification centre or as an email notification. When it is sent as an email we're determining a couple of things. Is the user currently online, in what language would they want to receive the email, is the user blocked, etc.

Email notifications are personalized. For one it contains the salutation, but also a direct link to the user's settings page.

Sending a single email isn't a very big issue, it's fast. The problem is in sending email notifications to a large group of people at once. E.g. a when someone posts a message in a group with >1000 members.

In order for us to speed up the internal process of email notifications, we need to look at how we can find common characteristics between the emails and limit the personalization in those emails.

## Solution
1. Add custom route for user edit page
2. Replace custom user edit link in Immediate emails
3. Sort on email frequency as quickly as possible
4. Prioritize direct emails over digests
5. For direct emails, identify who’s online and blocked before processing further
6. Split users by preferred language
7. Render the EmailActivityDestination::getSendEmailOutputText($message, $langcode); at once for all those users (saves up to 50 entity loads and renders)

## Issue tracker
https://getopensocial.atlassian.net/browse/DS-7541
https://www.drupal.org/project/social/issues/3208583

## How to test
- [ ] Create notifications by doing multiple activities.
- [ ] See that the particular targetted users are receiving the emails.

## Screenshots
Not applicable.

## Release notes
We have optimized direct email workflow to be more efficient in sending emails.

## Change Record
Not applicable.

## Translations
Not applicable.


modules/custom/activity_send/modules/activity_send_email/src/EmailFrequencyBase.php
```diff
@@ -39,7 +38,7 @@ public function getInterval() {
   /**
    * {@inheritdoc}
    */
-  public function processItem(Activity $activity, Message $message, User $target) {
+  public function processItem(Activity $activity, $body_text, User $target) {
```
ronaldtebrake, line 41: Unfortunately this is BC breaking, imagine there are already Queue items created (not finished) when people are updating their platform. 
The item being processed still has a message instead, so we need to think about how we can make it backwards compatible.

For example by adding a new optional variable `$body_text = NULL` so only when the body text is filled we can skip the entire rendering of the `$message` if it isnt we still do it the old way.

modules/custom/activity_send/modules/activity_send_email/src/Plugin/EmailFrequency/Immediately.php
```diff
@@ -25,15 +24,10 @@ class Immediately extends EmailFrequencyBase {
   /**
    * {@inheritdoc}
    */
-  public function processItem(Activity $activity, Message $message, User $target) {
-    if ($target->isBlocked()) {
```
ronaldtebrake, line 29: This needs to stay, sure to be sure. Also to ensure we have backwards compatibility and imagine in the meantime someone got blocked we also want to ensure we filter that out at the end.

modules/custom/activity_send/modules/activity_send_email/src/Plugin/EmailFrequency/Immediately.php
```diff
@@ -25,15 +24,10 @@ class Immediately extends EmailFrequencyBase {
...
-      return;
-    }
-
+  public function processItem(Activity $activity, $body_text, User $target) {
     // Continue if we have text to send and the user is currently offline.
     if (isset($activity->field_activity_output_text) && EmailActivityDestination::isUserOffline($target)) {
       $langcode = $target->getPreferredLangcode();
-      $body_text = EmailActivityDestination::getSendEmailOutputText($message, $langcode);
```
ronaldtebrake, line 36: This is where we can say:

``` if (!$body_text) {
      $body_text = EmailActivityDestination::getSendEmailOutputText($message, $langcode);
      }
      ```
      So by default for existing Queue Items $body_text will be NULL, but for those who have it passed along to the queue item, we can skip the render fase.
      
      That makes it backwards compatible :)

modules/custom/activity_send/modules/activity_send_email/src/Plugin/EmailFrequency/Immediately.php
navneet0693: Okay.

ronaldtebrake:
Looks good to me! 

To also confirm what I've tested and give some context.

This is what it looks like in 10.x in a Group on a multi lingual site. 2 users within that group have set their language to NL the others have En. When creating a topic in a Group the ActivitySendEmail actually used to send out e-mails one by one.

It grabs the first user from the batch, translates the message and sends as you can see below.

10 x - Group - create topic

After creating another topic I switched to this branch, to see if it was BC with existing queue items.

What you see below is that it now actually sorted the users in a better way. So now it groups users together by language, to it only has to translate the message once and can send that translated message to all users with that language. 
That's also why it now sends out messages to users with the same language first and not in a random order.

10 1 after changes

ronaldtebrake:
🍒  to 10.0.x

PR #2257, opened 2021-03-12
Issue #3203202 by vnech: Fatal error in "social_group/src/Plugin/EntityReferenceSelection/SocialGroupSelection.php"

nechai:
## Problem
We have a field "Group" in the entities that can be not a part of any group. This causes a fatal error if we visit the entity add/edit page. The error is raised in the file: modules/social_features/social_group/src/Plugin/EntityReferenceSelection/SocialGroupSelection.php

## Solution
Add additional checking to prevent query to an empty array.

## Issue tracker
- https://www.drupal.org/project/social/issues/3203202
- https://getopensocial.atlassian.net/browse/ECI-1131

## How to test
- [ ] Visit any entity (probably node) which is not a part of any group

### DIFF
--- a/modules/social_features/social_group/src/Plugin/EntityReferenceSelection/SocialGroupSelection.php
+++ b/modules/social_features/social_group/src/Plugin/EntityReferenceSelection/SocialGroupSelection.php
@@ -56,7 +56,7 @@ class SocialGroupSelection extends DefaultSelection {
     if ($excluded_group_types) {
       $query->condition(
         $this->entityTypeManager->getDefinition($configuration['target_type'])->getKey('bundle'),
-        array_diff($all_group_types, $excluded_group_types),
+        array_diff($all_group_types, $excluded_group_types) ?: [0],
         'IN'
       );
     }


ronaldtebrake:
Just saw the following feedback being provided on d.org (just forwarding it here)

Screenshot 2021-03-17 at 09 59 47

nechai:
@ribel Also I have one suggestion regarding changes in the PR. 
In "modules/social_features/social_group/src/Plugin/EntityReferenceSelection/SocialGroupSelection.php" I suggest to add these changes:
```diff
    if ($excluded_group_types) {
-     $diff = array_diff($all_group_types, $excluded_group_types);
+     $diff = array_diff($all_group_types, $excluded_group_types) ?: [0];
-      if (!empty($diff)) {
          $query->condition(
            $this->entityTypeManager->getDefinition($configuration['target_type'])->getKey('bundle'),
            $diff,
            'IN'
          );
-       }
    }
```
Thats because we need to add a subquery to check if the entity can be added to a group as content. Otherwise, we get a list of groups that can be attached and if we chose the one and submit we get the fatal error.

Of course, we hide a field "Group visibility" in this PR but anyway these changes add confidence to avoid breaking pages.

ribel (approved):
I have tested changes and it looks good to me.

ribel:
🍒 picked to 10.0.x

PR #2258, opened 2021-03-12
Issue #3201182 by Kingdutch: [GraphQL] Introduce an Actor interface

Kingdutch:
Currently we always expect things to be done by users. However, with the introduction of third-party applications and server-to-server OAuth authentication we may also see other systems perform actions (like creating content or triggering actions that perform notifications). Additionally for some data Open Social itself may be the cause.

An example where we already have something like this in our current UI is commenting as anonymous users. It would be useful to distinguish in the API between anonymous users (i.e. we have no data about them) and anonymized users (i.e. registered platform users acting anonymously). This can also help link actions by the same user, for example in anonymized discussions.

## Solution
This commit introduces the `Actor` interface. An `Actor` indicates a
person or machine that has a representation on the platform and can
perform actions.

The `User` type is a real person implementation of an `Actor` indicating
a human user on the platform.

An example of a future `Actor` may be a `Bot` which could be used for
automated chat interactions or an `ExternalUser` which may be used to
post content from other platforms without having an account on the
platform itself.

This commit introduces a `DecoratableTypeResolver` which should be used
for all type resolvers in Open Social's GraphQL API. This allows other
modules to extend the types that are returned for interfaces or unions
by decorating existing type resolvers.

## Issue tracker
https://www.drupal.org/project/social/issues/3201182

## How to test
Automated tests should still pass.

## Screenshots
N/a

## Release notes
Unsupported internal changes. N/a

## Change Record
Unsupported internal changes. N/a

## Translations
N/a

### DIFF
--- a/modules/custom/social_graphql/graphql/open_social.graphqls
+++ b/modules/custom/social_graphql/graphql/open_social.graphqls
@@ -22,7 +22,14 @@ schema {
 The schema's entry-point for queries. This acts as the public, top-level API
 from which all queries must start.
 """
-type Query
+type Query {
+  """
+  Get information about the currently authenticated user.
+
+  NULL if you're not authenticated.
+  """
+  viewer : Actor
+}
 
 """
 The schema's entry-point for subscriptions. This acts as the public, top-level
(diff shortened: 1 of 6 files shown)


Kingdutch:
The only failing test is the famous file upload test. PHPUnit tests that cover the affected code are green :)

PR #2260, opened 2021-03-12
Issue #3203499: As a developer I want to have precise control over how cronjobs and queue items are processed 

ronaldtebrake:
## Problem
At this point, all tasks are executed on every cron run. 
This isn't necessarily an issue as most of these tasks check for themselves if they have been recently executed.
However, there's no control over these tasks other than what they enforce on themselves.

In order for us to be sure queue items are processed quickly, stable, and consistently, priorities and timings need to be configured. This is not possible by default with Drupal Core.

Another issue is that cron runs cannot run in parallel. 
A cron run can take a long time to process items. depending on what needs to happen. When cron runs cannot run in parallel they can block each other for a certain amount of time (or theoretically indefinitely if it's stale or stuck).

We need to fix this and provide opportunities to investigate what is happening with our Queues.

## Solution
- We have added `ultimate_cron` using its default values, to allow configuration of Cron Jobs, and to get insights on it's performance.  
- We have added `advancedqueue` to ensure we have an improved Queue API rather than Drupal Core's default. This allows us to migrate our Queue workers to more advanced Jobs. In this PR we have already started to migrate one of these QueueWorkers to a Job. We have added a new `social_advanced_queue` module which ensures sending Mass e-mails runs on the new queue API.

See screenshots below for more explanation.

## Issue tracker
https://www.drupal.org/project/social/issues/3203499

## How to test
- [ ] See that on install / update both ultimate_cron as well as social_advanced_queue is enabled.
- [ ] Turn off your social_cron
- [ ] As a admin visit `admin/config/system/queues/jobs` and `admin/config/system/cron/jobs` to see the current status of your cron / queue
- [ ] As a SM/CM you can't visit these pages, as they are too much developer focussed in a follow up we can create a SM friendly version
- [ ] Send a mass email message through admin people to multiple users.
- [ ] See a Job appear in the Jobs list and see its queued
- [ ] Run the Advanced queue cron to see it works
- [ ] Run regular cron to see it works
- [ ] Turn on social_cron to see it works

## Screenshots
Screenshot 2021-03-12 at 15 04 16
Screenshot 2021-03-12 at 15 48 03

## Release notes
We have added ultimate cron and advanced queue to your Drupal installations, this should help you get a grip on any issues related to our cron and queue. As our communities are heavily dependent on it (for emails / activity stream etc) this is a good base to start our performance improvements on.


bramtenhove (approved):
Tested functionally as well.

Thanks @ronaldtebrake, great work. Really nice improvements!

PR #2262, opened 2021-03-15
Issue #3203588 by laykin: Fix ability to invite members in custom groups (like Challenges)

diakivvasyl:
## Problem
The option to invite members in custom groups (Ex: Advanced Course, Basic Course, Challenge, Secret challenge) has disappeared. 
image

## Solution
Add new cases in the social group invite module for the possibility send invites.

## Issue tracker
https://www.drupal.org/project/social/issues/3203588

## How to test
In custom groups, after the manager clicked tab "Manage members" there must be to appear a button "Add members"

image


ronaldtebrake:
Hi guys, would be great to write this PR/Story as something understandable for those who haven't got challenges enabled as that is an extension and most likely this adds different benefits to our users.

Does this mean this fixes issues for all custom group type for example?

diakivvasyl:
Hi, @ronaldtebrake!
I updated the title and description for PR.

modules/social_features/social_group/modules/social_group_invite/social_group_invite.module
```diff
@@ -110,25 +110,44 @@ function social_group_invite_preprocess_views_view(&$variables) {
   /** @var \Drupal\views\ViewExecutable $view */
   $view = &$variables['view'];
 
+  if (!\Drupal::moduleHandler()->moduleExists('social_group_invite')) {
```
ribel, line 113: @diakivvasyl This condition doesn't make sense. It will never work since we are already in the `social_group_invite` module. It means code in this file will be already executed only when the module is enabled.

ribel (changes requested):
@diakivvasyl Please check and fix comments.

ribel (approved):

ribel:
🍒 picked to 9.x and 10.x

PR #2264, opened 2021-03-15
Issue #3203608 by tBKoT: New module that provide the KPI Analytics: Tag Followers block

dmytriikaiun:
## Problem
As a CM+ I want to see an overview over users following content tags

## Solution
Create a block using KPI Analytics

## Issue tracker
https://www.drupal.org/project/social/issues/3203608
https://getopensocial.atlassian.net/browse/YANG-3982

## How to test
- [ ] Enable the  KPI Analytics module
- [ ] Follow some tags
- [ ] Go to the `/admin/reports/social-kpi`
- [ ] You should see the new block `Follow Tag`
Or
- [ ] Add a new `KPI Analytics: Tag Followers` block on the dashboard

## Screenshots
N/A

## Release notes
Now CM+ is allowed to add the Follow Tag block on dashboards.

## Change Record
N/A

## Translations
N/A


dmytriikaiun:
@ronaldtebrake can I add a new rule to `phpstan-baseline.neon` to fix PHPStan checks? As I can see there is already exist a similar rule https://github.com/goalgorilla/open_social/blob/10.1.x/phpstan-baseline.neon#L1689

ribel:
@ronaldtebrake As you can see nature of those warning comes from fact that we use service from the `kpi_analytics` module, which is not part of the Distro. In this PR we adding a new submodule to Social Follow Taxonomy with dependency from `kpi_analytics`, the same as we did before for `social_node_statistics`, `social_discussion_kpi` or `social_challenge_kpi`.
So to resolve this warning we have two options:

1. Add ignore rules the same as it was done earlier for `social_node_statistics`
2. Move `social_follow_tag_kpi` to the `kpi_analytics`  module, with dependency from `Social Follow Taxonomy`

What approach is better for you?
cc @dmytriikaiun

ribel (approved):

ronaldtebrake:
Thanks for getting back to me @ribel !

2 makes most sense. It probably needs it's own composer.json file so you can specify it only works with Open Social 10.0.0 and higher.

Especially since `Social_follow_tag` is an optional module in the Distro, so having that with dependencies to a contrib module which we are not installing by default. Just means we bloat the software for a limited amount of users that require it and composer won't be installing kpi_analytics correctly if we enable this feature (which we can, using drush which downloads the package itself). 

So would be better to not have these optional modules in the Distro in these circumstances.

And just to make sure, social_discussion_kpi / social_challenge_kpi / social_node_statistics are also not part of the distro.

ribel:
@ronaldtebrake Thanks for the quick reply!
@dmytriikaiun Then please create new issue and Merge Request with this submodule in https://www.drupal.org/project/social_kpi_lite/

PR #2269, opened 2021-03-22
Issue #3204775 - Update patch for ajax comments based on beta 3

ronaldtebrake:
## Problem
There is no restriction in the version of ajax_comments, this resulted in it automatically updating to beta 3 which ensured our patch wasn't applying anymore

## Solution
Update patch to ensure it applies to beta3

## Issue tracker
https://www.drupal.org/project/social/issues/3204775#comment-14037093

## How to test
- [x] Enable social_ajax_comments
- [x] See that it still works as expected
- [x] https://www.drupal.org/project/ajax_comments/releases/8.x-1.0-beta3 test beta 3 release to check if this release is applicable for OS or we need to lock it to beta2
- [x] See that the patch still works for `Ajax not working when using non-default view mode.`

## Release notes
Ajax comments is now updated to it's newest version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -99,7 +99,7 @@
                 "Make sure lazy doesnt do anything on cron": "https://www.drupal.org/files/issues/2019-07-30/3071331-lazy-cron-empty-path-2.patch"
             },
             "drupal/ajax_comments": {
-                "Fix display mode issue": "https://www.drupal.org/files/issues/2021-02-19/ajax_comments-ajax_not_working_when_using_non_default_view_mode-2896916-41-beta1.patch"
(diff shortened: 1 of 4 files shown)


ronaldtebrake:
Hi @navneet0693 if you have some time would be great to get your review :D

navneet0693:
@ronaldtebrake This patch works with the beta3 release of the module, but for existing websites, it will be important for them to upgrade the module while upgrading to the version of OS which contains this change.

navneet0693 (approved):

navneet0693:
🍒 -picked to 10.0.x and 8.x-9.x

PR #2271, opened 2021-03-23
Issue #3204322 by tBKoT: Allow flexible tagging structure for profiles tags

dmytriikaiun:
## Problem
We need to expand profile tagging functionality and make it look like social tagging.

## Solution
- SM is able to allow category split for profile tags
- SM is able to allow LU to add/edit profile tags
- Change field widget for the `Profile tag` field in the user profile form
- Change field widget for the `Profile tag` field on the user search filters
- Update profile templates to show the split structure of tags

## Issue tracker
https://www.drupal.org/project/social/issues/3204322
https://getopensocial.atlassian.net/browse/YANG-5091

This PR should be used with patches from theme issues:
- https://www.drupal.org/project/socialbase/issues/3205263
- https://www.drupal.org/project/socialblue/issues/3205264

## How to test
- [ ] Log in as SM+
- [ ] Add some profile tags (`/admin/structure/taxonomy/manage/profile_tag/overview`) with 2 level structure
- [ ] Go to the profile settings -> `admin/config/people/social-profile`
- [ ] Enable the `Allow category split`
- [ ] Go to the user profile edit page or to the user search page.
- [ ] Check changed the profile tag field widget\structure

## Screenshots
N/A

## Release notes
Now profile tags have the same structure as social content tags. Profile tags can be split to provide a more flexible structure. Additionally SM can decide if a user is allowed to add profile tags to the profile, so it keeps backward compatibility with previous behaviour.
So users can still use the Content tags on a Profile but can now also use separated dedicated Profile tags.

## Change Record
N/A

## Translations
N/A


modules/social_features/social_profile/src/Plugin/Field/FieldWidget/SocialSplitTagWidget.php
```diff
@@ -0,0 +1,147 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\select2\Plugin\Field\FieldWidget\Select2EntityReferenceWidget;
+use Drupal\social_profile\SocialProfileTagServiceInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Class SocialSplitTagWidget.
+ *
+ * @FieldWidget (
+ *   id = "social_split_tag",
```
ribel, line 15: @dmytriikaiun From the code, I see this widget works only for the profile tag field.
So maybe it's better to rename the machine name and class to something with a Social Profile prefix.

modules/social_features/social_profile/src/Plugin/views/filter/SocialSearchApiSplitTerms.php
```diff
@@ -0,0 +1,151 @@
...
+use Drupal\search_api\Plugin\views\filter\SearchApiTerm;
+use Drupal\social_profile\SocialProfileTagServiceInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Defines a filter for filtering on taxonomy term references.
+ *
+ * @ingroup views_filter_handlers
+ *
+ * @ViewsFilter("social_search_api_split_terms")
```
ribel, line 15: @dmytriikaiun From the code, I see this filter works only for the profile tag field.
So maybe it's better to rename the machine name and class to something with a Social Profile prefix.

modules/social_features/social_profile/config/install/social_profile.settings.yml
```diff
@@ -1,2 +1,5 @@
 social_profile_show_email: true
 social_profile_show_language: true
+enable_profile_tagging: true
```
ribel, line 3: @dmytriikaiun To keep backward compatibility I think we need to set FALSE by default for all those settings.

modules/social_features/social_profile/social_profile.install
```diff
@@ -433,3 +433,14 @@ function social_profile_update_8803() {
   // Output logged messages to related channel of update execution.
   return $updateHelper->logger()->output();
 }
+
+/**
+ * Updates profile configuration.
+ */
+function social_profile_update_8804(&$sandbox) {
+  $config = \Drupal::configFactory()->getEditable('social_profile.settings');
+  $config->set('enable_profile_tagging', TRUE);
```
ribel, line 442: @dmytriikaiun To keep backward compatibility I think we need to set FALSE by default for all those settings.

modules/social_features/social_profile/social_profile.links.menu.yml
```diff
@@ -1,5 +1,5 @@
 social_profile.settings:
- title: 'Profile privacy settings'
- description: 'Privacy settings for Social Profile module.'
+ title: 'Profile settings'
+ description: 'Profile settings for Social Profile module.'
```
ribel, line 3: @dmytriikaiun I see the point of changing the Title since it now not only privacy settings.
But I think we can keep old description since `Profile settings for Social Profile module` is a bit clumsy as well ;)

ribel (changes requested):
@dmytriikaiun Please check my comments.

modules/social_features/social_profile/social_profile.install
dmytriikaiun: @ribel I guess it better to remove hook_update

ribel (approved):
I have tested changes and all works as expected

ronaldtebrake (changes requested):
Thanks for your work guys! Some remarks from my side.

1. Could you ellaborate a bit more in your PR what changes you actually made? It seems you not only split the profile tags, you also added a new option to allow users to add profile tags as well, while this used to be a CM+ only function.

2. The update path isn't working, did you guys test this?

I have added a profile tag as SM to chris hall:

Screenshot 2021-03-31 at 12 31 34

checked out to your branch, no updates pending (eventhough schema files changed..)

As SM i'm going to profile edit, and the tag is not there anymore.

Screenshot 2021-03-31 at 12 32 59

modules/social_features/social_profile/src/Form/SocialProfileSettingsForm.php
```diff
@@ -98,6 +98,41 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#title' => $this->t('Tag settings'),
+      '#open' => TRUE,
+    ];
+
+    $form['tagging']['enable_profile_tagging'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Allow users to tag in profile.'),
+      '#required' => FALSE,
+      '#default_value' => $config->get('enable_profile_tagging'),
+      '#description' => $this->t('Determine whether users are allowed to tag profiles, view tags and filter on tags in profiles.'),
```
ronaldtebrake, line 112: > Determine whether users are allowed to tag profiles, view tags and filter on tags in profiles.

Don't you mean filter on tags in search?

Also would be great to directly link to the profile tag vocabulary, so a SM / CM can add tags directly. Otherwise they are searching around where to do this.

ribel:
Hi, guys (@ronaldtebrake, @jochemvn and @bramtenhove),
I agree that we need to give a bit more explanation to the description of the PR, but you can already found a bit more context from the acceptance criteria of linked stories. But let me also give some more understanding here.
First of all, we already have Social Tagging support for Profiles, which we added in version 10.0 here: #2137

But as we see in practice it doesn't always work for our customers, since they usually need to have a completely different set of taxonomies on Profiles and Content.

We were also discussing this feature during the ECI refinement with Moritz, where he first also proposed to define what exactly tags from social tagging would be available for what entity types, but then we all agreed that it would be better to actually just have the same abilities to split tags in profile tagging.

Imagine that for Site Managers it already may be complex to manage taxonomies in a current way - is it a good idea to make it even more complex? Isn't it better to re-use the same patterns in social tagging and profile tagging for consistency?

Last by not least, we received this request as very urgent in our team, from the customer from another team who needed to launch platform with this feature asap, which may also explain why implemented it so fast. 
But we probably don't need to agree on that next time with our sales and managers next time and verify all acceptance criteria with all interested teams and especially core team before implementation.

dmytriikaiun:
@ronaldtebrake I've added changes after your remarks
1. Now on the settings page SM can directly go to the profile tag vocabulary page (added link in description).
2. About allowing the user to add tags, it's still allowed only for CM+ users (changed text of setting).
3. For social_tagging you are not allowed to add the first-level tags so I copy this behavior too, but for profile tags, it should be possible to cover backward compatibility, so I've fixed it.

Please, chck changes.

modules/social_features/social_profile/src/Form/SocialProfileSettingsForm.php
```diff
@@ -98,6 +99,47 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#title' => $this->t('Tag settings'),
+      '#open' => TRUE,
+    ];
+
+    // Get profile vocabulary overview page link.
+    $profile_tags = Link::createFromRoute('profile tags', 'entity.taxonomy_vocabulary.overview_form', ['taxonomy_vocabulary' => 'profile_tag']);
+
+    $form['tagging']['enable_profile_tagging'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Allow to tag in profile'),
```
ronaldtebrake, line 113:
```suggestion
      '#title' => $this->t('Allow profiles to be tagged'),
```

modules/social_features/social_profile/src/Form/SocialProfileSettingsForm.php
```diff
@@ -98,6 +99,47 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#type' => 'checkbox',
+      '#title' => $this->t('Allow category split'),
+      '#default_value' => $config->get('allow_category_split'),
+      '#required' => FALSE,
+      '#description' => $this->t("Determine if the main categories of the vocabulary will be used as separate tag fields or as a single tag field when using tags on profile."),
+    ];
+
+    $form['tagging']['use_category_parent'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Allow use a parent of category'),
```
ronaldtebrake, line 132:
```suggestion
      '#title' => $this->t('Allow parents to be used as tag'),
```

ronaldtebrake:
Hi @dmytriikaiun, thanks for all the changes they look very good to me! 
Could you update the PR to explain to everybody what it is you've developed, that would be super helpful for others reading these changes. I think you've created even more value, not by splitting tags. No by separating the profile tags from the content tags. 

People can still use the content tags on a Profile but can now also use your newly created separated dedicated Profile tagging.

PR #2272, opened 2021-03-24
Issue #3205376 - The activity notification filter for follow taxonomy filters out content related to group_content / event_enrollment

ronaldtebrake:
## Problem
The activity notification filter for follow taxonomy filters out content related to group_content / event_enrollment as it only checks for certain entity and filters out the rest.

Activities like:
- User has enrolled to the event you are organizing
- Your membership has been approved.
(and more)

Are being filtered out on the notifications page.

This was a regression created in #2201 where the goal was to filter out notifications for content created based on tags you are following. This could be that you saw content which was placed in a group you are not a member of, but because it had a tag you're following it showed up in the notification list. 

## Solution
For this feature we should also care about entities such as event enrollment which should only be checked on the `recipient_user` our context ensures only the EO is a recipient so it will only show up for those users.
Same with `group_content`, which is a little more tricky. For GM we also have to check `recipient_user`, but for some cases we should also check on the group itself, thats because if the recipient is a group and not a user it dictates that all members of the group can receive it.

## Issue tracker
https://www.drupal.org/project/social/issues/3205376

## How to test
This filter is only added to the notification centre, so no other impact is required to be tested.
- [x] Please check all activities as per the notification list to see if they now end up in the notification centre
- [x] Make sure you don't get any notifications to content you cant access
- [x] Ensure all tests pass as ive created test coverage to check for event enrollment & group_content related notifications

## Release notes
We have fixed a regression in 10.0.0 which filtered out activities related to `group_content` & `event_enrollments` in the notification centre. 

## Screenshots
These were created on the same database, the only difference is my updated code.
Before you saw the notification icon showed there were 25 new notifications but it only showed:

Screenshot 2021-04-05 at 11 11 57

After switching branch to my code, clearing cache the following showed:
Screenshot 2021-04-05 at 11 10 42
Screenshot 2021-04-05 at 11 10 49


modules/custom/activity_viewer/src/Plugin/views/filter/ActivityNotificationVisibilityAccess.php
```diff
@@ -183,17 +183,58 @@ public function query() {
...
-      $vote_access->condition('activity__field_activity_recipient_user.field_activity_recipient_user_target_id', $account->id());
-
+      $vote_access->condition('activity__field_activity_recipient_user.field_activity_recipient_user_target_id', (string) $account->id());
       $or->condition($vote_access);
     }
 
+    // For an enrollment we match the recipient uid for enrollments
+    // activities are only created for those with the OrganizerActivityContext
+    // see getRecipientOrganizerFromEntity.
+    if ($account->isAuthenticated()) {
```
navneet0693, line 198: Just a small suggestion: We should add `$authenticated = $account->isAuthenticated();`, this will avoid multiple function calls at different lines.

navneet0693 (approved):

navneet0693:
🍒 to 10.0.x

PR #1994, opened 2020-09-22
Issue #3174544 and #3174547 by Kingdutch: Implement basic GraphQL support for users and profile data

Kingdutch:
## Problem
Open Social is slowly moving towards an API first approach.

This issue serves as a meta issue for community discussion and to track to work that will be required to build the new APIs as well as what can be done in the existing themes to utilise the new API structure.

## Solution
Open Social has selected version 4 of <a href="https://www.drupal.org/project/graphql">the GraphQL module</a> as it allows us to precisely control the shape of the Graph that we want to present. This allows us to better decouple a clear data graph from the internal data representation.

This pull request seeks to implement the base GraphQL module as well as the initial exposure of profile and user data. No authentication for external applications is implemented yet. The current implementation requires being authenticated with the default Drupal browser session.

The included GraphQL module is hidden and experimental for Open Social's internal use and evaluation at the moment and no support is yet provided until the covered parts of the Graph are deemed stable.

## Issue tracker
- [Issue 3174544: [GraphQL] Implement base GraphQL module](https://www.drupal.org/project/social/issues/3174544)
- [Issue 3174547: [GraphQL] Implement basic user and profile querying in the Graph](https://www.drupal.org/project/social/issues/3174547)
- [Issue 3191701: User creation breaks if no profile type exists](https://www.drupal.org/project/social/issues/3191701)
- [Issue 3191702: Provided user schema is incorrect and unused](https://www.drupal.org/project/social/issues/3191702)
- [Issue 3191703: Allow attaching libraries to AccountHeader elements](https://www.drupal.org/project/social/issues/3191703)
- [Issue 3191704: Open Social should only run its own PHPUnit tests](https://www.drupal.org/project/social/issues/3191704)

## How to test
- [x] Enable the social_grahpql module
- [x] As a LU go to `/admin/config/graphql/servers/manage/open_social_graphql/explorer`
- [x] You should be able to query user and profile data.

## Documentation
A walk-through of how this all works is available at https://www.alexandervarwijk.com/blog/2020-11-18-graphql-in-the-open-social-drupal-distribution/

## Screenshots
afbeelding

### Issue 3174544 & 3174547
Open Social now includes an experimental GraphQL module. This module is in active development and no support will be provided for it at this time.

### Issue 3191701
User creation in tests no longer fails if no profile types have been configured.

### Issue 3191702
The schema for the social_user module has been fixed and true/false values are now proper booleans.

### Issue 3191703
[...]

PR #2273, opened 2021-03-25
Issue #3205576 by robertragas: Change the header icon condition to bools

robertragas:
## Problem
On the top right we have an account header menu where we can add some optional icons defined on
admin/config/opensocial/navigation-settings

These settings do not work anymore on a 10.x install.

## Solution
In an older PR https://github.com/goalgorilla/open_social/pull/1994/files changes got made to replace the value being set to a bool. The check itself in the header menu does not take this new bool into account, always resulting in returning.
We also need to update this value to check for the bool.

## Issue tracker
https://www.drupal.org/project/social/issues/3205576

## How to test
- [ ] Install Open Social on 10.x
- [ ] Enable social_private_message
- [ ] Go to admin/config/opensocial/navigation-settings and see both are enabled
- [ ] See they are not showing in the header menu
- [ ] Check out to this branch and check again

## Screenshots

## Release notes
We have solved a bug where a fresh install on 10.x would not show the header icons of private messages and groups correctly when enabled.

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -2098,7 +2098,7 @@ function social_group_social_user_account_header_create_links($context) {
  * site manager.
  */
 function social_group_social_user_account_header_items(array $context) {
-  if (\Drupal::config('social_user.navigation.settings')->get('display_my_groups_icon') !== 1) {
+  if (\Drupal::config('social_user.navigation.settings')->get('display_my_groups_icon') !== TRUE) {
     return [];
   }
 
--- a/modules/social_features/social_private_message/social_private_message.module
+++ b/modules/social_features/social_private_message/social_private_message.module
@@ -493,7 +493,7 @@ function social_private_message_activity_send_email_notifications_alter(array &$
  * it's enabled by the site manager.
  */
 function social_private_message_social_user_account_header_items(array $context) {
-  if (\Drupal::config('social_user.navigation.settings')->get('display_social_private_message_icon') !== 1) {
+  if (\Drupal::config('social_user.navigation.settings')->get('display_social_private_message_icon') !== TRUE) {
     return [];
   }
 


modules/social_features/social_user/social_user.install
```diff
@@ -206,3 +206,18 @@ function social_user_update_8810() {
     'view blocked user',
   ]);
 }
+
+/**
+ * Migrate navigation settings into the correct format.
+ */
+function social_user_update_8811() {
```
Kingdutch, line 213: I think config updates are better in post_update hooks, since we rely on schemas having stabilised? I'm never quite sure though.

modules/social_features/social_user/social_user.install
```diff
@@ -206,3 +206,18 @@ function social_user_update_8810() {
...
+  $config = \Drupal::configFactory()->getEditable('social_user.navigation.settings');
+  // If someone manually changed the value in the form then skip it.
+  if (!is_bool($config->get('display_my_groups_icon'))) {
+    $config->set('display_my_groups_icon', $config->get('display_my_groups_icon') === 1 ? TRUE : FALSE);
+  }
```
Kingdutch, line 218: I like the fact that we're being explicit but this may be the perfect place for type-casting :D 

```suggestion
  $config->set('display_my_groups_icon', (bool)$config->get('display_my_groups_icon'));
```

This even works for strings:
```php
echo '"1": '; var_dump((bool)"1");
echo '"0": '; var_dump((bool)"0");

echo '1: '; var_dump((bool)1);
echo '0: '; var_dump((bool)0);
```

outputs
```
"1": bool(true)
"0": bool(false)
1: bool(true)
0: bool(false)
```

modules/social_features/social_user/social_user.install
```diff
@@ -206,3 +206,18 @@ function social_user_update_8810() {
...
+  if (!is_bool($config->get('display_social_private_message_icon'))) {
+    $config->set('display_social_private_message_icon', $config->get('display_social_private_message_icon') === 1 ? TRUE : FALSE);
+  }
```
Kingdutch, line 221:
```suggestion
  $config->set('display_social_private_message_icon', (bool)$config->get('display_social_private_message_icon'));
  }
```

Kingdutch (commented):
Whoops! That's one I should've caught and fixed earlier.

Great improvement! :D

nachosalvador (approved):

modules/social_features/social_user/social_user.install
robertragas: yeah think you are right.

PR #2276, opened 2021-03-27
Issue #3206028: UI improvements according UX design for 10.x (for Event)

nechai:
## Problem
Change field titles for Event node type according to design.

## Issue tracker
- https://www.drupal.org/project/social/issues/3206028
- https://getopensocial.atlassian.net/browse/ECI-1157


modules/social_features/social_event/social_event.module
```diff
@@ -1243,3 +1243,11 @@ function social_event_preprocess_fieldset(&$variables) {
     }
   }
 }
+
+/**
+ * Implements hook_social_core_compatible_content_forms_alter().
+ */
+function social_event_social_core_compatible_content_forms_alter(array &$compatible_content_type_forms) {
```
ribel, line 1250: We don't need this hook for core content types, since they already added in the `social_core_form_node_form_alter()`.

ribel:
I think we can close this PR for now as it has a low prio. 
Also such changes as from Title to Name, etc needs to be consistent for all content types.

PR #2280, opened 2021-04-01
Issue #3203225: revert the package name to goalgorilla/open_social and use an updated travis config

ronaldtebrake:
This PR is in favor of #2278

## Problem
Resolves an issue introduced in #2259 where standardizing our package name caused issues with drupal's infrastructure, as well as the fact we have widened `goalgorilla/open_social` replace range caused composer to think that 10.1.x-dev can
replace 10.0.3, which the replace key says is true but it's not what we want. 

## Solution
Revert the package name to `goalgorilla/open_social` whilst maintaining some of our travis improvements. Also we make sure we add `drupal/social` to our composer.json replace attribute using the `self.version` to make sure external dependencies can still depend on drupal/social versions

## Issue tracker
https://www.drupal.org/project/social/issues/3203225

## How to test
- [ ] Test should pass.
- [ ] You should be able to install open social again through our social template once 

## Change Record
We can postpone https://www.drupal.org/node/3203229 untill we can finally standardize our package name.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "drupal/social",
+    "name": "goalgorilla/open_social",
     "description": "Open Social is a distribution for building social communities and intranets.",
     "type": "drupal-profile",
     "license": "GPL-2.0-or-later",
@@ -204,7 +204,7 @@
         "mglaman/drupal-check": "^1.0"
     },
     "replace": {
-      "goalgorilla/open_social": ">=10.0.0 <11.0.0"
+      "drupal/social": "self.version"
     },
     "conflict": {
       "goalgorilla/open_social": "<10.0.0 || >=11.0.0"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -204,7 +204,7 @@
         "mglaman/drupal-check": "^1.0"
     },
     "replace": {
-      "goalgorilla/open_social": ">=10.0.0 <11.0.0"
+      "drupal/social": "self.version"
     },
     "conflict": {
       "goalgorilla/open_social": "<10.0.0 || >=11.0.0"
```
Kingdutch, line 210:
```suggestion
```

We should remove this :D

Kingdutch (changes requested):

composer.json
ronaldtebrake: Good point!

ronaldtebrake:
RTBC'ed and approved by @Kingdutch

PR #2281, opened 2021-04-01
#2880361 Move social_post update hooks

harlor:
This workaround moves the social_post update hooks into a new submodule to avoid naming convention issue with post update hooks.

See https://www.drupal.org/project/social/issues/2880361


ronaldtebrake:
Thanks for your efforts here @harlor, it is greatly appreciated

For now we will try and get it into Drupal core, as it's definitely not an Open Social architectural flaw but the way Drupal deals with getting the available updates & post_updates.

See for more details here: https://www.drupal.org/project/drupal/issues/3236316 and a proposed solution in this merge request: https://git.drupalcode.org/project/drupal/-/merge_requests/1215

In this case the direction chosen will also be hell to maintain. As update helper stores all their update files still in `social_post`, the `hook_update_dependencies` doesn't match at all.

So eventhough this might work on the short term, the longterm effects added are unfortunately not something we can work with and we rather focus our efforts in making Drupal capable of handling it.

PR #2282, opened 2021-04-02
Issue #3200113 by rolki: Enable visibility states only for the flexible group

rollki:
## Problem
There are some new issues with saving visibilities for other group types when we use `hook_social_group_settings_alter`, so it means states from flexible group can't be reused for other group types yet.

## Solution
Revert changes for group visibility states

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-5116
https://www.drupal.org/project/social/issues/3200113

## How to test
- [ ] Add `hook_social_group_settings_alter` to other group type
- [ ] Go to edit or create this group page
- [ ] Select some visibilities
- [ ] After saving visibility saved with the wrong values

## Screenshots
image


ribel (approved):
This PR fixes the regression made in PR #2248, which was not released yet.

PR #2283, opened 2021-04-02
Issue #3206847 by tBKoT: Allow event organizers to disable event enrolments

dmytriikaiun:
## Problem
There is no longer a way to hide enroll option on events.

## Solution
As designed here https://www.figma.com/proto/eFn9UYCaSnerZgx9vxsmbR/Open-Social-Design-System?node-id=6630%3A0&viewport=739%2C1292%2C1&scaling=min-zoom we need to add a new option to enable/disable enrollment globally on the platform and new settings in an event to enable/disable enrollment per event.

## Issue tracker
https://getopensocial.atlassian.net/browse/YANG-4146
https://www.drupal.org/project/social/issues/3206847

## How to test
*For example*
- [ ] Go to the event settings page `admin/config/opensocial/event` as a sitemanager
- [ ] You should be able to enable/disable option `Allow event organizers to disable event enrolments`
- [ ] If disable all event enrollment buttons should not display
- [ ] If enabled event organizers can decide to enable/disable event enrollment

## Screenshots
N/A

## Release notes
Now SM can disable event enrollments platform-wide and event organizers could disable event enrollment if allowed to.

## Change Record
N/A

## Translations
N/A


ribel (approved):
I have tested these changes from OS 10.1.x branch and all works as expected for me.

ribel:
Hi, @ronaldtebrake,
Not sure if the ability to disable the event enrollment button was removed by intense, but our customer reported it as a critical issue: https://getopensocial.atlassian.net/browse/YANG-4146

> It looks like with the new update there is no way to have an event content type that does not have the enroll option.
> This is an issue for events that may use a different RSVP/enrollment system, or simply don't have a need to track enrollments.
> Can we please add this back?

We added this option back but would like you to hear your input and approval on that.

modules/social_features/social_event/social_event.install
```diff
@@ -1796,3 +1797,55 @@ function social_event_update_8915() {
     }
   }
 }
+
+/**
+ * Add the 'enable event enrolment' field.
+ */
+function social_event_update_8916() {
+  $config_path = drupal_get_path('module', 'social_event') . '/config/install';
```
ronaldtebrake, line 1805: Please use static files instead for update hooks.

modules/social_features/social_event/src/Form/EventSettingsForm.php
```diff
@@ -50,31 +102,44 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
 
+    $form['event_enrolment'] = [
+      '#type' => 'details',
+      '#title' => $this->t('Event enrollment settings'),
+      '#open' => TRUE,
+    ];
+
+    $form['event_enrolment']['allow_to_disable_enroll'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Allow event organizers to disable event enrolments'),
```
ronaldtebrake, line 139: I think Enrollment is standard in American English, whereas enrolment is standard in British English, so in this case we should stick to the double ll like u used in the other parts as well.

dmytriikaiun:
@ronaldtebrake I've added changes regarding the comments. Please, check.

ronaldtebrake:
Thanks @dmytriikaiun for the update, one more remark I had after the functional review.

https://www.loom.com/share/3e0cac964c114dcfb0dc7c83b31abddc

1. After installing this, as a event organizer decide to disallow users to enroll to your event. This works as expected.
2. As a SM go to the event settings page, uncheck the Event enrollment settings. This works as expected.
3. As a event organizer go to your event, the enrollment methods are now gone and there is no way for them to make changes anymore. We need to ensure in that case users go back to normal.

dmytriikaiun:
@ronaldtebrake about 3rd point. It was made like this because of our AC
>  should be possible for SM to decide if 
> - enrolment can be disabled or not per event base
> - enrolment is disabled platform-wide

And for these behaviors, we have only one checkbox on the design of the settings page.
A possible solution to make it more flexible is:
1. Add a new option on the settings page `Disable enrolment for all events`
2. When we uncheck the Event enrolment setting we also uncheck settings in the event and make it disabled from editing.

Which behavior is more preferred for you or maybe you have your own vision about that?

ronaldtebrake:
Ah ok I see thanks for the clarification @dmytriikaiun i misread that in the PR itself and from the HTT steps, sorry.

I think in that case it does make sense to keep it like it is but we should rename the setting a bit and maybe add some description as this is quite a heavy setting without it being clear from the settings itself.

The actual value of this setting is not for a SM to decide if a Event Organizer can decide it per event but they can actually disable event enrollments for all events platform wide with that setting, so we should state that and add a description stating, if it's enabled the event organizers can decide per event what they prefer. 

Does that make sense to you too? @dmytriikaiun

dmytriikaiun:
@ronaldtebrake Yes, we can add a description to this setting to make it more clear for a SM

ribel:
@ronaldtebrake so looks like all is done here? Can it be merged then?

modules/social_features/social_event/src/Form/EventSettingsForm.php
```diff
@@ -50,31 +102,45 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
 
+    $form['event_enrolment'] = [
+      '#type' => 'details',
+      '#title' => $this->t('Event enrollment settings'),
+      '#open' => TRUE,
+    ];
+
+    $form['event_enrolment']['allow_to_disable_enroll'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Allow event organizers to disable event enrollments'),
```
ronaldtebrake, line 139:
```suggestion
      '#title' => $this->t('Disable all event enrollments on your community.'),
```

modules/social_features/social_event/src/Form/EventSettingsForm.php
```diff
@@ -50,31 +102,45 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#description' => $this->t('If enabled, event organizers can disable event enrollment. In another case, enrollment will be disabled for all events as default'),
```
ronaldtebrake, line 140:
```suggestion
      '#description' => $this->t('If disabled, event organizers can decide to disable or enable event enrollments when creating or editing an event.'),
```

ronaldtebrake:
Hi @ribel @dmytriikaiun I've proposed some small changes to make the value more clear as to what the checkbox really does for a SM. After that if you also agree i'm all good with it!

dmytriikaiun:
@ronaldtebrake I've added your proposals, check please

modules/social_features/social_event/social_event.post_update.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+/**
+ * @file
+ * Post update functions for the Social Course Quiz module.
```
navneet0693, line 5: Copy-paste errors 🙈

modules/social_features/social_event/social_event.post_update.php
```diff
@@ -0,0 +1,48 @@
...
+ * Update existing events.
+ */
+function social_event_post_update_update_events(&$sandbox) {
+  /** @var \Drupal\node\NodeStorageInterface $node_storage */
+  $node_storage = \Drupal::entityTypeManager()->getStorage('node');
+
+  if (!isset($sandbox['total'])) {
+    // Get all event ids.
+    $sandbox['ids'] = $node_storage
+      ->getQuery()
```
navneet0693, line 18: `$node_storage->getQuery()->accessCheck(FALSE)->condition('type', 'event')->execute();` would have been a very nice idea as this query might collection only public events.

PR #2290, opened 2021-04-12
Issue #3203750 by robertragas: set max height to the photo album pop-…

robertragas:
…up to avoid pictures being cut off

## Problem
In 10.x the photo album feature got introduced which opens images in a pop-up. When a picture has a portrait ratio it has the chance of being cut off only showing a part of the image.

## Solution
Set a max-height of 100vh on the pop-up img class so it will never become higher than the actual screen.

## Issue tracker
https://www.drupal.org/project/social/issues/3203750

## How to test
- [ ] Upload a portrait image to the stream that means the height must be bigger than the width.
- [ ] Click on the image in the stream where it will open in a pop-up
- [ ] See the image is being cut off and the top of the image cannot be seen
- [ ] Check out to this branch and check again
- [ ] Also try it out with different format of images to make sure they also work

## Screenshots
Before
Screen Shot 2021-04-12 at 10 21 27 PM

After
Screen Shot 2021-04-12 at 10 21 52 PM

## Release notes
We have resolved an issue where the photo album pop-up would cut off a portrait image resulting in only showing it partially. It should now respect the max height of the browser window.


robertragas:
Hey @nachosalvador could you test this PR?

@agami4 I also added you as a reviewer, as you could maybe just check the file diff in github to see if you think my solution is good enough.

nachosalvador (approved):

BSimon47:
May I suggest moving the resulting image to the center?
robertregas-album-height-fix-revised

Also, IMO, we could survive reducing the margins all around by no less than 1/2 (as image properties allow)
Thanks

robertragas:
> May I suggest moving the resulting image to the center?
> robertregas-album-height-fix-revised
> 
> Also, IMO, we could survive reducing the margins all around by no less than 1/2 (as image properties allow) - in this case top and bottom!
> 
> Thanks

Hello @BSimon47,

About the centering of the image, this should be correct but the screenshot I took was not of the full screen. See below for how it looks.
Screen Shot 2021-04-13 at 5 33 43 PM

I will check if removing the padding at the top and bottom is a possibility.

PR #2291, opened 2021-04-13
[Internal] Change link from getopensocial.com/services to Review Open Social on Capterra

sjoerdvandervis:
## Problem
The link in the social_lets_connect_contact module to `Services` leads to a 404 on getopensocial.com . We would like to change this to Review Open Social on Capterra.

## Solution
Change the link in said module to link (Distro) users to reviewing Open Social on Capterra.

## Issue tracker
https://getopensocial.atlassian.net/browse/SAL-21

## How to test
- [ ] As an SM, see that the link in the social_lets_connect_contact module now points to the review page on Capterra.

### DIFF
--- a/modules/custom/social_lets_connect/modules/social_lets_connect_contact/social_lets_connect_contact.links.menu.yml
+++ b/modules/custom/social_lets_connect/modules/social_lets_connect_contact/social_lets_connect_contact.links.menu.yml
@@ -4,10 +4,10 @@ social_lets_connect_contact.extensions:
   url: https://www.getopensocial.com/extensions
   weight: -6
 
-social_lets_connect_contact.services:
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
Looks good to me @sjoerdvandervis 
just wondering if it's really the idea to have it there for all users? Everybody who downloads the open source distro gets an admin account so regardless of where they are in the funnel will be able to add that review. 
That could also mean you get reviews from people who have issues with maintenance/hosting instead of product reviews etc and we have no way to verify what they are encountering as its on their hosting.

If thats what sales/marketing wants, you can definitely merge it, just wanted to ensure the team understands that potential impact :)

ronaldtebrake (approved):

PR #2294, opened 2021-04-14
Issue #3208745 by agami4: Update GIN theme to newer version

agami4:
## Problem
The Gin theme has not the latest version.

## Solution
Improve the version of the gin theme on the composer.json file on the profile.

## Issue tracker
https://www.drupal.org/project/social/issues/3208745
https://getopensocial.atlassian.net/browse/DS-7587

## Release notes
The Gin theme was updated to the latest version.


navneet0693 (approved):

composer.json
```diff
@@ -142,7 +142,7 @@
         "drupal/field_group": "3.1",
         "drupal/file_mdm": "1.1",
         "drupal/flag": "4.0-alpha3",
-        "drupal/gin": "3.0-alpha20",
+        "drupal/gin": "^3.0",
```
navneet0693, line 145: We need to fix this version to avoid accidental updates of theme in future.

navneet0693 (changes requested):

composer.json
ronaldtebrake, line 145: Thanks for updating @agami4 & @ribel 
However I think in this case it's not a good idea to have this caret constrained.
Gin is still in alpha, which means that whenever someone updates / installs Open Social it will grab the latest available 3.0 version without us knowing.
Since it's in alpha it can potentially mean there are still potential backwards compatible breaking changes. 

So imagine that happen when someone runs a `composer update goalgorilla/open_social -W` that means their site will break.
Ideally we lock this on the alpha version which you test it, and caret it once its in a RC / stable release.

ronaldtebrake (changes requested):

composer.json
agami4: Hey @ronaldtebrake, thank you for the response. I've added the latest alpha version of the gin theme.

navneet0693 (approved):

ronaldtebrake:
Restarted tests, if it turns green with the approved review by Navneet it can be merged, thanks :D !

PR #2298, opened 2021-04-20
Issue #3209766 by tBKoT: Improve access for unpublished content

dmytriikaiun:
## Problem
AN/LU users have access to the unpublished content.
This bug was found by Robert when he tested scheduled publishing.

## Solution
Needs logic improvements in the `entity_access_by_field` module. Or rather in the class that is responsible for access to the content by visibility fields (EntityAccessHelper).

## Issue tracker
https://www.drupal.org/project/social/issues/3209766
https://getopensocial.atlassian.net/browse/YANG-5479

## How to test
- [ ] Using the latest version of Open Social
- [ ] Create an unpublished topic with public/community visibility settings
- [ ] Go to the crated topic page directly as AN/LU user
- [ ] You have access to this page but shouldn't

## Screenshots
N/A

## Release notes
N/A

## Change Record
N/A

## Translations
N/A


ronaldtebrake:
Hi @dmytriikaiun, thanks for this!
Can we cover this in a test? 
Seems very useful for us to understand what went wrong and to ensure we are not creating that regression. 
The PR itself is a bit confusing, it seems you're now returning something for users with administer concent permissions, however the PR text is about LU/AN and unpublished and the HTT seems to talk about creating content without it even having to be unpublished? 

So I'm not sure how this compares to what you are solving here and what I need to test / check for this..

ronaldtebrake:
Hi @dmytriikaiun,

Thanks for your comment :) that clears it up a bit.
Would it be possible to write a small test to cover this scenario? To ensure we don't create this regression anymore?

After that I'd be happy to have this in :D

ribel (approved):

ribel:
🍒 picked to 10.1.x and 10.0.x

PR #2301, opened 2021-04-23
Issue #3210386 by Ressinel: Fix - Users see hidden First name and Last name.

Ressinel:
## Problem
If you enable social_profile_privacy module and set the First name or Last name as hidden, users will see Fullname anyway in all places where it shows.

## Solution
Add additional checks in social_profile_privacy module

## Issue tracker
- https://www.drupal.org/project/social/issues/3210386

## How to test
- [ ] Enable social_profile_privacy module
- [ ] Go to the settings page `/admin/config/people/social-profile`
- [ ] Select "Hide for others" for First name or Last name
- [ ] As LU you see Fullname (First name + Last name) for all users

## Screenshots
N/A

## Release notes
N/A

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_profile/modules/social_profile_privacy/social_profile_privacy.module
+++ b/modules/social_features/social_profile/modules/social_profile_privacy/social_profile_privacy.module
@@ -316,18 +316,112 @@ function social_profile_privacy_entity_field_access($operation, FieldDefinitionI
  * users that are allowed to see a full name even when there's a nickname.
  */
 function social_profile_privacy_social_user_name_display_suggestions_alter(array &$suggestions, AccountInterface $account) {
+  $current_user = \Drupal::currentUser();
   $config = \Drupal::config('social_profile_privacy.settings');
 
   if (
     $config->get('limit_search_and_mention')
     && isset($suggestions['full_name'], $suggestions['nickname'])
-    && \Drupal::currentUser()->hasPermission('social profile privacy always show full name')
+    && $current_user->hasPermission('social profile privacy always show full name')
   ) {
     $suggestions['nickname_with_full_name'] = [
       'weight' => -PHP_INT_MAX,
       'name' => $suggestions['nickname']['name'] . ' (' . $suggestions['full_name']['name'] . ')',
     ];
   }
+
+  // If the current user has whether permission to view full name or hidden
+  // fields then do nothing here.
+  if ($current_user->hasPermission('social profile privacy always show full name') || $current_user->hasPermission('social profile privacy view hidden fields')) {
+    return;
+  }
+
+  $uid = $account->id();
+  $fields = social_profile_privacy_private_fields_list($uid);
+
+  // If First name and Last name are hidden then suggestion with check access
+  // for those fields doesn't add for Full name.
+  // @see social_profile_privacy_social_user_name_display_suggestions()
+  // So we need to remove the Full name suggestion if all of those fields are
+  // hidden.
+  if (isset($suggestions['full_name'])) {
+    // Profile fields which use for composing the full name.
+    $account_name_fields = [
+      'field_profile_first_name',
+      'field_profile_last_name',
+    ];
+    $account_name_hidden = TRUE;
+    // If at least one field is not hidden then we don't need to remove the
+    // Full name suggestion.
+    foreach ($account_name_fields as $account_name_field) {
+      if (!in_array($account_name_field, $fields)) {
+        $account_name_hidden = FALSE;
+        break;
+      }
+    }
+    if ($account_name_hidden) {
+      unset($suggestions['full_name']);
+    }
+  }
+
+  // Remove nickname suggestion if module social_profile_fields are installed
+  // and field is hidden.
+  if (\Drupal::moduleHandler()->moduleExists('social_profile_fields') && isset($suggestions['nickname']) && in_array('field_profile_nick_name', $fields)) {
+    unset($suggestions['nickname']);
+  }
+}
+
+/**
+ * Implements hook_social_user_name_display_suggestions().
+ *
+ * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+ * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+ */
+function social_profile_privacy_social_user_name_display_suggestions(AccountInterface $account) : array {
+  $current_user = \Drupal::currentUser();
+  // If the current user has whether permission to view full name or hidden
+  // fields then do nothing here.
+  if ($current_user->hasPermission('social profile privacy always show full name') || $current_user->hasPermission('social profile privacy view hidden fields')) {
+    return [];
+  }
+
+  $suggestions = [];
+
+  /** @var \Drupal\profile\ProfileStorageInterface $storage */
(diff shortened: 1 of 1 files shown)


Kingdutch:
Hey @Ressinel ,

It looks like we're duplicating quite a bit of functionality here (basically re-implementing `social_profile_social_user_name_display_suggestions`) to fix a very specific issue. I have a suggestion for an alternative solution, that's more flexible and simpler at the same time.

The real problem here is that `social_profile_social_user_name_display_suggestions` which creates the `full_name` suggestion doesn't implement field level access handling. That can be solved by applying the following diff:

```diff
diff --git a/modules/social_features/social_profile/social_profile.module b/modules/social_features/social_profile/social_profile.module
index a20209f50..457d359e4 100644
--- a/modules/social_features/social_profile/social_profile.module
+++ b/modules/social_features/social_profile/social_profile.module
@@ -729,16 +729,29 @@ function social_profile_get_default_image() {
 function social_profile_social_user_name_display_suggestions(AccountInterface $account) : array {
   $suggestions = [];
 
-  /** @var \Drupal\profile\ProfileStorageInterface $storage */
-  $storage = \Drupal::entityTypeManager()->getStorage('profile');
+  $entitTypeManager = \Drupal::entityTypeManager();
 
-  if ($user_profile = $storage->loadByUser($account, 'profile', TRUE)) {
-    $accountname = trim($user_profile->get('field_profile_first_name')->value . " " . $user_profile->get('field_profile_last_name')->value);
-
-    if (strlen($accountname) > 0) {
+  /** @var \Drupal\profile\ProfileStorageInterface $storage */
+  $storage = $entitTypeManager->getStorage('profile');
+
+  if ($user_profile = $storage->loadByUser($account, 'profile')) {
+    /** @var \Drupal\profile\ProfileAccessControlHandler $accessControlHandler */
+    $accessControlHandler = $entitTypeManager->getAccessControlHandler('profile');
+    $first_name_field = $user_profile->get('field_profile_first_name');
+    $last_name_field = $user_profile->get('field_profile_last_name');
+
+    $first_name = $accessControlHandler->fieldAccess('view', $first_name_field->getFieldDefinition())
+      ? $first_name_field->value
+      : "";
+    $last_name = $accessControlHandler->fieldAccess('view', $last_name_field->getFieldDefinition())
+      ? $last_name_field->value
+      : "";
+    $accountname = trim($first_name . " " . $last_name);
+
+    if ($accountname !== '') {
       // Add the full name with a default weight.
       $suggestions['full_name'] = [
-        'name' => trim($accountname),
+        'name' => $accountname,
       ];
     }
   }

```

The remaining task is to implement `hook_entity_field_access` in the `social_profile_privacy` module to apply a permission based access check. The benefit of doing it that way is that we don't need to write code to hide the first name/last name in every scenario (e.g. the GraphQL API in the future) but can use Drupal's access system.

Ressinel:
Hi @Kingdutch 
Thanks for your response! Sounds good! I Will check and try to implement your suggestion.

Ressinel:
@Kingdutch 
When we use check field access by the field definition in `social_profile_social_user_name_display_suggestions` we always receive `$items = NULL` for `social_profile_privacy_entity_field_access`. And we can't get user ID which we need to get hidden fields by `social_profile_privacy_private_fields_list($uid)`. 
Are there any thoughts on this?

Kingdutch:
@Ressinel We have the fields when invoking `fieldAccess` and its signature is `public function fieldAccess($operation, FieldDefinitionInterface $field_definition, AccountInterface $account = NULL, FieldItemListInterface $items = NULL, $return_as_object = FALSE)` so we can pass in the items when calling it? :)

Ressinel:
Hi @Kingdutch 

Can you check PR updates, please?

Thanks!

modules/social_features/social_profile/modules/social_profile_privacy/social_profile_privacy.module
```diff
@@ -316,18 +316,59 @@ function social_profile_privacy_entity_field_access($operation, FieldDefinitionI
...
+    if ($account_name_hidden) {
+      unset($suggestions['full_name']);
+    }
+  }
+
+  // Remove nickname suggestion if module social_profile_fields are installed
+  // and field is hidden.
+  if (\Drupal::moduleHandler()->moduleExists('social_profile_fields') && isset($suggestions['nickname']) && in_array('field_profile_nick_name', $fields)) {
+    unset($suggestions['nickname']);
+  }
```
Kingdutch, line 371: This shouldn't be needed, this should be implemented by the field level access check for those two fields. However, I'm missing that in this PR.

Kingdutch:
@Ressinel @ronaldtebrake Given the nature of this issue we need to add test coverage so that we can catch regressions in the future. That should be added as part of this issue. The test coverage also helps to show what is the desired behaviour (so it's easier to debate whether access checks are implemented correctly).

Scenarios I can come up with:

- [ ] View own display name (e.g. in a post) as AN
- [ ] View own display name (e.g. in a post) as LU
- [ ] View own display name (e.g. in a post) as CM
- [ ] View own display name (e.g. in a post) as SM
- [ ] View own profile as AN
- [ ] View own profile as LU
- [ ] View own profile as CM
- [ ] View own profile as SM
- [ ] View other's display name (e.g. in a post) as AN
- [ ] View other's display name (e.g. in a post) as LU
- [ ] View other's display name (e.g. in a post) as CM
- [ ] View other's display name (e.g. in a post) as SM
- [ ] View other's profile as AN
- [ ] View other's profile as LU
- [ ] View other's profile as CM
- [ ] View other's profile as SM

Repeat above tests for different profile settings (e.g. "show first name" or "hide first name", etc.) It's probably easiest to group the tests by role (e.g. "As an AN when I view my own post then I expect and when I view my own profile then I expect" etc.)

Theoretically we could also test for permissions instead of specific roles. However, I think in Open Social's case it's a good idea to test for the individual roles so that the permission assignment as we intended it is also captured in our tests.

modules/social_features/social_profile/modules/social_profile_privacy/social_profile_privacy.module
Ressinel: Hi, @Kingdutch  
Yes, you're right, I have removed it and add a new check for the nickname field into the `social_profile_fields` module. Check it, please.

Ressinel:
Hi, @Kingdutch Thanks for your response!

Hi, @ronaldtebrake Let me know if all tests needed to implement.

Thanks!

ronaldtebrake:
Hi @Ressinel 

We already have `tests/behat/features/capabilities/profile/restricted-full-name.feature` basically we need to check what is missing there. We should extend it with a test covering the exact scenario we have fixed. But it's already partially covering the tests Alexander mentioned.

Ressinel:
@Kingdutch @ronaldtebrake I have added new scenarios to `tests/behat/features/capabilities/profile/restricted-full-name.feature`.

ronaldtebrake (approved):
Very nicely done! Also with checking the non-happy flow in the tests and implementing all the feedback. Very happy with the result @Ressinel :D

Ressinel:
@ronaldtebrake 

Nice to hear it 😁
Thanks!

PR #2302, opened 2021-04-23
Issue #3210411 by chmez: When visiting and clicking on a notification in the Notification Center it does not go away

ghost:
## Problem
When visiting and clicking on a notification in the Notification Center it does not go away.

## Solution
Mark new notifications as read when the current user visits the notifications page a second time.

## Issue tracker
- https://www.drupal.org/project/social/issues/3210411
- https://getopensocial.atlassian.net/browse/YANG-4086

## How to test
- [ ] Log in as "User 1" and create a post on a stream of the "User 2" user.
- [ ] When you log in as "User 2" from a mobile device and navigate to the "Notification Center" page (/notifications) then you should see a notification in the 1st position as unread (notification should have grey background).
- [ ] When you refresh the page then you should see the notification in the 1st position as read (notification shouldn't have grey background).

## Release notes
Mark new notifications as read when the current user visits the notifications page a second time.

### DIFF
--- a/modules/custom/activity_creator/activity.page.inc
+++ b/modules/custom/activity_creator/activity.page.inc
@@ -129,6 +129,7 @@ function template_preprocess_activity(array &$variables) {
 
       if (\Drupal::routeMatch()->getRouteName() === 'view.activity_stream_notifications.page_1') {
         $activity_notification_service->changeStatusOfActivity([$activity->id()], \Drupal::currentUser(), ACTIVITY_STATUS_SEEN);
+        $variables['#cache']['max-age'] = 0;
       }
 
       break;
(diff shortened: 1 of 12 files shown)


modules/custom/activity_creator/activity.page.inc
```diff
@@ -122,6 +126,12 @@ function template_preprocess_activity(array &$variables) {
   switch ($status) {
     case ACTIVITY_STATUS_RECEIVED:
       $status_class = 'bg-gray-lightest';
+
```
ribel, line 129: @OH17L Please add some comment with the explanation why do we need max-age 0 here.

OH17L: Done.

ribel (approved):

ribel:
🍒 picked to 10.0.x and 10.1.x

PR #2307, opened 2021-04-29
Issue #3171966 - Update search_api to latest version to allow usage of current solr submodule

slowflyer:
## Problem
With current version 1.15 of search_api it is not possible to use suggested or latest version of the solr submodule.

## Solution
Update to search_api 1.19

## Issue tracker
https://www.drupal.org/project/social/issues/3171966

## How to test
Check serach configuration and use search function on your install


Kingdutch (changes requested):
Thanks so far @slowflyer ! Looks like a good improvement. We may be able to get rid of even more code and probably need to update our install config for good measure :) 

Based on a look at the Search API release notes this needs a bit of work: https://www.drupal.org/project/social/issues/3171966#comment-14090657

We also may need to include some of the things from the Search API release notes in our own release notes (see my comment on the issue).

slowflyer:
After some testing yesterday evening, I'm convinced that **_social_search_retrigger_profile** and **social_search_user_update** can be removed, and I removed both functions. 
Creating a new user, blocking a user, enabling a user and updating user profile works fine without.

After having a short look at **social_search_resave_search_indexes** which is only called form social_tagging.install, it seems that this can be removes as well. But want to have a second opinion on this.

slowflyer:
I guess, I need a hint, why this Travis error occurs:

--- Failed scenarios:
    /var/www/html/profiles/contrib/social/tests/behat/features/capabilities/private-message/create-private-message.feature:8

No idea, how this is related to search api.

ronaldtebrake:
Hi @slowflyer its usually a false negative, we will have to see if we can stabilize that one as it indeed doesn't seem to be related to your work. I'll restart the test to see if we can get it passed.

Thanks for all your hard work!!

modules/social_features/social_search/config/install/search_api.index.social_all.yml
```diff
@@ -138,8 +136,10 @@ field_settings:
       view_mode:
         'entity:group':
           closed_group: teaser
+          flexible_group: teaser
```
ronaldtebrake, line 139: flexible/secret are optional modules, most likely it's enabled in your environment but we shouldn't have it in here as they might be disabled for the users.

navneet0693 (approved):

PR #2306, opened 2021-04-26
Issue #3210717 - As a LU, I want my group membership count to show only active & correct group memberships

ronaldtebrake:
## Problem
Right now in the Profile statistics block we count a users group memberships according to

```$variables['profile_groups'] = count(\Drupal::service('social_group.helper_service')->getAllGroupsForUser($user->id()));```

This however doesn't take in to account what kind of groups / memberships there are.

For example a Group type could be filtered which makes sure the membership count isn't correctly reflected compared to the Groups overview.
For example a user could get an invite or request to join a Group, there aren't full memberships untill accepted so they shouldn't be part of the count.

## Solution
The current count isn't flexible enough, it counts all occurrences of group plugins but that doesn't allow us to filter out group types, or plugins easily without breaking BC or cache.

So instead we:
1. Created a separate helper function to count a users group membership
2. Add a query tag so people can alter the query
3. Add an alter hook, so people can filter out certain group types

## Issue tracker
https://www.drupal.org/node/3210717

## How to test
- [x] Enable SKY as a theme so the Profile statistics block is shown
- [x] As a sitemanager, invite user Chris Hall to a new group
- [x] Go to the profile -> Groups overview of Chris Hall
- [x] See that invites are removed from the Group count and the group count is the same as the joined groups in the groups overview

## Release notes
We ensured the count of groups in the profile statistics block show the correct count and allows for better altering if certain groups should not be reflected there.

PR #2308, opened 2021-04-30
Issue #3210717 - As a LU, I want my group membership count to show only active & correct group memberships [backport of PR #2306]

navneet0693:
## Original PR
https://github.com/goalgorilla/open_social/pull/2306

## Problem
Right now in the Profile statistics block, we count users group memberships according to

```$variables['profile_groups'] = count(\Drupal::service('social_group.helper_service')->getAllGroupsForUser($user->id()));```

This however doesn't take into account what kind of groups/memberships there are.

For example, a Group type could be filtered which makes sure the membership count isn't correctly reflected compared to the Groups overview.
For example, a user could get an invite or request to join a group, there aren't full memberships until accepted so they shouldn't be part of the count.

## Solution
The current count isn't flexible enough, it counts all occurrences of group plugins but that doesn't allow us to filter out-group types, or plugins easily without breaking BC or cache.

So instead we:
1. Created a separate helper function to count a users group membership
2. Add a query tag so people can alter the query
3. Add an alter hook, so people can filter out certain group types

## Issue tracker
https://www.drupal.org/node/3210717

## How to test
- [x] Enable SKY as a theme so the Profile statistics block is shown
- [x] As a site manager, invite user Chris Hall to a new group
- [x] Go to the profile -> Groups overview of Chris Hall
- [x] See that invites are removed from the Group count and the group count is the same as the joined groups in the group's overview

## Release notes
We ensured the count of groups in the profile statistics block shows the correct count and allows for better altering if certain groups should not be reflected there.

### DIFF
--- a/modules/social_features/social_group/social_group.api.php
+++ b/modules/social_features/social_group/social_group.api.php
@@ -26,6 +26,18 @@ function hook_social_group_types_alter(array &$social_group_types) {
 }
 
 /**
+ * Hide group types used in open social.
+ *
+ * @param array $hidden_types
+ *   List of group type id's which you want to see removed.
+ *
+ * @ingroup social_group_api
+ */
+function hook_social_group_hide_types_alter(array &$hidden_types) {
+  $hidden_types[] = 'challenge';
+}
+
+/**
  * Provide a method to alter the default content visibility for a group type.
  *
  * @param string $visibility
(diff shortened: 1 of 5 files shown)


ronaldtebrake:
Think we should use this as patch and not merge it back like this looking at the changes that might be a bit too impactful potentially. What do you think @navneet0693 ?

Close this and use it as a patch for those still on 9.x?

navneet0693:
@ronaldtebrake Yes, we can use it as a patch.

PR #2315, opened 2021-05-06
Create dependabot.yml

ronaldtebrake:
## Problem
Even if it's a seemingly small task, it's a tedious job for people to create PR's with updates using a fork versus the ease of using something that can do this automated.

## Solution
Use a dependency update manager. 
In this case we went for GitHub's built in dependabot.

With configuration:
- For now focus on updating github actions and composer packages
- Do github actions once per week, nothing fancy
- For Drupal we do it every night after midnight, so we are not clogging our github actions & CI/QA tools
- Assign Nucleus team members by default
- Dismiss npm/other non drupal packages for now
- Have only 5 PR's opened at once, we are running behind on our updates and we can finish them one at a time. Once we are finished 5 should be enough as well.

## Issue tracker
No drupal.org ticket will be created, we can create separate drupal.org tickets for the updates that will be created by Dependabot.

## How to test
We can merge it after getting an OK of team Nucleus, this will only add 5 PR's with updates

## Release notes
Not needed, Internal process update.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,39 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions - Update once every week.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "saturday"
+    labels:
+      - "type: update"
+      - "team: nucleus"  
+
+  # Maintain dependencies for Composer.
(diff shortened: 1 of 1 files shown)


ronaldtebrake:
[note to self] Lets create a nucleus team in github so we dont have to maintain the reviewers in the config file but rather as a team :D

navneet0693 (approved):

PR #2324, opened 2021-05-07
[Internal] - Run PHPUnit as action and move PHP Stan + CS for performance

ronaldtebrake:
## Problem
- We want to run PHPUnit as github action
- Having composer install 3 times for Phpunit / phpstan / phpcs isn't necessary, we can preserve the composer caches for all of these across actions
- We might want to run our code on multiple PHP versions to see if we're compatible with all supported versions

## Solution
- Created one workflow file
- Use https://github.com/php-actions/composer#caching-dependencies-for-faster-builds 
- Migrate phpstan/phpcs actions in this one (will remove those later)

## Issue tracker
[Internal]

## How to test
Watch test run!

## Release notes
Not needed.


ronaldtebrake:
Closing this in favor of #2376

PR #2335, opened 2021-05-14
Issue #3214049 by chmez: As a LU I want to see all fields split by sections

ghost:
## Problem
New fields placed outside of any section on the content creation/edit page.

## Solution
Add new fields to the "Settings" section if others not set.

## Issue tracker
- https://www.drupal.org/project/social/issues/3214049
- https://www.drupal.org/node/2842405
- https://getopensocial.atlassian.net/browse/YANG-5605

## How to test
- [ ] Log in as administrator and navigate to the "/admin/structure/types/manage/event/form-display" page.
- [ ] Move the "Promoted to front page" field from the "Disabled" section to the section above.
- [ ] When you log in as a regular user and navigate to the event creation page then you should see the "Promoted to front page" field in the "Settings" section.

## Release notes
Add new fields placed outside of any section on the content creation/edit page to the "Settings" section.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,8 @@
                 "Can't specify the language in TermStorage::loadTree": "https://www.drupal.org/files/issues/2020-09-08/drupal-termstorage-loadTree-lang-3123561-6-d8.patch",
                 "Issue #3188258: Aggregation queries fail across entity references": "https://www.drupal.org/files/issues/2020-12-18/drupal-3188258-aggregation-across-entity-reference-fail-2.patch",
                 "Pagination does not work correctly for comment fields that are rendered using #lazy_builder": "https://www.drupal.org/files/issues/2020-12-22/pagination-does-not-work-with-lazy-builder-3189538-2.patch",
-                "Providing default route value for entity forms is not possible": "https://www.drupal.org/files/issues/2020-12-29/2921093-18.patch"
+                "Providing default route value for entity forms is not possible": "https://www.drupal.org/files/issues/2020-12-29/2921093-18.patch",
+                "Empty language (langcode) field wrapper inserted into contact form": "https://www.drupal.org/files/issues/2020-03-06/empty-langcode-field-wrapper-2842405-17.patch"
             },
             "drupal/flag": {
                 "Add relationship to flagged entities when Flagging is base table": "https://www.drupal.org/files/issues/2723703_31.patch"
--- a/modules/social_features/social_core/social_core.module
+++ b/modules/social_features/social_core/social_core.module
@@ -100,7 +100,10 @@ function social_core_entity_form_display_alter(EntityFormDisplayInterface $form_
   $fields = array_diff(array_keys($form_display->getComponents()), $fields);
 
   if ($fields) {
-    $field_groups['group_settings']['children'] += $fields;
+    $field_groups['group_settings']['children'] = array_merge(
+      $field_groups['group_settings']['children'],
+      $fields
+    );
 
     $form_display
       ->setThirdPartySetting('field_group', 'group_settings', $field_groups['group_settings'])
--- a/modules/social_features/social_core/social_core.module
+++ b/modules/social_features/social_core/social_core.module
@@ -7,6 +7,7 @@
 
 use Drupal\block\Entity\Block;
 use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Entity\Display\EntityFormDisplayInterface;
 use Drupal\Core\Field\FieldItemList;
 use Drupal\Core\File\FileSystemInterface;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -44,7 +44,8 @@
                 "Can't specify the language in TermStorage::loadTree": "https://www.drupal.org/files/issues/2020-09-08/drupal-termstorage-loadTree-lang-3123561-6-d8.patch",
                 "Issue #3188258: Aggregation queries fail across entity references": "https://www.drupal.org/files/issues/2020-12-18/drupal-3188258-aggregation-across-entity-reference-fail-2.patch",
                 "Pagination does not work correctly for comment fields that are rendered using #lazy_builder": "https://www.drupal.org/files/issues/2020-12-22/pagination-does-not-work-with-lazy-builder-3189538-2.patch",
-                "Providing default route value for entity forms is not possible": "https://www.drupal.org/files/issues/2020-12-29/2921093-18.patch"
+                "Providing default route value for entity forms is not possible": "https://www.drupal.org/files/issues/2020-12-29/2921093-18.patch",
+                "Empty language (langcode) field wrapper inserted into contact form": "https://www.drupal.org/files/issues/2020-03-06/empty-langcode-field-wrapper-2842405-17.patch"
```
ribel, line 48: @OH17L Could you explain how this patch is related? And if it is related - could you link both issues on drupal.org?

modules/social_features/social_core/social_core.module
```diff
@@ -80,6 +81,36 @@ function social_core_modules_installed($modules) {
...
+  if (!isset($field_groups['group_settings'])) {
+    return;
+  }
+
+  $fields = [];
+
+  foreach ($field_groups as $field_group) {
+    $fields = array_merge($fields, $field_group['children']);
+  }
+
```
ribel, line 99: @OH17L Please add some comments with explanations here.

modules/social_features/social_core/social_core.module
```diff
@@ -865,6 +896,22 @@ function social_core_form_node_form_alter(&$form, FormStateInterface $form_state
...
+        foreach (['promote', 'sticky'] as $field) {
+          if (isset($form[$field])) {
+            $form['options'][$field] = $form[$field];
+            unset($form[$field]);
+            $has_option_fields = TRUE;
+          }
+        }
+
+        if ($has_option_fields) {
```
ribel, line 910: @OH17L If some developers want to use Promoted and Sticky options, and enable it by purpose, will it be displayed in form? Please add a comment on what happens here.

ribel (changes requested):
@OH17L Looks good, but need more comments with the explanation of changes.

composer.json
OH17L: This patch is needed to completely hide the language field in the "Settings" section when multilingual functionality is disabled. But I don't think that an issue of the patch should be linked to the issue of the current pull request because the patch is partially related to this task.

modules/social_features/social_core/social_core.module
OH17L: Done.

modules/social_features/social_core/social_core.module
OH17L: Done.

ribel (approved):

ribel:
🍒 picked to 10.0.x and 10.1.x

modules/social_features/social_core/social_core.module
```diff
@@ -80,6 +81,38 @@ function social_core_modules_installed($modules) {
...
+
+  if ($fields) {
+    $field_groups['group_settings']['children'] = array_merge(
+      $field_groups['group_settings']['children'],
+      $fields
+    );
+
+    $form_display
+      ->setThirdPartySetting('field_group', 'group_settings', $field_groups['group_settings'])
+      ->save();
```
navneet0693, line 112: We shouldn't do a save in an alter function.

modules/social_features/social_core/social_core.module
```diff
@@ -80,6 +81,38 @@ function social_core_modules_installed($modules) {
   }
 }
 
+/**
+ * Implements hook_entity_form_display_alter().
+ */
+function social_core_entity_form_display_alter(EntityFormDisplayInterface $form_display, array $context) {
```
navneet0693, line 87: @chmez Is it a good idea to simply add "Settings" field group during the install of "social_core" module?

modules/social_features/social_core/social_core.module
OH17L: You're right. We'll need to improve it.

modules/social_features/social_core/social_core.module
OH17L: Indeed. We'll need to improve it as well. Also, please, mention me through my work account.

PR #2270, opened 2021-03-22
Issue #3204850 by rolki: Handling the visibility of a the about / members tab

rollki:
## Problem
Group visibility options don't work for groups

## Solution
Add some extra checks to the FlexibleGroupContentAccessCheck

## Issue tracker
- https://www.drupal.org/project/social/issues/3204850
- https://getopensocial.atlassian.net/browse/YANG-5116

## How to test
- [ ] Go to the flexible group creation page
- [ ] Select the "Group members only (secret)" option
- [ ] Save the group
- [ ] Go to this group under anonymous
- [ ] You should see "Access denied" instead of a group

## Screenshots
Screenshot (1)
Снимок экрана 2021-03-22 в 10 46 15

PR #2336, opened 2021-05-17
Issue #3204850: Make sure SM / Admin can still visit flexible groups marked as secret

ronaldtebrake:
## Problem
In #2270 a regression was introduced where SM / Admin can't visit flexible groups which have the "secret" selected.

## Solution
Make sure we check users with manage all groups permissions before we do the field value checks, this is more performant for them and also makes sure we can escape this access check when users have these powerfull global permissions.

## Issue tracker
https://www.drupal.org/project/social/issues/3204850

## How to test
- [x] Login as LU - Create a Flexible group with Secret group visibility
- [x] Login as SM - See you get an access denied
- [x] Go to this branch and repeat, see the access denied is fixed for SM / Admin but is there for non members.

## Release notes
As a SM you get an access denied whilst visiting a flexible group with the "Secret" configuration. That is now fixed as a SM should be able to see this always.

### DIFF
--- a/tests/behat/features/capabilities/group/group-create-flexible.feature
+++ b/tests/behat/features/capabilities/group/group-create-flexible.feature
@@ -233,8 +233,8 @@ Feature: Create flexible Group
     Then I should see the link "Stream"
     And I should see the link "About"
     And I should see the link "Members"
-    And I should not see the link "Events"
-    And I should not see the link "Topics"
+    And I should see the link "Events"
+    And I should see the link "Topics"
 
     # Test flexible group with community visibility and members only / invite for closed group.
     Given I am logged in as "GivenUserOne"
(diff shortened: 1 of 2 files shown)


navneet0693 (approved):

Ressinel (approved):
Tested, works properly to me.

tests/behat/features/capabilities/group/group-create-flexible.feature
```diff
@@ -219,11 +220,22 @@ Feature: Create flexible Group
...
 
+    # Check this as an user with SM permissions.
+    Given I am logged in as "SiteManagerOne"
+    When I am on "all-groups"
+    Then I should see "Flexible group - Secret option"
+    When I click "Flexible group - Secret option"
+    Then I should see the link "Stream"
+    And I should see the link "About"
+    And I should see the link "Members"
+    And I should not see the link "Events"
```
navneet0693, line 236: @ronaldtebrake Are you sure that a SiteManger - should not see Events and Topics?

navneet0693:
🍒 - to 10.1.x, 10.0.x ;-)

PR #2341, opened 2021-05-18
Issue #3214521 by vnech: Change "Mentions Filter" filter type

nechai:
## Problem
**"Mentions filter"** doesn't allow to use tokens in `href` attribute in **Full HTML** text format.

For example, if I add this string:
```html
<p><a href="[node:author:url:absolute]">This is a link</a></p>
```

.....
the filter will returns:
```html
<p><a href="absolute]">This is a link</a></p>
```

## Solution
Change filter type from `Drupal\filter\Plugin\FilterInterface::TYPE_HTML_RESTRICTOR` to `Drupal\filter\Plugin\FilterInterface::TYPE_MARKUP_LANGUAGE`

## Issue tracker
- https://www.drupal.org/project/social/issues/3214521
- https://getopensocial.atlassian.net/browse/YANG-5657

## How to test
- [ ] Enable "Mentions filter" in "Full HTML" text format
- [ ] Create any entity has text field with CKEditor
- [ ] In the entity in the text field choose "full html" format and type button "source"
- [ ] Add the string - `<a href="[node:author:url:absolute]">This is a link</a>`
- [ ] Save and click "edit" the entity

## Release notes

### DIFF
--- a/modules/custom/mentions/src/Plugin/Filter/MentionsFilter.php
+++ b/modules/custom/mentions/src/Plugin/Filter/MentionsFilter.php
@@ -24,7 +24,7 @@ use Drupal\filter\Entity\FilterFormat;
  * id = "filter_mentions",
  * title = @Translation("Mentions Filter"),
  * description = @Translation("Configure via the <a href='/admin/structure/mentions'>Mention types</a> page."),
- * type = Drupal\filter\Plugin\FilterInterface::TYPE_HTML_RESTRICTOR,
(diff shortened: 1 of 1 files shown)


ribel:
In Distro by default we don's use mentions filter on Full HTML text format, so this is not an issue their atm.
We had a special case or one o our customer where we need to enable mentions for content Description due to t historically migrated content.

So @ronaldtebrake do you think this change is still interesting for Distro, or can I close it?
In general, I think we need to be careful with filter types, since it may have a security impact.
See more examples here: https://www.drupal.org/node/1817474

PR #2347, opened 2021-05-20
Issue #3174548 by rolki: Add basic node scope coverage

rollki:
## Problem
Any user can submit a request and access private data using `GraphQL`, unless a `query_access` entity handler is added.
So it means I can send something like this via GraphQL and received data about users:
```
query {
  users(first: 5) {
    nodes {
      displayName
      mail
    }
  }
}
```

## Solution
Add a scope implementation, that is, an unauthenticated user should not receive any data after sending the request. Only if the user or application has the appropriate permissions (scopes), and he is authenticated and authorized, only then give access to the data to which he has access.

Probably we can use entity API to implement this (see https://www.drupal.org/project/entity/issues/2909970).

## Issue tracker
- https://www.drupal.org/project/social/issues/3174548
- https://getopensocial.atlassian.net/browse/YANG-5435

## How to test
*For example*
- [ ] As a sitemanager

## Screenshots

## Release notes

## Change Record

## Translations

### DIFF
--- a/modules/social_features/social_node/social_node.module
+++ b/modules/social_features/social_node/social_node.module
@@ -5,9 +5,11 @@
  * The social node module alterations.
  */
 
+use Drupal\Core\Database\Query\SelectInterface;
+use Drupal\social_node\Entity\Access\NodeQueryAccessHandler;
 use Drupal\social_node\Entity\Node;
-use Drupal\social_node\SocialNodeForm;
 use Drupal\social_node\NodeViewBuilder;
+use Drupal\social_node\SocialNodeForm;
 
 /**
  * Implements hook_entity_type_alter().
@@ -20,6 +22,78 @@ function social_node_entity_type_alter(array &$entity_types) {
 
     foreach (['default', 'edit'] as $operation) {
       $entity_types['node']->setFormClass($operation, SocialNodeForm::class);
+    }
+    $entity_types['node']->setHandlerClass('query_access', NodeQueryAccessHandler::class);
+  }
+}
+
+/**
+ * Implements hook_query_TAG_alter().
+ */
+function social_node_query_entity_query_alter(SelectInterface $query) {
+  if ($query->getTables()['base_table']['table'] == 'node') {
+    $account = \Drupal::currentUser();
+
+    if ($account->isValidOauth2Account()) {
+      /** @var \Drupal\oauth2_server\OAuth2Storage $storage */
+      $storage = \Drupal::service('oauth2_server.storage');
+      $client_scope = $storage->getClientScope($account->getAccountAccessToken()['client_id']);
+      $request_scope = $account->getAccountAccessToken()['scope'];
+
+      $request_scopes = explode(' ', $request_scope);
+      $client_scopes = explode(' ', $client_scope ?? '');
+
+      foreach ($request_scopes as $scope) {
+        if (in_array($scope, $client_scopes)) {
+          $scope_with_access[] = $scope;
+        }
+      }
+
+      $node_types = \Drupal::entityTypeManager()
+        ->getStorage('node_type')
+        ->getQuery()
+        ->execute();
+
+      $public_bundles = [];
+      $community_bundles = [];
+      foreach ($scope_with_access ?? [] as $element) {
+        $scopes = explode(':', $element);
+        [$bundle, $visibility] = $scopes;
+
+        if ($visibility === 'public' && in_array($bundle, $node_types) && !in_array($bundle, $community_bundles)) {
+          $public_bundles[] = $bundle;
+        }
+        elseif ($visibility === 'community' && in_array($bundle, $node_types) && !in_array($bundle, $public_bundles)) {
+          $community_bundles[] = $bundle;
+        }
+      }
+
+      if (!empty($public_bundles) || !empty($community_bundles)) {
+        $query->innerJoin('node__field_content_visibility', 'nfv');
+
+        $and1 = $query->andConditionGroup();
+        $and2 = $query->andConditionGroup();
+
+        $or = $query->orConditionGroup();
+        if (!empty($public_bundles)) {
+          $and1->condition('nfv.field_content_visibility_value', 'public');
+          $and1->condition('node_field_data.type', $public_bundles, 'IN');
+
+          $or->condition($and1);
+        }
+        if (!empty($community_bundles)) {
+          $and2->condition('nfv.field_content_visibility_value', ['public', 'community'], 'IN');
+          $and2->condition('node_field_data.type', $community_bundles, 'IN');
+
+          $or->condition($and2);
+        }
+
+        $query->condition($or);
+        $query->where('node_field_data.nid=nfv.entity_id');
+      }
+      else {
+        $query->where('1 = 0');
+      }
     }
   }
 }
--- a/modules/social_features/social_node/src/Entity/Access/NodeQueryAccessHandler.php
+++ b/modules/social_features/social_node/src/Entity/Access/NodeQueryAccessHandler.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Drupal\social_node\Entity\Access;
+
+use Drupal\Core\Session\AccountInterface;
+use Drupal\entity\QueryAccess\QueryAccessHandlerBase;
+
+/**
+ * Controls query access for node entities.
+ *
+ * @see \Drupal\entity\QueryAccess\QueryAccessHandler
+ */
+class NodeQueryAccessHandler extends QueryAccessHandlerBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConditions($operation, AccountInterface $account) {
+    $conditions = parent::buildConditions($operation, $account);
+
+    // The user doesn't have access to any entities by default, but in this case
+    // anonymous can view some content, so then following logic added below.
+    // See social_node_query_entity_query_alter.
+    if ($account->isValidOauth2Account() && $conditions->isAlwaysFalse()) {
+      return $conditions->alwaysFalse(FALSE);
+    }
+
+    return $conditions;
+  }
+
+}
--- a/modules/social_features/social_user/social_user.services.yml
+++ b/modules/social_features/social_user/social_user.services.yml
@@ -17,3 +17,9 @@ services:
     arguments: ['@current_route_match', '@current_user', '@config.factory']
     tags:
       - { name: event_subscriber }
+  social_user.current_user:
+    class: Drupal\social_user\SocialAccountProxy
+    decorates: current_user
+    decoration_priority: 5
+    public: false
+    arguments: ['@event_dispatcher', '@oauth2_server.oauth_helper', '@request_stack', '@oauth2_server.storage']
--- a/modules/social_features/social_user/src/SocialAccountProxy.php
+++ b/modules/social_features/social_user/src/SocialAccountProxy.php
@@ -1,0 +1,118 @@
+<?php
+
+namespace Drupal\social_user;
+
+use Drupal\Core\Session\AccountProxy;
+use Drupal\oauth2_server\OAuth2HelperInterface;
+use Drupal\oauth2_server\OAuth2StorageInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Class for adding new methods to the user account to check if this user comes
+ * from Oauth2.
+ *
+ * Decorates the current_user service.
+ */
+class SocialAccountProxy extends AccountProxy {
+
+  /**
+   * The token data.
+   *
+   * @var array
+   */
+  private array $token;
+
+  /**
+   * The OAuth2Helper service.
+   *
+   * @var \Drupal\oauth2_server\OAuth2HelperInterface
+   */
+  protected $oauth2Helper;
+
+  /**
+   * The related request stack.
+   *
+   * @var \Symfony\Component\HttpFoundation\Request
+   */
+  protected $request;
+
+  /**
+   * The OAuth2Storage.
+   *
+   * @var \Drupal\oauth2_server\OAuth2StorageInterface
+   */
+  protected $oauth2Storage;
+
+  /**
+   * SocialAccountProxy constructor.
+   *
+   * @param \Symfony\Component\EventDispatcher\EventDispatcherInterface|null $eventDispatcher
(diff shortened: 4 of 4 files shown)


rollki:
For `authorization` and `authentication` using `Oauth2`, it is necessary to send a `scope` in the request, this is a kind of distribution of access, that is, we determine what the data user or application will have access to in the Drupal system.

The first problem is the definition of the `scope` pattern (this problem was discussed [here](https://getopensocial.atlassian.net/browse/YANG-5435)), and there are not many options here, it looks like this:
- `topic:public:read`
- `topic:community:read`
- `discussion:public:read`
- `discussion:community:read`

So far we are talking only about `node`, and in this example, programmatically without hardcode, we define the type of entity, after which we define the scope, respectively, this should map to our content visibility system, and the last is the type of action, while we are only considering reading data.

For groups, the pattern can be the same. The only thing I don’t know how to implement using this pattern is `comments`, since we do not have a visibility field for comments, but there is a `comment_permissions` module that allows showing comments for certain types of content for anonymous users, it’s hard to cover this via code.

Another problem is test coverage, since in this implementation we check `isValidOauth2Account()`, and if so, add the query. The problem is that in this case we are working with request, and in the request we have the token and data about `Oauth2` authentication, and when we run the test, we have no means to emulate `Oauth2` authentication, the only available option that I see is to test authentication with a `password` or `client credentials` grand types.

This implementation proposal is implemented using the entity API (see https://www.drupal.org/project/entity/issues/2909970).
Since it is impossible to break it down into `query_access`, since there is no access to the query, the `hook_query_TAG_alter` hook was used, and in it, in accordance with the scope to which the user or application has access, the data is accessed.

**So, this is the idea about scope coverage and basic implementation for this scope coverage and access to the node data via this scope. If anyone has another vision about that, please don't hesitate to let us know about that.** 

_P.S. Technical implementation can be change and should be changed because it is probably should fork fine with other entity types (groups, comments and etc.)._

modules/social_features/social_node/social_node.module
```diff
@@ -21,5 +23,77 @@ function social_node_entity_type_alter(array &$entity_types) {
...
       $entity_types['node']->setFormClass($operation, SocialNodeForm::class);
     }
+    $entity_types['node']->setHandlerClass('query_access', NodeQueryAccessHandler::class);
+  }
+}
+
+/**
+ * Implements hook_query_TAG_alter().
+ */
+function social_node_query_entity_query_alter(SelectInterface $query) {
```
bramtenhove, line 33: @rollki Could you add some more comments here?

Kingdutch:
Hi Roman,

Thanks for the very clear write-up of your proposal!

> The first problem is the definition of the scope pattern (this problem was discussed here), and there are not many options here, it looks like this:
>
>    topic:public:read
>    topic:community:read
>    discussion:public:read
>    discussion:community:read

I hadn't thought about the visibility yet. I wonder if that's a distinction we want to make for applications or whether we just want to say "You can give applications access to all content of a certain type". I'm not sure it works for groups with the eye on moving towards flexible groups (since there's no pre-defined visibility anymore, but it differs per group).

How would this look when adding `write`? Would we also do that per visibility?

How do these scopes work if an application is trying to access content within a group? Would they need both the `<content_type>:read` and `group:read` permission?

> Another problem is test coverage, since in this implementation we check isValidOauth2Account(), and if so, add the query. The problem is that in this case we are working with request, and in the request we have the token and data about Oauth2 authentication, and when we run the test, we have no means to emulate Oauth2 authentication, the only available option that I see is to test authentication with a password or client credentials grand types.

In tests we're able to manipulate the current user. So our tests should test with the different possibilities of OAuth users. That would solve our test coverage right?

The thing I had in mind was that we would always provide an OAuth enabled user proxy for every Open Social request. In case a user is using Drupal directly then we would have a "fake" OAuth application that's on-behalf of the currently logged in (or anonymous) user and has all scopes. Whenever a request is made by an OAuth enabled application instead we would use that application. We should be able to implement this securely.

> Since it is impossible to break it down into query_access, since there is no access to the query, the hook_query_TAG_alter hook was used, and in it, in accordance with the scope to which the user or application has access, the data is accessed.

Could you clarify this? Why is there no access to the query? I think part of this story may be to alter/replace/decorate access handlers to support our needs.

----

It looks like your proposal only covers the client credentials grant type (i.e. an application without user) at the moment. How would scopes look for the authorization code grant type (i.e. an application on-behalf of a user). In the latter case the application would only gain access to content the user they're acting for has access within the scopes of the application (so the access is even narrower).

modules/social_features/social_node/social_node.module
rollki: Sorry @bramtenhove, didn't have time yesterday to add inline comments here, but today we will have an internal meeting with the team where we will have a brainstorm and discuss some improvements (add group coverage, resolve comments issue), and I think I will try to add that improvements and at that time comment the code.

rollki:
@Kingdutch 
Thank you for so detailed feedback!

> I hadn't thought about the visibility yet. I wonder if that's a distinction we want to make for applications or whether we just want to say "You can give applications access to all content of a certain type". I'm not sure it works for groups with the eye on moving towards flexible groups (since there's no pre-defined visibility anymore, but it differs per group).

I'm afraid that we can not say "You can give applications access to all content of a certain type", since we have a requirement from the client that they want to get only public information from the platform via API (data that can only be accessed by the anonymous user).

> How would this look when adding write? Would we also do that per visibility?

So, not sure about that, I just saw how it is implemented for other platform APIs and just did similar, and it seems it is more clear for clients, but as I said I'm not sure, probably we can scrap it now because we do not touch this part of the scope in the code.

> How do these scopes work if an application is trying to access content within a group? Would they need both the <content_type>:read and group:read permission?

I thought about that, and I also thought about group content implementation in the schema design. And in my opinion group content should be part of group schema design, we do not add it now, because we start from a minimum (see schema proposal https://www.drupal.org/project/social/issues/3217363), but in the future, we probably can add it and in case if it will be part of group schema, then I think it will be enough only one `group:read`.

BTW in the current implementation for node entities, we don't discriminate whether they are in a group or not, and I don’t think we should distinguish them.

> Would it make sense to nest discussion, topic, etc. under content so we have content:topic:read and content:discussion:read so that providing an application access to all content is easier (i.e. grant content:read)?

I think it is a very good idea to get all content with one scope `content:read`, but I still think that we should keep our visibility, so it will be `content:public:read`, since we have request from the client that they want to get only public data, or if there is any other idea how we can get only public data, then we probably can scrap current proposal for visibility in scope and try to implement another one.

> In tests we're able to manipulate the current user. So our tests should test with the different possibilities of OAuth users. That would solve our test coverage right?

Yeah, it solves our problem, but I tried, and I failed to manipulate the user in the test, that is, imitate the Oauth2 user.

> The thing I had in mind was that we would always provide an OAuth enabled user proxy for every Open Social request. In case a user is using Drupal directly then we would have a "fake" OAuth application that's on-behalf of the currently logged in (or anonymous) user and has all scopes. Whenever a request is made by an OAuth enabled application instead we would use that application. We should be able to implement this securely.

It is indeed what we want, but I'm not sure how we can properly implement this.

> Could you clarify this? Why is there no access to the query? I think part of this story may be to alter/replace/decorate access handlers to support our needs.

Because `QueryAccessHandlerBase` works with conditions, permissions, and account and we need that handler, but we can not join to other tables there and change the query. And in our case, we need to change the query, since as I said we have a request to get only public data, so we should join visibility tables.

> It looks like your proposal only covers the client credentials grant type (i.e. an application without user) at the moment. How would scopes look for the authorization code grant type (i.e. an application on-behalf of a user). In the latter case the application would only gain access to content the user they're acting for has access within the scopes of the application (so the access is even narrower).

Actually no, this proposal covers also an application with the user (`authorization_code` grand type). It is doesn't really matter, so Drupal already know who is authorized if it will be an application without a user, then, the current user will be with ID 0, and if an application with the user (`authorization_code` grand type) then in our `query_access` and `hook_query_TAG_alter` we will get the user account with some real ID.
It is a bit complicated to describe, it's better to show, if it is still not clear or I missed something, then I probably can record some video.


In terms of a technical implementation:

> Rather than testing whether we're dealing with OAuth client using isValidOAuth2Application, why not implement a method such as hasUserScopes(['profile:write']) and hasApplicationScopes(['users:read']) to base our logic on. That more directly asks for the required information (similar to hasPermission checks for users).

> A hasApplicationScopes should probably replace a hasPermission check for anonymous users has (e.g. $account->hasApplicationScopes('comment:read') || $account->hasPermission('view all comments')). While a hasUserScopes should be in addition to hasPermission checks for a logged in user (e.g. $account->hasUserScopes('comment:read') && $comment->access('view', $account) . In case you're using Drupal without an OAuth application then hasApplicationScopes would always return FALSE and hasUserScopes would always return TRUE so that you're back to using only the hasPermission calls.

Mm, that's interesting, so you can be right, I should think about that.

Kingdutch:
> I'm afraid that we can not say "You can give applications access to all content of a certain type", since we have a requirement from the client that they want to get only public information from the platform via API (data that can only be accessed by the anonymous user).

Okay, but that's not a requirement that has to be solved in the authentication layer. We can also create a GraphQL field that provides that data.

The only consideration for how we design scopes should be how granular we want the access to be in Open Social (e.g. do we foresee users always providing all visibility scopes to an application, or would having only one visibility cause issues with displaying group-content because not all group content may be visible? Or do we think there will be third-party applications that may not be used if a user can not limit access only to public data (e.g. because they may not fully trust the application's developer))

> So, not sure about that, I just saw how it is implemented for other platform APIs and just did similar, and it seems it is more clear for clients, but as I said I'm not sure, probably we can scrap it now because we do not touch this part of the scope in the code.

I think we still need to make a design decision on this to make sure we end up with consistent naming and predictable scopes (and also to check that our code can be extended to handle the case in the future).

> I think it is a very good idea to get all content with one scope content:read, but I still think that we should keep our visibility, so it will be content:public:read, since we have request from the client that they want to get only public data, or if there is any other idea how we can get only public data, then we probably can scrap current proposal for visibility in scope and try to implement another one.

Yeah so a `topicWithVisibility(visibility: Visibility!)` field would be a way to do this regardless of scope. We can still think about whether the scopes are useful in this way separately from this.

The way I see `content:*` is basically as a tree:

- content:*
-- topic:*
-- discussion:*
-- etc.

So granting `content:*` is not a separate scope but it's just as if you're giving the application all the scopes for the different content types.

An example of this can be seen in [the personal access token creation on GitHub](https://github.com/settings/tokens/new).
image

> Yeah, it solves our problem, but I tried, and I failed to manipulate the user in the test, that is, imitate the Oauth2 user.
> It is indeed what we want, but I'm not sure how we can properly implement this.

Okay lets plan some time together next week to look at this.

> Because QueryAccessHandlerBase works with conditions, permissions, and account and we need that handler, but we can not join to other tables there and change the query. And in our case, we need to change the query, since as I said we have a request to get only public data, so we should join visibility tables.

We can not create joins directly, but we can traverse tables using entity reference fields. So if `field_content_visibility` is a node field then we should be able to do `field_content_visibility.value` `=` `public` and that should create the join automatically. We do something similar in [an internal project](https://bitbucket.org/goalgorilla/social_chat/src/2ab387952d777d4cb35e6da0c7a5e232c8117d23/src/GraphQL/QueryHelper/ConversationParticipantMessagesQueryHelper.php?at=feature%2Fchat-groups#lines-82) where we start from a receipt entity, then move through the message to match on the conversation that them message was sent in.

> Actually no, this proposal covers also an application with the user (authorization_code grand type). It is doesn't really matter, so Drupal already know who is authorized if it will be an application without a user, then, the current user will be with ID 0, and if an application with the user (authorization_code grand type) then in our query_access and hook_query_TAG_alter we will get the user account with some real ID.
It is a bit complicated to describe, it's better to show, if it is still not clear or I missed something, then I probably can record some video.

Yes I think from a technical perspective that's true. However, I think we need to separate the scopes based on grant types.

E.g. an application that may access all topics for a user should not be able to access all topics without a user (or vice versa). See also my description of different methods for checking :)

rollki:
@Kingdutch 

> Okay, but that's not a requirement that has to be solved in the authentication layer. We can also create a GraphQL field that provides that data.
> 
> The only consideration for how we design scopes should be how granular we want the access to be in Open Social (e.g. do we foresee users always providing all visibility scopes to an application, or would having only one visibility cause issues with displaying group-content because not all group content may be visible? Or do we think there will be third-party applications that may not be used if a user can not limit access only to public data (e.g. because they may not fully trust the application's developer))

I like the solution where we can implement scopes without visibility options, also I guess it will be easier to implement getting only `public/community` data on query level instead of scopes. So it can be like `topicWithVisibility(visibility: Visibility!)`, the only thing that I thinking is group visibility, not sure if this is good idea to allow something like 
```
  topicWithVisibility(visibility: "group") {
    nodes {
      id
    }
  }
```
Why I think this is a bad idea? Because it is possible to be authorised as user or application, and user can has access to some group content but application probably not, but application can has access to all public and community content.

So we should think maybe allow to get only public and community content for now.

> The way I see content:* is basically as a tree:
> 
> content:*
> -- topic:*
> -- discussion:*
> -- etc.
> So granting content:* is not a separate scope but it's just as if you're giving the application all the scopes for the different content types.
> 
> An example of this can be seen in the personal access token creation on GitHub.

That is very good example, I think we can try to implement similar, and maybe think to implement also form with list of checkboxes for client?

So, the it can be something like:

```
- content:*
  -- topic
  -- discussion
  -- etc.
- comment
- group:*
  -- flexible
  -- open
  -- public

Or it can be another way for groups, like:

- group:*
  -- topic
  -- event
  -- membership
```

And also maybe we can think to use combined field to get e.g. topics by type, tags or visibility, like:
```
  topics(first: 10, visibility: "community", topicType: "Blog", tags: {category: "CoP", tags: ["test 1", "test 2"], }) {
    nodes {
      id
    }
  }
```
What do you think about this format? Only not sure about tags, since `social_tagging` is optional module and it will hard to describe what is `category`.
  

> We can not create joins directly, but we can traverse tables using entity reference fields. So if field_content_visibility is a node field then we should be able to do field_content_visibility.value = public and that should create the join automatically. We do something similar in an internal project where we start from a receipt entity, then move through the message to match on the conversation that them message was sent in.

Yeah, if we scrap visibility part from scope, and implement this visibility as argument then we can use this way.

> Yes I think from a technical perspective that's true. However, I think we need to separate the scopes based on grant types.

Yeah but then it can be too complicated, we also can think maybe for `client_credentials` grand type allow getting public and community content, but for ``

PR #2354, opened 2021-05-24
Issue #3215464 by vnech: Add ability for CM+ to remove "Group Admin" role for members

nechai:
## Problem
There is no possibility to remove the "Group Admin" group role for users who are group members and have CM+ role at the same time.

## Solution
Delete functions (hooks) handled force assign "Group Admin" role for members with CM+ role.

NOTE: This PR saves the ability to make CM+ users as group admins if the one creates a group.

## Issue tracker
- https://www.drupal.org/project/social/issues/3215464
- https://getopensocial.atlassian.net/browse/YANG-5715
- https://getopensocial.atlassian.net/browse/YANG-4990

## How to test

- [ ] Login as an admin
- [ ] Create a new group (flexible group, for example)
- [ ] Add a member with global role CM+ to the group
- [ ] Go to "manage members" page for the created group
- [ ] Try to remove the "Group admin" role for the added member

## Release notes
Add an ability to remove the "Group Admin" role for members with CM+ global roles.

### DIFF
--- a/modules/social_features/social_group/social_group.install
+++ b/modules/social_features/social_group/social_group.install
@@ -613,3 +613,30 @@ function social_group_update_8905() {
   // Output logged messages to related channel of update execution.
   return $updateHelper->logger()->output();
 }
+
+/**
+ * Make group creators as a group managers by default.
+ */
+function social_group_update_8906() {
+  $config_factory = \Drupal::configFactory();
+
+  $group_types = [
+    'open_group',
+    'closed_group',
+    'flexible_group',
+  ];
+
+  foreach ($group_types as $group_type) {
+    $config = $config_factory->getEditable('group.type.' . $group_type);
+    $config->set('creator_membership', TRUE);
+
+    $creator_roles = $config->get('creator_roles');
+    $group_role_id = $group_type . '-group_manager';
+    if (!in_array($group_role_id, $creator_roles)) {
+      $creator_roles[] = $group_role_id;
+      $config->set('creator_roles', $creator_roles);
+    }
+
+    $config->save(TRUE);
+  }
+}
--- a/modules/social_features/social_group/social_group.install
+++ b/modules/social_features/social_group/social_group.install
@@ -621,8 +621,10 @@ function social_group_update_8906() {
   $config_factory = \Drupal::configFactory();
 
   $group_types = [
+    'public_group',
     'open_group',
     'closed_group',
+    'secret_group',
     'flexible_group',
   ];
 
--- a/modules/social_features/social_group/social_group.install
+++ b/modules/social_features/social_group/social_group.install
@@ -624,8 +624,6 @@ function social_group_update_8906() {
     'public_group',
     'open_group',
     'closed_group',
-    'secret_group',
-    'flexible_group',
   ];
 
   foreach ($group_types as $group_type) {
--- a/modules/social_features/social_group/config/install/group.type.closed_group.yml
+++ b/modules/social_features/social_group/config/install/group.type.closed_group.yml
@@ -1,8 +1,6 @@
 langcode: en
 status: true
 dependencies: {  }
-_core:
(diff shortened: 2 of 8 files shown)


nechai:
@ronaldtebrake Hi! 
In the PR I deleted the function as a redundant - `social_group_group_insert(GroupInterface $group)` because [the issue](https://www.drupal.org/node/2702743) relates to it is already closed.
Can you please confirm this removal?

ronaldtebrake:
Hi @nechai 

I don't think it can be removed just yet.
We need to ensure we update the Group configuration accordingly, the issue ensures there is now a checkbox on the group types for this. If we remove this code, which is fine for me, we need to ensure the configuration for the creator getting the membership is done through config :)

nechai:
@ronaldtebrake Added related hook update.

ribel:
@ronaldtebrake It looks like @nechai added changes you asked for in the hook update, but I still not sure about few things, so maybe you can help to clarify.

**Currently**
- CM+ automatically receives role `group_manager` ✅ 
- CM+ automatically receives role  `group_admin` since they have permission `manage all groups` ✅ 
- CM+ is able to remove/add the `group_manager` role to themselves or other group members ✅ 
- CM+ is not able to remove the `group_admin` role, which was confusing and reported by customers as a bug. ⛔ 

After changes in this PR:
- CM+ automatically receives role `group_manager` ✅ 
- CM+ does not automatically receive role  `group_admin` ⛔ 
- CM+ is able to remove/add the `group_manager` role to themselves or other group members ✅ 
- CM+ is able to remove/add the `group_manager` role to themselves or other group members ✅ 

To keep the same behaviour I would recommend keeping adding role `group_admin` automatically for CM+, but not sure.

Also not clear why did we disable the ability to add/unset the `group_admin` role initially? Was some security reason for it?

ribel (approved):

ronaldtebrake:
Hi @ribel, good question! :) 

https://communitytalks.getopensocial.com/manual-features-and-site-settings/groups

Here you see that it's indeed expected a CM to be able to do so, so it makes sense to have these fixes in place.

> Site- and content managers always have the permissions to edit groups that are not created by themselves. This is independent of the fact that they joined the group or not.

> If they join a group, they are granted the role of group admin automatically. They have the rights to edit the group, but they are not perse the group manager. If they are also the group manager, this role should be granted to them. That will make them appear in the ' group managers' list, so members know who to contact for questions about the group.

modules/social_features/social_group/social_group.install
```diff
@@ -613,3 +613,30 @@ function social_group_update_8905() {
...
   return $updateHelper->logger()->output();
 }
+
+/**
+ * Make group creators as a group managers by default.
+ */
+function social_group_update_8906() {
+  $config_factory = \Drupal::configFactory();
+
+  $group_types = [
```
ronaldtebrake, line 623: don't forget we have more group types ^^

nechai: @ronaldtebrake I added these groups because these ones were used in the function I removed. Should I add "public" and "secret" group types as well?

ronaldtebrake: As long as you make sure it works for all group types i'm happy, can imagine it might not be needed but want to make sure you've tested this :D (i just did a quick glance at the code and it didnt become clear why haha)

nechai: @ronaldtebrake Seems like you were right - these two group types ("open_group" && "secret_group") also should be updated as I didn't find any related updates for them. Fixed in the last commit.

modules/social_features/social_group/social_group.install
```diff
@@ -613,3 +613,32 @@ function social_group_update_8905() {
...
+    'public_group',
+    'open_group',
+    'closed_group',
+    'secret_group',
```
ribel, line 627: @nechai I think it's better to clone this update hook and place relevant updates in 
`social_group_secret` and `social_group_flexible_group` modules.

PR #2355, opened 2021-05-27
Issue #3215978 by nkoporec: Administer social_footer settings permission is missing

nkoporec:
## Problem
The social_footer has a route requirement for a permission called 'administer social_footer_block settings' , but this permission does not exists in the module (there is no .permission.yml file for it). So only users that have the permission added programetically can access this route.

## Solution
We need to create a permission.yml file where this permission will be defined.

## Issue tracker
https://www.drupal.org/project/social/issues/3215978

## How to test
1. Go to /admin/people/permissions
2. Try to find the 'administer social_footer_block settings'

### DIFF
--- a/modules/social_features/social_footer/social_footer.permissions.yml
+++ b/modules/social_features/social_footer/social_footer.permissions.yml
@@ -1,3 +1,3 @@
-administer social_footer settings:
+administer social_footer_block settings:
   title: 'Administer Social Footer settings'
   description: 'Configure Social Footer'
--- a/modules/social_features/social_footer/social_footer.permissions.yml
+++ b/modules/social_features/social_footer/social_footer.permissions.yml
@@ -1,0 +1,3 @@
+administer social_footer settings:
+  title: 'Administer Social Footer settings'
+  description: 'Configure Social Footer'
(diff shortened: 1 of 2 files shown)


nachosalvador (approved):

modules/social_features/social_footer/social_footer.permissions.yml
```diff
@@ -0,0 +1,3 @@
+administer social_footer settings:
```
bramtenhove, line 1: If we're adding this can we also change the routing.yml file?

https://github.com/goalgorilla/open_social/blob/8ca7907fceb1c542f36b91a8d64bf1c6e2b8fb83/modules/social_features/social_footer/social_footer.routing.yml#L7

The permission is called `administer social_footer_block settings` there.

bramtenhove (changes requested):

modules/social_features/social_footer/social_footer.permissions.yml
navneet0693: @bramtenhove I have corrected it :-)

robertragas (approved):

robertragas:
Merging this now the permission name is correct.

PR #2313, opened 2021-05-05
Issue #3212327 by nkoporec: Implement AND condition when filtering with multiple terms

nkoporec:
## Problem
Currently the terms filtering works as an OR condition, this means that if you have a topic that is tagged to Term 1 and Term 2 and multiple topics that are tagged to Term 1 filtering by Term 1 and Term 2 produces multiple results, which is confusing.

## Solution
 Implement a settings that will enabled to use AND condition in the query for terms filtering. This way a user could decide what kind of tags filtering he wants on his page.

## Issue tracker
https://www.drupal.org/project/social/issues/3212327

## How to test
1. Install social_tagging
2. Create multiple root tags with children
3. Tag multiple topics with these new tags
4. Try to search for a specific topic that is tagged to multiple terms.

## Screenshots

_Learn more about the EAA_ is tagged to Term 1 and Term A

_Welcome to the EAA community_ is tagged to Term 1

Before:
Screenshot from 2021-05-05 13-04-12

When the setting is enabled:
Screenshot from 2021-05-05 13-04-40

PR #2360, opened 2021-05-28
Issue #3216240 by SV: Improve visibility of tags in content creation

volodymyr-sydor-lemberg:
## Problem
Tags are currently grouped together with other settings on the add/edit node pages. Also, these settings are collapsed by default, and as result users can miss them when creating or edit content.

## Solution
Move “Tags” from “Additional information” to a separate section on create/edit screen

## Issue tracker
- https://www.drupal.org/project/social/issues/3216240
- https://getopensocial.atlassian.net/browse/YANG-5649

## How to test
- [ ] Using the latest version of Open Social with the social_tagging module enabled
- [ ] As a sitemanager
- [ ] On creating or edit content page try to find the “Tags” field
- [ ] When creating/editing I expect the “Tags” field displayed in the separate section.

## Screenshots
**Before** 
is collapsed:
Screenshot from 2021-05-28 15-09-17

is mixed with other fields:
Screenshot from 2021-05-28 15-44-42

**After:** 
displays in separate fieldset
Screenshot-from-2021-05-28-15-11-13-png-777×786-

## Release notes
Now, "Tags" is much more visible and displays in the new separate fieldset

### DIFF
--- a/modules/social_features/social_tagging/social_tagging.module
+++ b/modules/social_features/social_tagging/social_tagging.module
@@ -5,6 +5,7 @@
  * Contains social_tagging.module.
  */
 
+use Drupal\Core\Entity\Display\EntityFormDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Field\BaseFieldDefinition;
@@ -180,14 +181,14 @@ function social_tagging_social_tagging_field_form_alter(array &$form, FormStateI
       ];
 
       // We want to move the tagging field in new fieldset
-      // "Additional information". Only when the theme settings are updated.
+      // "Tags". Only when the theme settings are updated.
       $use_social_content_forms = theme_get_setting('content_entity_form_style');
-      if (isset($form['#fieldgroups']['group_attachments']) &&
+      if (isset($form['#fieldgroups']['group_social_tags']) &&
         $use_social_content_forms === 'open_social') {
         $form['tagging']['#type'] = 'details';
-        $form['tagging']['#title'] = t('Tags');
-        $form['#fieldgroups']['group_attachments']->children[] = 'tagging';
-        $form['#group_children']['tagging'] = 'group_attachments';
+        $form['tagging']['#title'] = '';
+        $form['#fieldgroups']['group_social_tags']->children[] = 'tagging';
+        $form['#group_children']['tagging'] = 'group_social_tags';
       }
 
       // Add Tags in flexible groups if enabled.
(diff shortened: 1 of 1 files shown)


ribel (approved):

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -558,3 +559,33 @@ function social_tagging_preprocess_profile(array &$variables) {
     }
   }
 }
+
+/**
+ * Implements hook_entity_form_display_alter().
+ */
+function social_tagging_entity_form_display_alter(EntityFormDisplayInterface $form_display, array $context) {
```
navneet0693, line 566: @volodymyr-sydor-lemberg @ribel I am trying to understand why this function was removed in https://github.com/goalgorilla/open_social/pull/2450/files#diff-19de5cecbacf0e4c7d03f7e5f50073625b76b77849e19185b222a591a96e03dfL566 and then re-added in this?

modules/social_features/social_tagging/social_tagging.module
```diff
@@ -558,3 +559,33 @@ function social_tagging_preprocess_profile(array &$variables) {
...
+      "required_fields" => TRUE,
+      "id" => "tags",
+      "classes" => "card",
+      "label" => "Tags",
+    ],
+    "region" => "hidden",
+  ];
+
+  $form_display->setThirdPartySetting('field_group', 'group_social_tags', $settings);
+  $form_display->save();
```
navneet0693, line 590: We should not do a "save" in an alter function.

modules/social_features/social_tagging/social_tagging.module
volodymyr-sydor-lemberg: @navneet0693 
If we check the dates of merging those PR's/commits we can see that in current PR this function was added initially and in the second PR it was refactored/removed. I'm not sure why this function still exists, probably something went wrong during merging but now it should be removed.

modules/social_features/social_tagging/social_tagging.module
volodymyr-sydor-lemberg: @navneet0693 
Yes, I totally agree and it was fixed in that PR which you mentioned above

modules/social_features/social_tagging/social_tagging.module
navneet0693: @volodymyr-sydor-lemberg I am also trying to find out the reason, from the git blame I get this :-/ https://github.com/goalgorilla/open_social/commit/620754e51b850437b327704d52e66796f820f6cb

navneet0693: I think I found it, it got introduced in https://github.com/goalgorilla/open_social/pull/2313 probably due to branch merges.

PR #2363, opened 2021-05-31
Issue #3215464 by vnech: Missed translating support for "Book page" node type

nechai:
## Problem
In "Open Social" there is missed translation support for "Book page" node type.

## Solution
Add related patch for Drupal core.

## Issue tracker
- https://www.drupal.org/project/social/issues/3216539
- https://getopensocial.atlassian.net/browse/YANG-5635

## How to test
- [ ] Login as an admin
- [ ] Add "Book page" (/node/add/book) at least with one translation
- [ ] Visit translated created book view page

## Release notes
Missed translating support for "Book page" node type

## Change Record
Patches:
- drupal/core
  -  "Add Book page node type translation support": "https://www.drupal.org/files/issues/2020-12-16/2470896_237.patch"

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,8 @@
                 "Pagination does not work correctly for comment fields that are rendered using #lazy_builder": "https://www.drupal.org/files/issues/2020-12-22/pagination-does-not-work-with-lazy-builder-3189538-2.patch",
                 "Providing default route value for entity forms is not possible": "https://www.drupal.org/files/issues/2020-12-29/2921093-18.patch",
                 "Selecting the same day in a date between filter returns no results": "https://www.drupal.org/files/issues/2020-07-06/2842409-15.patch",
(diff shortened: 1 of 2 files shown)


ribel (changes requested):
@nechai See the [comment](https://www.drupal.org/project/drupal/issues/2470896#comment-14038900) from the original issue:

> With patch 233 (237 as well) we have a critical issue for D8:
> 
> - unpublished nodes were shown in the book navigation
> - unpublished nodes were exported when you are using the printer friendly version
> 
> This is a huge issue when you try to create customer documentation were unpublished nodes gets exported.

I have tested it and it looks valid. So I would not recommend to merged this patch until it will be fixed.
On other hand, we can just wait until we update to Drupal 9.2, where this issue is already fixed without the patch.

@ronaldtebrake do you know when we plan to update to Drupal 9.x?

PR #2365, opened 2021-05-31
Issue #3216573 by tBKoT: Fix notification issue for follow tag activity

dmytriikaiun:
## Problem
We have two issues with the following tag activity
1. User1 does not receive notification/email when user2 creates content with the tag that user1 followed. It because notifications sent only for content that have a group to the users that outside of it
2. The activity `Update node with following tag` created even if the node does not have tags

## Solution
1. Change the logic for the activity contexts `follow_taxonomy_activity_context` and `follow_tag_activity_context`
2. Add additional condition in FollowTaxonomyActivityContext::isValidEntity to check if entity definitely have some tags

## Issue tracker
https://www.drupal.org/project/social/issues/3216573
https://getopensocial.atlassian.net/browse/YANG-5609

### 1
- [ ] As user 1 create content with a tag
- [ ] As user 2 go to the content page and follow the tag
- [ ] As user 1 create another content with the same tag
- [ ] As user 2 you should receive notification about creating content with the tag you are following, but he/she didn't

### 2
- [ ] Update any topic/event without a tag
- [ ] User will receive notification that followed tag was added to the content, but shouldn't

## Screenshots
N/A

## Release notes
Users should be notified of content creation with followed tags regardless of the presence of the group in the content

## Change Record
N/A

## Translations
N/A


ribel (approved):

ribel (changes requested):
@dmytriikaiun Changes looks good to me, as we discussed.
But I see we don't have any tests for this functionality, so would be good to add it.

Kingdutch:

Link:
- https://pr2365-wd9weoddywtm1ed5tpha2pi0i4b7d0gc.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60c0b2f4757146cd20a11c3b

ribel (approved):

PR #2373, opened 2021-06-04
[GraphQL] Move addMutationResolver into trait for schema extension

Kingdutch:
Open Social creates its own ResolverRegistry so that it can make changes to the way the schema is built in case this is desirable. However, the actual schema extension plugins always expect a standardised ResolverRegistry as they must adhere to the schema extension interfaces. This means there's no easy way to call public methods from a schema extension on our custom resolver registry because it may lead to calling a non-existent function.

Instead of trying to provide the functionality for our schema extensions in the registry we should just make the function easily available to the extensions directly. This can easily be done by providing a trait that can be reused by any schema extension that wants this functionality.

To standardise schema extensions created by Open Social itself we include the trait on a custom base class for schema extension plugins. Making it a trait allows other developers to use the specific function even if they don't want to use the custom base class.

__Changes in this PR are based on #2207 which should be merged first__

## Issue tracker
https://www.drupal.org/project/social/issues/3217376

## How to test
- [ ] PHPUnit tests should pass

## Screenshots

## Release notes
`StandardisedMutationSchemaTrait` has been added which provides a standardised manner to handle GraphQL mutation input and processing.

## Change Record
N/a

## Translations
N/a

### DIFF
--- a/modules/custom/social_graphql/src/GraphQL/ResolverRegistry.php
+++ b/modules/custom/social_graphql/src/GraphQL/ResolverRegistry.php
@@ -2,7 +2,6 @@
 
 namespace Drupal\social_graphql\GraphQL;
 
-use Drupal\graphql\GraphQL\ResolverBuilder;
 use Drupal\graphql\GraphQL\ResolverRegistry as ResolverRegistryBase;
 
 /**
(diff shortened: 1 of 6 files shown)


rollki (approved):
The changes look good to me, I didn't apply them as a patch, that's why I didn't functionally test them, but visually everything looks good.

PR #2375, opened 2021-06-07
Issue #3217678 by vnech: Remove group admins as recipient for sending message when a new member is added

nechai:
## Problem
Don't send a message to group admins when a new member is added as this produces a lot of similar messages.

## Issue tracker
- https://www.drupal.org/project/social/issues/3217678
- https://getopensocial.atlassian.net/browse/YANG-4990

## How to test
- [ ] Login as SM
- [ ] Enable Welcome Message on some group
- [ ] Join this group as a "Group Admin"
- [ ] Add a new member to this group
- [ ] Member should receive Welcome message only from Group Managers, but not from Group Admin


ribel (approved):
I have tested changes and they works as expected

PR #2383, opened 2021-06-16
Issue #3219123 by Kingdutch: [GraphQL] Allow EntityConnectionQueryHelper to change sorting direction

Kingdutch:
In some cases the default sorting direction should not be ascending but instead descending. A possible scenario is for example sorting content by the number of likes.

Forward is currently always assumed to be "Ascending" and reversed is considered to be "Descending" with no way to change this.

Update <code>ConnectionQueryHelperInterface</code> to add <code>getForwardSortDirection</code> and <code>getReverseSortDirection</code> which can be called instead of the current hardcoded 'ASC' and 'DESC' values.

Introduce <code>ConnectionQueryHelperBase</code> that contains these functions with their current default values to make upgrading easier.

## Issue tracker
https://www.drupal.org/project/social/issues/3219123

## How to test

- [x] GraphQL Tests should pass

## Screenshots
N/a

## Release notes
N/a unsupported internal change

## Change Record
N/a unsupported internal change

## Translations
N/a

### DIFF
--- a/modules/custom/social_graphql/src/GraphQL/EntityConnection.php
+++ b/modules/custom/social_graphql/src/GraphQL/EntityConnection.php
@@ -234,7 +234,8 @@ class EntityConnection implements ConnectionInterface {
     // The order is descending if
     // - we want the first results in a reversed query
     // - we want the last results in a non reversed query.
-    $query_order = (!is_null($this->first) && !$this->reverse) || (!is_null($this->last) && $this->reverse) ? 'ASC' : 'DESC';
+    $field_query_order = (!is_null($this->first) && !$this->reverse) || (!is_null($this->last) && $this->reverse) ? $this->queryHelper->getForwardSortDirection() : $this->queryHelper->getReverseSortDirection();
+    $id_query_order = (!is_null($this->first) && !$this->reverse) || (!is_null($this->last) && $this->reverse) ? 'ASC' : 'DESC';
 
     // If a cursor is provided then we alter the condition to select the
     // elements on the correct side of the cursor.
@@ -274,13 +275,13 @@ class EntityConnection implements ConnectionInterface {
       $query->sortAggregate(
         $sort_field,
         $function,
-        $query_order
+        $field_query_order
       );
     }
     else {
       $query->sort(
         $sort_field,
-        $query_order
+        $field_query_order
       );
 
       // @todo https://www.drupal.org/project/social/issues/3191638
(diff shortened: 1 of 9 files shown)


Kingdutch:

Link:
- https://pr2383-uzwxpyanesahsowphn8c1jzzxyinnsx6.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60ca1209f3ee307ca2f0fea4

modules/custom/social_graphql/src/GraphQL/EntityConnection.php
```diff
@@ -234,7 +234,8 @@ protected function execute() : SyncPromise {
     // The order is descending if
     // - we want the first results in a reversed query
     // - we want the last results in a non reversed query.
-    $query_order = (!is_null($this->first) && !$this->reverse) || (!is_null($this->last) && $this->reverse) ? 'ASC' : 'DESC';
+    $field_query_order = (!is_null($this->first) && !$this->reverse) || (!is_null($this->last) && $this->reverse) ? $this->queryHelper->getForwardSortDirection() : $this->queryHelper->getReverseSortDirection();
+    $id_query_order = (!is_null($this->first) && !$this->reverse) || (!is_null($this->last) && $this->reverse) ? 'ASC' : 'DESC';
```
rollki, line 238: But these two field has the same result, but for `$field_query_order` uses methods from queryHelper and for the `$id_query_order` uses static values, why not only one variable, this is through https://www.drupal.org/project/social/issues/3191638?

Kingdutch: We need a fallback sort in case the value of the sorted field is the same. In such cases I think sorting by ID in ascending order (which means "older to newer") makes more sense.

For example in an overview of content sorted by number of likes. With the change in this PR we can decide that sorting by likes means sorting from most to fewest (so descending order) and the reverse sort would be from fewest to most likes (ascending order). In the case of two pieces of content have the same amount of likes I believe it makes more sense to show the oldest content first, rather than have this be decided on how we sort the field.

In the future I think it may make sense to allow choosing the fallback field (for example when sorting by user role we may want to fallback to sorting by name rather than ID) and order but for now just sorting by ID works fine.

In summary: we could use the same value, but we don't have to, so I think we shouldn't.

rollki: Okay. thank you for the detailed explanation!

rollki (approved):
Changes look good to me.

PR #2309, opened 2021-04-30
Issue #3211565 by Ressinel: Sort users in Group view.

Ressinel:
## Problem
GM can sort users by clicking on Member column (sort by First Name) and keep privacy in mind.

## Solution
Create a separate field for Profile name for sorting by it.

## Issue tracker

- https://getopensocial.atlassian.net/browse/DS-7546
- https://www.drupal.org/project/social/issues/3211565

## How to test
- [x] As GM goes to the group membership page
- [x] You must be able to sort members taking into account all the limitations of the module Social Profile Privacy.

## Screenshots
N/A

## Release notes
Group Manager will now be able to sort the members of their group in "manager members" tab by Members name.

## Change Record
N/A

## Translations
N/A

PR #2392, opened 2021-06-23
Issue #3211565 by Ressinel: Sort users in Group view (improvements).

Ressinel:
## Problem
1. Update all profiles by post_update take a lot of time
2. Sorting for users which have permission to view hidden fields doesn't works properly.

## Solution
1. Add the Profile name field value directly by database insert to reduce the update time of a large number of profiles.
2. Try to improve sorting code.

## Issue tracker
- https://www.drupal.org/project/social/issues/3211565
- https://getopensocial.atlassian.net/browse/DS-7546

## How to test
- [ ] As GM goes to the group membership page
- [ ] You must be able to sort members taking into account all the limitations of the module Social Profile Privacy.

## Screenshots
N/A

## Release notes
Some improvements and fixes:
1. Reduce the update time of a large number of profiles.
2. Sort members in group view for users which have permission to view hidden fields works fixed and works properly now.

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_profile/src/Plugin/views/field/ProfileEntitySortable.php
+++ b/modules/social_features/social_profile/src/Plugin/views/field/ProfileEntitySortable.php
@@ -125,8 +125,24 @@ class ProfileEntitySortable extends RenderedEntity {
           // We will have different expressions depending on is the Nickname
           // field provided or not.
           $field = in_array('field_profile_nick_name', $order_by_fields) ?
-            "CASE WHEN profile__field_profile_nick_name.field_profile_nick_name_value IS NULL THEN CONCAT(TRIM(profile__field_profile_first_name.field_profile_first_name_value), ' ', TRIM(profile__field_profile_last_name.field_profile_last_name_value)) ELSE TRIM(profile__field_profile_nick_name.field_profile_nick_name_value) END" :
-            "CONCAT(TRIM(profile__field_profile_first_name.field_profile_first_name_value), ' ', TRIM(profile__field_profile_last_name.field_profile_last_name_value))";
+            "CASE WHEN
+              profile__field_profile_nick_name.field_profile_nick_name_value IS NOT NULL
+            THEN
+              TRIM(profile__field_profile_nick_name.field_profile_nick_name_value)
+            WHEN
+              (profile__field_profile_nick_name.field_profile_nick_name_value IS NULL) AND ((profile__field_profile_first_name.field_profile_first_name_value IS NOT NULL) OR (profile__field_profile_last_name.field_profile_last_name_value IS NOT NULL))
+            THEN
+              CONCAT(TRIM(profile__field_profile_first_name.field_profile_first_name_value), ' ', TRIM(profile__field_profile_last_name.field_profile_last_name_value))
+            ELSE
+              TRIM(" . $this->view->relationship['profile']->tableAlias . ".name)
+            END" :
+            "CASE WHEN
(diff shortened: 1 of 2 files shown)


open-social-tugboat:

Link:
- https://pr2392-2xhvarzxiayyh062ofbyp2e81a3yn4ro.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60d34476c82560216d45616d

modules/social_features/social_profile/src/Plugin/views/field/ProfileEntitySortable.php
```diff
@@ -125,8 +125,24 @@ public function clickSort($order) {
           // We will have different expressions depending on is the Nickname
           // field provided or not.
           $field = in_array('field_profile_nick_name', $order_by_fields) ?
-            "CASE WHEN profile__field_profile_nick_name.field_profile_nick_name_value IS NULL THEN CONCAT(TRIM(profile__field_profile_first_name.field_profile_first_name_value), ' ', TRIM(profile__field_profile_last_name.field_profile_last_name_value)) ELSE TRIM(profile__field_profile_nick_name.field_profile_nick_name_value) END" :
-            "CONCAT(TRIM(profile__field_profile_first_name.field_profile_first_name_value), ' ', TRIM(profile__field_profile_last_name.field_profile_last_name_value))";
+            "CASE WHEN
```
navneet0693, line 128: Please provide more inline documentation with an explanation of what we are doing here.

Ressinel: Provided.

open-social-tugboat:

Link:
- https://pr2392-2xhvarzxiayyh062ofbyp2e81a3yn4ro.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60d34476c82560216d45616d

navneet0693 (approved):

navneet0693:
🍒 - picked to 10.2.x

Kingdutch:
This was a followup to #2309

czange:
This update creates issues for us - the update just does not run through - it says integrity violation and stops at a random user - if we delete this user, it stops on the next user etc. Could it be a memory problem or something alike

czange:
Okay - some interesting news about this: I tried to go as far as possible near the problematic update - so up to 10.2.0-
This worked - but then I was also able to go up to 10.2.2 - and I was wondering. Looking a bit more into the code, I became aware of the fact, that social_profile_post_update_10101_profile_names_update is different in 10.2.0 compared to 10.2.2 - but will no be rerun, if you upgrade from 10.2.0- to 10.2.2

can you make sense of this? It is solved now for me - but if course I wonder, if I miss something, if the social_profile_post_update_10101_profile_names_update is now not run in the form that is present in 10.2.2

Ressinel:
Hello @czange Thanks for your research!

In `10.2.2` we only added improvements to `social_profile_post_update_10101_profile_names_update` compared with `10.2.0`. So logic of functionality stayed the same. It's why we changed only code in `social_profile_post_update_10101_profile_names_update`.

Could you provide more information about your issue, error code, etc.?

czange:
Well - I tried to update from an older OS installation (9) - and the update would never go through - always stop - integrity error. So going to 10.2.0 and than 10.2.2 was the solution. We did not have a problem with our accounts - as we tried to fix it by deleting accounts, where it stopped - it did not change anything but would than stop at the next account etc etc.

czange:
So finally I just checked where this update came in - and went up to just before (10.2.0) - and oh wonder - the update went through. after than, going to 10.2.2 was a breeze. So it seems if you tried to make this better, you did not succeed for every case. Thanks for your consideration.

PR #2394, opened 2021-06-28
Issue #3221056 by nkoporec: Don't cache the number of event enrollments

nkoporec:
## Problem
When viewing the event enrollments on /group/*/events the number of enrollments is cached. This means that if after your first page load an additional members joins the event then the enrollment number will not be correct until the render cache is cleared.

## Solution
Don't cache the number.

## Issue tracker
https://www.drupal.org/project/social/issues/3221056

## How to test
*For example*
1. Find a group with an event
2. Go to /group/*/events and check the number of enrollments
3. Go to incognito as a different user and enroll into that event
4. As a previous user reload the /group/*/events page and check the enrollment count ( it will still be the same, even though a new user enrolled)

### DIFF
--- a/modules/social_features/social_event/social_event.module
+++ b/modules/social_features/social_event/social_event.module
@@ -285,6 +285,9 @@ function social_event_node_view_alter(array &$build, EntityInterface $entity, En
     $build['enrollments_count'] = [
       '#type' => '#text_field',
       '#markup' => $enrollments_count,
+      '#cache' => [
+        'max-age' => 0,
+      ],
     ];
 
     // Check if ongoing.
--- a/modules/social_features/social_event/social_event.module
+++ b/modules/social_features/social_event/social_event.module
@@ -285,9 +285,6 @@ function social_event_node_view_alter(array &$build, EntityInterface $entity, En
     $build['enrollments_count'] = [
       '#type' => '#text_field',
       '#markup' => $enrollments_count,
-      '#cache' => [
-        'max-age' => 0,
-      ],
     ];
 
     // Check if ongoing.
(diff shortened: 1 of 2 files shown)


nachosalvador (approved):

robertragas:
I wonder if there is another way to fix this without removing the cache. I see also in the first few lines of the view alter that we do set cache tags for $enrollment_tag. Isn't this something that we need to clear when someone enrolls?

navneet0693 (changes requested):
This has the potential to make the whole page uncacheable. It would be best to base it on a tag.

open-social-tugboat:

Link:
- https://pr2394-88z7rtnl7yh7r7xqldm8seb8pyymlfve.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611d10a6331d46de77638032

nkoporec:
@robertragas I changed the approach and now when user enrolls we simply clear that node cache tag. We can't use the $enrollment_tag because it's specific the the uid, so that won't work. We could add additional tag for enrollment, eq. `enrollment:nid` but since we are clearing the node cache tags when the anonymous user enrolls(in social_event_an_enroll module), I think it's best to use this approach, for consistency reasons.

cc: @navneet0693

open-social-tugboat:

Link:
- https://pr2394-88z7rtnl7yh7r7xqldm8seb8pyymlfve.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/611d10a6331d46de77638032

navneet0693:
This is interesting, I am not able to reproduce it 😓 

Steps.

1. As Chris Hall, I visited: /node/12
2. In incognito, as Susan Williams I enrolled on /node/12
3. As Chris, I refreshed, and i was able to see the enrollment.

navneet0693 (approved):

Ressinel (approved):

navneet0693:
🍒  - picked to 10.2.x and 10.1.x

PR #2395, opened 2021-06-28
Issue #3220809 by vnech: Refactor "Social Content Translation" module

nechai:
## Problem
We have the module "Social Language - Content Translation"(`social_content_translation`) module responsible for enabling content translations support in Distro. 

In general, this module has a settings `social_content_translation.settings` with a list of modules for what we enable loading special config overrides classes extended from abstract class `Drupal\social_core\ContentTranslationConfigOverrideBase`
We don't have any page where we can configure these settings. Seems this configuration is redundant as just enough to enable the module `social_content_translation` for loading all translations.

## Solution
Delete redundant config `social_content_translation.settings` and make translation support depending only from enabling/disabling the module `social_content_translation`.

## Issue tracker
- https://www.drupal.org/project/social/issues/3220809
- https://getopensocial.atlassian.net/browse/YANG-5904

## How to test
- [ ] Login as an admin
- [ ] Enable an extention "Social Language - Content Translation"
- [ ] Add a Flexible group
- [ ] Add a translation for created flexible group

## Release notes
*Make translations support depending only from enabling/disabling the module `social_content_translation`*

## Change Record
- Delete `social_content_translation.settings` configuration

### DIFF
--- a/modules/custom/social_language/modules/social_content_translation/config/install/social_content_translation.settings.yml
+++ b/modules/custom/social_language/modules/social_content_translation/config/install/social_content_translation.settings.yml
@@ -1,6 +1,0 @@
-# TODO: Overrides to these settings are currently unusable (See ContentTranslationDefaultsConfigOverride).
-social_book: true
-social_landing_page: true
-social_page: true
-social_event: true
-social_group_flexible_group: true
--- a/modules/custom/social_language/modules/social_content_translation/social_content_translation.install
+++ b/modules/custom/social_language/modules/social_content_translation/social_content_translation.install
@@ -90,4 +90,3 @@ function social_content_translation_update_8002() {
     ->getEditable('social_content_translation.settings')
     ->delete();
 }
-
--- a/modules/custom/social_language/modules/social_content_translation/social_content_translation.install
+++ b/modules/custom/social_language/modules/social_content_translation/social_content_translation.install
@@ -78,14 +78,16 @@ function social_content_translation_disable_field_translations() {
  * Add support translations for entities.
  */
 function social_content_translation_update_8001() {
-  $module_names = [
-    'social_group_flexible_group',
-  ];
-
-  $config = \Drupal::configFactory()->getEditable('social_content_translation.settings');
-
-  foreach ($module_names as $module_name) {
-    $config->set($module_name, TRUE);
-    $config->save(TRUE);
-  }
+  // Don't do anything.
 }
+
+/**
+ * Delete redundant configuration settings.
+ */
+function social_content_translation_update_8002() {
+  // Delete redundant settings.
+  \Drupal::configFactory()
+    ->getEditable('social_content_translation.settings')
+    ->delete();
+}
+
--- a/modules/custom/social_language/modules/social_content_translation/social_content_translation.install
+++ b/modules/custom/social_language/modules/social_content_translation/social_content_translation.install
@@ -84,7 +84,7 @@ function social_content_translation_update_8001() {
 /**
  * Delete redundant configuration settings.
  */
-function social_content_translation_update_8002() {
+function social_content_translation_update_10301() {
   // Delete redundant settings.
   \Drupal::configFactory()
     ->getEditable('social_content_translation.settings')
--- a/modules/social_features/social_core/src/ContentTranslationConfigOverrideBase.php
+++ b/modules/social_features/social_core/src/ContentTranslationConfigOverrideBase.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Drupal\social_content_translation;
+namespace Drupal\social_core;
 
 use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Config\ConfigFactoryOverrideInterface;
@@ -9,7 +9,7 @@ use Drupal\Core\Config\StorageInterface;
 /**
  * Provides a base class for configurable content translation config overrides.
  *
- * @package Drupal\social_content_translation
+ * @package Drupal\social_core
  */
 abstract class ContentTranslationConfigOverrideBase implements ConfigFactoryOverrideInterface {
 
--- a/modules/social_features/social_core/src/ContentTranslationConfigOverrideBase.php
+++ b/modules/social_features/social_core/src/ContentTranslationConfigOverrideBase.php
@@ -28,25 +28,12 @@ abstract class ContentTranslationConfigOverrideBase implements ConfigFactoryOver
   /**
    * Returns the module that provides the overrides.
    *
-   * This is used as the social_contant_translation.settings configuration key
-   * as well as in the cache suffix for the overrides.
+   * This is used as the cache suffix for the overrides.
    *
    * @return string
    *   The module name providing the overrides.
    */
   abstract protected function getModule();
-
-  /**
-   * Returns the display name for this set of configuration overrides.
-   *
-   * This can be used in a user interface to let sitemanagers determine which
-   * parts of Open Social should be translatable. For consistency when
-   * displaying this should always be a plural string.
-   *
-   * @return \Drupal\Core\StringTranslation\TranslatableMarkup|string
-   *   The (translatable) string that can be shown to the user.
-   */
-  abstract protected function getDisplayName();
 
   /**
    * {@inheritdoc}
@@ -56,10 +43,9 @@ abstract class ContentTranslationConfigOverrideBase implements ConfigFactoryOver
 
     // This setting can't be changed in an override because that would create
     // and endless loop in trying to apply the override.
-    $settings = \Drupal::configFactory()->getEditable('social_content_translation.settings');
-    $is_enabled = $settings->getOriginal($this->getModule(), FALSE);
+    $is_content_translations_enabled = \Drupal::moduleHandler()->getModule('social_content_translation');
 
-    if ($is_enabled) {
+    if ($is_content_translations_enabled) {
       $translation_overrides = $this->getTranslationOverrides();
 
       foreach ($translation_overrides as $name => $override) {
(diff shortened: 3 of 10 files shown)


open-social-tugboat:

Link:
- https://pr2395-t4qbv5ny9bjfjbrsa8zxqv6iyj8hmjqw.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60dc0b0357072d42ece74567

modules/custom/social_language/modules/social_content_translation/social_content_translation.install
```diff
@@ -78,14 +78,15 @@ function social_content_translation_disable_field_translations() {
...
+}
 
-  foreach ($module_names as $module_name) {
-    $config->set($module_name, TRUE);
-    $config->save(TRUE);
-  }
+/**
+ * Delete redundant configuration settings.
+ */
+function social_content_translation_update_8002() {
```
ribel, line 87: @nechai As we agreed better to start hook update with the target OS version, so it will be 10301

modules/custom/social_language/modules/social_content_translation/config/install/social_content_translation.settings.yml
```diff
@@ -1,6 +0,0 @@
-# TODO: Overrides to these settings are currently unusable (See ContentTranslationDefaultsConfigOverride).
```
ribel, line 1: @nechai I see the point why you decided to remove this config, since it just increases complexity for content translation support, especially when we need to add support for external entities, like courses for example.

But I think we need to get approval for this removal from other teams and make sure they are not used specifically on some projects.

@Kingdutch, @ronaldtebrake could you comment on that? Do you think we can remove this config, and then just enable content translation support for all supported entities when this module is enabled?

Also is it fine to include it in the 10.3 release? Because I am afraid it will be a problem for us if we postpone this to 11.0

modules/social_features/social_core/src/ContentTranslationConfigOverrideBase.php
```diff
@@ -56,10 +43,9 @@ public function loadOverrides($names) {
 
     // This setting can't be changed in an override because that would create
     // and endless loop in trying to apply the override.
-    $settings = \Drupal::configFactory()->getEditable('social_content_translation.settings');
-    $is_enabled = $settings->getOriginal($this->getModule(), FALSE);
+    $is_content_translations_enabled = \Drupal::moduleHandler()->moduleExists('social_content_translation');
 
-    if ($is_enabled) {
+    if ($is_content_translations_enabled) {
```
ribel, line 48: @nechai I think we can move this whole config to the `social_content_translation` module. Then we don't need this extra check and spaghetti code. 

@Kingdutch do you agree with that? Or was there some specific reason to have it in `social_core` rather than in `social_content_translation` ?

phpstan-baseline.neon
```diff
@@ -5080,3 +5080,7 @@ parameters:
 			count: 1
 			path: src/Installer/OptionalModuleManager.php
 
+		-
```
ribel, line 5083: @nechai If we move config override to the `social_content_translation` we will not need this rule anymore.

ribel (changes requested):

modules/custom/social_language/modules/social_content_translation/social_content_translation.install
nechai: resolved

modules/social_features/social_core/src/ContentTranslationConfigOverrideBase.php
nechai: @ribel I think we still need this check otherwise config overrides will be loaded even if the module "social_content_translate" be disabled (PSR-4 class autoloading) - class loading(detecting) doesn't require module enabling....

nechai: Seems like I was wrong on that point - services can't load the class from the disabled module. 

So this expression `use Drupal\social_content_translation\ContentTranslationConfigOverrideBase` will raise the error if module `social_content_translation` is disabled (and it's disabled by default in distro).
I tried to declare our config overrides services with ServiceProvider but got the same issue....

We need to leave the base class `ContentTranslationConfigOverrideBase` and live with the "spaghetti code" or create another approach to resolve the problem 🤔

Kingdutch:
Hey @ribel and @nechai !

The `social_content_translation` module acts as a feature flag. So its only job is to enabled content translation in our other modules. It's exact reason for existence and the way it's worked is documented in [the module's README](https://github.com/goalgorilla/open_social/tree/10.3.x/modules/custom/social_language/modules/social_content_translation#social-language---content-translation)

Moving the base override class to the `social_content_translation` module will break things because that module may not be enabled. The module also shouldn't be a dependency of each of our content type modules because we may want platforms without content translation. That's why the override module is in `social_core` (so it's always available) but checks for `social_content_translation` (because that module acts as feature flag).

Let me know if the README answers your questions or you have any other questions afterwards. :)

open-social-tugboat:

Link:
- https://pr2395-t4qbv5ny9bjfjbrsa8zxqv6iyj8hmjqw.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60dc0b0357072d42ece74567

open-social-tugboat:

Link:
- https://pr2395-t4qbv5ny9bjfjbrsa8zxqv6iyj8hmjqw.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60dc0b0357072d42ece74567

open-social-tugboat:

Link:
- https://pr2395-t4qbv5ny9bjfjbrsa8zxqv6iyj8hmjqw.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60dc0b0357072d42ece74567

ribel:
@ronaldtebrake, @Kingdutch Do you think we can delete `social_content_translation.settings` configuration, which will simplify content translation for external modules? Or are there strict reasons why we should keep it?

ribel (approved):

Kingdutch:
> Do you think we can delete social_content_translation.settings configuration, which will simplify content translation for external modules? Or are there strict reasons why we should keep it?

It depends on what the plan is for translatability.

1) Does the new structure allow distribution users to use Drupal's tools without the `social_content_translation` to configure content translatability.
2) Does the `social_content_translation` module provide a sensible default for all our own (SaaS) platforms?
3) Do we need Site Managers to change what content can be translated on their platform? Do we have an adequate UI to do this?

As per the README in the `social_content_translation` module.

> [The] module exists to make an alternative UI possible for sitemanagers because the content translation UI exposes too much in certain scenarios.

So the settings are there to configure what things can be translated in content translation. The defaults have been chosen in such a way that they make sense for our platforms. However, developer can alter them by changing settings. A UI has not been created since.

If we want more flexibility for (SaaS) site managers then the solution suggested a few years ago would've been to build a UI to change the settings for `social_content_translation`. If we now have a different way of accomplishing that, then that will determine whether the settings can be removed.

ribel:
Thanks for the input @Kingdutch!
Taking into account the points raised above we changed the milestone to 11.0 and made it low prio.

We will use the existing approach for now, since we need to release the project with content translation support in a few weeks.

It means we don't need immediate actions and input here, and you or the Nucleus team (cc @ronaldtebrake) can decide if we want to proceed with such changes at all or just close this pull request.

For the extensions, we will proceed with an approach that will not depend on the `social_core` or `social_content_translate` module, but just on Drupal `content_translate`. You can see an example in the [kpi_analytics](https://git.drupalcode.org/project/kpi_analytics/-/merge_requests/5/diffs#230bf48e99a2c4032d7b7a0793ea8c7d9d9b2a31) module.

Kingdutch:
Hey @ribel ,

Okay! Regarding what you linked in KPI Analytics, we shouldn't use config overrides as done in 
`src/ContentTranslationDefaultsConfigOverride.php`. Since it depends on a core module the configuration in the database should just be updated as needed. Making a config override on a core feature causes no-one using the module to be able to change these settings, which is undesirable because someone may want content translation in their custom project without having this translatable.

This is the reason we have such a specific (and seemingly useless) module in the distribution. So that we can enable a specific translation configuration for the sites we manage (and let the community opt-in to the same behaviour) but if that's undesired they can ignore our `social_content_translation` module and use the `content_translation` module with configuration directly. The config override added in the KPI analytics change you linked prevents anyone from changing the content translation config for that module altogether.

ribel:
Hi @Kingdutch 
We just be update configurations with `translatable: true` where this is possible, as for example in [field.field.block_content.kpi_analytics.field_kpi_chart_labels.yml](https://git.drupalcode.org/project/kpi_analytics/-/merge_requests/5/diffs#230bf48e99a2c4032d7b7a0793ea8c7d9d9b2a31)
But as I understand from @nechai in some cases this is not possible if we for example just don't have such configs until the translation is not enabled (as `language.content_settings.block_content.kpi_analytics`).
So that's why we need to use config override there. 

I see your point that other developers might use `content_translation` in a different way. Should we then change check from  `if moduleExists('content_translation')` to the `if moduleExists('social_content_translation')`? 
Then this config override will work only if the optional social module is enabled, and other developers still will be able to implement content translation support in a different way.

nechai:
> Hi @Kingdutch
> We just update configurations with `translatable: true` where this is possible, as for example in [field.field.block_content.kpi_analytics.field_kpi_chart_labels.yml](https://git.drupalcode.org/project/kpi_analytics/-/merge_requests/5/diffs#230bf48e99a2c4032d7b7a0793ea8c7d9d9b2a31)
> But as I understand from @nechai in some cases this is not possible if we for example just don't have such configs until the translation is not enabled (as `language.content_settings.block_content.kpi_analytics`).
> So that's why we need to use config override there.
> 
> I see your point that other developers might use `content_translation` in a different way. Should we then change check from `if moduleExists('content_translation')` to the `if moduleExists('social_content_translation')`?
> Then this config override will work only if the optional social module is enabled, and other developers still will be able to implement content translation support in a different way.

Kingdutch:
@ribel Yeah that's what we did for other content translation things :) 

The `social_content_translation` is basically "This is how it should work in Open Social without customisation" and if it should work differently then you can change the configuration as needed. If we need more customisation in Open Social for site managers we probably have to dive deeper into the translation configuration system, because a single flag may not be enough, but I don't think such a broad config override based on a core module is friendly to third-party developers using Open Social :)

PR #2343, opened 2021-05-19
Issue #3185523 by rolki: Implement basic topic querying in the Graph

rollki:
## Problem
With the intention to allow external websites to display teasers for topics. Those teasers should have access to the author, title, created data, topic type, image, contents and comment data. Additionally it should be able to provide a link to the topic so that it can be viewed on the Open Social website.

The updated date is currently omitted because this information is not generally displayed on an Open Social platform itself.

It should be possible to

- Select a topic by its UUID
- Select an overview of all topics and paginate through results according to the Relay connection specification

Querying and searching for data should respect existing privacy measures (such as visibility settings) according to platform configuration.

## Solution
Implement the required GraphQL Schema and DataProducer plugins in `social_topic` that extend the schema defined on `social_graphql`

## Issue tracker
- https://www.drupal.org/project/social/issues/3185523
- https://getopensocial.atlassian.net/browse/YANG-5684
- https://github.com/drupal-graphql/graphql/issues/1219

## How to test
- [ ] Enable `social_graphql` module
- [ ] Apply https://github.com/goalgorilla/open_social/pull/2400 as a patch
- [ ] Use `Altair` or some similar tool to send a request
- [ ] Create some events on the platform
- [ ] Send request, like:
```
  topics(first: 2) {
    nodes {
      id
      title
    }
  }
```

PR #2400, opened 2021-07-05
Issue #3222148: Fix schema errors in social_topic, social_event, social_node and social_group

rollki:
## Problem
During adding kernel test coverage we should install different modules and at this moment we faced with invalid configurations or missing schemas.
So, e.g. when we trying to install `social_event` then we faced with `Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for views.view.events with the following errors: views.view.events:display.events_overview.display_options.sorts.event_date.granularity missing schema`, it means that in our configuration we used custom `plugin_id` and we do not have correct schema for that `plugin_id`.

## Solution
It is too complicated to fix all cases with incorrect configurations or missing schema but we can start from things that most important for us and for our API. So, my proposal to start from `social_topic` and `social_events` since we should implement our API there, and they are used very often. Also here's will be some fixed for some extensions (`discussion` at the moment 😁).

## Issue tracker
- https://www.drupal.org/project/social/issues/3222148

## How to test
- [ ] Use these changes together with https://github.com/goalgorilla/open_social/pull/2343
- [ ] Run some kernel tests from `social_event` module, like `/var/www/scripts/social/run-tests.sh html/profiles/contrib/social/modules/social_features/social_topic/tests/src/Kernel/GraphQL/QueryTopicsTest.php`
- [ ] Everything should be green 🤗

## Screenshots
Not needed.

## Release notes
Not needed.

## Change Record
Not needed.

## Translations
No.


open-social-tugboat:

Link:
- https://pr2400-cqw313tufal4iripduj0symtc5o9mn7m.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60e30757f543e766366a4369

open-social-tugboat:

Link:
- https://pr2400-cqw313tufal4iripduj0symtc5o9mn7m.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60e30757f543e766366a4369

modules/social_features/social_event/config/schema/social_event.schema.yml
```diff
@@ -11,3 +11,31 @@ social_event.settings:
...
       type: boolean
       label: 'Disable all event enrollments on your community'
+    show_user_timezone:
+      type: boolean
+      label: 'Show user timezone'
+    langcode:
+      type: string
+      label: 'Language code'
+    _core:
+      type: _core_config_info
```
Kingdutch, line 21: The `_core` config info doesn't need a type I believe. This should be handled entirely by core.

Kingdutch (changes requested):
Looks like good changes! Also good catches on the schemas that existed but were in the incorrect paths.

Could you check if this also fixes the issues that I added as related issues on Drupal.org?

The only thing this PR needs is an update hook to fix the types of the configuration in the database and a check on where the configuration is used to see if we need to adjust the types of checks (e.g. `1 === "1"` would be false). 🎉

modules/social_features/social_event/config/schema/social_event.schema.yml
rollki: Hmm, that is interesting because I faced this error during running the test, but now I tested it without this `_core` and it works, so, yeah, definitely we can remove it.

rollki: I checked related issues on drupal.org and it seems that they are not fixed within this PR.

open-social-tugboat:

Link:
- https://pr2400-cqw313tufal4iripduj0symtc5o9mn7m.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60e30757f543e766366a4369

modules/custom/entity_access_by_field/entity_access_by_field.install
```diff
@@ -47,7 +47,7 @@ function entity_access_by_field_update_8003() {
       foreach ($data['bundles'] as $bundle) {
         $config_factory
           ->getEditable("field.field.{$entity_type_id}.{$bundle}.{$field_name}")
-          ->set('default_value', NULL)
+          ->set('default_value', [])
```
Kingdutch, line 50: The update hook you're changing here was [added 4 years ago](https://github.com/goalgorilla/open_social/commit/f71cc2eb7bb9c5875a6ec10e0090d932313fa627) That's not going to run on any platform. This is not how update hooks work. Please add a new update hook instead.

Kingdutch, line 50: The update hook you're changing here was [added 4 years ago](https://github.com/goalgorilla/open_social/commit/f71cc2eb7bb9c5875a6ec10e0090d932313fa627) That's not going to run on any platform. Please add a new update hook instead.

Kingdutch (changes requested):
The update hooks should be created with the update_helper module. We only want to update the configuration if it still has the old default value. If the configuration has been changed by the site builder then we don't change the configuration.

open-social-tugboat:

Link:
- https://pr2400-cqw313tufal4iripduj0symtc5o9mn7m.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60e30757f543e766366a4369

rollki (commented):
Hey @Kingdutch,
PR was updated, a new configuration was now updated via the `update_helper` module.

modules/custom/entity_access_by_field/entity_access_by_field.install
rollki: I know that this update hook was added 4 years ago, I did changes there with the thought that someone might copy it in the future and repeat the same mistake 😔
But yeah we can add a new hook in this module, that will set the correct value.

open-social-tugboat:

Link:
- https://pr2400-cqw313tufal4iripduj0symtc5o9mn7m.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60e30757f543e766366a4369

open-social-tugboat:

Link:
- https://pr2400-cqw313tufal4iripduj0symtc5o9mn7m.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60e30757f543e766366a4369

modules/custom/entity_access_by_field/entity_access_by_field.install
Kingdutch: Ah that's something I hadn't considered! Definitely a good point! 

One thing we can also do if this update hook is more than 2 major versions old is remove the update hook and implement `hook_update_last_removed` instead to make sure people no longer copy it :D (We should wait with deleting 9.x update hooks until 11.0 though, not sure where this one falls in)

Kingdutch (approved):
Looks great like this! The update helper updates are very specific so they should always work and if they don't we can be very confident that a user updated the value purposefully :D 

We do still need to check why PHPStan and Travis are failing before merging this, not sure if that's caused by this PR or something in 10.3.x

rollki:
> We do still need to check why PHPStan and Travis are failing before merging this, not sure if that's caused by this PR or something in 10.3.x

It seems all PRs to 10.3.x are failed with the same problems (PHPStan and Travis).

open-social-tugboat:

Link:
- https://pr2400-cqw313tufal4iripduj0symtc5o9mn7m.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/60e30757f543e766366a4369

Kingdutch:
@goalgorilla/maintainers Even though I classify this as  a bugfix, I've not cherry-picked this to 10.2.x because it creates merge conflicts around install hooks that I don't know how to resolve safely (I don't think renaming the install hook is safe, and keeping the same name would cause issues)

PR #2411, opened 2021-07-12
Issue #3223235: Social embed iframe display issues.

nkoporec:
## Problem
When embedding videos via CKeditor or adding a media URL to the stream, the video/embed is not displayed correctly.

This is similar to: https://github.com/goalgorilla/open_social/pull/2204, but it also fixes the issues with node pages and not just stream.

## Solution
I think the issue is with the html markup, when embedding videos, the iframe element is placed in the body, but this makes it very hard to theme it, so it can be responsive, because it does not have any wrapper/parent element. I think we should wrap all iframe elements within a div, so we can then style it easier.

## Issue tracker
https://www.drupal.org/project/social/issues/3223235

## How to test
1. Enable social_embed
2. Create a new node and embed a YouTube video.
3. See the video, notice it's not displayed correctly (width, height).
4. Switch to this branch and re-test it.
5. Also test out other embeds (Spotify, Instagram ...) and also test it on the /stream page.

## Screenshots

Basic page before:
Screenshot from 2021-07-12 12-39-41

Basic page after:
Screenshot from 2021-07-12 12-37-26

Stream before:
Screenshot from 2021-07-12 12-40-14

Stream after:
Screenshot from 2021-07-12 12-37-04

## Release notes
1. Fixed an issue where embeds were not displayed correctly.

### DIFF
--- a/modules/social_features/social_embed/css/iframe.css
+++ b/modules/social_features/social_embed/css/iframe.css
@@ -1,0 +1,16 @@
+.iframe-wrapper {
+  position: relative;
+  overflow: hidden;
+  height: 0;
+  padding-bottom: 56.25%;
+}
+
+iframe {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  max-width: 100%;
+  border-radius: .5rem;
+}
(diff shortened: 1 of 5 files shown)


nachosalvador (approved):

navneet0693:
@nkoporec Do we still require this PR after https://www.drupal.org/project/socialbase/issues/3228070 has been fixed? Please try with latest release of socialblue and socialbase.

navneet0693:
This is not required anymore after the latest version of themes. I am closing this. Please re-open if you think this is a mistake.

PR #2455, opened 2021-08-06
Issue 3227098 by tBKoT: Redirect to the user's preferred language.

dmytriikaiun:
## Problem
This problem arises only then user's preferred language is different from the default language.
It's because when we use multiple language detections for the interface:
- by URL
- by preferred language

Url detection checks prefix in requested Uri and compares it with language prefixes in configuration, but default language does not have any prefix so Url negotiation returns false instead of language on all pages except the home page. In that case, Drupal goes to the next available negotiation and gets the user's preferred language. And sins it is not equal to default language we cannot switch back.
## Solution
Turn off the user language detection if it already was enabled.
Add redirection to the front page with the user's selected language.

## Issue tracker
https://www.drupal.org/project/social/issues/3227098
https://getopensocial.atlassian.net/browse/YANG-6094

## How to test
*For example*
- [ ] Enable the Social Language module
- [ ] Add any additional language
- [ ] Enable the User language detection
- [ ] Change your preferred language
- [ ] Go any page, except home
- [ ] Switch language to newly added and switch back to default

#### Changed to Dutch
зображення

#### Switch back to English
зображення

## Release notes
Now the user is able to select a preferred language when the site has at least two languages, even if User language detection is not enabled.

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/custom/social_language/config/update/social_language_update_10201.yml
+++ b/modules/custom/social_language/config/update/social_language_update_10201.yml
@@ -1,0 +1,8 @@
+language.types:
+  expected_config: { }
+  update_actions:
+    delete:
+      negotiation:
+        language_interface:
+          enabled:
+            language-user: { }
--- a/modules/custom/social_language/social_language.install
+++ b/modules/custom/social_language/social_language.install
@@ -122,6 +122,20 @@ YAML;
 }
 
 /**
+ * Disable user language detection.
+ */
+function social_language_update_10201() {
+  /** @var \Drupal\update_helper\Updater $updateHelper */
+  $updateHelper = \Drupal::service('update_helper.updater');
+
+  // Execute configuration update definitions with logging of success.
+  $updateHelper->executeUpdate('social_language', 'social_language_update_10201');
+
+  // Output logged messages to related channel of update execution.
+  return $updateHelper->logger()->output();
+}
+
+/**
  * Allow "Content Manager" role to manage translations.
  */
 function social_language_update_10301() {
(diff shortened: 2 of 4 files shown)


open-social-tugboat:

Link:
- https://pr2455-srs2u0viel1s44roazselagodgvzp7gw.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/610d36af378c5f009e794f9c

open-social-tugboat:

Link:
- https://pr2455-srs2u0viel1s44roazselagodgvzp7gw.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/610d36af378c5f009e794f9c

ribel (changes requested):
@dmytriikaiun 
So the problem occurs when the user has the preferred language in user settings and then if language switcher is also used it doesn't override user setting language, and looks like broken for LU, right?

I don't think this is a good solution just disable user language detection. It might be important for some existing sites, where language switcher is not enabled for LU.

Maybe it makes more sense to change the weight of the URL method to higher than User?

Also would like to hear the opinion of @chmez and @Kingdutch since I see they worked on this module.

image

Kingdutch:
This is a tricky one. Drupal provides a lot of flexibility because the needs are different for all sites.

As far as I'm aware we still don't have an overarching view of how multi-lingual content should work in an Open Social installation and how the various features we have should work together. (We should press Moritz on this).

My suggestion in the case where a user can set their preference in the profile then we should: always show the interface in the preferred language, show content in the preferred language if available and fall back to platform default if not, and not show the language switcher in the menu. This means that on these platforms we only show the language switcher to anonymous users.

Moving the url to a higher weight is not an option. That would result in the interface language changing when a user is working on a translation, which is undesirable.

ribel:
I agree with the approach to only show the language switcher to anonymous users. In this case, I would propose to close this PR for now.

PR #2459, opened 2021-08-10
Internal - Lock versions of our contrib modules & themes

ronaldtebrake:
## Problem
We have seen some recent releases of modules, which versions we didn't lock in Open Social, to create critical issues with the product. When someone is installing or updating using composer, the latest stable release of a given constraint will be used even if it's untested. This is usually a positive, but right now we have seen more issues coming out of this than good things.

## Solution
We lock the versions in order to have more stability & structure.

Dependabot will ensure a new PR is created once there is an update available and this will ensure we first get to see an environment, our test suite is running before someone is installing open social suddenly with a newly released versions that might not be backwards compatible.

So there might be a little more maintenance required before it reaches production, but right now we are choosing for stability & quality.

## Issue tracker
[TBD]

## How to test
- [ ] Test to see if composer is using the latest versions, nothing should be updated or downgraded or changed for that matter.

## Release notes
Nothing changes, just latest versions are locked.


open-social-tugboat:

Link:
- https://pr2459-yzgmjut8gimzhrdsgtjbdeil5wz2wzua.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6112382db52cf96a23b6c04e

Ressinel (approved):
I have checked it and see that the composer is using the latest versions, nothing downgraded.

PR #2479, opened 2021-08-27
[GraphQL] Implement community name querying in the Graph

nachosalvador:
## Problem
Different applications (mobile app, web app, etc) wants to consume the community name of Open Social. This allows them to adapt their own look and feel to that of the configured application, providing a smoother experience for the user.

## Solution
Expose the community name through the GraphQL API. This requires extending plaftorm branding schema and add CommunityAbout type.

A dedicated module (social_branding) is extended to put this GraphQL field. This allows see the community name and open for extension in the future we want provide more information about community.

## Issue tracker
https://www.drupal.org/project/social/issues/3229605
https://getopensocial.atlassian.net/browse/DE-451

## How to test

- [ ]  Enable social_branding module.
- [ ]  Check the endpoint:

```graphql
query  {
  about {
    name
  }
  branding {
    logoUrl
    colorScheme {
      accentBackground {
        css
        hexRGB
        rgba {
          alpha
          blue
          green
          red
        }
      }
    }
  }
  preferredFeatures {
    machineName
  }
}
```

## Screenshots
Screenshot at 2021-08-27 13-50-10

## Release notes
Implement community name querying in the Graph and restructure&renaming community branding data.

## Change Record
Changed how to query.
Before:
```graphql
extend type Query {
  """
  Branding information for this Open Social platform.
  """
  platformBranding: PlatformBranding!
}
```

After:
```graphql
extend type Query {
  """
  Information about the community itself.
  """
  about: CommunityAbout!
  
  """
  The brand identity for the community.
  """
  branding: CommunityBranding!

  """
  The preferred features for this community   
  """
  preferredFeatures: [Feature!]!
}
```

### DIFF
--- a/modules/social_features/social_branding/graphql/social_branding_schema_extension.base.graphqls
+++ b/modules/social_features/social_branding/graphql/social_branding_schema_extension.base.graphqls
@@ -2,6 +2,16 @@
 #                              Type Information                                #
 ################################################################################
 """
+Information about the community
+"""
+type CommunityAbout {
+  """
+  The human readable name of the community.
+  """
+  name: String!
+}
+
+"""
 Information about the branding of the Platform.
 
 This provides access to branding information such as the logo url or 
--- a/modules/social_features/social_branding/graphql/social_branding_schema_extension.base.graphqls
+++ b/modules/social_features/social_branding/graphql/social_branding_schema_extension.base.graphqls
@@ -30,14 +30,6 @@ type CommunityBranding {
   NULL if the community has not configured a branding color scheme.
   """
   colorScheme: BrandingColorScheme
-  """
-  The preferred features for this platform in order from most important to least important.
-
-  Provides focus about the focus of the platform and lets
-  external applications know what actions they should persent
-  to users in primary navigation.
-  """
-  preferredFeatures: [PlatformFeature!]!
 }
 
 """
@@ -192,9 +184,9 @@ type Color {
 }
 
 """
-A feature on the platform.
+A feature on the community.
 """
-type PlatformFeature {
+type Feature {
   """
   The machine name of this feature.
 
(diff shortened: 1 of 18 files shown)


Kingdutch (approved):
Performed a cursory review.

The proposed schema looks good and I think it's definitely an improvement of what we had. The example query that queries all fields can be understood by reading the query (i.e. I know from looking at the query what data I'm getting without necessarily knowing the schema itself).

Functionally this is covered by tests using PHPUnit so I've performed no further manual verification.

Other changes have been to rename things by removing the "platform" or "community" specific naming in our internal code. That's a good change as it's context not required for developers working on the internal code (but only for the external schema).

Technically most changes here are breaking changes. However, they're for a feature that's currently in internal testing and not supported for external users so it's safe to break this in 10.3.0 and perform no backport.

Looks good to me. Lets get this merged. 👍

Kingdutch:
@nachosalvador @goalgorilla/maintainers I think we may need to revert this. My review does not override the need for our automated checks to pass unfortunately.

nachosalvador:
Where I can see the log with the test broken?

Kingdutch:
Once a PR is merged GitHub hides the automated processes and their statuses. In this case you can see "3 of 4 checks passed" and then click view details.

https://user-images.githubusercontent.com/327697/131535870-937cd351-e7d0-4b55-b7a7-2ee6e1f712c1.mov

I think in this case I think we're okay. You can ask the Nucleus team if they can trigger a test run for the 10.3.x branch to be sure. However, I don't see any failing tests in commits after this PR is merged, so we're probably good :) 

In the future we should make sure there's a positive review for the code _and_ the CIs are green as well. (If there's an issue in the CI we should explicitly address why we're ignoring the CI in a comment before merging and explain in what alternative way we tested instead, but we didn't do that here) :)

PR #2481, opened 2021-08-30
Issue #3236262 by tBKoT: Create a new module for management overview

dmytriikaiun:
## Problem
Ideally, we want to move away dashboard page from the social_saas_core module and give the Distro the possibility (optional) to set itself up similar to our default SaaS install.

## Solution
- Create a new module with the management overview page
- Create a menu links file to add the management overview page to the toolbar
- Add to menu links list of categories and core pages
- Create/edit menu links to add categories/links provided by modules in Open social

## Issue tracker
https://www.drupal.org/project/social/issues/3236262
https://getopensocial.atlassian.net/browse/ECI-1238

## How to test
N/A

## Screenshots
зображення
зображення

## Release notes
Now we have our own management page without using a solution from the `social_saas_core` module.

## Change Record
Previously, if we want to add/edit links on the admin dashboard page, which is provided from the `social_saas_core` module, we need to add changes also to this module. These are now possible through the creation of the default Drupal core feature `MODULE.menu.links.yml`. Every category is a separate menu link item with the `<nolink>` route, it acts as a parent for other menu links. The module has the following categories as default:
- social_management_overview.overview_page.content_management (Content management)
- social_management_overview.overview_page.menu_management (Menu management)
- social_management_overview.overview_page.configuration (Configuration)
- social_management_overview.overview_page.appearance (Appearance)
- social_management_overview.overview_page.people (People)

#### Examples
To create a new category for the management overview page you need to create (if not exist) the `MODULE.menu.links.yml`. Then, add to it a new link. For best practice use pattern `MOODULE.overview_page.CATEGORY`. As a parent use `social_management_overview.overview_page` it's linked to the main management overview page, as `route_name` can be used `<nolink>` in this case, we do not need to create a separate route in the `MODULE.routing.yml` file
```
MOODULE.overview_page.CATEGORY:
  title: CATEGORY_TITLE
  route_name: <nolink>
  parent: social_management_overview.overview_page
  weight: WEIGHT
```

Creating a new link inside a specific category also requires adding a link in the `MODULE.menu.links.yml` file. For best practice use the `MOODULE.overview_page` prefix for that kind of link. As a parent, you can use one of the listed above or your own
```
MOODULE.overview_page.WHATEVER_YOU_WANT:
  title: TITLE
  description: DESCRIPTION
  route_name: EXISTING_ROUTE
  parent: PARENT
  weight: WEIGHT
```

Check docs for more information about the [Module-defined menu links](https://www.drupal.org/docs/drupal-apis/menu-api/providing-module-defined-menu-links)

## Translations
N/A


open-social-tugboat:

Link:
- https://pr2481-87izyw24ls6e7geotyx3ixemt2ubl9cs.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/612cf50abaa70d8d0392f86c

open-social-tugboat:

Link:
- https://pr2481-87izyw24ls6e7geotyx3ixemt2ubl9cs.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/612cf50abaa70d8d0392f86c

open-social-tugboat:

Link:
- https://pr2481-87izyw24ls6e7geotyx3ixemt2ubl9cs.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/612cf50abaa70d8d0392f86c

dmytriikaiun:
@Kingdutch @ronaldtebrake please check if this solution works for you

modules/custom/alternative_frontpage/src/Plugin/SocialManagementOverviewItem/AlternativeFrontpageItem.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Drupal\alternative_frontpage\Plugin\SocialManagementOverviewItem;
+
+use Drupal\social_management_overview\Plugin\SocialManagementOverviewItemBase;
+
+/**
+ * Provides a new overview item "Alternative frontpage settings".
+ *
+ * @SocialManagementOverviewItem(
+ *   id = "alernaive_frontpage_item",
```
ronaldtebrake, line 11:
```suggestion
 *   id = "alternative_frontpage_item",
```

modules/custom/entity_access_by_field/src/Plugin/SocialManagementOverviewItem/EntityAccessByFieldSettingsItem.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Drupal\entity_access_by_field\Plugin\SocialManagementOverviewItem;
+
+use Drupal\social_management_overview\Plugin\SocialManagementOverviewItemBase;
+
+/**
+ * Provides a new overview item "Public visibility settings".
+ *
+ * @SocialManagementOverviewItem(
+ *   id = "enity_access_by_field_settings_item",
```
ronaldtebrake, line 11:
```suggestion
 *   id = "entity_access_by_field_settings_item",
```

ronaldtebrake:
Hi @dmytriikaiun I didn't have the chance to fully review this. 
But I truly like the first stab at this and also thanks for the details in the PR! 👏 

It seems this will give us a lot of flexibility and seems like a good start.

What would be interesting to see if you can easily alter these plugins as well.
Lets say you want to disable one of these in an extension or re-order / update the weight. 

How would you explain this to other developers (thats great documentation to have ;) )

> If a route does not exist or the user does not have access to it link will not be shown.

Thats pretty cool! Does this also work with all cache contexts? I saw you've used `user.permission`, but there could well be different contexts that influence this. 
Maybe we need something similar as the route parameters there just for future extendibility (I don't see one right now) lets say there is a special admin link for a certain very hidden group you need to be a member of regardless of permissions. (Just thinking out loud)

Kingdutch:
Hey @dmytriikaiun ! Thanks for taking an initial stab at this.

I feel like the original dashboard feature, and the way we're re-implementing it here duplicates a lot of functionality that Drupal gives us out of the box.

Have we evaluated what the menu module system gives us? What are the shortcomings to adding normal links to the menu system (using `module.menu.links.yml`). In a properly functioning Drupal installation that should also hide links that are inaccessible for a user. (I do suspect we have some places where we check access in the controller rather than at the route level, which could break this for some pages).

For example by going to `/admin` I get the following dashboard.
image

image

In the second screenshot I'm logged in as Ben Florez and I do see the link "Mail System" which indicates to me there's a bug somewhere because I get an access denied when trying to visit that page.

These pages are generated in `\Drupal\system\Controller\SystemController` which performs an access check for every element. The fact that this is broken in Open Social is probably the core reason why we started creating a separate dashboard.

```
      // Only render accessible links.
      if (!$element->access->isAllowed()) {
        continue;
      }
```

If we require a separate dashboard for other reasons than the above bug, my suggestion would still be to copy the work that is already present in the Drupal menu module and use `module.links.dashboard.yml` files along with the corresponding hooks, rather than using a plugin based system.

ronaldtebrake:
Good input @Kingdutch ! :)

Kingdutch:
@ronaldtebrake @dmytriikaiun I was able to track down the problem to the `link to any page` permission.

This permission is granted by the `social_landing_page` module and was [added in 2020](https://github.com/goalgorilla/open_social/commit/759db948a345ecb17bbc5c0a6134f2ba2ca684fe). So it's not the original reason for our internal Dashboard, but it's definitely a WTF.

`DefaultMenuLinkTreeManipulators` skips access checks in `menuLinkCheckAccess` for users with the permission. This is why SM can see all the possible config pages but still can't access them.

open-social-tugboat:

Link:
- https://pr2481-87izyw24ls6e7geotyx3ixemt2ubl9cs.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/612cf50abaa70d8d0392f86c

dmytriikaiun:
@Kingdutch I've fully updated the structure of this module and used default Drupal corer features. Check if this way is more preferred for you please.

As for the bug with user access to links, I'm going to look at it in the current sprint, but I'll create a separate PR for it.

dmytriikaiun:
Hello @Kingdutch 
I've created a [PR](https://github.com/goalgorilla/open_social/pull/2486) that fixes an issue with visible links to which CM+ does not have access.

open-social-tugboat:

Link:
- https://pr2481-87izyw24ls6e7geotyx3ixemt2ubl9cs.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/612cf50abaa70d8d0392f86c

ribel:
@dmytriikaiun I have tested the module and all works like expected. Great job!
The only thing - I think it makes sense to give `access management overview` permissions for CM and SM by default during installing this module. Also please create a ticket on drupal.org and update PR template.

Also not sure if we need to install this module by default for new sites.
Would like to hear the input of @ronaldtebrake and @Kingdutch on it and do a final check.

open-social-tugboat:

Link:
- https://pr2481-87izyw24ls6e7geotyx3ixemt2ubl9cs.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/612cf50abaa70d8d0392f86c

ribel (approved):
All looks good to me now.

ronaldtebrake:
Hi guys,

Thank you so much for your efforts on this, I can see this becoming a good part of Open Social.
I think we still have some work to do here.

So with your PR in #2486 You have basically ensured our CM / SM's only get to see the Menu links they are allowed to see.
That also means, that when they open up Open Social, they will likely have quite some duplicates between the management overview / dashboard and the regular Drupal menu.
For example:

Screenshot 2021-09-20 at 13 35 43
Screenshot 2021-09-20 at 13 35 50

Basically what we would like to achieve with this is.

Place the menu links under the already existing Menu structure.
So for example why not put this under the already existing People menu.
Screenshot 2021-09-20 at 13 46 05

Screenshot 2021-09-20 at 13 46 09

Whenever we have migrated all the SaaS permission, that should basically allow us to install the Distro in a SaaS like versions, with only menu links the SM / CM have access to and that should properly show only menu links that they should see, if more is shown we likely need to fix permissions instead.

Otherwise with every contrib module, extension or anything we are adding to Open Social we always have to duplicate a menu in the new Management overview which results in extra maintenance / work.

Also noticed some issues with copy, I think we should also include Jessica here for improvements there in a next iteration.

Screenshot 2021-09-20 at 13 49 52

So I think planning wise this is likely only going to make it for 11.0.0

Let me know what you think!

dmytriikaiun:
Closed this PR regarding the discussion in [Slack](https://opensocial.slack.com/archives/C01CGBMLSJX/p1632131082014900).

PR #2498, opened 2021-09-14
Issue #3232812 - Add core version requirements in all info.yml files of all modules in Open Social and remove the old core: key.

ronaldtebrake:
## Problem
As per https://www.drupal.org/node/3070687

> The new `core_version_requirement` key in *.info.yml files for modules, themes, and profiles now supports semantic versioning as implemented by the Composer project. This allows modules, themes, and profiles to also specify that they are compatible with multiple major versions of Drupal core.

## Solution
Add the new `core_version_requirement` key in *.info.yml of all modules in Open Social and remove the old `core` key.

We can remove the core: key because we only accept Drupal ^8.8

> The core: is required here because Drupal Core versions before 8.7.7 do not recognize the core_version_requirement: key.

And we apply to this section

> Most modules however will have to remove deprecated code to be compatible with Drupal 9. Therefore they will not able to be compatible with all versions of Drupal 8.

For example a module that is compatible with Drupal 8 versions after Drupal 8.8.0 and also Drupal 9 will need a info.yml file like this:

```
name: My Module
type: module
core_version_requirement: ^8.8 || ^9
```

## Issue tracker
https://www.drupal.org/project/social/issues/3232812

## How to test
- [ ] Run the https://www.drupal.org/project/upgrade_status module to see that all warnings are removed.
- [ ] See that all modules are installed correctly on install & update.

## Release notes
Drupal 9 requirement: We have added the `core_version_requirements` to all our info.yml files of all our modules.

### DIFF
--- a/modules/custom/activity_send/modules/activity_send_email/activity_send_email.info.yml
+++ b/modules/custom/activity_send/modules/activity_send_email/activity_send_email.info.yml
@@ -1,7 +1,7 @@
 name: Activity Send Email
 type: module
 description: Used to send activity notifications by Email
-core: 8.x
+core_version_requirement: ^8.8 || ^9
 package: Social
 dependencies:
   - social:activity_send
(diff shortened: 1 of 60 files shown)


open-social-tugboat:

Link:
- https://pr2498-xmcwqtvhokgs1frxbcabzx9bvfgoojmd.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/61406172b8fde5506cabca29

navneet0693 (approved):

PR #2502, opened 2021-09-15
Issue #3233038 - Social_event should set time field as optional for all day events

lussoluca:
## Problem/Motivation
When creating an event, if I flag "All Day", the time field disappears but remains required, and the browser refuses to submit the form.

## Steps to reproduce
Create a new event with "All Day" selected and try to submit the form.

## Proposed resolution
Add a state to set the time field as non-required when "All Day" is selected.

## Remaining tasks
Implement it.

## User interface changes
None.

## API changes
None.

## Data model changes
None.


navneet0693:
I am not able to reproduce it on the latest version of Open Social. Please feel free to re-open if required.

PR #2516, opened 2021-09-23
Use own copy of `GraphQLTestBase` in tests

Kingdutch:
## Problem
The `GraphQLTestBase` of the graphql module creates a user during set-up
and installs languages. This may be problematic in case we have a module
that acts on this (e.g. through hook_user_insert) and then relies on
their own configuration that may not yet be installed. This happens
because all modules are installed before `setUp` is called (enabling
their hooks) but the GraphQL module's `setUp` is called before some of
our own config is installed (to ensure dependent config is installed in
the proper order).

## Solution
The easiest way to solve this is to duplicate everything we need but
remove the offending lines. We already expected tests to be explicit
about setting up their test user in the class. We choose to keep a
separate class so that it's clear what is part of the `graphql` module's
dependencies and what is needed for Open Social's added functionality.
This should also help keep the class up-to-date during module updates.

## Issue tracker
No ticket, unsupported internal testing change.

## How to test

- [ ] PHPCS should pass
- [ ] PHPStan should pass
- [ ] Travis should pass

## Screenshots
N/a

## Release notes
N/a, unsupported internal testing change.

## Change Record
N/a

## Translations
N/a

### DIFF
--- a/modules/custom/social_graphql/tests/src/Kernel/GraphQLTestBase.php
+++ b/modules/custom/social_graphql/tests/src/Kernel/GraphQLTestBase.php
@@ -1,0 +1,135 @@
+<?php
+
+namespace Drupal\Tests\social_graphql\Kernel;
+
+use Drupal\Core\Cache\Cache;
+use Drupal\graphql\GraphQL\ResolverBuilder;
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\Tests\graphql\Traits\DataProducerExecutionTrait;
+use Drupal\Tests\graphql\Traits\MockingTrait;
+use Drupal\Tests\graphql\Traits\HttpRequestTrait;
+use Drupal\Tests\graphql\Traits\QueryFileTrait;
+use Drupal\Tests\graphql\Traits\QueryResultAssertionTrait;
+use Drupal\Tests\graphql\Traits\SchemaPrinterTrait;
+use Drupal\Tests\user\Traits\UserCreationTrait;
+
+/**
+ * Modified version of graphql's GraphQLTestBase class.
+ *
+ * For GraphQL tests in Open Social use `SocialGraphQLTestBase` instead.
+ *
+ * This version of the class removes the language and user creation from the
+ * setUp step. This allows code in Open Social to rely on its own configuration.
+ *
+ * In the previous version of the class it could happen that a module was
+ * installed causing its hooks to be invoked but its configuration was not yet
+ * present, causing errors. This is not a scenario that would occur in an actual
+ * Drupal installation since config is installed when the module is enabled
+ * (before content is created).
+ *
+ * With the exception of the above outlined modification, this class should be
+ * kept in sync with the grahpql module's version of this class.
+ *
+ * @internal
+ */
+abstract class GraphQLTestBase extends KernelTestBase {
+  use DataProducerExecutionTrait;
+  use HttpRequestTrait;
+  use QueryFileTrait;
+  use QueryResultAssertionTrait;
+  use SchemaPrinterTrait;
+  use MockingTrait;
+  use UserCreationTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'system',
+    'user',
+    'language',
+    'node',
+    'graphql',
+    'content_translation',
+    'entity_reference_test',
+    'field',
+    'menu_link_content',
+    'link',
+    'typed_data',
+  ];
+
+  /**
+   * The resolver builder used for our server's schema.
+   *
+   * @var \Drupal\graphql\GraphQL\ResolverBuilder
+   */
+  protected $builder;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    $this->installConfig('system');
+    $this->installConfig('graphql');
+    $this->installEntitySchema('node');
+    $this->installEntitySchema('user');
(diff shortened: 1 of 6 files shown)


open-social-tugboat:

Link:
- https://pr2516-eijkuqx00h42phaetllmh5w7doyjdsvf.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/614c3afdd7000a409c490b08

modules/custom/social_graphql/tests/src/Kernel/GraphQLTestBase.php
```diff
@@ -42,7 +42,16 @@ abstract class GraphQLTestBase extends KernelTestBase {
...
+   * Modules to enable.
+   *
+   * The test runner will merge the $modules lists from this class, the class
+   * it extends, and so on up the class hierarchy. It is not necessary to
+   * include modules in your list that a parent class has already declared.
+   *
+   * @see \Drupal\Tests\KernelTestBase::enableModules()
+   * @see \Drupal\Tests\KernelTestBase::bootKernel()
+   *
+   * @var string[]
```
Kingdutch, line 54: Here we overwrote the `inheritdoc` because Drupal core actually just has `@var array` (this would make a good issue to file against Drupal to improve).

We could maybe combine `{@inheritdoc}` with `@var string[]` but I haven't tried that.

modules/custom/social_graphql/tests/src/Kernel/SocialGraphQLTestBase.php
```diff
@@ -43,7 +44,9 @@ protected function setUp() : void {
     $this->installConfig("social_graphql");
 
     // Set up the schema and use the Open Social GraphQL server in queries.
-    $this->server = Server::load("open_social_graphql");
+    $server = Server::load("open_social_graphql");
+    self::assertInstanceOf(ServerInterface::class, $server, "Test set-up failed: 'open_social_graphql' server is not installed.");
```
Kingdutch, line 48: It might feel pedantic that PHPStan complains about the assignment to `$this->server` but it helps us for two reasons.

1. Drupal's `Server::load` falls through to `EntityInterface::load` whose return type is `EntityInterface`. So from a type annotation perspective we're not actually guaranteed a `Server` (or `ServerInterface`) instance. We all know that's how Drupal works, but that's only if implementations are well behaved.
2. Our test would break if `open_social_graphql` was not installed and `Server::load` returned NULL. This could show up as a "Can't assign NULL to non-nullable object" or "call to function on NULL" depending on the type of `$this->server`. Neither are particularly helpful. An explicit check would tell a developer that their test setup has been broken.

modules/social_features/social_profile/tests/src/Kernel/GraphQLUsersEndpointTest.php
```diff
@@ -99,7 +99,7 @@ static function (User $u) {
   public function testUserProfileFieldsPresence() : void {
     // Test as the admin users, this allows us to test all the fields that are
     // available in an all-access scenario.
-    $this->setCurrentUser(User::load(1));
+    $this->setUpCurrentUser([], [], TRUE);
```
Kingdutch, line 102: This achieves the same and no longer relies on user 1. The problems solved are the same as for our `Server::load` call.

modules/social_features/social_user/tests/src/Kernel/GraphQL/GraphQLViewerEndpointTest.php
```diff
@@ -25,8 +26,8 @@ class GraphQLViewerEndpointTest extends SocialGraphQLTestBase {
    * It loads the current user.
    */
   public function testViewerLoadsCurrentUser() : void {
-    $user = $this->createUser();
-    $this->setCurrentUser($user);
+    $user = $this->setUpCurrentUser();
+    self::assertInstanceOf(UserInterface::class, $user, "Test setup failure: could not create test user.");
```
Kingdutch, line 30: setUpCurrentUser creates a user and returns it. However, it can also return NULL so we must still add the assertion similarly as we've done for `Server::load` so that our typechecker can keep track of what type it actually is. 

If the assertion fails that'd be a bug in the `setUpCurrentUser` which should never happen. Then again, we're programmers, so we're paid to guard against the things that should never happen (but still do) :D

open-social-tugboat:

Link:
- https://pr2516-eijkuqx00h42phaetllmh5w7doyjdsvf.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/614c3afdd7000a409c490b08

modules/custom/social_graphql/tests/src/Kernel/GraphQLTestBase.php
Kingdutch: Please note that `@var` should be above `@see` in the above example (apparently Drupal Core also didn't do this correctly). This was fixed in a new commit.

open-social-tugboat:

Link:
- https://pr2516-eijkuqx00h42phaetllmh5w7doyjdsvf.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/614c3afdd7000a409c490b08

open-social-tugboat:

Link:
- https://pr2516-eijkuqx00h42phaetllmh5w7doyjdsvf.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/614c3afdd7000a409c490b08

Kingdutch:
Merged without waiting for Travis since these changes no longer impact Travis.

PR #2277, opened 2021-03-31
Emails and Notification consistency.

navneet0693:
## Problem
*Describe the problem you're trying to solve, this should motivate why the changes you're proposing are needed.*

## Solution
*Describe the solution you've created, elaborate on any technical choices you've made. Why is this the right solution and is a different solution not the right one?*

## Issue tracker
*Paste a link to the drupal.org issue queue item. If any other issue trackers were used, include links to those too.*

## How to test
*For example*
- [ ] Using version X.Y.Z of Open Social with the example module enabled
- [ ] As a sitemanager
- [ ] Try to enable the option B on screen c/d/e
- [ ] When saving I expect the result to be F but instead see G.
- [ ] The expected result F is attained when repeating the steps with this fix applied.

## Screenshots
*If this Pull Request makes visual changes then please include some screenshots that show what has changed here. A before and after screenshot helps the reviewer determine what changes were made.*

## Release notes
*A short summary of the changes that were made that can be included in release notes.*

## Change Record
*If this Pull Request changes the way that developers should do things or introduces a new API for developers then a change record to document this is needed. Please provide a draft for a change record or a link to an unpublished change record below. [Existing change records](https://www.drupal.org/list-changes/social) can be consulted as example.*

## Translations
*Translatable strings are always extracted from the latest development branch. To ensure translations remain available for platforms running older versions of Open Social the original string should be added to `translations.php` when it's changed or removed.*
- [ ] Changed or removed source strings are added to the `translations.php` file.

PR #2529, opened 2021-09-29
New email call to actions are not translatable

Kingdutch:
## Problem
In #2277 we created a new service that accepted strings in a `getCtaButton` method. Meanwhile Drupal has great types for the things that we are processing. Using string types caused two problems:

- Call to action texts are not translatable (since the string literals were not wrapped in a translatable class or function, static analysis can not find them)
- URLs may be empty or invalid. Since we were relying on path literals rather than Drupal routes, a change in paths could break emails being sent (e.g. because someone changes the path structure or because a language on the site requires a localised path).

## Solution
Switch over to using `Url` and `TranslatableMarkup` classes.

Looking at the goal of the function being called we should be able to go a step further and use `Link` classes, maybe even eliminating the service and template that were added. However, that'll take more time than is currently available and does not solve any pressing issues.

## Issue tracker
No issue created yet

## How to test

- [ ] Check that the upgrade from old version and install of this branch of Open Social is succesful.
- [ ] Trigger an activity which triggers an email on platform.
- [ ] See that the email received contains proper context, preview of the content in action and a CTA button which links to the entity.
- [ ] See that the back-compatibility is maintained.
- [ ] Check the message templates with destination notification and email are upgraded as per new pattern.

## Release notes
Unreleased change so the release notes of #2277 are sufficient.

### DIFF
--- a/modules/social_features/social_activity/social_activity.tokens.inc
+++ b/modules/social_features/social_activity/social_activity.tokens.inc
@@ -6,6 +6,7 @@
  */
 
 use Drupal\Core\Render\BubbleableMetadata;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\group\Entity\GroupContentInterface;
 use Drupal\node\NodeInterface;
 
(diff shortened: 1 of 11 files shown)


open-social-tugboat:

Link:
- https://pr2529-6ckklwmvomii8wpedimkvmbm5lrss44u.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6154202ecc9cc4519a591a0b

navneet0693 (approved):

PR #2569, opened 2021-10-28
Issue #3246384 by vnech: Move unapplied code from Album config override

nechai:
## Problem
In the config overrides exists code that never applied:
- `modules/social_features/social_post/modules/social_post_album/src/SocialPostPhotoAlbumConfigOverride.php`

Also,  during updating OS to 10.3 a run time error is rising (see screenshot).

## Solution
- Move unapplied code from config overrides to appropriate hooks etc.

## Issue tracker
- https://www.drupal.org/project/social/issues/3246384

## How to test
- [x] Login as SM
- [x] Create "photo" post
- [x] Go to canonical post page
- [ ] Make sure the images in post using `social_x_large` image style

## Screenshots
image

## Release notes
*Move unapplied code from config overrides to appropriate hooks etc.*

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_post/modules/social_post_album/social_post_album.module
+++ b/modules/social_features/social_post/modules/social_post_album/social_post_album.module
@@ -6,6 +6,7 @@
  */
 
 use Drupal\Component\Serialization\Json;
+use Drupal\Core\Entity\Display\EntityFormDisplayInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Url;
 use Drupal\social_post\Entity\PostInterface;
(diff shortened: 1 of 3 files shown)


ribel (approved):
This PR reverts fatal errors after the update to 10.2.6 or 10.3.0 which was introduced in #2493 and solves it in a different way.

Ressinel:
Hi, @nechai Thanks for your fixes!!!

Could you add the HTT list?

Do we need to wait for fixes for the two first config overrides before review since I see fixes for the last one only?
 - modules/social_features/social_event/modules/social_event_invite/src/SocialEventInviteConfigOverride.php
- modules/social_features/social_group/modules/social_group_invite/src/SocialGroupInviteConfigOverride.php

nechai:
@Ressinel I updated the PR description, please check. 
For [these two configs overrides services](https://github.com/goalgorilla/open_social/pull/2569#issuecomment-959923576) would be great to add fixes as each one uses `requestStack` service and I'm not sure that these changes are applying as configs are cached 🤔... But as they don't cause any issues I think fixes for them can be postponed and added in another PR.

Ressinel (approved):
Looks good to me.

ribel:
🍒 picked to 10.3.x and 10.2.x

PR #2570, opened 2021-10-28
Issue #3246398 by tbsiqueira: Social Tag Settings not working on a new installation

tbsiqueira:
Initially the content types didn't have a default settings, this caused an issue where on a clean installation, users would need to re-save the social tagging in order to work. Also added a hook update to enable it by default on all already installed content types, since this is the expected behaviour

## Problem
On a clean installation, user should be able to create and add tags to the following content types as a default behaviour:

- Book page
- Event
- Basic page
- Topic

## Solution
- Enable tags by default for the following content types: Book page, event, basic page, topic
- Add a hook update to social_tag to enable tags on all installed content types as default true

## Issue tracker
https://www.drupal.org/project/social/issues/3246398

## How to test
- [ ] Fresh install Open Social
- [ ] Go to "admin/structure/taxonomy/manage/social_tagging/overview" and add a couple of tags with a parent a child
- [ ] Go to "/node/add/event" and look for the field set "Tags"
- [ ] You should see the option to add tags

## Screenshots
N/A

## Release notes
N/A

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_tagging/config/install/social_tagging.settings.yml
+++ b/modules/social_features/social_tagging/config/install/social_tagging.settings.yml
@@ -2,4 +2,8 @@ enable_content_tagging: true
 allow_category_split: true
 use_category_parent: true
 use_and_condition: false
+tag_node_type_book: true
+tag_node_type_event: true
 tag_node_type_landing_page: false
+tag_node_type_page: true
+tag_node_type_topic: true
--- a/modules/social_features/social_tagging/config/install/social_tagging.settings.yml
+++ b/modules/social_features/social_tagging/config/install/social_tagging.settings.yml
@@ -2,7 +2,7 @@ enable_content_tagging: true
 allow_category_split: true
 use_category_parent: true
 use_and_condition: false
-tag_node_type_book: true
+tag_node_type_book: false
 tag_node_type_event: true
 tag_node_type_landing_page: false
 tag_node_type_page: true
--- a/modules/social_features/social_tagging/social_tagging.install
+++ b/modules/social_features/social_tagging/social_tagging.install
@@ -121,3 +121,22 @@ function social_tagging_update_8004() {
     \Drupal::logger('social_tagging')->info($e->getMessage());
   }
 }
+
+/**
+ * Create, if not set, default settings to enable tagging on content.
+ */
+function social_tagging_update_11001(): void {
+  // Get the editable social tagging configuration.
+  $tag_settings = \Drupal::getContainer()->get('config.factory')->getEditable('social_tagging.settings');
+
+  // Get the multiple types of nodes.
+  /** @var \Drupal\node\NodeTypeInterface[] $node_types */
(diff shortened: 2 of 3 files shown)


modules/social_features/social_tagging/social_tagging.install
```diff
@@ -121,3 +121,22 @@ function social_tagging_update_8004() {
...
+/**
+ * Create, if not set, default settings to enable tagging on content.
+ */
+function social_tagging_update_11001(): void {
+  // Get the editable social tagging configuration.
+  $tag_settings = \Drupal::getContainer()->get('config.factory')->getEditable('social_tagging.settings');
+
+  // Get the multiple types of nodes.
+  /** @var \Drupal\node\NodeTypeInterface[] $node_types */
+  $node_types = \Drupal::service('entity_type.manager')->getStorage('node_type')->loadMultiple();
```
navneet0693, line 134: There can be a possibility in an already existing environment that more content types (node types) than what we expect are already present. This will ignore our intended node types, i.e, book, event, landing_page, page, and topic.

Also, we cannot assume that it's true for all. This is a little ambiguous decision to make it happen on already existing environment. May be can only focus on fresh installations?

navneet0693 (changes requested):

modules/social_features/social_tagging/social_tagging.install
tbsiqueira: Hi Navneet, thank you for your comment. 

Can you please elaborate a bit more over the first comment? The idea of the hook update is to create settings for content types that don't have a valid settings entry. 

About the update for existing installations, when you go to the settings page the default value is "true", this is why I added this hook update with the default value as "enabled".

Please let me know your thoughts.

navneet0693: This is for already existing environments, which don't have this problem. It is only on the first installation. Site managers might have also customized it according to their needs. So, it is safe to fix it only for "first installation".

modules/social_features/social_tagging/config/install/social_tagging.settings.yml
```diff
@@ -2,4 +2,8 @@ enable_content_tagging: true
 allow_category_split: true
 use_category_parent: true
 use_and_condition: false
+tag_node_type_book: true
+tag_node_type_event: true
 tag_node_type_landing_page: false
```
tbsiqueira, line 7: Change this value to false

modules/social_features/social_tagging/social_tagging.install
tbsiqueira: Remove this hook update

open-social-tugboat:

Link:
- https://pr2570-dcjlilt4gzbhijpawubohlct9it0f0vt.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/617a9de59468f8274521699e

navneet0693 (approved):

navneet0693:
🍒 -picked to 10.3.x, 10.2.x and 10.1.x

PR #2571, opened 2021-10-28
Issue #3246407 by tBKoT: Check if parent term exist on the Follow Tag on the Landing Page

dmytriikaiun:
## Problem
In this [PR](https://github.com/goalgorilla/open_social/pull/2566) I've missed one critical thing. If we add parent term to the following block on the Landing page then we get an error:
`Error: Call to a member function getName() on bool in social_follow_landing_page_preprocess_paragraph() (line 61 of profiles/contrib/social/modules/social_features/social_follow_taxonomy/modules/ social_follow_landing_page/social_follow_landing_page.module)`

## Solution
As even first level tags have a parent with a "0" id we need to check if the term does not have any "real" parent terms.

## Issue tracker
https://www.drupal.org/project/social/issues/3246407
https://getopensocial.atlassian.net/browse/YANG-6521

## How to test
- [x] Make sure that the "Allow category split." option is enabled
- [x] Create/edit Landing Page
- [x] Add tag section with first level tag
- [x] Save Landing Page

## Screenshots
N/A

## Release notes
N/A

## Change Record
N/A

## Translations
N/A


ribel (approved):
Fix was tested and resolves the regression.

ribel:
🍒 picked to 10.3.x, 10.2.x and 10.1.x

PR #2589, opened 2021-11-05
Issue #3247742 by tBKoT: Disallow to use unpublished content tags

dmytriikaiun:
## Problem
It's possible to add unpublished content tags to the content which is not logical.

## Solution
Change the process how we can taxonomy terms for the select options

## Issue tracker
https://www.drupal.org/project/social/issues/3247742
https://getopensocial.atlassian.net/browse/YANG-6546

## How to test
- [x] Create some aren't and child content tags
- [x] Created new content and tagged it with some tags
- [x] Go to Content Tags and un-published tag which we add to the content in the previous step
- [x] Edit content again
- [x] You are still able to add the unpublished tag

#### When all tags published
зображення
#### After unpublishing term
зображення

## Release notes
Now content creators can not use unpublished Content tags for content tagging.

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_tagging/social_tagging.module
+++ b/modules/social_features/social_tagging/social_tagging.module
@@ -8,6 +8,7 @@
 use Drupal\Core\Entity\Display\EntityFormDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Entity\FieldableEntityInterface;
 use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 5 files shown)


open-social-tugboat:

Link:
- https://pr2589-8dkclnda71ihjnxlpujdrionga3y1xlh.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/61852b2be368658569364df8

ribel (approved):

ribel:
Hi @ronaldtebrake,
Our customer reported described this as a bug. But just want to confirm with you or other @goalgorilla/maintainers if you don't have any objections about the proposed solution?

ronaldtebrake:
It makes sense, thanks for raising this @ribel 
Just one question, what happens if there is content connected to a tag you unpublish?

ribel:
@ronaldtebrake We discussed such use cases, but decided to go with such an approach:

If the tag was added to the content, and then the tag was unpublished it will still remain added to the content, but:

1. It will be not available on the content edit form anymore, so after the second edit it will be removed from the content
2. Unpublished tag already is hidden from content details (follow tag block)

dmytriikaiun:
@ribel Yes, all correct. We will not be able to use unpublished tags on creating/editing content, and it's will not be shown on the details pages(nodes, groups, etc.)

cc @ronaldtebrake

Ressinel (approved):
@dmytriikaiun Thanks for the fixes. Looks good to me.

ribel:
There was a conflict in `phpstan-baseline.neon` during cherry-pick, so I cherry-pick it only to 10.3.x, also the issue was reported in this version.

PR #2612, opened 2021-11-19
Issue #3250345 by tBKoT: Changes the process of converting a term name to a machine-readable name

dmytriikaiun:
## Problem
We can not use taxonomy terms with special characters as a filter field for a content search

## Solution
Change the function which converted term names to the field names to replace any of the special characters

## Issue tracker
https://www.drupal.org/project/social/issues/3250345
https://getopensocial.atlassian.net/browse/YANG-6663

## How to test
- [x] Enable `social_tagging` module
- [x] Make sure that the `Allow category split` option is enabled on the Tag settings page (`admin/config/opensocial/tagging-settings`)
- [x] Create parent tag in Content tags vocabulary with special character, for example `|`
- [x] Create child tag(s) for newly created parent tag
- [x] Add tags to new/existing content
- [x] Reindex content search index
- [x] Go to search content page
- [x] Try to filter by tag

## Screenshots
зображення
зображення

## Release notes
The process of replacing the term with machine-readable has changed. Now all non-alphanumeric characters are replaced with underscores

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_tagging/social_tagging.module
+++ b/modules/social_features/social_tagging/social_tagging.module
@@ -563,10 +563,7 @@ function social_tagging_form_views_exposed_form_alter(&$form, FormStateInterface
  *   A machine name so it can be used pogrammatically.
  */
 function social_tagging_to_machine_name($text) {
-  $text = strtolower($text);
-  $text = str_replace(' ', '', $text);
-
-  return $text;
+  return preg_replace('@[^a-z0-9-]+@', '_', strtolower($text));
 }
 
 /**


ribel (approved):
Changes were tested and work as expected on the new site where this issue appears.

ribel:
@dmytriikaiun Since we change the logic of function `social_tagging_to_machine_name()`, which is used in a few places in code, I am wondering if changes here are backwards compatible for existing sites? 
Especially for the search index. Should we write some additional update hook for re-indexing if `search_tagging` were enabled?

Also, wondering what @goalgorilla/maintainers think?

dmytriikaiun:
@ribel I guess we don't need to add any updates. This function needs mostly for the form to create a hierarchy of tags when the split category is enabled. All data is saved without any relation to converted tern names and as you can check search works as expected without any reindexing.

ronaldtebrake:
What about existing terms using this? Isn't there a way to still allow these and rather make the filters work? Potentially encoding/decoding these characters or some other variant?
It seems like we are now masking an issue

ribel:
@ronaldtebrake It actually works for existing terms. As you can see in the screenshot, currently, if you have tags with characters like `|` filters fill not work. But after this change, it will work without changes to taxonomy itself. I was just wondering about the search index, but as @dmytriikaiun explained it will not impact that.

ribel:
🍒 picked to the 10.3.x

PR #2613, opened 2021-11-19
Issue #3250368 by rolki: Add the ability to alter conditions in CommentQueryAccessHandler class

rollki:
## Problem
We have `CommentQueryAccessHandler` class that is implement `query_access` for comment entity type, and there's no way to alter `$conditions` for that class.

E.G. Drupal has a few modules to extend permission system (or custom one). And in case if we have custom permissions to display comments for anonymous users e.g. for only one node bundle, then we have to add specific conditions to `query_access` (if current user has some custom permission).

Also it can affect out `GraphQL API`, in case if we have case as I said before, when we have custom permission to display comments e.g. only for topics for anonymous users and accordingly we have check access method then sure anonymous user will see comments for topics, but when we will try to get comments for topics via `GraphQL`, also under anonymous request then we will have empty result because of `query_access`.

## Solution
Add the ability to alter conditions in CommentQueryAccessHandler class.

See an example how this change can be used: 
* https://bitbucket.org/goalgorilla/comment_permission/pull-requests/6/yang-6671-fix-issue-with-wrong-permissions
* https://bitbucket.org/goalgorilla/social_discussion/pull-requests/111/yang-6671-alter-conditions-for-comment

## Issue tracker
- https://www.drupal.org/project/social/issues/3250368
- https://getopensocial.atlassian.net/browse/YANG-6671

## How to test
- [ ] Create custom permission to display comments for anonymous users
- [ ] Assign that permission to anonymous role
- [ ] You should have check access method
- [ ] Go to the node with comments
- [ ] You should see comments
- [ ] You should not see comments via GraphQL API

## Screenshots
No.

## Release notes
It will be possible to alter conditions in `CommentQueryAccessHandler` class.

## Change Record
No.

## Translations
No.

### DIFF
--- a/modules/social_features/social_comment/social_comment.api.php
+++ b/modules/social_features/social_comment/social_comment.api.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * @file
+ * Hooks provided by the Social Comment module.
+ */
+
+/**
+ * @addtogroup hooks
+ * @{
+ */
+
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\entity\QueryAccess\ConditionGroup;
+
+/**
+ * Provides a method to alter query access conditions for comments.
+ *
+ * @param \Drupal\Core\Session\AccountInterface $account
+ *   The account to check access for.
+ * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
+ *   The entity type.
+ * @param \Drupal\entity\QueryAccess\ConditionGroup|null $conditions
+ *   The access conditions or NULL.
+ *
+ * @ingroup social_comment_api
+ */
+function hook_social_comment_query_access_alter(AccountInterface $account, EntityTypeInterface $entity_type, ConditionGroup &$conditions = NULL) {
+  $conditions = new ConditionGroup('OR');
+  $conditions->addCacheContexts(['user.permissions']);
+  $conditions->alwaysFalse();
+}
+
+/**
+ * @} End of "addtogroup hooks".
+ */
--- a/modules/social_features/social_comment/social_comment.api.php
+++ b/modules/social_features/social_comment/social_comment.api.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @file
  * Hooks provided by the Social Comment module.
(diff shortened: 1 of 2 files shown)


open-social-tugboat:

Link:
- https://pr2613-mafgmvkhrwmsj8j7sp1jb0ui5w07t9vz.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6197b1f8908ae1285b5ea28a

rollki:
[See also discussion here](https://bitbucket.org/goalgorilla/comment_permission/pull-requests/6/yang-6671-fix-issue-with-wrong-permissions)

ribel (approved):
@rollki Changes here looks good to me.
But since @dmytriikaiun is currently working on a related issue, I would like to receive his approval as well and also see if it needs to be combined and merged together with some other improvements for comment visibility.

dmytriikaiun (approved):

dmytriikaiun:
@ribel this PR does not affect and is not necessary for my changes in the distro, but it needs for the [PR](https://bitbucket.org/goalgorilla/comment_permission/pull-requests/7) in the comment_permission module.

rollki:
Hey, @andriiholub, @navneet0693
What do you think about this change? Can we merge it to OS 11.0.0?

Ressinel (approved):

Kingdutch (changes requested):
Hey @rollki ,

`QueryAccessHandlerBase` already implements such an alter hook (check line 91 and 92)
```
    // Allow other modules to modify the conditions before they are used.
    $event = new QueryAccessEvent($conditions, $operation, $account, $entity_type_id);
    $this->eventDispatcher->dispatch("entity.query_access", $event);
    $this->eventDispatcher->dispatch("entity.query_access.{$entity_type_id}", $event);
```

Why do those events (which is the standard for all QueryAccessHandlers, e.g. [the group module implements them too](https://git.drupalcode.org/project/group/-/blob/8.x-1.x/src/Entity/Access/QueryAccessHandlerBase.php#L90)) not solve the problem that we have?

My suggestion here would be to close this as "Won't fix" and use the `entity.query_access` or `entity.query_access.comment` events instead.

rollki:
Hi @Kingdutch 
Not sure how I missed that 🙈
You are right, I'll decline this PR.

PR #2617, opened 2021-11-24
Add validation for users don't have the right roles

NickDaelemans:
## Problem
When adding people to a flexible group that is restricted to specific roles, people without the right role could be added to that group.
## Solution
Added an extra validation that checks if the selected users have the right roles. If they don't, an error is thrown.

## Issue tracker
https://www.drupal.org/project/social/issues/3250886
## How to test

1. Create a flexible group that is restricted on role.
2. Try to add a user (Add directly) to that newly created group that doesn't have the role.
3. You should get an error message, that you are not allowed to add the user, since he hasn't got the right role.

## Translations
- [ ] Changed or removed source strings are added to the `translations.php` file.`


open-social-tugboat:

Link:
- https://pr2617-idiym7qtt7k2faejdc8m89gkwwtvtfwn.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/619e29bfe3686578e6459290

open-social-tugboat:

Link:
- https://pr2617-idiym7qtt7k2faejdc8m89gkwwtvtfwn.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/619e29bfe3686578e6459290

open-social-tugboat:

Link:
- https://pr2617-idiym7qtt7k2faejdc8m89gkwwtvtfwn.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/619e29bfe3686578e6459290

modules/social_features/social_group/src/Element/SocialGroupEntityAutocomplete.php
```diff
@@ -61,9 +61,30 @@ public static function validateEntityAutocomplete(array &$element, FormStateInte
         ];
 
         $account = User::load($match);
```
Kingdutch, line 63: I think it makes sense to change this to

```suggestion
        $account = User::load($match);
        if ($account === NULL) {
          throw new \RuntimeException("Input matched a user that was supposed to exist but couldn't be loaded. This indicates an implementation error where we may be mixing up entity types.");
        }
```

That avoids your individual `$account !== NULL` changes.  

The functionality in this function relies on vaildating some kind of user, so if we don't have that user we can't continue.

Additionally in your current imlpementation we have `if ($account !== NULL && !$user_has_role) {` as potential error. In this case if we don't have a user we know for sure we don't have the role which would allow them to be added. So I would expect some kind of error to come up. However, because we have no user, we're not providing an error. That indicates an issue with our implementation that throwing a clear error makes explicit.

modules/social_features/social_group/src/Element/SocialGroupEntityAutocomplete.php
```diff
@@ -61,9 +61,30 @@ public static function validateEntityAutocomplete(array &$element, FormStateInte
...
+            $form_state->setError($element, $message);
+            return;
+          }
+        }
         // User is already a member, add it to an array for the Form element
         // to render an error after all checks are gone.
-        if ($group->getMember($account)) {
+        if ($group !== NULL && $account !== NULL && $group->getMember($account)) {
           $duplicated_values[] = $account->getDisplayName();
         }
```
Kingdutch, line 89: It could be a good idea to pull the `if ($group !== NULL)` clause out of these two if statements. That way we can also add a comment explaining what the condition means, for example.

```
// We expect to have a group for the group entity autocomplete. However, in case
//  this is used outside of a group context we can still act as a normal user
// autocomplete. In that case it makes no sense to do group based validation since
// we know that a user isn't a member of a group we don't have.
```

translations.php
```diff
@@ -32,6 +32,7 @@
         - There has been more than one failed login attempt for this account. It is temporarily blocked. <br>
         - Too many failed login attempts from your computer (IP address). This IP address is temporarily blocked. <br> <br>
         To solve the issue, try using different login information, try again later, or <a href=":url">request a new password</a>');
+new TranslatableMarkup("@username doesn't have the required role. You can't add them. Please contact your community manager.");
```
Kingdutch, line 35: There's no need to add translations that are available on the `main` branch to the `translations.php` file. Those translations will be picked up by our translation extraction system automatically.

The `translations.php` file exists for strings that are modified or removed in `main` but are still used in older versions. If we removed them from the codebase entirely, they'd also be removed from our translations system and sites running older versions would not be able to load the translation. By keeping those strings from older versions in `translations.php` we ensure that the translations remain available until all our sites have upgraded past that old version (at which point we can clean up this file).

```suggestion
```

modules/social_features/social_group/src/Element/SocialGroupEntityAutocomplete.php
NickDaelemans: Couldn't we also check if the group is NULL and throw an exception as your suggestions for User? Because if we don't have a group, it doesn't make sense to further execute this validation function.

translations.php
NickDaelemans: Okay, thanks for clarifying, that was very confusing for me. I did it because it was mentioned in the default PR description :)

Kingdutch: Understandable! We can probably improve the template. If you have ideas on better wording then feel free to open a PR against [the PR template file](https://github.com/goalgorilla/open_social/blob/main/.github/PULL_REQUEST_TEMPLATE.md) :D

modules/social_features/social_group/src/Element/SocialGroupEntityAutocomplete.php
NickDaelemans: Thanks, I added this change

open-social-tugboat:

Link:
- https://pr2617-idiym7qtt7k2faejdc8m89gkwwtvtfwn.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/619e29bfe3686578e6459290

open-social-tugboat:

Link:
- https://pr2617-m0u5pqypah7r7hoidmskydzaubxt8sho.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/62454e239413e8006b3d9a01

Ressinel:
HI @NickDaelemans 
Rebase/merge the latest main branch, please.

tbsiqueira:
Closing this PR as outdated

PR #2655, opened 2021-12-16
Issue #3254684 by nachosalvador: Add featured view mode for flexible …

nachosalvador:
## Problem
The images for the flexible group spaces displayed on the Social Featured Content are no longer showing up:

## Solution
Create featured view mode for flexible group type.

## Issue tracker
https://www.drupal.org/project/social/issues/3254684

## How to test

- [ ] Enable social_featured_content module
- [ ] Create flexible group and set a group image
- [ ] Create a featured content (from a dashboard for example) for the previous created flexible group

## Screenshots
**Before**
Screenshot at 2021-12-15 18-07-48

**After**
Screenshot at 2021-12-15 19-14-41

## Release notes
None

## Change Record
None

## Translations
None

### DIFF
--- a/modules/social_features/social_featured_content/config/install/core.entity_view_display.group.flexible_group.featured.yml
+++ b/modules/social_features/social_featured_content/config/install/core.entity_view_display.group.flexible_group.featured.yml
@@ -1,0 +1,49 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - core.entity_view_mode.group.featured
+    - field.field.group.flexible_group.field_group_address
+    - field.field.group.flexible_group.field_group_description
+    - field.field.group.flexible_group.field_group_image
+    - field.field.group.flexible_group.field_group_location
+    - group.type.flexible_group
+    - image.style.social_featured
+  module:
+    - address
+    - image
+id: group.flexible_group.featured
+targetEntityType: group
+bundle: flexible_group
+mode: featured
+content:
+  field_group_address:
+    type: address_plain
+    weight: 1
+    region: content
+    label: hidden
+    settings: {  }
+    third_party_settings: {  }
+  field_group_image:
+    type: image
+    weight: 0
+    region: content
+    label: hidden
+    settings:
+      image_style: social_featured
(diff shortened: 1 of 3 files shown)


open-social-tugboat:

Link:
- https://pr2655-4gm7ysoopu1dpgqiboenhyxy0bpgawwb.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/61baeded7499db6522b09ac7

nkoporec (changes requested):
I can't reproduce this locally. 
Steps:
 1. Installed the distro and enabled the social_featured_content, social_landing_page and social_dashboard
 2. Created a new landing page and dashboard
 3. Created a new flexible group with an featured image
 4. Added a content block on landing page and dashboard
 5. In both cases the image showed up, without this patch
 
So, I'm not sure if this is needed, maybe it was fixed already ? Can you add some more steps that I can try ?

nachosalvador:
Hey @nkoporec now I see what happen, when you enable the social_landing_page the social_flexible_group_featured is enable as well and this last one provides that featured view mode.

It's not happens if you only enable the social_dashboard extension :(

I'm still wondering if this view mode configuration should be social_featured_content or just add social_flexible_group_featured to social_dashboard. @ronaldtebrake what do you think?

ronaldtebrake:
HI @nachosalvador 

Interesting question.
Flexible group is an optional module for now, and it will be very difficult to figure out which module is enabled first on our customer sites. You only want this view mode to be there if both flexible groups & dashboard is enabled, but when you enable dashboard you can't be sure flexible group is already enabled, if it isn't you don't want this configuration to be there. That is why we can't add this to `social_dashboard` 

I think in this case it requires some documentation, to ensure that we manually enable that module if we have both Flexible groups / Dashboard enabled and that didn't get enabled before 12.0.0 hits. 

From 12.0.0 of Open Social we will have Flexible Group enabled by default for everybody so that will make our lifes way easier, than we can add something to Dashboard, but for now that won't work and we need to do it manually or similar to our.

PR #2450, opened 2021-08-02
Issue #3226318 by tBKoT: Update possibly translated field group labels

dmytriikaiun:
## Problem
We noticed incorrect translation for form labels when switching languages. It happens because some of `hook_entity_form_display_alter` have the save process for form displays and it saves translated string to the default configuration.

## Solution
- Fix `hook_entity_form_display_alter` in all modules. Override language for configurations before saving. 
- Update all possibly broken displays.

## Issue tracker
https://www.drupal.org/project/social/issues/3226318
https://getopensocial.atlassian.net/browse/YANG-6094

## How to test
- [ ] Enable `social_tagging` and `social_content_translation` modules
- [ ] Add additional language for site
- [ ] Add translation for any field group label
- [ ] Go to any add/edit content form where used translated field group label
- [ ] Switch to added language
- [ ] Switch back to the default language

## Screenshots
image

## Release notes
N/A

## Change Record
N/A

## Translations
N/A

PR #2664, opened 2021-12-21
[Internal] Remove un-required code.

navneet0693:
## Problem
1. Remove accidentally added function from code social_tagging_entity_form_display_alter().
2. Remove $form_display->save() call in hook_entity_form_display_alter().

See: https://github.com/goalgorilla/open_social/pull/2450
See: https://github.com/goalgorilla/open_social/pull/2335#discussion_r772869928
See: https://github.com/goalgorilla/open_social/pull/2360#discussion_r772869513

## Solution
Remove the code.

## How to test
- [ ] Follow the HTT steps in PRs mentioned above

## Screenshots
N.A

## Release notes
N.A

## Change Record
N.A

## Translations
N.A

### DIFF
--- a/modules/social_features/social_core/social_core.module
+++ b/modules/social_features/social_core/social_core.module
@@ -107,6 +107,11 @@ function social_core_entity_form_display_alter(EntityFormDisplayInterface $form_
       $fields
     );
 
+    // This is intended to move the already existing "other fields" in the
+    // Settings section, only if they are not part of any field group.
+    // As this happens in an alter, we don't need to save our
+    // ThirdPartySettings.
+    // @see \Drupal\Core\Entity\EntityInterface::save().
     $form_display->setThirdPartySetting('field_group', 'group_settings', $field_groups['group_settings']);
   }
 }
--- a/modules/social_features/social_core/social_core.module
+++ b/modules/social_features/social_core/social_core.module
@@ -107,9 +107,7 @@ function social_core_entity_form_display_alter(EntityFormDisplayInterface $form_
       $fields
     );
 
-    $form_display
-      ->setThirdPartySetting('field_group', 'group_settings', $field_groups['group_settings'])
-      ->save();
+    $form_display->setThirdPartySetting('field_group', 'group_settings', $field_groups['group_settings']);
   }
 }
 
--- a/modules/social_features/social_tagging/social_tagging.module
+++ b/modules/social_features/social_tagging/social_tagging.module
@@ -590,36 +590,6 @@ function social_tagging_preprocess_profile(array &$variables) {
 }
 
 /**
- * Implements hook_entity_form_display_alter().
- */
-function social_tagging_entity_form_display_alter(EntityFormDisplayInterface $form_display, array $context) {
-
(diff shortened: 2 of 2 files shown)


open-social-tugboat:

Link:
- https://pr2664-9l3hyqkq5hoe7ag3d3pgdcyspyeozugs.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/61c1b6e25f46ed13d84c3de0

modules/social_features/social_core/social_core.module
```diff
@@ -107,9 +107,7 @@ function social_core_entity_form_display_alter(EntityFormDisplayInterface $form_
       $fields
     );
 
-    $form_display
-      ->setThirdPartySetting('field_group', 'group_settings', $field_groups['group_settings'])
-      ->save();
+    $form_display->setThirdPartySetting('field_group', 'group_settings', $field_groups['group_settings']);
```
ronaldtebrake, line 110: @navneet0693 could you elaborate a bit on why this is necessary?
so whenever a next person comes here they are not trying to chain `->save();` 

Also wondering if we would need to setComponents on the form_display instead of third party settings, see example below:

```
  if ($context['form_mode'] === 'layout_builder') {
    $form_display->setComponent('moderation_state', [
      'type' => 'moderation_state_default',
      'weight' => -900,
      'settings' => [],
    ]);
  }
```

Although if it work's I'd be happy with that but right now I have too little information to understand this :)

navneet0693: > could you elaborate a bit on why this is necessary?

From Drupal core's `html/core/lib/Drupal/Core/Entity/EntityInterface.php` class.

```
/**
   * Saves an entity permanently.
   *
   * When saving existing entities, the entity is assumed to be complete,
   * partial updates of entities are not supported.
   *
   * @return int
   *   Either SAVED_NEW or SAVED_UPDATED, depending on the operation performed.
   *
   * @throws \Drupal\Core\Entity\EntityStorageException
   *   In case of failures an exception is thrown.
   */
  public function save();
```

which says that it will save the entities permanently in the database. Calling `->save()`in an alter function can lead to the following:

1. Saving unwanted settings permanently
2. Saving unwanted translations in the database

If someone wants to do it, he/she should do it directly in a hook_update_N.

> Also wondering if we would need to setComponents on the form_display instead of third party settings

This is not required as the code is only intended to move the already existing "other fields" in the Settings section, only if they are not part of any field group.

ronaldtebrake: Sorry, I meant inline in code, so the next developer looking at that code also understands why :D

ronaldtebrake: (But thanks for the elaboration, that makes it clear :) ! )

Ressinel (approved):

open-social-tugboat:

Link:
- https://pr2664-9l3hyqkq5hoe7ag3d3pgdcyspyeozugs.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/61c1b6e25f46ed13d84c3de0

navneet0693:
🍒 -picked to 11.0.x

PR #2724, opened 2022-01-28
Issue #3261093 by Kingdutch: Fix out of the box swiftmailer config

Kingdutch:
<h3 id="summary-problem-motivation">Problem/Motivation</h3>
The default swiftmailer config is to use sendmail with the <code>bs</code> option.

<blockquote>Stand-alone SMTP server mode. Read SMTP commands from standard input, and write responses to standard output.  In stand-alone SMTP server mode, mail relaying and other access
controls are disabled by default. To enable them, run the process as the mail_owner user.
This mode of operation is implemented by running the smtpd(8) daemon.
</blockquote>

The issue is that this mode isn't allowed on some more secure hosting providers (such as Platform.sh). Instead they will have their PHP settings configured to use the <code>t</code> option.

<blockquote>
Extract recipients from message headers. These are added to any recipients specified on the command line.<
</blockquote>

This is also indicated in an issue open for Swiftmailer itself which will probably not get fixed: https://www.drupal.org/project/swiftmailer/issues/3174215#comment-13845855

<h3 id="summary-proposed-resolution">Proposed resolution</h3>
Adjust the <code>social_swiftmail_install</code> hook to update the <code>sendmail_mode</code> in the <code>swiftmailer.transport</code> config based on the PHP settings. This ensures that the <code>t</code> mode is properly used on supported platforms so that emailing in Open Social works out of the box.

## Issue tracker
https://www.drupal.org/project/social/issues/3261093

## How to test
*For example*
- [ ] Install Open Social on a server that has <code>sendmail_path</code> configured to <code>/usr/sbin/sendmail -t -i</code> in their php.ini and try to send an email.

## Release notes
The swiftmail_transport will now be set to `t` during installation if PHP is configured to use `sendmail` in that mode.

PR #2725, opened 2022-01-28
Backport #2724 - Fix out of the box swiftmailer config

Kingdutch:
Backport of #2724


ronaldtebrake:
Cleared the milestone here and using reference from #2724 so we have all information from that ticket going towards 10.3.x / 11.0.x and 11.1

PR #2772, opened 2022-02-16
Issue 3263649 by Corn696: Fix Event enrollment, decline message

Corn696:
## Problem
If the user id is out of sync with the profile id the wrong name is displayed in the in the decline request confirmation and the success message. The reason is that the profile entity is loaded directly with the user id of the user.

## Solution
Instead of loading the profile directly by id use EntityTypeManagerInterface loadByProperties to get the profile for the specific user.

## Issue tracker
https://www.drupal.org/project/social/issues/3263649


Ressinel:
@Corn696  I have created a separate PR https://github.com/goalgorilla/open_social/pull/2794. Let's continue there.

PR #2785, opened 2022-02-24
(10.3.x) Issue #3260572: Filter groups on access on all-groups page.

NickDaelemans:
## Problem
When you look at the 'All Groups' overview (/all-groups) you can also see community groups. The view is not filters sufficiently on group access.

## Solution
Adding a ViewsFilter and applying it using config override class should solve this issue.

## Issue tracker
https://getopensocial.atlassian.net/browse/SUP-2034
https://www.drupal.org/project/social/issues/3260572

## How to test
- [ ] Create a community group (open_group or flexible with right visibility)
- [ ] Go to /all-groups as anonymous user
- [ ] You are not able to see that new group (since you don't have access to it)
- [ ] Feel free to also check this with closed_group or flexible (member groups).

## Release notes
Groups are now filtered on right access when viewing the all-groups page. You will not be able to see closed or community groups when being anonymous or not in the right group.


NickDaelemans:
Closing as this PR is only being used for patch

PR #2837, opened 2022-03-28
Issue #3272028 by tBKoT: Removes the 'View' action on the 'Manage members' page

dmytriikaiun:
## Problem
On the `Manage members` page, we have actions like `Edit`, `Delete`, `View`, etc. As we can click on the member's avatar or name, so the `View` action actually not needed. Also, when we click on `View` it does not work and redirects back to the `Manage members` page.

## Solution
Remove the `View` action from the `Manage members` page.

## Issue tracker
 - https://www.drupal.org/project/social/issues/3272028
 - https://getopensocial.atlassian.net/browse/YANG-1680

## Theme issue tracker
N/A

## How to test
- [ ] Using the latest version of Open Social
- [ ] As a group manager
- [ ] Go to the group's `Manage members` page
- [ ] Pres on the `View` action in the table of members.
- [ ] You will be redirected to the same page

### Before merge
- [ ] If applicable (I.E. new hook_updates) the update path from previous versions (major and minor versions) are tested

### After merge

## Screenshots
зображення

## Release notes
N/A

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -2328,6 +2328,11 @@ function social_group_entity_operation_alter(array &$operations, EntityInterface
     ];
 
     foreach ($operations as $key => &$operation) {
+      // Remove users' view page from actions.
+      if ($operation['url']->getRouteName() === 'entity.user.canonical') {
+        unset($operations[$key]);
+        continue;
+      }
       if (isset($titles[$key])) {
         $operation['title'] = $titles[$key];
       }


open-social-tugboat:

Link:
- https://pr2837-toc1ximwgzj9gcx1y0wz9x6bocxnmd9g.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/6241c5719413e8006b1e8f75

rollki (approved):
LGTM, @ronaldtebrake what do you think about this change?

ronaldtebrake:
@rollki Interesting, it does make sense, the user should have access to the operation so revoking that isn't a good idea. 
Also I think this makes it more consistent with for example Events.

Does this change also make sense from a perspective of things like Courses & Challenges? 
Might be a good idea to double check this with the design team, what do you think? Alternatively we could of course fix the actual "view" link but I'm not sure from a UX perspective if that is something users expect to have or not.

dmytriikaiun:
@ronaldtebrake we have the same list of actions o the `Manage enrollment` page and the `View` action is absent as we can click on the user's name or avatar for view.
зображення

ronaldtebrake:
> @ronaldtebrake we have the same list of actions o the `Manage enrollment` page and the `View` action is absent as we can click on the user's name or avatar for view. зображення

Thanks for the info @dmytriikaiun 

What I think is important is that there are two options here.

1. Fixing the view option and create a follow up issue to add View to all the other overviews to make it consistent
2. Removing the view option which you've done and seems fair from what you've mentioned.

From a UX perspective I'm not sure if the actual "view" action is needed or not, I don't think as developers we should be making that decision.

For now I'm OK with removing it as it makes it at least more consistent as you've showed but would be good to verify these things with UX 😊

dmytriikaiun:
Hi @ronaldtebrake 
I have confirmation from Jessica to remove the view action

> You're absolutely right that the “view” CTA is not needed if we can click on the user profile, so I agree with the solution to remove the “view” CTA and instead rely on clicking the name/picture. However, we must make sure this is consistent on all member lists. So that it is always the same interaction and navigation :)

dmytriikaiun:
@ronaldtebrake can we merge this PR as we already have confirmation from Jessica?

ronaldtebrake (approved):

Ressinel:
🍒-picked to `11.3.x`

PR #2937, opened 2022-05-19
get correct node for correct group on second way

slowflyer:
Follow up on https://github.com/goalgorilla/open_social/pull/1997 with new solution
## Problem
OLD:
Just figured out, that I can not access the group manage members tab, after one of the last upgrades. fe. /group/34/membership
Even with user 1 I receive a "You are not authorized to access this page."
NEW:
When using ajax comments and you need to figure out the group during comment insert _social_group_get_current_group() will not return the group

## Solution
OLD:
This seems to be a "multilayer issue" on most installs I got rid of the behaviour by rebuilding node access permission.
But on one install I still got "You are not authorized to access this page." after rebuilding the permission.
NEW:
Get the correct node (to get correct group) on second way by using other parameters delivered by \Drupal::routeMatch().

See issue https://www.drupal.org/project/social/issues/3173334

## Issue tracker
https://www.drupal.org/project/social/issues/3173334

## How to test
*For example*
- [ ] As a sitemanager

## Screenshots

## Release notes

## Change Record

## Translations

### DIFF
--- a/modules/social_features/social_group/social_group.module
+++ b/modules/social_features/social_group/social_group.module
@@ -1839,6 +1839,14 @@ function _social_group_get_current_group($entity = NULL): ?SocialGroupInterface 
           ->load($gid_from_entity);
       }
     }
+    else {
+      $entity_type = \Drupal::routeMatch()->getParameter('entity_type');
+      $entity = \Drupal::routeMatch()->getParameter('entity');
+      if ($entity_type == 'node') {
+        $group = _social_group_get_current_group($entity);
+        $nid = $entity->id();
+      }
+    }
   }
 
   // If we have a cache key we store the value.


Ressinel:
Hi, @slowflyer 

Thanks for your contribution!
I cannot reproduce this issue. Could you provide steps in the **How to test** section?

modules/social_features/social_group/social_group.module
```diff
@@ -1839,6 +1839,14 @@ function _social_group_get_current_group($entity = NULL): ?SocialGroupInterface
           ->load($gid_from_entity);
       }
     }
+    else {
+      $entity_type = \Drupal::routeMatch()->getParameter('entity_type');
+      $entity = \Drupal::routeMatch()->getParameter('entity');
+      if ($entity_type == 'node') {
+        $group = _social_group_get_current_group($entity);
+        $nid = $entity->id();
```
robertragas, line 1847: This variable $nid isn't being used anywhere else further in the code.

robertragas:
@slowflyer I also seem to have issues with reproducing the error. Like Ressinel asked, could you provide some additional information.

I tried with multilanguage, autopath and ajax comment. In what usecase does this go wrong? Is this some custom implementation or all within Open Social? 

How can I reproduce going into this function where I have no entity, and only these routeMatch parameters that you describe "entity_type" and "entity"?

PR #2962, opened 2022-05-30
Added GraphQL OAuth module

bojanbogdanovic:
## Problem
We want to be able to limit access on GraphQL types and fields via the schema using OAuth2 scopes; so that scope validation can be performed at query validation time, so requests with insufficient scopes aren't even executed.

## Solution
Extend the "open_social" schema with "allowUser", "allowBot" and "allowAll" GraphQL directives, which are applicable on GraphQL type and field definitions. The directives accept one argument "requiredScopes": the scope id's (used on the OAuth2 authenticate/token request). Restrict access to the field/type that is being requested by collecting all scope id’s from the the OAuth2 directives by and performing scope and grant type checks.

## Issue tracker
https://www.drupal.org/project/social/issues/3174548

## How to test
- [ ] Enable 'graphql_oauth' module.
- [ ] Configure Simple OAuth2 on "admin/config/people/simple_oauth" see: https://git.drupalcode.org/project/simple_oauth/-/blob/6.0.x/README.md.
- [ ] Enable a social feature module, where a schema extension is available, e.g: 'social_user'.
- [ ] Add an "allow" directive on a field or type definitions on the enabled schema extension, see: "modules/custom/graphql_oauth/graphql/oauth_directives.graphqls".
- [ ] Perform GraphQL request with OAuth2 Authorization via a http client (e.g: Postman).

### Before merge
- [ ] If applicable (I.E. new hook_updates) the update path from previous versions (major and minor versions) are tested

### After merge
N/A

## Screenshots
N/A

## Release notes
Introducing "GraphQL OAuth2" module: extends the "open_social" schema with OAuth2 directives.

## Change Record
N/A

## Translations
N/A


open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

Ressinel:
Hi, @bojanbogdanovic 

I'd suggest moving this submodule into the `social_graphql/modules/social_graphql_oauth2` folder as it is for almost all submodules in the distro. Are there reasons why we can't do this?

Thanks!

bojanbogdanovic:
> 

There is no actual standard for it that I'm aware of, but for consistency reasons I agree with your point.

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

composer.json
```diff
@@ -181,6 +181,7 @@
         "drupal/search_api": "1.23.0",
         "drupal/select2": "1.13.0",
         "drupal/shariff": "1.7",
+        "drupal/simple_oauth": "6.0.x-dev",
```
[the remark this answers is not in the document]
bojanbogdanovic: I hope to have soon an alpha version on simple_oauth, but in the mean time it's no issue to lock this to a specific commit.

Kingdutch (changes requested):
# Module name and position
I saw the discussion around moving it into `social_graphql` as submodule. I would however actually like to propose going the other way. I think we should maybe already drop the `social_` prefix and move it into `custom/graphql_scopes` or `custom/graphql_oauth` (we can debate the name :P).

To open that discussion my current question is: "What is our current dependency in the module on Open Social? Can we remove that dependency?".

Changing the name and ensuring we don't have an Open Social dependency makes giving this to the broader Drupal GraphQL community a lot easier which I think is our goal in the long term already. It also ensures our own upgrade path is easier because it requires only `composer.json` changes rather than also changing IDs of things.

I think the only real dependency on Open Social right now is the fact that we target our schema to the `open_social` schema. Solving this may be a bit more work than we want to do in this PR but it's also something we can solve in a BC manner (dropping the direct dependency on `social_graphql` in the future). 

# Dependency on Open Social
## Cause
Our dependency on Open Social (and the need for `.extension.graphqls`) is caused by the following.

The GraphQL Schema (which contains Directive definitions) is set-up in (GraphQL Module) `Server::configuration()` which calls `$plugin->getSchema()` and sets the returned `Schema` (GraphQL library) in the `ServerConfig` (GraphQL library). The `$plugin` in this case is our instance of `SchemaPluginInterface` (GraphQL module) which lives in `social_graphql`.

When `getSchema` is called on `SdlSchemaPluginBase` it uses Drupal's plugin system to load schema extensions (`SchemaExtensionPluginInterface` from GraphQL module). 

We use `SdlSchemaExtensionPluginBase` which requires the `base` and `extension` file to be present (`SdlSchemaExtensionPluginBase::getBaseDefinition` and `SdlSchemaExtensionPluginBase::getExtensionDefinition`). Our implemented plugin also has the annotation to indicate to the plugin manager for what schema it should be loaded.

(Within Open Social all schema extensions should use `SchemaExtensionPluginBase` from `social_graphql` because it can contain goodies like `StandardisedMutationSchemaTrait` to make our lives easier. But that doesn't apply to our directive extension here.)

## Solution
We should implement `SchemaExtensionPluginInterface` without using `SdlSchemaExtensionPluginBase`. These are only three methods. `registerResolvers` for us is empty since we overwrite resolvers on a lower level. `getBaseDefinition` would return the directives (hard coding this in the class is probably fine since it avoids a file load). `getExtensionDefinition` would return `NULL`.

We should also remove the `schema` from the `@SchemaExtension` annotation. This will cause the extension not to be loaded automatically anymore. 

This means that the Open Social schema itself would now have to opt-in to using the extension (which is also the right dependency hierarchy when moving the module out of the distribution). 

For this we can look at the [`ComposableSchema`](https://github.com/drupal-graphql/graphql/blob/8.x-4.x/src/Plugin/GraphQL/Schema/ComposableSchema.php) which is included in the GraphQL module.

We don't need the `ConfigurableInterface` here because we don't want to add any configurability. We do want to override `getExtensions()`:

1. Always add the oauth2 directives extension (by `createInstance` with the ID)
2. Call `parent::getExtensions()` to load any GraphQL extensions that indicate they're extending the `open_social` schema, this keeps our existing schema extensions (e.g. in `social_topics` working).
3. Merge the results so we still return a single array :) 

# Requested Changes
I think we should do the name, filesystem position and ID changes before we merge this. Since all of those are breaking changes.

The refactor of the Schema Extension and change of our base schema class can happen in this PR if they're deemed simple enough, or they can be moved to a follow-up PR since they're not a breaking change (IDs would remain the same, the directives remain available for anyone building on the open_social schema, it just swaps graphql_oauth2 -> social_graphql dependency to social_graphql -> graphql_oauth2).

modules/custom/social_graphql/modules/social_graphql_oauth2/src/Entity/Server.php
```diff
@@ -0,0 +1,167 @@
...
+   *   The scopes on the token.
+   *
+   * @throws \GraphQL\Error\UserError
+   */
+  private function checkRequiredScopes(array $required_scopes, array $token_scopes): void {
+    foreach ($required_scopes as $scope) {
+      if (!in_array($scope, $token_scopes)) {
+        throw new UserError(sprintf("The '%s' scope is required.", $scope));
+      }
+    }
```
Kingdutch, line 140: Do we think it'd be a DX improvement to not exit early but report all missing scopes? Or do we think that's difficult because there may be different required scopes for `allowAll` or `allowType` where only requesting a single scope could solve all errors?

We could also log this as future improvement.

modules/custom/social_graphql/modules/social_graphql_oauth2/src/Entity/Server.php
```diff
@@ -0,0 +1,167 @@
...
+   *   Returns the required scopes.
+   */
+  private function getRequiredScopes(string $directive_name, array $directive_values): array {
+    $required_scopes = [];
+
+    if (isset($directive_values[$directive_name]['requiredScopes'])) {
+      $required_scopes = $directive_values[$directive_name]['requiredScopes'];
+    }
+    if (isset($directive_values['allowAll']['requiredScopes'])) {
+      $required_scopes = empty($required_scopes) ? $directive_values['allowAll']['requiredScopes'] : array_merge($directive_values[$directive_name]['requiredScopes'], $required_scopes);
```
Kingdutch, line 161: Since `$required_scopes` is always an array, should we just simplify this line and call `array_merge` directly?

It's possible it's not even slower because it avoids an if-statement. Even if it is a miniscule bit slower, it's more readable :) 
```suggestion
      $required_scopes = array_merge($directive_values[$directive_name]['requiredScopes'], $required_scopes);
```

phpstan-baseline.neon
```diff
@@ -18374,3 +18374,7 @@ parameters:
 			count: 2
 			path: modules/social_features/social_user_export/src/Plugin/Action/ExportUser.php
 
+		-
+			message: "#^Parameter \\#2 \\$node of static method GraphQL\\\\Executor\\\\Values\\:\\:getDirectiveValues\\(\\) expects GraphQL\\\\Language\\\\AST\\\\EnumValueDefinitionNode\\|GraphQL\\\\Language\\\\AST\\\\FieldDefinitionNode\\|GraphQL\\\\Language\\\\AST\\\\FieldNode\\|GraphQL\\\\Language\\\\AST\\\\FragmentSpreadNode\\|GraphQL\\\\Language\\\\AST\\\\InlineFragmentNode, GraphQL\\\\Language\\\\AST\\\\TypeDefinitionNode given\\.$#"
+			count: 1
+			path: modules/custom/social_graphql/modules/social_graphql_oauth2/src/Entity/Server.php
```
Kingdutch, line 18380: We should no longer be adding things to the baseline. In PRs the baseline should only get smaller.

We can use an `assert` in the code instead if we expect tests to fail when the condition no longer holds.

modules/custom/social_graphql/modules/social_graphql_oauth2/src/Entity/Server.php
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        // Get directive values on the type.
+        $type_directive_values = $type->astNode !== NULL ? Values::getDirectiveValues(
+          $directive_def,
+          $type->astNode
+        ) : NULL;
+
+        // Get directive values on the field.
+        $field_directive_values = $info->fieldDefinition->astNode ? Values::getDirectiveValues(
```
Kingdutch, line 84: Prefer explicit checks on non-boolean values over using PHP's truthy behavior (I thought this was a lint rule somwhere 🤔 )

```suggestion
        $field_directive_values = $info->fieldDefinition->astNode !== NULL ? Values::getDirectiveValues(
```

Kingdutch (changes requested):

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

phpstan-baseline.neon
bojanbogdanovic: Asserting won't work here, the second parameter on 'Values::getDirectiveValues()' is missing the 'TypeDefinitionNode' in the docblock comment, which is triggering the PHPStan error. We can remove this from the baseline when webonyx/graphql-php introduces a new release, for more info see: https://github.com/webonyx/graphql-php/pull/1155

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

modules/custom/social_graphql/modules/social_graphql_oauth2/src/Entity/Server.php
```diff
@@ -0,0 +1,167 @@
...
+          $directive_def,
+          $info->fieldDefinition->astNode
+        ) : NULL;
+
+        $values = $type_directive_values ?: $field_directive_values;
```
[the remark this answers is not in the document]
bojanbogdanovic: > This means "type directive" trumps "field directive" right?
Yes, actually we need to support both, that's why I have updated the logic in my last commit.

Regarding your other points, I have updated the error messaging to be more descriptive and added tests that cover the mentioned scenarios.

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

modules/custom/social_graphql/modules/social_graphql_oauth2/src/Entity/Server.php
bojanbogdanovic: The UserError only accepts a single string, it would be perhaps messy to communicate all errors in a single string. I agree we could indeed log this as future improvement.

phpstan-baseline.neon
Kingdutch: Then rather than putting it in the baseline lets move the error into `phpstan.neon` itself as ignored error with a link to the issue and PR, that:

- Ensures the issue is automatically muted in all places where we occur (since we know it's an invalid error and not a single unfixed instance)
- Makes it clearer that it's safe to remove this error if PHPStan complains it's no longer present :)

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

modules/custom/graphql_oauth/src/Entity/Server.php
```diff
@@ -0,0 +1,177 @@
...
+          if ($token === NULL) {
+            throw new MissingDataException('There is no OAuth2 token to work on.');
+          }
+          $this->checkAccess($allow_directive_values, $directive_name, $token_scopes);
+        }
+        return $parent($value, $args, $context, $info);
+      };
+    }
+    return $parent;
+  }
```
Kingdutch, line 58: On of the things that is slightly dangerous in the current set-up is that a future developer who does not understand the empty values of `$directive_name` and `$token` are linked may accidentally change the cases in which `$token` is changed while leaving `$directive_name` empty. That would cause `checkAccess` to be called with an invalid `$directive_name` value.

We can re-arrange the code by simply returning a different function depending on whether we have an account proxy or not. That can make our code simpler and more clearly communicate the intention to future developers.

```suggestion
  protected function getFieldResolver(ResolverRegistryInterface $registry) {
    $parent = parent::getFieldResolver($registry);
    if (is_callable($parent)) {
      $account_proxy = \Drupal::currentUser()->getAccount();

      if (!$account_proxy instanceof TokenAuthUserInterface) {
        // If the current user is not authenticated with OAuth then our field resolver 
        // simply denies access to any protected field because there is never a token.
        // That has the scope.
        return function ($value, $args, ResolveContext $context, ResolveInfo $info) use ($parent) {
          $allow_directive_values = $this->getAllowDirectiveValues($info);
          if (!empty($allow_directive_values)) {
              throw new MissingDataException('Token with valid scope was required but no token was presented.');
          }
          return $parent($value, $args, $context, $info);
          };
      }

      $token = $account_proxy->getToken();
      $scopes_field = $token->get('scopes');
      assert($scopes_field instanceof Oauth2ScopeReferenceItemListInterface);
      $token_scopes = array_map(function (Oauth2ScopeInterface $scope) {
          return $scope->getName();
      }, $scopes_field->getScopes());
      // `auth_user_id` is only set for tokens created for specific users.
      // Application ("bot") tokens are not linked to specific users but their user is 
      // determined by the user id configured for the consumer entity.
      $directive_name = !$token->get('auth_user_id')->isEmpty() ? 'allowUser' : 'allowBot';

      return function ($value, $args, ResolveContext $context, ResolveInfo $info) use ($parent, $token, $directive_name, $token_scopes) {
        $allow_directive_values = $this->getAllowDirectiveValues($info);
        if (!empty($allow_directive_values)) {
          $this->checkAccess($allow_directive_values, $directive_name, $token_scopes);
        }
        return $parent($value, $args, $context, $info);
      };
    }
    return $parent;
  }
```

phpstan.neon
```diff
@@ -63,4 +63,7 @@ parameters:
       - '#Unsafe usage of new static\(\)#' # See - https://github.com/mglaman/drupal-check/pull/187
       # The accessCheck test seems to be horribly broken and providing false positives
       # Disable until https://github.com/mglaman/phpstan-drupal/issues/396 is fixed.
-      - "#^Missing explicit access check on entity query\\.$#"
+      - "#^Missing explicit access check on entity query\\.$#" # See: https://github.com/webonyx/graphql-php/pull/1155
```
Kingdutch, line 66:
```suggestion
      - "#^Missing explicit access check on entity query\\.$#"
      # See: https://github.com/webonyx/graphql-php/pull/1155
```

Kingdutch (commented):
Changes look good to me! <3 

I spotted two places where I think we can change the code slightly to reduce the risk of a developer hurting themselves in the future :) 

Let me know what you think!

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

Kingdutch (approved):

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

open-social-tugboat:

Link:
- https://pr2962-t9qa8ghvijwccjmy5tfxbrzu4n5icn7l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6294d316ee0732728f5a7810

PR #3045, opened 2022-07-20
Patch Drupal 9.2.x with backported fixes for drupal core SA-CORE-2022

ronaldtebrake:
## Problem
Core released:

https://www.drupal.org/sa-core-2022-012
https://www.drupal.org/sa-core-2022-013
https://www.drupal.org/sa-core-2022-014
https://www.drupal.org/sa-core-2022-015

But since we still support versions 9.2.x in Open Social we need to backport this.

## Solution
Backport issues as patch, I'm trying directly to see if this applies to our 11.3.x branch.

## Issue tracker
https://drupal.org/node/3298688

## How to test
See that all our tests pass.

PR #3046, opened 2022-07-20
Issue #3298713: Remove patch for empty langcode so we dont create regression

ronaldtebrake:
## Problem
The patch for:

```
"Empty language (langcode) field wrapper inserted into contact form": "https://www.drupal.org/files/issues/2022-01-29/empty-langcode-field-wrapper-2842405-23.patch"
```
Created regression as we are revoking access, while all we want is to remove it from the render process in certain occassions.

## Solution
This is a 2 step solution.

1. Remove the patch.
2. Find a way to fix the use case we were trying to fix.

This PR is to fix number 1, so we don't cause regression and have this waiting for too long.
Also good to note is that the patch was already removed as part of #3045 in the 11.3.x, 11.2.x, 11.1.x branches.

## Issue tracker
https://www.drupal.org/project/social/issues/3298713

## How to test
This doesn't need a HTT, step 2 does when we fix it. 
See: https://www.drupal.org/project/drupal/issues/2842405#comment-14261236

## Release notes
Not necessary as this doesn't fix anything, nor does it close the ticket. It's just step one and to align it with #3045.


ronaldtebrake:
Cherry picked to 11.4.x 🍒

PR #3078, opened 2022-08-25
Issue #3305919 by nkoporec: User does not receive email notification when added directly to a event

nkoporec:
## Problem
When a person is directly added to a event, they do not receive a notification via email.

## Solution
 There is a message template ` member_added_by_event_organiser` that should be sent to the user, but this never happens. The issue is that in `ActivitySendEmailJobType.php `we always check if the recipient has access to the entity, before sending out an email. This is not the case, when adding users directly to the event, since the owner of the event enrollment is not the recipient but the event manager, so the condition evaluates to false

We need to allow view access to event enrollment entity if the user is a recipient, since it make sense that the user has access to its own event enrollment.

## Issue tracker
https://www.drupal.org/project/social/issues/3305919

## Theme issue tracker
*[Required if applicable] Paste a link to the drupal.org theme issue queue item, either from [socialbase](https://www.drupal.org/project/socialbase) or [socialblue](https://www.drupal.org/project/socialblue). If any other issue trackers were used, include links to those too.*

## How to test
1. Make sure you have emails sending working
2. Create an event
3. Go to the event and click the Manage enrollments tab
4. Open the Add enrollees menu and click the Add directly option
5. Select a random authenticated user to add
7. Run cron 
8. Check if the user received an email that it has been added to the event.

### Before merge
- [x] Code/peer review is completed
- [x] All commit messages are [clear and clean](https://open-social.slite.com/app/docs/DnmermZDIx_0OQ). If applicable a rebase was performed
- [ ] All automated tests are green
- [ ] Functional/manual tests of the acceptance criteria are approved
- [ ] All acceptance criteria were met
- [ ] New features or changes to existing features are covered by tests, either unit (preferably) or behat
- [ ] Update path is tested. New hook_updates should respect update order, right naming convention and consider hook_post_update code
- [ ] Module can be safely uninstalled. Update/implement hook_uninstall and make sure that removed configuration or dependencies are removed/uninstalled
- [ ] This pull request has all required labels (team/type/priority)
- [ ] This pull request has a milestone
- [ ] This pull request has an assignee (if applicable)
- [ ] Any front end changes are tested on all major browsers
- [ ] New UI elements, or changes on UI elements are approved by the design team
- [ ] New features, or feature changes are approved by the product owner

### After merge
- [ ] Code is tested on all branches that it has been cherry-picked
- [ ] Update hook number might need adjustment, make sure they have the correct order
- [ ] The Drupal.org ticket(s) are updated according to this pull request status

## Screenshots
[...]

PR #3080, opened 2022-08-25
Issue #3305919: User does not receive email notification when added directly to a event

tbsiqueira:
## Problem
When a person is directly added to a event, they do not receive a notification via email.

## Solution
 There is a message template ` member_added_by_event_organiser` that should be sent to the user, but this never happens. The issue is that in `ActivitySendEmailJobType.php `we always check if the recipient has access to the entity, before sending out an email. This is not the case, when adding users directly to the event, since the owner of the event enrollment is not the recipient but the event manager, so the condition evaluates to false

We need to allow view access to event enrollment entity if the user is a recipient, since it make sense that the user has access to its own event enrollment.

## Issue tracker
https://www.drupal.org/project/social/issues/3305919

## Theme issue tracker
*[Required if applicable] Paste a link to the drupal.org theme issue queue item, either from [socialbase](https://www.drupal.org/project/socialbase) or [socialblue](https://www.drupal.org/project/socialblue). If any other issue trackers were used, include links to those too.*

## How to test
1. Make sure you have emails sending working
2. Create an event
3. Go to the event and click the Manage enrollments tab
4. Open the Add enrollees menu and click the Add directly option
5. Select a random authenticated user to add
7. Run cron 
8. Check if the user received an email that it has been added to the event.

### Before merge
- [x] Code/peer review is completed
- [x] All commit messages are [clear and clean](https://open-social.slite.com/app/docs/DnmermZDIx_0OQ). If applicable a rebase was performed

### After merge

## Screenshots
*[Required if new feature, and if applicable] If this Pull Request makes visual changes then please include some screenshots that show what has changed here. A before and after screenshot helps the reviewer determine what changes were made.*

## Release notes
Fixed an issue with email sending when adding users directly to the event.

## Change Record
*[Required if applicable] If this Pull Request changes the way that developers should do things or introduces a new API for developers then a change record to document this is needed. Please provide a draft for a change record or a link to an unpublished change record below. Existing change records can be consulted as example. Please provide a draft for a change record or a link to an unpublished change record below. [Existing change records](https://www.drupal.org/list-changes/social) can be consulted as example.*

## Translations
*[Optional]Translatable strings are always extracted from the latest development branch. To ensure translations remain available for platforms running older versions of Open Social the original string should be added to `translations.php` when it's changed or removed.*

### DIFF
--- a/modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
+++ b/modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
@@ -68,7 +68,7 @@ function social_event_managers_activity_send_email_notifications_alter(array &$i
 /**
  * Implements hook_entity_access().
  */
-function social_event_managers_event_enrollment_access(EventEnrollment $event_enrollment, $operation, AccountInterface $account) {
+function social_event_managers_event_enrollment_access(EventEnrollment $event_enrollment, string $operation, AccountInterface $account): AccessResult {
   $access_helper = new SocialEventManagersAccessHelper();
   return $access_helper->eventEnrollmentAccessCheck(
     $event_enrollment,
--- a/modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
+++ b/modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
@@ -69,12 +69,31 @@ function social_event_managers_activity_send_email_notifications_alter(array &$i
  * Implements hook_entity_access().
  */
 function social_event_managers_event_enrollment_access(EventEnrollment $event_enrollment, string $operation, AccountInterface $account): AccessResult {
-  $access_helper = new SocialEventManagersAccessHelper();
-  return $access_helper->eventEnrollmentAccessCheck(
-    $event_enrollment,
-    $operation,
-    $account
-  );
+  // This allows view access to event_enrollment entities for users which are
+  // the recipients of the event enrollment but not the owner of the entity.
+  // For example a site manager can create an enrollment for a specific user.
+  if ($operation !== 'view') {
+    // If we are doing different operations than viewing, then let other
+    // access checks to determine the access.
+    return AccessResult::neutral();
+  }
+
+  $enrollment_status = $event_enrollment->get('field_enrollment_status')->getString();
+  if (!(bool) $enrollment_status) {
+    // Return neutral and let other access checks acts on this.
+    return AccessResult::neutral();
+  }
+
+  $owner = $event_enrollment->getOwner();
+  if (!$owner->hasPermission('view published event enrollment entities')) {
+    // Return neutral and let other access checks acts on this.
+    return AccessResult::neutral();
+  }
+
+  $field_account_id = $event_enrollment->get('field_account')->getString();
+
+  // If the user is a recipient allow access to event enrollment.
+  return AccessResult::allowedIf((int) $field_account_id === (int) $account->id());
 }
 
 /**
--- a/modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
+++ b/modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
@@ -69,44 +69,12 @@ function social_event_managers_activity_send_email_notifications_alter(array &$i
  * Implements hook_entity_access().
  */
 function social_event_managers_event_enrollment_access(EventEnrollment $event_enrollment, $operation, AccountInterface $account) {
-  // This allows view access to event_enrollment entities for users which are the
-  // recipients of the event enrollment but not the owner of the entity. For
-  // example a site manager can create an enrollment for a specific user.
-  if ($operation !== 'view') {
-    // If we are doing different operations than viewing, then let other access
-    // checks to determine the access.
-    return AccessResult::neutral();
-  }
-
-  if ((bool) $event_enrollment->get('field_enrollment_status')->getString() !== 1) {
-    // Return neutral and let other access checks acts on this.
-    return AccessResult::neutral();
-  }
-
-  $owner = $event_enrollment->getOwner();
-  if (!$owner->hasPermission('view published event enrollment entities')) {
-    // Return neutral and let other access checks acts on this.
-    return AccessResult::neutral();
-  }
-
-  $field_account = $event_enrollment->get('field_account')->referencedEntities();
-  if (empty($field_account)) {
-    // This shouldn't happen, but just in case we return access neutral
-    // so other parts can determine access if needed.
-    return AccessResult::neutral();
-  }
-
-  $field_account = end($field_account);
-  if ($field_account->id() != $account->id()) {
-    // If current user is not the recipient then let other
-    // part of the system determine the access, by default it should
-    // be access denied, but it could be overwritten, so we don't make
-    // any assumptions here.
-    return AccessResult::neutral();
-  }
-
-  // The user is a recipient, so we allow access to event enrollment.
-  return AccessResult::allowed();
+  $access_helper = new SocialEventManagersAccessHelper();
+  return $access_helper->eventEnrollmentAccessCheck(
+    $event_enrollment,
+    $operation,
+    $account
+  );
 }
 
 /**
--- a/modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
+++ b/modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
@@ -66,6 +66,40 @@ function social_event_managers_activity_send_email_notifications_alter(array &$i
 }
 
 /**
+ * Implements hook_entity_access().
+ *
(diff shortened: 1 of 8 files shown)


tbsiqueira:
This PR recreates the followin: https://github.com/goalgorilla/open_social/pull/3078

open-social-tugboat:

Link:
- https://pr3080-j1aaxzsrthyhbkjm3oafnpe76xnlf4bs.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/63078034a982dc5767b9b2ae

modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
```diff
@@ -65,6 +65,40 @@ function social_event_managers_activity_send_email_notifications_alter(array &$i
   }
 }
 
+/**
+ * Implements hook_entity_access().
+ *
+ * This allows view access to event_enrollment entities for users which are the
```
tbsiqueira, line 71: This comment is true for now, but might change in the future, how about moving it inside the function so it will be more visible if something changes?

modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
```diff
@@ -65,6 +65,40 @@ function social_event_managers_activity_send_email_notifications_alter(array &$i
...
+ * recipients of the event enrollment but not the owner of the entity. For
+ * example a site manager can create an enrollment for a specific user.
+ */
+function social_event_managers_event_enrollment_access(EventEnrollment $event_enrollment, $operation, AccountInterface $account) {
+  if ($operation != 'view') {
```
tbsiqueira, line 76: Please use strict checks !==

modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
```diff
@@ -65,6 +65,40 @@ function social_event_managers_activity_send_email_notifications_alter(array &$i
...
+    return AccessResult::neutral();
+  }
+
+  if ($event_enrollment->get('field_enrollment_status')->getString() != "1") {
```
tbsiqueira, line 80: Boolean field? How about casting it to bool ? It will make it more readble

modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
tbsiqueira, line 76: Please add some comments on why we are returning `AccessResult::neutral();` if we are viewing the content

modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
```diff
@@ -65,6 +65,40 @@ function social_event_managers_activity_send_email_notifications_alter(array &$i
...
+    return AccessResult::neutral();
+  }
+
+  $owner = $event_enrollment->getOwner();
+  if (!$owner->hasPermission('view published event enrollment entities')) {
```
tbsiqueira, line 85: Same comment here, please add the reason why this return is neutral

modules/social_features/social_event/modules/social_event_managers/social_event_managers.module
```diff
@@ -65,6 +65,40 @@ function social_event_managers_activity_send_email_notifications_alter(array &$i
...
+    return AccessResult::neutral();
+  }
+
+  $field_account = $event_enrollment->get('field_account')->referencedEntities();
```
tbsiqueira, line 89: Same comment here, please add the reason why this return is neutral

tbsiqueira:
Hi @nkoporec thank you for your work. 

Is this covered by a behat test? If so please make sure to cover this change on a test.

nkoporec:
@tbsiqueira This is ready for another review.

modules/social_features/social_event/modules/social_event_managers/src/SocialEventManagersAccessHelper.php
```diff
@@ -63,4 +64,42 @@ public static function getEntityAccessResult(NodeInterface $node, $op, AccountIn
...
+    }
+
+    $owner = $event_enrollment->getOwner();
+    if (!$owner->hasPermission('view published event enrollment entities')) {
+      // Return neutral and let other access checks acts on this.
+      return AccessResult::neutral();
+    }
+
+    $field_account_id = $event_enrollment->get('field_account')->getString();
+    if ($field_account_id != $account->id()) {
```
NickDaelemans, line 93: Shouldn't we use strict checks?

modules/social_features/social_event/modules/social_event_managers/src/SocialEventManagersAccessHelper.php
```diff
@@ -63,4 +64,42 @@ public static function getEntityAccessResult(NodeInterface $node, $op, AccountIn
...
+    if ((int) $field_account_id !== (int) $account->id()) {
+      // If current user is not the recipient then let other
+      // part of the system determine the access, by default it should
+      // be access denied, but it could be overwritten, so we don't make
+      // any assumptions here.
+      return AccessResult::neutral();
+    }
+
+    // The user is a recipient, so we allow access to event enrollment.
+    return AccessResult::allowed();
```
robertragas, line 102: @nkoporec it works, but wondering if there is something we can rewrite for this? In general I get a bit scared when I see an access check that end with allowed() because if something would change in the flow in the future it could possible create a security issue.

If it would be allowedIf() I would already feel a bit better.

nkoporec: hmm I get your point, but I don't think that there is a way to rewrite this without having the allowed() since it needs to override the default behavior. We could add the `allowedIf((int) $field_account_id === (int) $account->id())` and get rid of the if before that , if that works better for you ?

robertragas: Yeah, I think that sounds like a good plan, if you can do that I will approve and merge.

Kingdutch:
👋 Spotted this while looking at the state of our CI queue. I think the issue's title is a lot more clear about the "Why" (since our code already documents "What") so I've mirrored the issue title to the PR :D 

Loving the addition of a unit test for our access check here <3 Is there any reason we're moving the new function into a Helper rather than keeping it in the access hook directly?

Not having that indirection gives us a few benefits:
- It's easier to see what's happening. The function contains comments of what's happening but the hook implementation doesn't give any information on why it exists.
- We ensure the logic isn't re-used other than through Drupal's access system so we don't accidentally duplicate all the checks in another hook or bypass other hooks by calling the helper function directly.
- Not adding to the helper makes it easier to eliminate these "catch-all" classes in the future.

We should be able to still unit test when all the code is in the hook! 🤩 I'm not sure if we can call the function directly (e.g. our autoloader already loads .module files) or we can `include` the `.module` file that we're testing within our test function :)

robertragas (approved):

robertragas:
> 👋 Spotted this while looking at the state of our CI queue. I think the issue's title is a lot more clear about the "Why" (since our code already documents "What") so I've mirrored the issue title to the PR :D
> 
> Loving the addition of a unit test for our access check here <3 Is there any reason we're moving the new function into a Helper rather than keeping it in the access hook directly?
> 
> Not having that indirection gives us a few benefits:
> 
> * It's easier to see what's happening. The function contains comments of what's happening but the hook implementation doesn't give any information on why it exists.
> * We ensure the logic isn't re-used other than through Drupal's access system so we don't accidentally duplicate all the checks in another hook or bypass other hooks by calling the helper function directly.
> * Not adding to the helper makes it easier to eliminate these "catch-all" classes in the future.
> 
> We should be able to still unit test when all the code is in the hook! 🤩 I'm not sure if we can call the function directly (e.g. our autoloader already loads .module files) or we can `include` the `.module` file that we're testing within our test function :)

Good one, will see if we can relocate the code.

robertragas:
So I ran into an issue with rewriting the unit test as we need to include the .module file for the hook_entity_access. I tried a few different methods like the moduleHandler and setting containers but didn't get it to work.

This morning I spoke with @Kingdutch and he was talking to just include it like you can see in the code with require(). To move a bit away from Drupal and more towards PHP. Also as we just need to test this specific function. Maybe he can explain it in more detail.

cc @navneet0693 @ronaldtebrake if you have a certain opinion, feel free to state them.

robertragas:
I did found one more thing when I go for main -> install to this branch -> update.

>  [notice] Update started: social_calendar_update_11501
>  [notice] Update completed: social_calendar_update_11501
>  [warning] setDisplay() called with invalid display ID "rest_export_events". ViewExecutable.php:805
>  [warning] setDisplay() called with invalid display ID "rest_export_events". ViewExecutable.php:805
 [success] Finished performing updates.

PR #3094, opened 2022-09-07
Issue #3308363 by rolki: Improve access checking and fix possible performance issue for custom content list block

rollki:
## Problem
After processing of all plugins in `social_content_block/ContentBuilder`, we have an array of entities, after which we sort them and limit the number of results using `$query->range(0, $block_content->field_item_amount->value);` 
The problem is that after limiting, we load the objects entities and check whether the user has access to them, this is not correct as the selection is already limited to, for example, 5 items, but the user has access to only 2, in this case, even though 5 is selected in the block settings, only 2 will be displayed.

Another problem is that some sites, to solve the problem above, simply remove this range and add `array_slice` after the objects are loaded and checked for access - and this is even worse since `5000+` (depends on platform) entities can be loaded and checked for access. As a result, a serious performance problem arises.

## Solution
Access checking should be performed at the range and at the database level.

## Issue tracker

- https://www.drupal.org/project/social/issues/3308363
- https://getopensocial.atlassian.net/browse/PROD-22116

## Theme issue tracker
NONE.

## How to test

- [ ] A custom content list block was added to the site, and to ensure that a lot of data was included in the sample
- [ ] Entities with different visibility options should be displayed
- [ ] Log in as a user who cannot see all entities

### Before merge

### After merge

## Screenshots
NONE.

## Release notes
Improve access checking and fix possible performance issue for custom content list block.
Now access is checked at the earliest stage, using the `ENTITY_access` tag

## Change Record
NONE.

## Translations
NONE.

### DIFF
--- a/modules/social_features/social_content_block/src/ContentBuilder.php
+++ b/modules/social_features/social_content_block/src/ContentBuilder.php
@@ -73,7 +73,7 @@ class ContentBuilder implements ContentBuilderInterface {
   /**
    * {@inheritdoc}
    */
-  public static function trustedCallbacks() {
+  public static function trustedCallbacks(): array {
     return ['getEntities', 'build'];
   }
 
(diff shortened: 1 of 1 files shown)


ribel (approved):
Changes looks good to me.

tbsiqueira (approved):

PR #3108, opened 2022-09-13
Internal: Move Behat into a single workflow

Kingdutch:
## Problem
Our current test parallelisation is a bit arbitrary and we can probably run more things in parallel faster.

## Solution
We use the matrix function to run our different tests. Additionally we switch from our Travis `stability` split up to just running all individual features in separate jobs.

We take some inspiration from
https://tomasvotruba.com/blog/2020/11/16/how-to-make-dynamic-matrix-in-github-actions/ to discover all the feature files and then create a job for each of them. This creates more noise in the list of tasks for a PR but it also makes it a lot easier to figure out which task failed.

We also do our set-up in a way so that we only need to install Open Social twice.

Because some tests were never ran in the previous set-up we add the `@disabled` tag to those tests and exclude them from being executed. This allows us to revamp the way tests are executed but fix the tests one at a time. Since all tests now run in isolation but this wasn't the case previously, some tests depended on other tests running before them. Those tests have been disabled as well:

### Disabled tests
administration/flood-edit.feature fails because the Open Social SM menu also has a “configuration” link which leads to a difference screen. This means that the required form elements can’t e found.

book/book-create.feature was never actually run in the previous test set-up.

gdpr/user-consent.feature has some undeclared assumptions that previously worked because tests were not run in isolation. This means a set-up step is needed which is not actually in the test itself.

group/group-create-closed.feature is disabled because the Chrome Driver tries to click "Delete" using screen coordinates. However, between finding the coordinates of the button and clicking them the button moves due to styles or DOM being rendered, this causes the click to fail.

install/install.feature was not actually run since it has `@stability` but it does not have `@stability-*`. I suspect this is broken due to the way the Chrome Driver tries to find the installer element and we must change the assertion.

landing-page/featured-profile-create.feature (without optional modules) is broken because it doesn’t enable the layout builder module which is used for testing. The module that is enabled is a dependency and does not itself depend on the layout builder.

notifications/notification-emails.feature has only a `@stability` and not a `@stability-*` tag. It seems to fail in clicking a delete button which has been an issue other Drupal developers have seen with ChromeDriver too. The way it clicks elements is not always resistant to page shifts due to loading content.

post/post-album.feature seems to require AJAX but does not actually run JavaScript it looks like.

post/post-modal-comments.feature has a poorly written "Wait for AJAX" and causes instability. In the rewrite we should wait for the result that the user can see instead.

private-message/create-private-message.feature is not entirely clear why its failing.

profile/click-avatar.feature does not have the correct tags so it was not yet executed before we changed our testing set-up.

security/private-file-uploads.feature has a `@stability` but not a `@stability-*` tag so it was not executed in the previous testing set-up. It appears to fail because it requires JavaScript but doesn’t enable this for the test.

sharing/share.feature does not have a `@stability` tag so it was not executed in the previous set-up.

topic/topic-create.feature requires JavaScript but does not declare this in the test.

## Issue tracker
Internal change

## How to test
- [ ] Behat tests should pass

### Before merge

### After merge


Kingdutch:
This is now ready for review. After discussing with Tiago he's agreed that we will disable any tests that are failing because of the migration. Almost all of these tests were not running previously so there's only a very minimal change in test coverage. This way we can benefit from the improved speed in all other PRs sooner.

https://github.com/goalgorilla/open_social/pull/3110 showed that fixing all tests at once creates something that's difficult to review, so by disabling failing tests first we can fix and re-enable them in one easy to review PR at a time.

tbsiqueira (approved):

PR #3139, opened 2022-09-30
Issue #3313206 by navneet0693, zanvidmar: Email subjects are receive in default language of a multilingual website

zanvidmar:
## Problem
Accessing translated configuration

Drupal by default always uses the language selected for the page to load configuration with. So if you are viewing a Spanish page, all configuration is loaded with Spanish translations merged in. If you are viewing the same page in Hungarian, the same code will now receive Hungarian translated configuration. This means normally you don't need to do anything special to access configuration in the language needed.
 
However, there are cases, when you want to load the original copy of the configuration or ask for a specific language. Such as when sending emails to users, you will need configuration values in the right language. The following code is a slightly adapted excerpt from user_mail() to illustrate loading configuration in the preferred language of $account to compose an email:
 

```
 <?php
   $language_manager = \Drupal::languageManager();
   $language = $language_manager->getLanguage($account->getPreferredLangcode());
   $original_language = $language_manager->getConfigOverrideLanguage();
   $language_manager->setConfigOverrideLanguage($language);
   $mail_config = \Drupal::config('user.mail');
     // ...
     // ...Compose (and send) email here...
     // ...
   $language_manager->setConfigOverrideLanguage($original_language);
 ?>
 ```
 
Note that you are setting values on the language manager and not the configuration system directly. The configuration system is override-agnostic and can support overrides of different kinds with various conditions. It is the job of the overrides to manage their conditions, in this case to allow changing the language used. The same pattern can be used to load configuration entities in specific languages as well.

[Source](https://www.hojtsy.hu/blog/2014-may-26/drupal-8-multilingual-tidbits-16-configuration-translation-development)

## Solution
Create a service that will change language so configuration can be processed correctly.

Usage:
1. configOverrideLanguageStart($langcode)
2. Load configuration in correct language
3. configOverrideLanguageEnd()

## Issue tracker
https://www.drupal.org/project/social/issues/3313206

## Theme issue tracker
N/A

## How to test
- [ ] Install Open Social
- [ ] Enable the Social Language module
- [ ] Enable the Social Content Translation module.
- [ ] Add two language apart from English.
- [ ] Add two users with different preferred languages.
- [ ] Translate the email message template in those languages.
- [ ] Send a private message from User 1 to User 2.
- [ ] Given the User 2 is offline, it should receive an Email notification on Cron run.
- [ ] The email subject should be in the user's preferred language of the website.

### Before merge
- [ ] Code/peer review is completed
[...]

PR #3142, opened 2022-10-03
Issue #3313206 by navneet0693, zanvidmar: Email subjects are receive …

navneet0693:
Backport of #3139


zanvidmar (changes requested):
This PR is not covering the fact that emails can be send via social_swiftmail_sendgrid as well. Additional PR have to be crated for social_swiftmail_sendgrid as this is where subject (and some other data) is processed.

navneet0693:
@zanvidmar Extensions and Open Social distribution profile are independent of each other. This PR along with the parent PR are only focusing on Open Social distribution.

zanvidmar:
@navneet0693 Yes that is correct, as agreed on call this are two separated issues and both of them will coexist. Thank you for clarification.

PR #3154, opened 2022-10-13
Issue #3315177 by chmez: Extend group selection widget for any entity type

ghost:
## Problem
The "social_group_selector_widget" field widget which is used for selecting a group on the event/topic creation page supports only node entity type.

## Solution
Add a new method (the "types" method) to a class of this widget which provides a list of supported entity types. By default, this list contains only the node entity type to be backward compatible.

## Issue tracker
- https://www.drupal.org/project/social/issues/3315177
- https://getopensocial.atlassian.net/browse/PROD-22260

## How to test
- [ ] Create a field widget based on the "social_group_selector_widget" widget and extend the returned list of the "types" method by the "media" item.
- [ ] Enable the [Group Media](https://www.drupal.org/project/groupmedia) module and create the "File" bundle with the "Group" field (similar to a field with the same name from topics and events) based on created field widget.
- [ ] Navigate to the "/admin/group/types/manage/flexible_group/content" page and install the "Group media (File)" plugin.
- [ ] Add the "hook_social_group_linked_type" hook to your module. This hook should return the name of the entity type - media.
- [ ] Log in as "User 1" and create the "Group 1" flexible group with the "Group members" visibility.
- [ ] Create the "Group 2" flexible group with the "Community" visibility.
- [ ] Log in as "User 2" and create the "Group 3" flexible group.
- [ ] When you navigate to the creation page of the media entity type of the "File" bundle then you should see the "Group" field.
- [ ] When you open the values drop-down list of the "Group" field then you should see the "Group 2" item and you shouldn't see the "Group 1" item.

## Release notes
Support any entity type by group selection widget.


modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
```diff
@@ -111,79 +112,103 @@ public static function create(ContainerInterface $container, array $configuratio
...
+    if ($options_provider !== NULL) {
+      $options = $options_provider->getSettableOptions($account);
+
+      $storage = $this->entityTypeManager->getStorage('group');
+
+      // Check for each group type if the content type is installed.
+      foreach ($options as $key => $optgroup) {
+        // Groups are in the array below.
+        if (is_array($optgroup)) {
+          $group = $storage->load(array_keys($optgroup)[0]);
```
ksis, line 171: Wouldn't it be better to use the [key](https://www.php.net/manual/en/function.key.php) function?

ribel (approved):
Changes looks good to me.

navneet0693:
@OH17L Please update the PR description and HTT both on Github and D.org as well. Currently, it is ambiguous and HTT is not clear enough to guide us on what shall we test.

OH17L:
@navneet0693, done. Please, take a look at this again.

ribel:
Hi @tbsiqueira and @navneet0693 
This PR is ready for the review, could you check it please?

modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
```diff
@@ -111,79 +112,103 @@ public static function create(ContainerInterface $container, array $configuratio
...
+          $entity,
+        );
+      }
+    }
+    else {
+      $options = [];
     }
 
     // Add an empty option if the widget needs one.
     if ($empty_label = $this->getEmptyLabel()) {
```
navneet0693, line 202: What happens if `$options` is [] here?

modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
```diff
@@ -232,8 +257,11 @@ public function formElement(FieldItemListInterface $items, $delta, array $elemen
     }
 
     $change_group_node = $sg_settings->get('allow_group_selection_in_node');
-    /** @var \Drupal\Core\Entity\EntityInterface $entity */
-    $entity = $form_state->getFormObject()->getEntity();
+
+    /** @var \Drupal\Core\Entity\EntityFormInterface $form_object */
```
navneet0693, line 261: We should assert if the form object is returned, before getting entity.

navneet0693:
This is good to go, just needs some tiny adjustments.

modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
OH17L: When you will try to create an entity inside a group and this array isn't filled in by the "options_list" alter hook below then you should see only the "- None -" item in the "Group" field and when you press "Save" then the entity will be created outside of the group. An existing entity will be moved outside the group when you will edit the first one. I don't think that that array will be empty at any time. I've implemented it just to support a possible empty value returned by the "getOptionsProvider" method and avoid the new ignoring rule for PHPStan.

modules/social_features/social_group/src/Plugin/Field/FieldWidget/SocialGroupSelectorWidget.php
OH17L: Since it's part of a field widget which can be used only for entity form so I'd say that it's guaranteed that the form object will be returned in all cases.

OH17L:
Looks like everything can be kept as is.

navneet0693 (approved):

PR #3103, opened 2022-09-08
Issue #3308595: Remove social_profile theme suggestions alter

Kingdutch:
This change was split out from https://github.com/goalgorilla/open_social/pull/2478

<h3 id="summary-problem-motivation">Problem/Motivation</h3>
The suggestions are provided <a href="https://git.drupalcode.org/project/profile/-/blame/8.x-1.x/profile.module#L242">by the profile module itself</a> in [#2694265]

<h3 id="summary-proposed-resolution">Proposed resolution</h3>
Remove our duplication

## Issue tracker
https://www.drupal.org/project/social/issues/3308595

## How to test
- [ ] Check that profile pages and previews are still correct

### Before merge
- [ ] Code/peer review is completed
- [ ] All commit messages are [clear and clean](https://open-social.slite.com/app/docs/DnmermZDIx_0OQ). If applicable a rebase was performed
- [ ] All automated tests are green
- [ ] Functional/manual tests of the acceptance criteria are approved
- [ ] All acceptance criteria were met
- [ ] New features or changes to existing features are covered by tests, either unit (preferably) or behat
- [ ] Update path is tested. New hook_updates should respect update order, right naming convention and consider hook_post_update code
- [ ] Module can be safely uninstalled. Update/implement hook_uninstall and make sure that removed configuration or dependencies are removed/uninstalled
- [ ] This pull request has all required labels (team/type/priority)
- [ ] This pull request has a milestone
- [ ] This pull request has an assignee (if applicable)
- [ ] Any front end changes are tested on all major browsers
- [ ] New UI elements, or changes on UI elements are approved by the design team
- [ ] New features, or feature changes are approved by the product owner

### After merge
- [ ] Code is tested on all branches that it has been cherry-picked
- [ ] Update hook number might need adjustment, make sure they have the correct order
- [ ] The Drupal.org ticket(s) are updated according to this pull request status

## Release notes
social_profile no longer duplicates theme hooks from the profile module.

PR #3164, opened 2022-10-20
Issue #3316226 by navneet0693, tbsiqueira: Username not showing on admin/people page

navneet0693:
## Problem
Username is not showing on admin/people page, this is affecting only Open Social 11.5.x and above.

In PR #3103, we removed the hook_theme() which caused the username to disappear.

See: https://github.com/goalgorilla/open_social/pull/3103#issuecomment-1284277075

## Solution
Add the hook_theme back.

## Issue tracker
https://www.drupal.org/project/social/issues/3316226

## Theme issue tracker
N.A

## How to test
- [ ] Install Open Social 11.5.x
- [ ] Navigate to admin/page
- [ ] You will see that the column "Username" is empty
- [ ] Checkout to this branch
- [ ] Clear cache
- [ ] Refresh the page
- [ ] You will see that the username is back.

### Before merge

### After merge

## Screenshots
**Before**

Screenshot 2022-10-20 at 11 33 01 AM

**After**

Screenshot 2022-10-20 at 11 33 09 AM

## Release notes
N.A

## Change Record
N.A

## Translations
N.A

### DIFF
--- a/modules/social_features/social_profile/social_profile.module
+++ b/modules/social_features/social_profile/social_profile.module
@@ -315,6 +315,19 @@ function social_profile_preprocess_form_element_label(&$variables) {
 }
 
 /**
+ * Implements hook_theme().
+ */
+function social_profile_theme(): array {
+  return [
+    'profile__profile__name' => [
+      'template' => 'profile--profile--name',
+      'render element' => 'elements',
+      'base hook' => 'profile',
+    ],
+  ];
+}
+
+/**
  * Implements hook_preprocess_HOOK().
  *
  * Allows modifying link attributes by modules instead of using hardcoded


Kingdutch:
So it looks like in #3103 I deleted a hook too many. The goal was to remove the alter hooks already added by the profile and nothing more. So in that sense removing `social_profile_theme` was a mistake.

However, rather than just placing back `social_profile_theme` you actually provide some interesting insights in your comment and we can look at why `social_profile_theme` was there in the first place and whether it's needed.

Can we figure out why we need the template in the first place?

From the original form of the template it looks like we were just using it to render a name. But as you pointed out in your comment, views should be perfectly capable of doing that for us.

Since then we've added a bit more to add the organization tag, which could maybe not be done through views directly. If that's the case then we would want to add this after.

However, rather than creating a new theme hook and hardcoding it in a template, the more flexible Drupal solution would be to see if we can fix this with a view mode, since this appears to be a way of showing Drupal users in a lot of places (i.e. is what we're showing here different than the name + organization tag on a teaser, and can't we reuse that?)

I don't have a preference for any solution. They depend on how much time we have now to fix the problem, what actually works given the constraints of the view, and how much value we attribute to being flexible here and using Drupal rather than just using Twig.

tbsiqueira (approved):

navneet0693:
🍒 -picked to 11.5.x

PR #3129, opened 2022-09-26
Issue #3311777 by hitesh-jain: Fix broken notifications to add account exists validation

hitesh-jain:
## Problem

Notifications are broken for the recipient users where the activity user owners are removed from the system manually causing notification loading issue on PHP 8.0 version.

`TypeError: Drupal\profile\ProfileStorage::loadByUser(): Argument #1 ($account) must be of type Drupal\Core\Session\AccountInterface, null given, called in /var/www/html/docroot/profiles/contrib/social/modules/custom/activity_creator/activity.page.inc on line 88 in Drupal\profile\ProfileStorage->loadByUser() (line 33 of /var/www/html/docroot/modules/contrib/profile/src/ProfileStorage.php)
`

## Solution
Add appropriate validation to check for User Account exists before trying to load that particular user account.

## Issue tracker
https://www.drupal.org/project/social/issues/3311777

### Before merge
- [ ] Code/peer review is completed
- [ ] All commit messages are [clear and clean](https://open-social.slite.com/app/docs/DnmermZDIx_0OQ). If applicable a rebase was performed
- [ ] All automated tests are green
- [ ] Functional/manual tests of the acceptance criteria are approved
- [ ] All acceptance criteria were met
- [ ] New features or changes to existing features are covered by tests, either unit (preferably) or behat
- [ ] Update path is tested. New hook_updates should respect update order, right naming convention and consider hook_post_update code
- [ ] Module can be safely uninstalled. Update/implement hook_uninstall and make sure that removed configuration or dependencies are removed/uninstalled
- [ ] This pull request has all required labels (team/type/priority)
- [ ] This pull request has a milestone
- [ ] This pull request has an assignee (if applicable)
- [ ] Any front end changes are tested on all major browsers
- [ ] New UI elements, or changes on UI elements are approved by the design team
- [ ] New features, or feature changes are approved by the product owner

### After merge
- [ ] Code is tested on all branches that it has been cherry-picked
- [ ] Update hook number might need adjustment, make sure they have the correct order
- [ ] The Drupal.org ticket(s) are updated according to this pull request status

## Screenshots
*[Required if new feature, and if applicable] If this Pull Request makes visual changes then please include some screenshots that show what has changed here. A before and after screenshot helps the reviewer determine what changes were made.*

## Release notes
*[Required if new feature, and if applicable] A short summary of the changes that were made that can be included in release notes.*

## Change Record
[...]

PR #3071, opened 2022-08-16
Issue #3303840 by navneet0693: Actions on event enrollment pages gets removed

navneet0693:
## Problem

The override given in https://github.com/goalgorilla/open_social/blob/11.1.x/modules/social_features/social_event/modules/social_event_an_enroll_enrolments_export/src/SocialEventAnEnrollEnrolmentsExportOverrides.php causes the pre-configured "actions" on the manage enrollment page to be removed once the `social_event_an_enroll_enrolments_export` module is enabled.

This is an un-required override as the necessary actions are already being performed in https://github.com/goalgorilla/open_social/blob/11.1.x/modules/social_features/social_event/modules/social_event_an_enroll_enrolments_export/social_event_an_enroll_enrolments_export.install#L17

So, we are removing this override. Along with that to differentiate in names on "Export" options in views page, we have re-named the anonymous export actions.

## Solution
1. Remove the override.

2. Update the name of actions in [open_social/ExportAllEnrolments.php at 11.1.x · goalgorilla/open_social](https://github.com/goalgorilla/open_social/blob/11.1.x/modules/social_features/social_event/modules/social_event_an_enroll_enrolments_export/src/Plugin/Action/ExportAllEnrolments.php#L20) to be more specific to “anonymous users” as it matches with [open_social/ExportEnrolments.php at 11.1.x · goalgorilla/open_social](https://github.com/goalgorilla/open_social/blob/11.1.x/modules/social_features/social_event/modules/social_event_enrolments_export/src/Plugin/Action/ExportEnrolments.php#L15)

## Issue tracker
https://www.drupal.org/project/social/issues/3303840

## Theme issue tracker
N.A

## How to test
**For existing installation**
- [ ] Using version "main" of Open Social with module `social_event_managers` enabled.
- [ ] Create an event and enroll few people.
- [ ] As an event manager, go to manage enrollment page.
- [ ] You should see the "Actions" button and pre-configured options of "Remove", "Export" and "Email"
- [ ] Enable `social_event_an_enroll_enrolments_export` module.
- [ ] Go back to the "manage enrollment" page of the event you created.
- [ ] You will see that the action of "Remove" and "Email" has gone.
- [ ] Checkout to this branch.
- [ ] Rebuild the cache containers.
- [ ] You should enable the "required" actions back.

**For new installation**
- [ ] Using version "this branch" of Open Social with module `social_event_managers` enabled.
- [ ] Create an event and enroll few people.
- [ ] As an event manager, go to manage enrollment page.
[...]

PR #3169, opened 2022-10-22
Issue #3316206 by ronaldtebrake: Enabling AN enrolments export install removes actions for managing enrollees

ronaldtebrake:
## Problem
On `hook_install()` of `social_event_an_enroll_enrolments_export` the module tries to replace the Action provided by `social_event_enrolments_export` because this module only takes care of users with actual Drupal accounts. 
The one provided by `social_event_an_enroll_enrolments_export` also takes care of AN enrollees.

However, in doing to, it incorrectly removed other actions because the logic of detecting which action should be replaced.

Screenshot 2022-10-19 at 11 08 49

You can see it in this screenshot above.
Because of it's multidemensional indexed array, instead of removing the array with index 2, it removes the one with key 2, which is exactly what we dont need.

You can see an example here of what is going wrong in an associative/indexed array when trying to remove the "green" value from the array. This is exactly what happened.

Screenshot 2022-10-19 at 11 38 01

## Solution
Two things:
- First we ensure it's now indexed before we remove it. This way we can be sure the position of the pointer is grabbed correctly.
- Additionally I've added the enabling of this module in to an existing test, it actually happens on install and thats all, that removes the Email selected attendeed action. You can see this fail when I run it in main:

Screenshot 2022-10-22 at 11 07 21
 
## Issue tracker
https://www.drupal.org/project/social/issues/3316206

## How to test
* In main branch*
- [x] Enable `social_event_an_enroll_enrolments_export`
- [x] As a sitemanager / Event Manager / Admin
- [x] Login, go to an event and visit the manage enrollments
- [x] See that the Actions incorrectly show two Exports and Email seems to be removed
* In this branch*
- [x] Reinstall Open Social and perform the exact same steps
- [x] See that now the correct actions are shown (Email / Export / Remove) users

### Before merge

### After merge

## Release notes
Enabling of the `social_event_an_enroll_enrolments_export` module, disabled / removed certain actions on "Event's -> Manage Enrollment" page. 

For existing websites, the options will have to be added manually by "administrators" by editing the view at `/admin/structure/views/view/event_manage_enrollments`. We didn't want to "add back" those actions by code as that might be un-required for many users. Also, it might have overridden their customizations.

For new installations, this won't be able a problem anymore.


ronaldtebrake:
The PHPStan issue is coming from #3129 
Can confirm it's all clear in my branch when that is solved

Screenshot 2022-10-22 at 11 25 53

navneet0693 (approved):

navneet0693:
@ronaldtebrake It works great.

**Before**

Screenshot 2022-10-24 at 8 16 57 AM


**After**

Screenshot 2022-10-24 at 8 20 48 AM

But I am wonder that our existing installations after https://github.com/goalgorilla/open_social/pull/3071 which landed in Open Social 11.3.9 will not be having this issue, but any new install will have this problem as you mentioned.

Is there anything we can do about our existing installations after Open Social 11.3.9 ?

ronaldtebrake:
Thats a good question @navneet0693 

With this code performed I can't be sure what update hook to write.
Imagine an additional action is added to the same array, it means instead of the export enrollees item maybe the email item got removed and I don't know what to put back. 

I'll have a think about what to do, if you have any ideas let me know!

navneet0693:
Certainly, that's a problem. Is it possible to add it back manually into views? If so, we can write release notes to do so via Views UI for 'admin' ? This will be safer option.

We also did the same for #3071 

Screenshot 2022-10-24 at 11 43 48 AM

ronaldtebrake:
That's a nice idea :) 

The alternative is that we would re-run update hooks in all modules that have these actions to ensure the action is put back, but even then it needs to be checked for all extensions as well indeed.

So i'll update the release notes section with that 👍

navneet0693:
🍒 - picked to 11.5.x, 11.4.x and 11.3.x

PR #3910, opened 2024-05-28
Hotfix: inline images converted to secret file system

Kingdutch:
## Problem / Solution
Due to caching issues in the CKEditor, inline images added through the CKEditor were omitted from using the secret filesystem. However in the update path those images from the `file_managed` table were not excluded. This has caused `social_core_update_130002` to say that inline images that should be served by the private file system should be served by the secret file system. However, this is then invalid, causing issues.

This commit adds an update hook that reverts the incorrectly changed inline images. We can be sure that anything that matches was coming from private, since we didn't touch public images.

### DIFF
--- a/modules/social_features/social_core/social_core.install
+++ b/modules/social_features/social_core/social_core.install
@@ -495,3 +495,14 @@ function social_core_update_130004(): void {
 function social_core_update_130005(): void {
   user_role_grant_permissions('sitemanager', ['administer taxonomy']);
 }
+
+/**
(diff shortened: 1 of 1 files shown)


open-social-tugboat:

Link:
- https://pr3910-3lcx6fambjg0t4yudjzxro2wacz292qr.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6655f64a352ceca600406b98

ronaldtebrake (approved):

PR #3942, opened 2024-06-20
PROD-29703: Provide external identifier field

zanvidmar:
## Problem
As a developer I want a re-usable externalId implementation so that I can easily apply this to different entity types.

SCIM provides [the concept of externalId](https://datatracker.ietf.org/doc/html/rfc7643#section-3.1) which can be used to identify a resource with an identifier that is known to the SCIM client. This prevents the SCIM client from having to store a mapping between its internal data and our ID, instead putting that burden on us as a service provider.

Although not covered by SCIM itself, it provides various [recommendations for multi-tenancy](https://datatracker.ietf.org/doc/html/rfc7644#section-6) where multiple external systems may want to manage their own user data. Given our intention of targeting Enterprise users it’s likely that they’ll want to connect multiple systems to Open Social (e.g. managing users within their organization). We’ll want to combine this with SCIM’s [privacy recommendations](https://datatracker.ietf.org/doc/html/rfc7643#section-9.3) and determine how we want to handle the externalId.

We’ll want to decide whether we want to support multiple external systems managing the same user. This may be desired for migrations or when we need to pull data from multiple systems that might overlap and want to de-duplicate users. We use this ourselves in our Private Packagist where a developer can login with either Bitbucket or GitHub and user groups are managed through both (until we fully move to GitHub).

Regardless of our support for shared management of resources we’ll want to make sure that our externalId is tied to the system that created the user. This ensures that if multiple systems can only manage their own users they don’t run into conflicts (e.g. because they both have an externalId of user:5 which represent their own 5th users).

**Acceptance Criteria**
- An externalId field type which has sub-data of
  - external_id (string)
  - ~external_owner (entity_reference → blank by default)~
  - external_owner_target (string)
  - external_owner_id (int)
- Multi-valued (cardinality unlimited)
- Optional ~(default NULL)~ (default is empty, no record in database)
- The field type exists in the distribution
- When SCIM is enabled external_owner reference config is changed to allow consumer as target (this can be configured via field storage configuration)
- The field type is added to the user entity in ~the distribution~ enterprise version of Open Social.
- It’s easy to re-use this field on other entity types (e.g. group or organization)

## Solution
- Create new field type with 3 subfields:
  - External ID (external_id, string)
  - External Owner Target type (external_owner_target_type, string)
  - External Owner ID (external_owner_id, integer)
- In the field storage settings, it can be determined which entity types can be referenced.
- Field provides method `getExternalOwnerEntity()`, which returns an entity object based on External Owner Target type and External Owner ID

Decision records why we did not used dynamic entity reference / core entity reference field are in PROD-29703

## Issue tracker
- [3457487](https://www.drupal.org/project/social/issues/3457487)
- PROD-29703
- IOSAP-457

## How to use field
- Add external field type to any entity as any other regular field
- In field storage settings define which entities can be referenced as external owners
- in case that field values are managed programatically, do not forget to use `validate()` method on field/entity before saving, because most of the business logic needed for correct behaviour is validated with constraints and constraints are not triggered without field/entity validation.

## Theme issue tracker
N/A

## How to test
- [ ] As a developer I should be able to apply/use this field on any entity type.
- [ ] As a developer I should be able to apply single or multiple external ides per entity type.
- [ ] As a developer I should be able to select which target entity types (external owners) are supported per each field.
- [ ] As a developer I should be able to use this field via UI or as hidden field.
- [ ] As a user or developer I should be able to save this fields with all subfields values empty
- [ ] As a user or developer I should not be able to save this field if any subfield value is missing (except if all are empty)

## Screenshots
Screenshot 2024-06-27 at 08 18 41
Screenshot 2024-06-27 at 08 18 52
Screenshot 2024-06-27 at 08 19 29
Screenshot 2024-06-27 at 08 19 50

## Release notes
Provides re-usable externalId implementation, that can be easily applied to different entity types.

## Change Record
N/A

## Translations
N/A

### DIFF
--- a/modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
+++ b/modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
@@ -1,0 +1,227 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\social_core\Plugin\Field\FieldType;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Field\FieldItemBase;
+use Drupal\Core\Field\FieldStorageDefinitionInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\Core\TypedData\DataDefinition;
+use Drupal\Core\TypedData\DataReferenceTargetDefinition;
+
+/**
+ * Defines the 'social_external_identifier' field type.
+ *
+ * Supported settings (below the definition's 'settings' key) are:
+ * - target_type: The entity type to reference. Required.
+ *
+ * @FieldType(
+ *   id = "social_external_identifier",
+ *   label = @Translation("External Identifier"),
+ *   description = @Translation("Define external Identifier and external Owner/Consumer."),
+ *   default_widget = "social_external_identifier_default_widget",
+ *   default_formatter = "social_external_identifier_default_formatter",
+ *   constraints = {
+ *     "ExternalIdentifierEmptySubfieldsConstraint" = {},
+ *     "ExternalIdentifierExternalOwnerTargetTypeConstraint" = {},
+ *     "ExternalIdentifierExternalOwnerIdConstraint" = {},
+ *     "ComplexData" = {
+ *       "external_id" = {
+ *         "Length" = {"max" = 225}
+ *        }
+ *      }
+ *   }
+ * )
+ */
+final class ExternalIdentifierItem extends FieldItemBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isEmpty(): bool {
+    // Business rules:
+    // - Field is considered empty if all subfield values are empty.
+    // - Field is considered invalid (by constraints) if subfield values are
+    // just partly provided (except if all subfield values are empty).
+    // - Field is considered valid if all subfield values are provided (but it
+    // is also valid if none of the subfield value is provided)
+    //
+    // Subfields:
+    // - external_id,
+    // - external_owner_target_type
+    // - external_owner_id
+    //
+    // Note, even that external_owner_id is int, we are checking is as string
+    // because of some Drupal bugs that int values ane not matching type hints.
+    // See: https://www.drupal.org/project/drupal/issues/3441689
+    // See: https://www.drupal.org/project/drupal/issues/3224376
+    //
+    // Also note that the ExternalIdentifierEmptySubfieldsConstraintValidator
+    // prevents the field from being saved if all subfields do not have set
+    // values.
+    foreach ($this->getValue() as $value) {
+      if ($value !== '' && $value !== NULL) {
+        return FALSE;
+      }
+    }
+
+    return TRUE;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function propertyDefinitions(FieldStorageDefinitionInterface $field_definition): array {
+    $properties['external_id'] = DataDefinition::create('string')
+      ->setLabel(t('External ID'))
+      ->setRequired(TRUE);
+    $properties['external_owner_target_type'] = DataReferenceTargetDefinition::create('string')
+      ->setLabel(t('Target Entity Type'))
+      ->setRequired(TRUE);
+    $properties['external_owner_id'] = DataDefinition::create('integer')
+      ->setLabel(t('External Owner'))
+      ->setRequired(TRUE);
+
+    return $properties;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function storageSettingsSummary(FieldStorageDefinitionInterface $storage_definition): array {
+    $summary = [];
+    $target_types = $storage_definition->getSetting('target_types');
+    if (!empty($target_types)) {
+      $entity_type_labels = [];
+      foreach ($target_types as $target_type) {
+        $target_type_info = \Drupal::entityTypeManager()->getDefinition($target_type);
+        if (!empty($target_type_info)) {
+          $entity_type_labels[] = $target_type_info->getLabel();
+        }
+      }
+      $summary[] = new TranslatableMarkup('Reference type: @entity_type_labels', [
+        '@entity_type_labels' => implode(', ', $entity_type_labels),
+      ]);
+
+    }
+
+    return $summary;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function schema(FieldStorageDefinitionInterface $field_definition): array {
+    return [
+      'columns' => [
+        'external_id' => [
+          'type' => 'varchar',
+          'description' => 'External ID',
+          'length' => 255,
+          'not null' => TRUE,
+        ],
+        'external_owner_target_type' => [
+          'type' => 'varchar',
+          'description' => 'External Owner Target Type',
+          'length' => 255,
+          'not null' => TRUE,
+        ],
+        'external_owner_id' => [
+          'type' => 'int',
+          'description' => 'External Owner ID',
+          'unsigned' => TRUE,
+          'not null' => TRUE,
+        ],
+      ],
+      'indexes' => [
+        'external_id' => ['external_id'],
+        'external_owner_target_type' => ['external_owner_target_type'],
+        'external_owner_id' => ['external_owner_id'],
+      ],
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setValue($values, $notify = TRUE): void {
+    if (is_array($values)) {
+      $this->set('external_id', $values['external_id'] ?? NULL);
+      $this->set('external_owner_target_type', $values['external_owner_target_type'] ?? NULL);
+      $this->set('external_owner_id', $values['external_owner_id'] ?? NULL);
+    }
+    else {
+      parent::setValue($values, $notify);
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultStorageSettings() {
+    return [
+      'target_types' => [],
+    ] + parent::defaultStorageSettings();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function storageSettingsForm(array &$form, FormStateInterface $form_state, $has_data) {
+    $element['target_types'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Allowed external owner target types'),
+      '#description' => $this->t('Entity type of item to reference (Example: consumer)'),
+      '#default_value' => $this->getSetting('target_types'),
+      '#multiple' => TRUE,
+      // For field to work properly, at least one target type must be available,
+      // as this is hard requirement, If requirement can not be fulfilled, field
+      // will return an error if non-empty value will try to be applied.
+      // Despite all those facts, we still keep this option as optional, because
+      // by default, no target types are predefined.
+      '#required' => FALSE,
+      '#disabled' => $has_data,
+      '#size' => 10,
+    ];
+
+    // Only allow the field to target entity types that have an ID key. This
+    // is enforced in ::propertyDefinitions().
+    $entity_type_manager = \Drupal::entityTypeManager();
+    $filter = function (string $entity_type_id) use ($entity_type_manager): bool {
+      /** @var \Drupal\Core\Entity\EntityTypeInterface $entity_type */
+      $entity_type = $entity_type_manager->getDefinition($entity_type_id);
+      assert($entity_type instanceof EntityTypeInterface, "EntityTypeManager::getDefinition should throw an exception if the entity type does not exist rather than returning NULL");
+      return $entity_type->hasKey('id');
+    };
+    $options = \Drupal::service('entity_type.repository')->getEntityTypeLabels(TRUE);
+    foreach ($options as $group_name => $group) {
+      $element['target_types']['#options'][$group_name] = array_filter($group, $filter, ARRAY_FILTER_USE_KEY);
+    }
+    return $element;
+  }
+
+  /**
+   * Loads the external owner entity.
+   *
+   * @return \Drupal\Core\Entity\EntityInterface|null
+   *   Returns external owner entity if exists.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   * @throws \Drupal\Core\TypedData\Exception\MissingDataException
+   */
+  public function getExternalOwnerEntity(): ?EntityInterface {
+    $entity_type = $this->get('external_owner_target_type')->getValue();
+    $entity_id = $this->get('external_owner_id')->getValue();
+
+    if ($entity_type && $entity_id) {
+      return \Drupal::entityTypeManager()->getStorage($entity_type)->load($entity_id);
+    }
+    return NULL;
+  }
+
+}
--- a/modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierEmptySubfieldsConstraint.php
+++ b/modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierEmptySubfieldsConstraint.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Drupal\social_core\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Provides a custom constraint for external identifier field.
+ *
+ * Prevents field values from being saved if not all required subfield values
+ * are set. This can be either that all subfield values are empty, or that all
+ * subfield values are provided but partially filled subfield values are not
+ * allowed.
+ *
+ * @Constraint(
+ *   id = "ExternalIdentifierEmptySubfieldsConstraint",
+ *   label = @Translation("Makes sure that all required subfields values are set", context = "Validation"),
+ *   type = {"field"}
+ * )
+ */
+class ExternalIdentifierEmptySubfieldsConstraint extends Constraint {
+
+  /**
+   * The error message when all required subfield values are not set.
+   *
+   * @var string
+   */
+  public $requiredSubfieldsAreNotSet = 'Not all required subfields have been set. Please insert values for: %empty_required_subfield_list.';
+
+}
--- a/modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierExternalOwnerIdConstraint.php
+++ b/modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierExternalOwnerIdConstraint.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Drupal\social_core\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Provides a custom constraint for external identifier field.
+ *
+ * Prevents field values from being saved if the external owner entity does not
+ * exist.
+ *
+ * @Constraint(
+ *   id = "ExternalIdentifierExternalOwnerIdConstraint",
+ *   label = @Translation("Makes sure that External Owner Entity exist", context = "Validation"),
+ *   type = {"field"}
+ * )
+ */
+class ExternalIdentifierExternalOwnerIdConstraint extends Constraint {
+  /**
+   * The error message when entity does not exist.
+   *
+   * @var string
+   */
+  public $nonexistentIdMessage = 'The entity of type "%entity_type" and ID "%entity_id" does not exist.';
+
+}
--- a/modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierExternalOwnerTargetTypeConstraint.php
+++ b/modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierExternalOwnerTargetTypeConstraint.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Drupal\social_core\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Provides a custom constraint for external identifier field.
+ *
+ * Prevents field values from being stored if the referenced target type is not
+ * on the list of allowed ones. The list of allowed ones is defined by the field
+ * storage configuration and may contain more than one allowed target type.
+ *
+ * @Constraint(
+ *   id = "ExternalIdentifierExternalOwnerTargetTypeConstraint",
+ *   label = @Translation("Makes sure that target type exist and is allowed", context = "Validation"),
+ *   type = {"field"}
+ * )
+ */
+class ExternalIdentifierExternalOwnerTargetTypeConstraint extends Constraint {
+
+  /**
+   * The error message when target type is not valid.
+   *
+   * @var string
+   */
+  public $invalidTargetTypeMessage = 'Target type "%invalid_target_type" is not valid. Valid target types are: "%allowed_target_types".';
+
+  /**
+   * The error message when target type does not exist.
+   *
+   * @var string
+   */
+  public $nonexistentTargetTypeMessage = 'The entity type "%entity_type" does not exist.';
+
+  /**
+   * The error message, when there are no target type is supported.
+   *
+   * @var string
+   */
+  public $noAvailableTargetTypes = 'Currently, there are no available target types (allowed entity types). Please contact the system administrator to enable at least one target type.';
+
+}
--- a/modules/social_features/social_core/tests/modules/social_core_test_entity_provider/social_core_test_entity_provider.info.yml
+++ b/modules/social_features/social_core/tests/modules/social_core_test_entity_provider/social_core_test_entity_provider.info.yml
@@ -1,0 +1,8 @@
+name: "Social Core Entity Provider"
+type: module
+description: "A provider of an entity type."
+package: Testing
+core_version_requirement: ^10 || ^11
+hidden: TRUE
+dependencies:
+  - drupal:entity_test
--- a/modules/social_features/social_core/tests/src/Kernel/ExternalIdFieldTypeTest.php
+++ b/modules/social_features/social_core/tests/src/Kernel/ExternalIdFieldTypeTest.php
@@ -1,0 +1,399 @@
+<?php
+
+namespace Drupal\Tests\social_core\Kernel;
+
+use Drupal\Core\Field\FieldStorageDefinitionInterface;
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\node\Entity\Node;
+use Drupal\node\Entity\NodeType;
+use Drupal\social_core_test_entity_provider\Entity\ExternalOwnerEntityTestProvider;
+
+/**
+ * Tests the social_external_identifier field type.
+ *
+ * @group social_external_identifier
+ */
+class ExternalIdFieldTypeTest extends KernelTestBase {
+
+  /**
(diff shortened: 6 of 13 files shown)


open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierEmptySubfieldsConstraint.php
```diff
@@ -0,0 +1,25 @@
...
+ * @Constraint(
+ *   id = "ExternalIdentifierEmptySubfieldsConstraint",
+ *   label = @Translation("Makes sure that all required subfields values are set", context = "Validation"),
+ *   type = {"field"}
+ * )
+ */
+class ExternalIdentifierEmptySubfieldsConstraint extends Constraint {
+
+  /**
+   * The error message when all required subfields values not are set.
```
tbsiqueira, line 19: Small grammar suggestion: 

`The error message when all required subfield values are not set`

modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierEmptySubfieldsConstraint.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Drupal\social_core\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Provides a custom constraint for fields.
```
tbsiqueira, line 8: We should a better explanation for this class.

modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierExternalOwnerIdConstraint.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Drupal\social_core\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Provides a custom constraint for fields.
```
tbsiqueira, line 8: We should a better explanation for this class.

modules/social_features/social_core/src/Plugin/Validation/Constraint/ExternalIdentifierExternalOwnerTargetTypeConstraint.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Drupal\social_core\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Provides a custom constraint for fields.
```
tbsiqueira, line 8: We should a better explanation for this class.

tbsiqueira (changes requested):

modules/social_features/social_core/social_core.module
```diff
@@ -5,22 +5,24 @@
  * The Social core module.
  */
 
+use Drupal\block\BlockInterface;
```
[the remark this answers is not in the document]
zanvidmar: It was only the change of order by automatic phpcs fix => https://github.com/goalgorilla/open_social/pull/3942/files/70c007107d6d6931e0dbf3a4a54a8f53769f29a3#diff-85dda32ba341a00ea6ac3524f7ff8f964fdaba3370c6f9604b910b77ba69b41bL33

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
```diff
@@ -0,0 +1,221 @@
...
+   */
+  public static function propertyDefinitions(FieldStorageDefinitionInterface $field_definition): array {
+    $properties['external_id'] = DataDefinition::create('string')
+      ->setLabel(t('External ID'))
+      ->setRequired(TRUE);
+    $properties['external_owner_target_type'] = DataReferenceTargetDefinition::create('string')
+      ->setLabel(t('Target Entity Type'))
+      ->setRequired(TRUE);
+    // @todo See if we can improve this subfield to use
+    //   DataReferenceDefinition or DataDynamicReferenceDefinition.
```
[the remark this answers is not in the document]
zanvidmar: short answer, we might want to improve this in future,
long answer, decision records https://open-social.slite.com/app/docs/pA97W-b4Z7PkiK

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
```diff
@@ -0,0 +1,221 @@
...
+ *     "ExternalIdentifierExternalOwnerIdConstraint" = {},
+ *     "ComplexData" = {
+ *       "external_id" = {
+ *         "Length" = {"max" = 225}
+ *        }
+ *      }
+ *   }
+ * )
+ */
+final class ExternalIdentifierItem extends FieldItemBase {
```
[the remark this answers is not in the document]
zanvidmar: nice catch, PR description fixed

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
```diff
@@ -0,0 +1,221 @@
...
+      ->setRequired(TRUE);
+
+    return $properties;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function storageSettingsSummary(FieldStorageDefinitionInterface $storage_definition): array {
+    $summary = parent::storageSettingsSummary($storage_definition);
```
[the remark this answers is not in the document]
zanvidmar: this is actually redundant since we are not extending anything now in this field type - will remove it.

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

modules/social_features/social_core/social_core.module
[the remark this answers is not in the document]
tbsiqueira: Let's revert this, it's not part of the PR and we should not commit this together.

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isEmpty(): bool {
+    // If at least one of 3 subfields is not empty, field is not empty.
+    // Subfields are:
+    // - external_id,
+    // - external_owner_target_type
+    // - external_owner_id
```
Kingdutch, line 49: From a logic point of view I'm wondering if this is correct. Especially if we only expect the field to be valid when all 3 values are filled in, then shouldn't the field be considered empty if any of these values is empty (basically the inverse of the current logic)?

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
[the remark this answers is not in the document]
Kingdutch: I think a `@todo` should only survive a merge if it's linked to some Jira ticket so that we have an idea of what needs to happen and there's a non-zero chance it'll get picked up. The Jira issue is also a good place to point to the ADR and describe what/why we need this.

If we have a `@todo` in the code that's not linked to something outside of our code-base then it's likely to just sit there forever and only puzzle archeologists in the future :)

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setValue($values, $notify = TRUE): void {
+    if (is_array($values)) {
+      $this->set('external_id', $values['external_id'] ?? NULL);
+      $this->set('external_owner_target_type', $values['external_owner_target_type'] ?? NULL);
+      $this->set('external_owner_id', $values['external_owner_id'] ?? NULL);
```
Kingdutch, line 150: I realise we have a constraint that checks that all these fields are set later. However, I wonder if from a developer experience point of view it wouldn't be better to throw an `InvalidArgumentException` if any of those fields are not set.

For example, would we find an implementation that only sets some of those fields a correct implementation, or would we consider that to be a bug by whoever is using this field?

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
```diff
@@ -0,0 +1,221 @@
...
+      '#disabled' => $has_data,
+      '#size' => 10,
+    ];
+
+    // Only allow the field to target entity types that have an ID key. This
+    // is enforced in ::propertyDefinitions().
+    $entity_type_manager = \Drupal::entityTypeManager();
+    $filter = function (string $entity_type_id) use ($entity_type_manager): bool {
+      /** @var \Drupal\Core\Entity\EntityTypeInterface $entity_type */
+      $entity_type = $entity_type_manager->getDefinition($entity_type_id);
```
Kingdutch, line 191: Prefer using an `assert` statement which will fail in tests and more clearly demonstrate that if this doesn't happen there's a violation of base assumptions about the library we're using.

```suggestion
      /** @var \Drupal\Core\Entity\EntityTypeInterface $entity_type */
      $entity_type = $entity_type_manager->getDefinition($entity_type_id);
      assert($entity_type instanceof EntityTypeInterface, "EntitTypeManager::getDefinition should throw an exception if the entity type does not exist rather than returning NULL");
```

modules/social_features/social_core/tests/modules/social_core_test_entity_provider/social_core_test_entity_provider.info.yml
```diff
@@ -0,0 +1,9 @@
+name: "Social Core Entity Provider"
+type: module
+description: "A provider of an entity type."
+package: Testing
+core_version_requirement: ^10 || ^11
+php: 8.1
```
Kingdutch, line 6: I'm wondering if we should specify a php version here. I think it's used by Drupal's packager to put it into the composer.json but we don't use that. If we do have PHP requirements we should likely specify that in our `composer.json` directly instead :)

modules/social_features/social_core/tests/src/Kernel/ExternalIdFieldTypeTest.php
```diff
@@ -0,0 +1,325 @@
...
+   * Tests that the field can be added to a node and stores values correctly.
+   *
+   * TOC:
+   *  - Case 1: Create a node with the external identifier field.
+   * - Case 2: Create a node without the external identifier field.
+   * - Case 3: Create a node with the external identifier field with subfield
+   * values set as NULL.
+   * - Case 4: Create a node with the external identifier field with subfield
+   * values set as empty string.
+   */
```
Kingdutch, line 114: I think it's great that you've already split these into separate cases. Is there a reason that they're not actually separate tests? If a developer has to debug this then it's very difficult to know whether case 2 or case 3 has broken and thus where to start looking for errors. 

A data provider may help if the structure of the test is the same but the inputs are different: https://blog.martinhujer.cz/how-to-use-data-providers-in-phpunit/

Also don't be afraid to have some duplicated set-up in your tests if it helps a failure provide more information :) 

The same goes for your other test which described multiple cases.

Kingdutch (commented):
Looking good @zanvidmar ! I had a few questions about some business logic and the organisation of the tests which you can find below, but overall really great job 🙌

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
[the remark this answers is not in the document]
zanvidmar: Ok based on your comment @Kingdutch I will remove this todo and if we want to improve this we still have ADR where everything is documented.

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
zanvidmar: Actually no :) 
- Field is considered empty if all values are empty
- Field is considered invalid not all values are are provided (except if all values are empty) - to reframe (if values are partly provided, field is considered as invalid (by constraints).
- Field is considered valid if all values are provided (but it is also valid if non of the value is provided)

Let me know if this is better explained now.

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
zanvidmar: > I realise we have a constraint that checks that all these fields are set later.
Yes indeed

We could already throw `InvalidArgumentException` at `setValue` but IMHO constraints are IMHO more appropriate place to do it.

modules/social_features/social_core/tests/modules/social_core_test_entity_provider/social_core_test_entity_provider.info.yml
zanvidmar: I think we can remove this.

modules/social_features/social_core/tests/src/Kernel/ExternalIdFieldTypeTest.php
zanvidmar: Thank you for the feedback, will split them 💪

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
Kingdutch: I think it depends on what we want to achieve and how Drupal's constraint system fits into other data integrity decisions.

In my understanding, Drupal's constraint system is mostly there to help user input and runs only when `validate` is called on an entity. This means that programmatically we might still be able to create a field item that's invalid just because validation is bypassed. The question is then whether we want to allow this, or whether we want to prevent it and let a developer know "Hey you forgot to validate your inputs, that's not right!"

Throwing an `InvalidArgumentException` is one way to let a developer know their implementation is doing something wrong that they should change. So it depends on when `setValue` is called and where that exception ends up. If throwing an exception in `setValue` would mean that when a user submits an invalid form they no longer get a form validation error but instead the application crashes, then throwing an exception here is not the right answer. In that case we may want to think about "How do we warn a developer that they're trying to save an invalid field value because they (intentionally) skipped the validate step" and solve that problem.

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
Kingdutch: That makes sense :D I think that ties in to the other thread we're discussing and if we have clarity there then I believe we cover all cases. Thanks!

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
tbsiqueira: @Kingdutch , I just had a discussion with Zan where he explained that even if they are three different subfields, they come as a package, so we can't update just one part of that field, or have one NULL value for one of those subs. 

We can consider a wrong implementation/bug if we have a situation like that.

zanvidmar: @Kingdutch 

> In my understanding, Drupal's constraint system is mostly there to help user input and runs only when validate is called on an entity. This means that programmatically we might still be able to create a field item that's invalid just because validation is bypassed. The question is then whether we want to allow this, or whether we want to prevent it and let a developer know "Hey you forgot to validate your inputs, that's not right!"

Yes we also had this discussion with @tbsiqueira and my argument here is: we should never let any entity being saved without being validated (except if you really know what you are doing and accept all consequences of doing it so). I also believe that validate method is overlooked many times, so I understand your concerns, but still `setValue` nor `isEmpty` methods as their name says, are not places to validate stuff (IMHO). 
If we really want to prevent errors if someone forget to do `->validate()` than we basically need to put all constraints into `setValue` method (and I am still not suggesting this :D). 

The whole business logic relies on correct data (ATM handled by constraints), so throwing an InvalidArgumentException if any of those subfields are not set is basically just one piece of warning if someone forget to validate, and this will still not prevent errors and application will still crash if invalid data will be saved without validation. :)

> "How do we warn a developer that they're trying to save an invalid field value because they (intentionally) skipped the validate step" and solve that problem.

Can we somehow do automatic check for this - so that this is rather part of code quality than how we build stuff :)

modules/social_features/social_core/tests/modules/social_core_test_entity_provider/social_core_test_entity_provider.info.yml
zanvidmar: removed, marking as resolved.

modules/social_features/social_core/tests/src/Kernel/ExternalIdFieldTypeTest.php
zanvidmar: All split. I did not used data providers as all tests are different. I kept the TOC as we have 10 kernel tests methods now ;) Marking as resolved.

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
zanvidmar: I applied your suggestion and fix the small typo. Marking as resolved.

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
zanvidmar: Ok I will update my explanation in the code with bullet points I've put in my previous comment, but let's resolve the other thread first ;)

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

modules/social_features/social_core/src/Plugin/Field/FieldType/ExternalIdentifierItem.php
Kingdutch: > we should never let any entity being saved without being validated (except if you really know what you are doing and accept all consequences of doing it so). I also believe that validate method is overlooked many times, so I understand your concerns

I think in general that follows Drupal's own rules around validation. However, there's still valid cases where you want to bypass it (for example in one of our recent feature implementations for partial entities; or during migrations). So if we can help developers not forget the right thing (without eliminating their options) then we should try to do that.

> application will still crash if invalid data will be saved without validation

For me it's not necessarily about "preventing the application from crashing". If a developer implements a bug then that's one of the outcomes that may be unavoidable. However, our aim of exceptions should be that if we know a developer may make a mistake, we make it as easy as possible to know what that mistake was an how to fix them. So that's why I'm looking at "What happens if a developer doesn't validate and how difficult might it be to debug for them."

A contrived example would be something like:
```
/**
  * @param int[] $grades
  */
function averageGrades(array $grades) : float {
  return array_sum($grades) / count($grades);
}
```

We don't need to worry that a developer is passing a string because PHPStan will help us with that. However, in this case an empty array of grades would be invalid and would lead to the program crashing with a "division by 0" error, which might confuse the developer. To make debugging easier (because the division by 0 is not the actual problem) we can do:

```
/**
  * @param int[] $grades
  */
function averageGrades(array $grades) : float {
  if ($grades === []) {
     throw \InvalidArgumentException('An average can not be calculated without grades. $grades may not be an empty array.');
  }
  return array_sum($grades) / count($grades);
}
```

This has the same result if the developer does nothing: the program crashes with an exception. However, the information that the developer has is a lot easier to process.

--------

For me the rule of thumb that I try to use is: "Will not doing anything here have known errors down the line for certain inputs or will it cause unexpected behavior such as silently discarding a value". If the answer is "Yes" then we should try to capture those scenario's earlier and provide useful developer feedback. If the answer is "No" then we're all good and don't need to do anything else :)

zanvidmar: As discussed with @Kingdutch, in case developer save entity without `validate()` method, there is no good approach to inform the developer about mistake. In the setValue, we could check for constraints by calling `$this->validate()`, but if we would throw an exception we would break UI form. 

We also check how Drupal core is tackling this issue, but basically if you are not validating data, more or less, Drupal does not care and not trying prevent errors in this cases.

The only way (that I can think of now) if we need to throw an error for developer is to implement `hook_entity_presave` and trigger error there, by checking if validation was triggered or not - note via UI it will always trigger with validation. In case we will want to implement `hook_entity_presave` with described logic, we will do this in separated story/PR.

By this comment, I am resolving current conversation.

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

open-social-tugboat:

Link:
- https://pr3942-wfadybap5xevu4akgpd4nqes5mjtvfay.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/66740a2181747e9dde32f928

tbsiqueira (approved):

PR #4118, opened 2024-10-16
PROD-30967: Hux implementation

denis-getopensocial:
## Problem

As outlined in our decision record, we have identified the need to integrate Hux instead of relying on traditional hooks for managing interactions and events within our software. The use of hooks, while functional, has led to limitations in scalability, maintainability, and flexibility. 

## Solution

We will migrate from the current hook-based implementation to Hux. The Hux library offers a more modern and modular approach for handling events, making the system easier to maintain and extend in the long term.
Blog post on how to: https://www.previousnext.com.au/blog/hux-alternative-to-hooks
Drupal seems to go the same way: https://www.drupal.org/node/3442349

## Issue tracker

https://getopensocial.atlassian.net/browse/PROD-30967

## Change Record

Creation of a `current group service` to fetch the current group service from the context.

### DIFF
--- a/modules/social_features/social_group/social_group.services.yml
+++ b/modules/social_features/social_group/social_group.services.yml
@@ -49,3 +49,9 @@ services:
     arguments: []
     tags:
       - { name: cache.context }
+
+  social_group.current_group:
+    class: Drupal\social_group\CurrentGroupService
+    arguments:
+      - '@entity_type.manager'
+      - '@group.group_route_context'
--- a/modules/social_features/social_group/social_group.services.yml
+++ b/modules/social_features/social_group/social_group.services.yml
@@ -51,6 +51,5 @@ services:
       - { name: cache.context }
 
   Drupal\social_group\CurrentGroupService:
-    class: Drupal\social_group\CurrentGroupService
+    social_group.current_group_service: '@Drupal\social_group\CurrentGroupService'
     autowire: true
-    social_group.current_group: '@Drupal\social_group\CurrentGroupService'
--- a/modules/social_features/social_group/social_group.services.yml
+++ b/modules/social_features/social_group/social_group.services.yml
@@ -50,8 +50,7 @@ services:
     tags:
       - { name: cache.context }
 
-  social_group.current_group:
+  Drupal\social_group\CurrentGroupService:
     class: Drupal\social_group\CurrentGroupService
-    arguments:
-      - '@entity_type.manager'
-      - '@group.group_route_context'
+    autowire: true
+    social_group.current_group: '@Drupal\social_group\CurrentGroupService'
--- a/modules/social_features/social_group/src/CurrentGroupService.php
+++ b/modules/social_features/social_group/src/CurrentGroupService.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Drupal\social_group;
+
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Plugin\Context\ContextProviderInterface;
+use Drupal\group\Entity\GroupInterface;
+
+/**
+ * Centralized ways to get the current group.
+ *
+ * @package Drupal\social_group.
+ */
+class CurrentGroupService {
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  private EntityTypeManagerInterface $entityTypeManager;
+
+  /**
+   * The group route context.
+   *
+   * @var \Drupal\Core\Plugin\Context\ContextProviderInterface
+   */
+  private ContextProviderInterface $groupRouteContext;
+
+  /**
+   * Constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   * @param \Drupal\Core\Plugin\Context\ContextProviderInterface $group_route_context
+   *   The group route context.
+   */
+  public function __construct(
+        EntityTypeManagerInterface $entity_type_manager,
+        ContextProviderInterface $group_route_context,
+    ) {
+    $this->entityTypeManager = $entity_type_manager;
+    $this->groupRouteContext = $group_route_context;
+  }
+
+  /**
+   * Get group from runtime contexts.
+   *
+   * @return \Drupal\group\Entity\GroupInterface|null
+   *   The current group or NULL.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  public function fromRunTimeContexts(): ?GroupInterface {
+    $runtime_context = $this->groupRouteContext->getRuntimeContexts([]);
+    if (isset($runtime_context['group']) === FALSE) {
+      return NULL;
+    }
+
+    $group = $runtime_context['group']->getContextData()->getValue();
+    if ($group instanceof GroupInterface) {
+      return $group;
+    }
+
+    if (is_int($group) === TRUE) {
+      /** @var \Drupal\group\Entity\GroupInterface $loadedGroup */
+      $loadedGroup = $this->entityTypeManager
+        ->getStorage('group')->load($group);
+      return $loadedGroup;
+    }
+
+    return NULL;
+  }
+
+}
--- a/modules/social_features/social_group/src/CurrentGroupService.php
+++ b/modules/social_features/social_group/src/CurrentGroupService.php
@@ -2,8 +2,7 @@
 
 namespace Drupal\social_group;
 
-use Drupal\Core\Entity\EntityTypeManagerInterface;
-use Drupal\Core\Plugin\Context\ContextProviderInterface;
+use Drupal\Core\Plugin\Context\ContextRepositoryInterface;
 use Drupal\group\Entity\GroupInterface;
 
 /**
@@ -14,33 +13,22 @@ use Drupal\group\Entity\GroupInterface;
 class CurrentGroupService {
 
   /**
-   * The entity type manager.
+   * The context repository interface.
    *
-   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   * @var \Drupal\Core\Plugin\Context\ContextRepositoryInterface
    */
-  private EntityTypeManagerInterface $entityTypeManager;
-
-  /**
-   * The group route context.
-   *
-   * @var \Drupal\Core\Plugin\Context\ContextProviderInterface
-   */
-  private ContextProviderInterface $groupRouteContext;
+  private ContextRepositoryInterface $contextRepository;
 
   /**
    * Constructor.
    *
-   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
-   *   The entity type manager.
-   * @param \Drupal\Core\Plugin\Context\ContextProviderInterface $group_route_context
-   *   The group route context.
+   * @param \Drupal\Core\Plugin\Context\ContextRepositoryInterface $context_repository
+   *   The context repository.
    */
   public function __construct(
-        EntityTypeManagerInterface $entity_type_manager,
-        ContextProviderInterface $group_route_context,
+      ContextRepositoryInterface $context_repository,
     ) {
-    $this->entityTypeManager = $entity_type_manager;
-    $this->groupRouteContext = $group_route_context;
+    $this->contextRepository = $context_repository;
   }
 
   /**
@@ -48,29 +36,17 @@ class CurrentGroupService {
    *
    * @return \Drupal\group\Entity\GroupInterface|null
    *   The current group or NULL.
-   *
-   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
-   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
   public function fromRunTimeContexts(): ?GroupInterface {
-    $runtime_context = $this->groupRouteContext->getRuntimeContexts([]);
-    if (isset($runtime_context['group']) === FALSE) {
+    $group_runtime_context = $this->contextRepository->getRuntimeContexts(['@group.group_route_context:group']);
+
+    $group = $group_runtime_context['@group.group_route_context:group']->getContextData()->getValue();
+    if ($group === NULL) {
       return NULL;
     }
 
-    $group = $runtime_context['group']->getContextData()->getValue();
-    if ($group instanceof GroupInterface) {
-      return $group;
-    }
-
-    if (is_int($group) === TRUE) {
-      /** @var \Drupal\group\Entity\GroupInterface $loadedGroup */
-      $loadedGroup = $this->entityTypeManager
-        ->getStorage('group')->load($group);
-      return $loadedGroup;
-    }
-
-    return NULL;
+    assert($group instanceof GroupInterface, "The group context resolver returned a context value that is not a GroupInterface instance which violates the services contract.");
+    return $group;
   }
 
 }
--- a/modules/social_features/social_group/src/Tests/Unit/CurrentGroupServiceTest.php
+++ b/modules/social_features/social_group/src/Tests/Unit/CurrentGroupServiceTest.php
@@ -2,10 +2,8 @@
 
 namespace Drupal\Tests\social_group\Unit;
 
-use Drupal\Core\Entity\EntityStorageInterface;
-use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Plugin\Context\ContextInterface;
-use Drupal\Core\Plugin\Context\ContextProviderInterface;
+use Drupal\Core\Plugin\Context\ContextRepositoryInterface;
 use Drupal\Core\TypedData\TypedDataInterface;
 use Drupal\group\Entity\GroupInterface;
 use Drupal\social_group\CurrentGroupService;
@@ -20,142 +18,74 @@ use PHPUnit\Framework\MockObject\MockObject;
 class CurrentGroupServiceTest extends UnitTestCase {
 
   /**
-   * The mocked EntityTypeManagerInterface.
+   * The mocked context repository interface.
    *
-   * @var \Drupal\Core\Entity\EntityTypeManagerInterface|\PHPUnit\Framework\MockObject\MockObject
+   * @var \Drupal\Core\Plugin\Context\ContextRepositoryInterface|\PHPUnit\Framework\MockObject\MockObject
    */
-  private EntityTypeManagerInterface|MockObject $entityTypeManager;
-
-  /**
-   * The mocked ContextProviderInterface.
-   *
-   * @var \Drupal\Core\Plugin\Context\ContextProviderInterface|\PHPUnit\Framework\MockObject\MockObject
-   */
-  private ContextProviderInterface|MockObject $groupRouteContext;
+  protected ContextRepositoryInterface|MockObject $contextRepository;
 
   /**
    * The service under test.
    *
    * @var \Drupal\social_group\CurrentGroupService
    */
-  private CurrentGroupService $currentGroupService;
+  protected CurrentGroupService $currentGroupService;
 
   /**
-   * Set up the test case.
+   * {@inheritdoc}
    */
   protected function setUp(): void {
     parent::setUp();
+    $this->contextRepository = $this->createMock(ContextRepositoryInterface::class);
 
-    $this->entityTypeManager = $this->createMock(EntityTypeManagerInterface::class);
-    $this->groupRouteContext = $this->createMock(ContextProviderInterface::class);
-
-    $this->currentGroupService = new CurrentGroupService(
-      $this->entityTypeManager,
-      $this->groupRouteContext
-    );
+    $this->currentGroupService = new CurrentGroupService($this->contextRepository);
   }
 
   /**
-   * Helper function to create a mock context.
-   *
-   * @param mixed $value
-   *   The value returned by typeDataInterface.
-   *
-   * @return \Drupal\Core\Plugin\Context\ContextInterface
-   *   The updated mock context.
+   * Test that fromRunTimeContexts() returns a Group when a group is present.
    */
-  private function createContextWithValue(mixed $value): ContextInterface {
-    $typeDataInterface = $this->createMock(TypedDataInterface::class);
-    $typeDataInterface->expects($this->once())
+  public function testFromRunTimeContextsWithGroup(): void {
+    $group = $this->createMock(GroupInterface::class);
+    $this->mockContext($group);
+
+    // Call the method and assert the group is returned.
+    $result = $this->currentGroupService->fromRunTimeContexts();
+    $this->assertSame($group, $result, 'Group context should return a GroupInterface instance.');
+  }
+
+  /**
+   * Test that fromRunTimeContexts() returns NULL when group is not present.
+   */
+  public function testFromRunTimeContextsWithoutGroup(): void {
+    $this->mockContext(NULL);
+
+    // Call the method and assert NULL is returned.
+    $result = $this->currentGroupService->fromRunTimeContexts();
+    $this->assertNull($result, 'Group context should return NULL.');
+  }
+
+  /**
+   * Mock the context.
+   *
+   * @param \PHPUnit\Framework\MockObject\MockObject|\Drupal\group\Entity\GroupInterface|null $group
+   *   The mocked group or NULL.
+   */
+  private function mockContext(MockObject|GroupInterface|NULL $group): void {
+    $typedData = $this->createMock(TypedDataInterface::class);
+    $typedData->expects($this->once())
       ->method('getValue')
-      ->willReturn($value);
+      ->willReturn($group);
 
     $context = $this->createMock(ContextInterface::class);
     $context->expects($this->once())
       ->method('getContextData')
-      ->willReturn($typeDataInterface);
+      ->willReturn($typedData);
 
-    return $context;
-  }
-
-  /**
-   * Test that fromRunTimeContexts() returns NULL when no group context is set.
-   */
-  public function testFromRunTimeContextsNoGroupContext(): void {
-    $this->groupRouteContext
+    $this->contextRepository
       ->expects($this->once())
       ->method('getRuntimeContexts')
-      ->willReturn([]);
-
-    $group = $this->currentGroupService->fromRunTimeContexts();
-
-    $this->assertNull($group, 'No group context should return NULL.');
-  }
-
-  /**
-   * Test that fromRunTimeContexts() returns a GroupInterface when available.
-   */
-  public function testFromRunTimeContextsWithGroup(): void {
-    $group = $this->createMock(GroupInterface::class);
-    $context = $this->createContextWithValue($group);
-
-    $this->groupRouteContext
-      ->expects($this->once())
-      ->method('getRuntimeContexts')
-      ->willReturn(['group' => $context]);
-
-    $result = $this->currentGroupService->fromRunTimeContexts();
-
-    $this->assertSame($group, $result, 'Group should be returned when found in context.');
-  }
-
-  /**
-   * Test that fromRunTimeContexts() loads a group by ID.
-   */
-  public function testFromRunTimeContextsWithGroupId(): void {
-    $groupId = 1;
-    $group = $this->createMock(GroupInterface::class);
-
-    $context = $this->createContextWithValue($groupId);
-
-    $this->groupRouteContext
-      ->expects($this->once())
-      ->method('getRuntimeContexts')
-      ->willReturn(['group' => $context]);
-
-    $storage = $this->createMock(EntityStorageInterface::class);
-    $this->entityTypeManager
-      ->expects($this->once())
-      ->method('getStorage')
-      ->with('group')
-      ->willReturn($storage);
-
-    $storage
-      ->expects($this->once())
-      ->method('load')
-      ->with($groupId)
-      ->willReturn($group);
-
-    $result = $this->currentGroupService->fromRunTimeContexts();
-
-    $this->assertSame($group, $result, 'Group should be loaded by ID and returned.');
-  }
-
-  /**
-   * Test that fromRunTimeContexts() returns NULL when no valid group found.
-   */
-  public function testFromRunTimeContextsInvalidGroup(): void {
-    $invalidValue = 'invalid';
-    $context = $this->createContextWithValue($invalidValue);
-
-    $this->groupRouteContext
-      ->expects($this->once())
-      ->method('getRuntimeContexts')
-      ->willReturn(['group' => $context]);
-
-    $result = $this->currentGroupService->fromRunTimeContexts();
-
-    $this->assertNull($result, 'Invalid group value should return NULL.');
+      ->with(['@group.group_route_context:group'])
+      ->willReturn(['@group.group_route_context:group' => $context]);
   }
 
 }
--- a/modules/social_features/social_group/src/Tests/Unit/CurrentGroupServiceTest.php
+++ b/modules/social_features/social_group/src/Tests/Unit/CurrentGroupServiceTest.php
@@ -1,0 +1,161 @@
+<?php
+
+namespace Drupal\Tests\social_group\Unit;
+
+use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Plugin\Context\ContextInterface;
+use Drupal\Core\Plugin\Context\ContextProviderInterface;
+use Drupal\Core\TypedData\TypedDataInterface;
+use Drupal\group\Entity\GroupInterface;
+use Drupal\social_group\CurrentGroupService;
+use Drupal\Tests\UnitTestCase;
+use PHPUnit\Framework\MockObject\MockObject;
+
+/**
+ * Unit test for CurrentGroupService.
+ *
+ * @group social_group
+ */
+class CurrentGroupServiceTest extends UnitTestCase {
+
+  /**
+   * The mocked EntityTypeManagerInterface.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface|\PHPUnit\Framework\MockObject\MockObject
+   */
+  private EntityTypeManagerInterface|MockObject $entityTypeManager;
+
+  /**
+   * The mocked ContextProviderInterface.
+   *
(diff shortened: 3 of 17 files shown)


open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

modules/social_features/social_group/social_group.services.yml
```diff
@@ -49,3 +49,9 @@ services:
     arguments: []
     tags:
       - { name: cache.context }
+
+  social_group.current_group:
+    class: Drupal\social_group\CurrentGroupService
+    arguments:
```
denis-getopensocial, line 55: Another way to write arguments - i think it s more readable.

modules/social_features/social_post/social_post.module
```diff
@@ -15,45 +15,6 @@ use Drupal\group\Entity\GroupInterface;
 use Drupal\group\Entity\Group;
 use Drupal\user\UserInterface;
 
-/**
```
denis-getopensocial, line 18: Replaced hook.

modules/social_features/social_post/social_post.services.yml
```diff
@@ -11,3 +11,14 @@ services:
   social_post.helper:
     class: Drupal\social_post\Service\SocialPostHelper
     arguments: ['@entity_type.manager', '@current_user']
+
+
+  #hooks replacement
```
denis-getopensocial, line 16: Hook declaration. 
The `tags` part is used to be discovered by hux.

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,88 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Session\AccountProxyInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\hux\Attribute\Alter;
+use Drupal\social_group\CurrentGroupService;
+use Drupal\social_post\Service\SocialPostHelperInterface;
+
+
+final class SocialPostFormHooks {
```
denis-getopensocial, line 14: I initially wanted to have 1 class per hook replacement for better readability and maintenance but it will break the OS rule about not extra naming what we are doing. Because if I have 1 class per hook replacement the className would have the same name as the method.
IE: replacement of `social_post_form_post_form_alter`
ClassName `SocialPostFormPostFormAlterHook`, method called `SocialPostFormPostFormAlterHook`.
I'm really up for suggestion if we group hooks together or not in a class? (my preference is not) And then how do we name the class and the method?

I decided to name this class `SocialPostFormHooks ` meaning it will contains all hooks related to `form` but I initially wanted to have 1 class per hook

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,88 @@
...
+      $form['current_user_image'] = $content;
+    }
+
+    // Reset title display.
+    $form['field_post']['widget'][0]['#title_display'] = '';
+
+    // Set submit button caption to Post instead of Save.
+    $form['actions']['submit']['#value'] = t('Post', [], ['context' => 'Post button']);
+
+    if (empty($form['field_post']) || empty($form['field_post']['widget'][0])) {
```
denis-getopensocial, line 53: Clean Code Principles: return early

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,88 @@
...
+      return;
+    }
+
+    // Default value.
+    $form = $this->setFormTitleAndPlaceholder($form, t('Say something to the Community'));
```
denis-getopensocial, line 58: Clean Code Principles: Dont Reapeat Yourself

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * The method definition contains all standard parameters defined by the
+   * definition of the hook.
+   *
+   * @param array $form
+   * @param FormStateInterface $form_state
+   *
+   * @return array
+   */
+  #[Alter('form_post_form')]
```
denis-getopensocial, line 38: Hux use annotation to declare hooks.

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,88 @@
...
+  public function formPostFormAlter(array &$form, FormStateInterface $form_state): void {
+    $content = $this->socialPostHelper->buildCurrentUserImage();
```
denis-getopensocial, line 40: If we want to respect Clean Code principles, this method should be split into something like
```
$form = $this->buildUserImage($form);
$form = $this->setCaption($form);
$form =$this->setTitleAndPlaceholder($form);
```
Thoughts? Suggestion on how far CCP should be applied?

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Set form title and placeholder value.
+   *
+   * @param array $form
+   * @param TranslatableMarkup $displayedValue
+   *
+   * @return array
+   */
+  private function setFormTitleAndPlaceholder(array $form, TranslatableMarkup $displayedValue): array {
```
denis-getopensocial, line 82: Clean Code Principles: Use searchable and pronounceable value.

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

modules/social_features/social_group/src/CurrentGroupService.php
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * @return \Drupal\group\Entity\GroupInterface|null
+   *   The current group or NULL.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  public function fromRunTimeContexts(): ?GroupInterface {
+    $runtime_context = $this->groupRouteContext->getRuntimeContexts([]);
+    if (isset($runtime_context['group']) === FALSE) {
```
denis-getopensocial, line 57: Clean Code Principles:
- [Avoid negative conditionals](https://github.com/piotrplenik/clean-code-php?tab=readme-ov-file#avoid-negative-conditionals)

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,117 @@
...
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\hux\Attribute\Alter;
+use Drupal\social_post\Service\SocialPostHelperInterface;
+
+/**
+ * Replace hook: social_post_form_post_form_alter.
+ *
+ * @package Drupal\social_post\Hooks
+ */
+final class SocialPostFormHooks {
```
denis-getopensocial, line 17: Clean Code Principles: 
- [Prefer final classes](https://github.com/piotrplenik/clean-code-php?tab=readme-ov-file#prefer-final-classes)

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

phpstan-baseline.neon
```diff
@@ -9795,11 +9795,6 @@ parameters:
 			count: 1
 			path: modules/social_features/social_post/social_post.install
 
-		-
-			message: "#^Call to an undefined method Drupal\\\\Core\\\\Form\\\\FormInterface\\:\\:getEntity\\(\\)\\.$#"
-			count: 1
-			path: modules/social_features/social_post/social_post.module
-
```
denis-getopensocial, line 9802: Finally by replacing hooks we should be able to reduce this monster hidden under the carpet that contains plenty of non-compatible call that should be resolved.

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

modules/social_features/social_group/src/CurrentGroupService.php
```diff
@@ -0,0 +1,76 @@
...
+      return NULL;
+    }
+
+    $group = $runtime_context['group']->getContextData()->getValue();
```
Kingdutch, line 61: I think the context class provides you with a shorter way to do this that doesn't require traversing the data itself, by calling `getContextValue` directly on the context. Since that throws an exception if a required context isn't present (and group is required) we should check for `hasContextValue` first though.

```suggestion
    if (!$runtime_context['group']->hasContextValue()) {
      return NULL;
    }
    
    return $runtime_context['group']->getContextValue();
```

Since PHPStan likely can't entirely follow the types we can add an `assert` to help show the invariant.


```suggestion
    if (!$runtime_context['group']->hasContextValue()) {
      return NULL;
    }
    
    $group = $runtime_context['group']->getContextValue();
    assert($group instanceof GroupInterface, "The group context resolver returned a context value that is not a GroupInterface instance which violates the services contract".);
    return $group;
```

modules/social_features/social_group/src/CurrentGroupService.php
```diff
@@ -0,0 +1,76 @@
...
+    if ($group instanceof GroupInterface) {
+      return $group;
+    }
+
+    if (is_int($group) === TRUE) {
```
Kingdutch, line 66: There's one stylistic thing I disagree with and one functional issue in the code:

Stylistically: I'm unsure why the `=== TRUE` is added here since that's the default for if-statements. Especially since `is_int` returns an exact boolean and no truthy values. `if (is_int($group))` is less complex to view and equivalent so it should be preferred.

Functionally: This code will never be true. The group route service resolves the context and returns `\Drupal\group\Entity\GroupInterface|null` as type for the context service.

In the old code we would get the parameter from the route directly. That can be an integer in the case of views (which doesn't configure upcasting for its arguments). However, the group implementation converts that to `NULL` due to the following line of code in `GroupRouteContextTrait::getGroupFromRoute`: `if (($group = $route_match->getParameter('group')) && $group instanceof GroupInterface) {`.

Fixing that on our side is slightly more difficult because it means that things that are used in views (Actions/Blocks/Forms) should use the plugin context system or the form context system rather than getting the context from the route. That way they get the instance of the context that Views itself has converted into an entity. So basically anything that might need this service for this code-path, shouldn't use this service.

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
bramtenhove: My first instinct would be to split it in different classes. There will likely be cases where we it makes more sense to group them, e.g. when multiple hooks execute the same underlying code.

And I would ignore the rule about extra naming. Maybe it looks a bit weird, but while reading it you know what it does 🤷

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
bramtenhove: Splitting it up makes it easier to read and test.

It would be interesting to see the change in cyclomatic complexity too, I think it drops significantly.

modules/social_features/social_group/src/CurrentGroupService.php
```diff
@@ -0,0 +1,76 @@
...
+      /** @var \Drupal\group\Entity\GroupInterface $loadedGroup */
```
Kingdutch, line 67: A `@var` annotation should be needed here. There might be two reasons why you've added it:

1. PHPStan complained about having an `EntityInterface` rather than a `GroupInterface`. This requires updating the `phpstan.neon` configuration and providing an `entityMapping` for the group entity. That will make PHPStan Drupal smart enough to know that loading a group with the group storage is `GroupInterface|null` and not `EntityInterface|null`.
2. You're trying to suppress the `|null` part of a `GroupInterface|null` return. However, it's important we handle those null cases because they can occur in production. Especially in this case `NULL` would be a correct answer for a group that's provided in the URL but actually no longer exists, so removing the `@var` would provide better type information.

modules/social_features/social_group/src/Tests/Unit/CurrentGroupServiceTest.php
```diff
@@ -0,0 +1,161 @@
...
+  public function testFromRunTimeContextsWithGroupId(): void {
+    $groupId = 1;
+    $group = $this->createMock(GroupInterface::class);
+
+    $context = $this->createContextWithValue($groupId);
+
+    $this->groupRouteContext
+      ->expects($this->once())
+      ->method('getRuntimeContexts')
+      ->willReturn(['group' => $context]);
```
Kingdutch, line 124: This mock violates the contract of the group services' getRuntimeContexts` as explained in a previous comment.

modules/social_features/social_post/modules/social_post_photo/src/Plugin/Block/PostPhotoBlock.php
```diff
@@ -23,12 +25,14 @@ class PostPhotoBlock extends PostBlock {
    */
   public function __construct(
     array $configuration,
-    $plugin_id,
-    $plugin_definition,
+          $plugin_id,
+          $plugin_definition,
```
Kingdutch, line 29: Although this is a common IDE setting, I'd expect PHPCS to complain about this change 🤔

modules/social_features/social_post/social_post.services.yml
```diff
@@ -11,3 +11,12 @@ services:
   social_post.helper:
     class: Drupal\social_post\Service\SocialPostHelper
     arguments: ['@entity_type.manager', '@current_user']
+
+  #hooks replacement
+  social_post.form.hooks:
+    class: Drupal\social_post\Hooks\SocialPostFormHooks
+    arguments:
+    - '@social_post.helper'
+    - '@current_user'
```
Kingdutch, line 20: See my previous comment about Coding guideline [3.6 Use service autowiring and alias interfaces to concrete ID'd service definitions](https://open-social.slite.com/app/docs/dm052z21PPpHST#2e76f59c) :)

bramtenhove (commented):
@denis-getopensocial thanks a lot for the hard work!

I looked through all the code and I think this is definitely much better than the `form_alters` we've applied before. It's great to see additional unit testing is made possible and easier this way!

I left a few comments on files and your own remarks.

modules/social_features/social_post/modules/social_post_photo/src/Plugin/Block/PostPhotoBlock.php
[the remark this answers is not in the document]
denis-getopensocial: I think both are fine here.

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
Kingdutch: If I look at what the hook does in this case then it's preparing the post form to be personalised to the user and to the context?

What about `Drupal\social_post\Hooks\PostFormPersonalisation` as fully qualified name? It indicates it's doing something to the Post form and it's personalising. "Social" is omitted because that's included in the namespace already which IDEs also include in their class search.

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
Kingdutch: With where this if statement is positioned this will never be true anymore due to `$form['field_post']['widget'][0]['#title_display']`. Guard clauses should usually be at the top.

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
Kingdutch: I think this one is always hard and eventually boils down to "How can we make this easiest to understand". I think for me "setFormTitleAndPlaceholder" is more complicated than having the two assignment statements repeated a few times. At other places I'm already dealing with the `$form` array, so it's weird that a part of the code then can't deal with the `$form` array anymore but delegates it. That feels like it mixes multiple levels of abstraction. Usually in this case I point developers to https://kentcdodds.com/aha

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,118 @@
...
+   *   The form state.
+   */
+  #[Alter('form_post_form')]
+  public function formPostFormAlter(array &$form, FormStateInterface $form_state): void {
+    $content = $this->socialPostHelper->buildCurrentUserImage();
+    $form_object = $form_state->getFormObject();
+
+    if ($form_object instanceof ContentEntityForm === FALSE) {
+      return;
+    }
```
Kingdutch, line 67: I think if `$form_object` is not a `ContentEntityForm` then we have an implementation error right? So we probably want to make this an `assert` if we expect to catch this in testing or throw an actual exception if we think this might happen at run-time but shows an implementation error. The alternative would likely be a subtle and hard to debug bug (the post form not being altered) due to something that's wrong in an unrelated system (the post form no longer being a content entity form).

This if-statement also causes `$content` to be calculated but never used, so that statement should likely be after this guard clause.

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,118 @@
...
+
+    if ($form_object->getEntity()->isNew()
+      && $content !== NULL) {
+      $form['current_user_image'] = $content;
```
Kingdutch, line 71: `$content` is only ever used here so I would suggest inlining the variable. That has the advantage that if this is a post being edited we don't need to calculate `$content`.

modules/social_features/social_post/src/Hooks/SocialPostFormHooks.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    // $user_profile = $this->routeMatch->getParameter('user');
+    $user_profile = $form_state->get('recipientUser');
+    if ($user_profile !== NULL
+      && $user_profile->id() !== $this->currentUser->id()) {
+      $form = $this->setFormTitleAndPlaceholder($form, t('Leave a message to @name', [
+        '@name' => $user_profile->getDisplayName(),
+      ]));
+    }
```
Kingdutch, line 98: I wonder if we can re-arrange this logic to make sure we don't set the default value and then overwrite it but only set the default value if we haven't picked another value. In this case we know `setFormTitleAndPlaceholder` is a very light-weight function, but since it's abstracted, we really shouldn't need to know this and the function might be calculating pi or fetching expensive data from the database unnecessarily. We should not be afraid to use `else` if it makes sense :)

Kingdutch (commented):
Sorry this review took a while :D While I was writing this review, Bram also commented on some things so there may be some overlap :) 

I realise some of the things I'm reviewing might be debatable here, which is also why I think this PR might even be split into 3 PRs:

1. Add the hux dependency (that's already approved and non-controversial)
2. Rewrite the way to get the current group (we have a few smaller points there) which might even be used by the current hook implementation. It comes to light because we're using Hux but it's an improvement we may still want to make even if we wouldn't use Hux, so it's really an independent change
3. Rewrite the hook to Hux which is where the larger discussion is currently ongoing

I'm trying to get other developers to also think more about "What are the smallest independent changes that we can make".

modules/social_features/social_group/social_group.services.yml
```diff
@@ -49,3 +49,9 @@ services:
...
+      - '@entity_type.manager'
+      - '@group.group_route_context'
```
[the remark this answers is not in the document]
denis-getopensocial: We can create an Interface for every things we create but in that case I dont see the point. If i s required I'm happy to comply otherwise i think it s unnecessary artefact.

modules/social_features/social_group/src/CurrentGroupService.php
```diff
@@ -0,0 +1,76 @@
...
+   *   The entity type manager.
+   * @param \Drupal\Core\Plugin\Context\ContextProviderInterface $group_route_context
+   *   The group route context.
+   */
+  public function __construct(
+        EntityTypeManagerInterface $entity_type_manager,
+        ContextProviderInterface $group_route_context,
+    ) {
+    $this->entityTypeManager = $entity_type_manager;
+    $this->groupRouteContext = $group_route_context;
```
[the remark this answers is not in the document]
denis-getopensocial: tbh i was looking at how to, and the only way that was working was that.
I can try your approach.

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

modules/social_features/social_group/social_group.services.yml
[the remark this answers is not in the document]
Kingdutch: I agreed in my message that an interface may not be necessary here. We should still make sure that we use autowiring and define our services in a way so that this class can be used by another service that uses autowiring. That's what my suggested code snippet achieves.

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

modules/social_features/social_post/social_post.services.yml
denis-getopensocial: My understanding of AutoWiring was erroneous. 
Fixed

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

open-social-tugboat:

Link:
- https://pr4118-1i5zgqmcp3tuza7woeulwufzirnahp1f.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/670f775a8dfa8cf4259b702d

modules/social_features/social_group/social_group.services.yml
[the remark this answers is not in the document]
denis-getopensocial: Fixed

Kingdutch (approved):
Thanks for this Denis! Code looks good like this to me.

Once the pipelines are green this can be merged.

bramtenhove (approved):
Shout out to you @denis-getopensocial for moving this forward.

Not an easy process, but I do think this is really good! Congratulations 👏

modules/social_features/social_post/social_post.services.yml
```diff
@@ -11,3 +11,17 @@ services:
   social_post.helper:
     class: Drupal\social_post\Service\SocialPostHelper
     arguments: ['@entity_type.manager', '@current_user']
+
+  #hooks replacement
```
denis-getopensocial, line 15: the proper auto-wiring should looks like this _(credit Tiago)_

```
#hooks replacement
# use autowiring for arguments for `SocialPostFormHooks`
social_post.form.hooks:
  class: Drupal\social_post\Hooks\SocialPostFormHooks
  autowire: true
  tags:
  - { name: hooks }
# make `SocialPostFormHooks` injectable in another service (e.g. for decoration)
Drupal\social_post\Hooks\SocialPostFormHooks: '@social_post.form.hooks'

PR #4125, opened 2024-10-21
PROD-30967: Replace ModuleHandler usage with ModuleHandlerInterface

denis-getopensocial:
## Problem (for internal)

The PR to implement hux is too complex

## Solution (for internal)

Replace Modulehandler with ModuleHandlerInterface.

### DIFF
--- a/modules/custom/social_magic_login/src/Controller/MagicLoginController.php
+++ b/modules/custom/social_magic_login/src/Controller/MagicLoginController.php
@@ -5,7 +5,7 @@ namespace Drupal\social_magic_login\Controller;
 use Drupal\Component\Utility\Crypt;
 use Drupal\Core\Config\ConfigFactory;
 use Drupal\Core\Controller\ControllerBase;
-use Drupal\Core\Extension\ModuleHandler;
+use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Link;
 use Drupal\Core\Url;
 use Drupal\user\UserInterface;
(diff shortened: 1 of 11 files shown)


open-social-tugboat:

Link:
- https://pr4125-pju6nn4dcmatmcbjimbor2ftwa8d4rbg.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6715fa1a213f290841908ab9

open-social-tugboat:

Link:
- https://pr4125-pju6nn4dcmatmcbjimbor2ftwa8d4rbg.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6715fa1a213f290841908ab9

robertragas (approved):

Kingdutch:
Hey @denis-getopensocial and @robertragas,

Some feedback on this PR since I saw some uncertainty in Slack. 

Robert mentioned that tests might be needed. I would expect the functionality we're changing to be tested by our existing tests, but if we don't have that confidence, that might be good to document in the PR and we can decide to introduce test improvements separately or as part of the PR.

Most importantly I want to challenge the description of this PR. I think our goal in development should be to make small and incremental changes that are easy to comprehend and review. That's why I asked to split this change out into a separate PR rather than bundling it into a different change that depends on it. However, the change we're making doesn't exist because we're splitting it out, it exists because we weren't following Drupal best practices which we saw would lead to a problem.

I'd propose that this PR should have the following description instead.

> ## Problem
> Services in the Open Social distribution were typed to require the `ModuleHandler` class as arguments rather than the interface of that service (`ModuleHandlerInterface`). This can cause fatal errors in case another module tries to decorate the `ModuleHandler` service and provides a class that doesn't extend it.
>
> ## Solution
> As we don't actually need the original service class but are happy with any class that fulfills the role of the module handler, the `ModuleHandlerInterface` is actually what we care about and we should type the arguments as such.

Finally I'd like to caution for the unrelated changes to other properties that are being made in this PR (turning the PHPDoc untyped properties into strictly typed properties). While I understand we want to make this change eventually (and then remove the redundant `@var` annotation), it breaks any code that might extend a class and redeclare its properties. 

For example
```
class A {
    /**
     * @var string
     */
    protected string $foo;
}

class B extends A {
    protected $foo = "bar";
}
```
Produces the following
```
Fatal error: Type of B::$foo must be string (as in class A) on line 11
```

If the change is intentional then we should ensure it's documented in the PR that we don't expect this to happen (e.g. because we consider the classes internal and not part of our API) so that we have that information in case someone files a bug as a result.

PR #4190, opened 2024-11-20
PROD-31065: [IOSA-SUP] Reported content cannot be seen or deleted from the Report Overview Page

kashandarash:
## Problem (for internal)
Reported content cannot be seen or deleted from the Report Overview Page by a Site Admin role. Please see the attachment, there is no option to delete and also when clicked on the view comment it doesn't direct to the comment, the user stays at the page.  If we try to delete the reported content from the post itself, it gives the following message and do not delete.  {"message":"The "comment" parameter was not converted for the path "/ajax_comments/{comment}/delete" (route name: "ajax_comments.delete")"}  Reported contents can only be deleted from content overview page.

## Solution (for internal)
Rework the reports page to add a delete option, filters and fix errors on the view action.
The destination parameter was removed from action links because it breaks a redirect that exists on routes. 

Screenshot 2024-11-20 at 11 29 48

The Reports page (/admin/content/reports) was updated, direct link to remove content and filters by type and status were added.

## Issue tracker
https://getopensocial.atlassian.net/browse/PROD-31065

## How to test
- [x] Install Open Social
- [x] Report Content
- [x] Go to /admin/content/reports page
- [x] Filters and action links should work without errors

### DIFF
--- a/modules/social_features/social_content_report/config/install/views.view.report_overview.yml
+++ b/modules/social_features/social_content_report/config/install/views.view.report_overview.yml
@@ -8,16 +8,17 @@ dependencies:
     - flag.flag.report_node
     - flag.flag.report_post
     - system.menu.admin
-  enforced:
-    module:
-      - social_content_report
   module:
+    - better_exposed_filters
     - comment
     - flag
     - node
     - social_content_report
     - social_post
     - user
+  enforced:
+    module:
+      - social_content_report
 id: report_overview
 label: 'Report overview'
 module: views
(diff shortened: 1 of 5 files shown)


Orest-Solinskyi:
Tested! Reported issues were fixed. Also, the added improvements work well. Some FE works for button position is needed but it's a low prio issue now
Screenshot 2024-11-20 at 17 03 12

kashandarash:
> Tested! Reported issues were fixed. Also, the added improvements work well. Some FE works for button position is needed but it's a low prio issue now Screenshot 2024-11-20 at 17 03 12

Other admin pages have this issue as well, so I don't think we should fix it is the scope of this ticket 
Screenshot 2024-11-20 at 18 07 05

Kovalskiy266 (approved):
LGTM
Screenshot 2024-11-21 at 18 56 14
Screenshot 2024-11-21 at 18 53 17

PR #4201, opened 2024-11-25
PROD-30967: Add hux module as core dependency

denis-getopensocial:
## Problem (for internal)

As outlined in our decision record, we have identified the need to integrate Hux instead of relying on traditional hooks for managing interactions and events wit

## Solution (for internal)

We will migrate from the current hook-based implementation to Hux. The Hux library offers a more modern and modular approach for handling events, making the system easier to maintain and extend in the long term.
Blog post on how to: https://www.previousnext.com.au/blog/hux-alternative-to-hooks
Drupal seems to go the same way: https://www.drupal.org/node/3442349

## Change Record

Hook usage is becoming not recommended.
Hux annotation and services usage becomes the recommended way to implement a hook.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -147,6 +147,7 @@
         "drupal/graphql": "4.6.0",
         "drupal/graphql_oauth": "1.0.0-alpha3",
         "drupal/group": "2.3.0",
+        "drupal/hux": "1.6.0",
         "drupal/image_effects": "3.6.0",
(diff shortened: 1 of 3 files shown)


Kingdutch (approved):
Makes the module available in the project, enables it by default through a core dependency and ensures it's installed on existing sites. Looks good to me :D

PR #4204, opened 2024-11-26
Issue #3477793: Fix social_group_flexible_group_update_130006 update

rochek03:
## Problem (for internal)
The social_group_flexible_group_update_130006 was failing because of an already existing "verified" role.

## Solution (for internal)
The config name of the group role could be hashed if the length of the name is longer than 32 symbols, so we can't check configs by name, so we should check if such a role exists by GroupRole entity properties.

Internal.

## Issue tracker

- https://www.drupal.org/project/social/issues/3477793
- https://getopensocial.atlassian.net/browse/PROD-28608


ribel (approved):
Update hook was tested on site where it was failing before and now it works as expected.

Before:
image

After:
image

PR #4205, opened 2024-11-27
Issue #3490130: I want to see that a Book page is part of a Group

nkoporec:
## Problem (for internal)
When a topic or event are added to a Group, this information is visible in teasers and node page.
For book pages this information is missing.

## Solution (for internal)
Add general metadata info to book pages as well.

Updated the book display so it shows the author and group information on teaser and full view modes.

**Before:**
image
image

**After:**
image
image

## Issue tracker

- https://www.drupal.org/project/social/issues/3490130
- https://getopensocial.atlassian.net/browse/PROD-19343

## Theme issue tracker

Note: For teasers to work, you need this change:
- https://www.drupal.org/project/socialbase/issues/3490131#comment-15875719**

## How to test
1. Create a new topic and set it as part of the group
2. You should see the author information and the information that its part of a group below the title
3. Enable social_book and social_flexible_group_book modules
4. Create a new book and place it in a group
5. The info below the title about the author and a group is missing.

### DIFF
--- a/config/optional/node.type.book.yml
+++ b/config/optional/node.type.book.yml
@@ -19,4 +19,4 @@ description: '<em>Books</em> have a built-in hierarchical navigation. Use for ha
 help: ''
 new_revision: false
 preview_mode: 1
-display_submitted: false
+display_submitted: true
--- a/modules/social_features/social_book/social_book.install
+++ b/modules/social_features/social_book/social_book.install
@@ -113,3 +113,16 @@ function social_book_update_130000(): ?string {
   // Output logged messages to a related channel of update execution.
   return $updateHelper->logger()->output();
 }
+
+/**
+ * Enable author info on book pages.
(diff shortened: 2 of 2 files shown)


social.install
```diff
@@ -74,3 +74,18 @@ function _social_core_disable_secondary_gin_toolbar() {
 function social_update_11501() {
   _social_core_disable_secondary_gin_toolbar();
 }
+
+/**
+ * Enable author info on book pages.
```
ribel, line 79: @nkoporec I see we moved book-related config files here: https://github.com/goalgorilla/open_social/pull/1744

But anyway I think it makes more sense to add this update hook in the `social_book` module. Then we can also remove check if `moduleExists`.

ribel (approved):
Changes looks good and clear to me.
image

PR #4139, opened 2024-10-28
Issue #3476285: Improve usability of Taxonomy term creations

vcsvinicius:
## Problem (for internal)
Currently we have several taxonomy vocabularies and we display default options to SM.
Some of them are not relevant and should be hidden.

## Solution (for internal)
The fields are hidden using configuration, but don't config-modify or update-helper because some taxonomies doesn't have entity-form-display created.

So I created a logic to create entity-form-display when it doesn't exist or update when it already be created.

## Release notes (to customers)
Improved the user-experience to create taxonomy-terms, some unnecessary fields was hide.

## Issue tracker
[PROD-30250](https://getopensocial.atlassian.net/browse/PROD-30250)
[#3476285](https://www.drupal.org/project/social/issues/3476285)

## Theme issue tracker
N/A

## How to test
- [ ] Enable Social Content Report, Social Event Type, Social Event, Social Group Flexible, Social Profile Fields, Social Profile Organization Tag and Social Profile modules
- [ ] Access the which taxonimies mentioned below and check the fields as listed before

Taxonomies list:
- Content tags:
-- URL field visible

- Profile tag
-- URL field is not visible

- Event type:
-- URL field is not visible

- Report reason
-- URL field is not visible

- Topic type
-- URL field is not visible

- Nationality
-- URL field is not visible

- Interest
-- URL field is not visible

- Group type
-- URL field is not visible

- Expertise
-- URL field is not visible

- Organization
-- URL field is not visible

## Change Record
N/A

## Translations
N/A

[PROD-30250]: https://getopensocial.atlassian.net/browse/PROD-30250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4208, opened 2024-11-27
PROD-31405: Hide path field for nationality taxonomy

denis-getopensocial:
## Problem (for internal)

We need 

## Solution (for internal)

Hide path field for nationality taxonomy

Improvement of usability for taxonomy term nationality.

## Issue tracker

https://getopensocial.atlassian.net/browse/PROD-31405

## How to test

- turn on social_profile_fields
- go to admin/structure/taxonomy/manage/nationality/overview
- edit a term
- the field path is not shown.

### DIFF
--- a/modules/social_features/social_profile/modules/social_profile_fields/social_profile_fields.module
+++ b/modules/social_features/social_profile/modules/social_profile_fields/social_profile_fields.module
@@ -451,3 +451,11 @@ function social_profile_fields_preprocess_field(&$variables) {
     }
   }
 }
+
+/**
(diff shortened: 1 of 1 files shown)


vcsvinicius:
@denis-getopensocial When we discuss which approach would be used for it on PROD-30250, the Alexander adviced to do it by configuration, because the Distro is open source, so other installations can revert by configuration and avoid the necessity of code.

So I believe would be better use the same approach on https://github.com/goalgorilla/open_social/pull/4139

PR #4183, opened 2024-11-13
Issue #3487220: Fix user group list

zanvidmar:
## Problem (for internal)
In the database we have table "group_relationship_field_data" which is the data table for group content entities. In this table, we can see the columns type, group_type and plugin_id.

The pattern for the group membership type is usually <group_type>-<plugin>. The type (column) is limited to 32 characters, and for example closed_challenge-group_membership (33 characters) is just too long.

Group includes a way to generate names that fit the limits imposed by Drupal core, and it does this consistently in its own code. However, this means that where a <bundle>-<plugin> concatenation would exceed the limit, it instead converts it to 'group_content_type_' . md5($preferred_id) and truncates it. This is done in GroupRelationshipTypeStorage::getRelationshipTypeId. The problem is that our code is naively looking at the concatenation and doesn't handle the "too long" case, which is handled by the group module.

## Solution (for internal)
To solve the problem, we replaced the "type" condition in the database query with the "plugin_id" condition, because we do not need the relationship "type", which is unique per plugin_id and group_type. Since group_type is always the same per group entity (and we filter per entity), filtering by "plugin_id" instead of "type" gives the expected result.

## Release notes (to customers)
Fix user groups page to include all groups (group types with longer machine names were excluded).

## Issue tracker
[3487220](https://www.drupal.org/project/social/issues/3487220)

## Theme issue tracker
N/A

## How to test
- [ ] Install Open Social
- [ ] Go to /admin/group/types/add add group with machine name longer than 16 characters (like "Group with long name" -> "group_with_long_name")
- [ ] Add newly created group /group/add/group_with_long_name and become a member
- [ ] Go to /user/1/groups and you will not see your newly created group on the list
- [ ] Apply fix (and clear the cache)
- [ ] Go to /user/1/groups and you will see your newly created group on the list

Before:
Screenshot 2024-11-13 at 13 20 47

After
Screenshot 2024-11-13 at 13 22 06

## Change Record
N/A

## Translations
N/A

[PROD-31226]: https://getopensocial.atlassian.net/browse/PROD-31226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4212, opened 2024-11-28
Issue #3490022 - Revert #4183 from `12.4.x` to fix `Column not found: 1054 Unknown column 'gc.plugin_id'`

ronaldtebrake:
## Problem (for internal)
#4183 
Was cherry-picked to 12.4.x but contains fixes using Group version 2 changes, that isn't part of 12.4.x only 13.x so we can't cherry-pick this.

This caused:

```
Drupal\Core\Database\DatabaseExceptionWrapper: Exception in User Groups[groups]: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'gc.plugin_id' in 'where clause': SELECT "groups_field_data"."created" AS "groups_field_data_created", "groups_field_data"."label" AS "groups_field_data_label", "groups_field_data"."id" AS "id", "flagging_groups_field_data"."id" AS "flagging_groups_field_data_id" FROM "groups_field_data" "groups_field_data" LEFT JOIN "flagging" 
```
For those on 12.4.x

## Solution (for internal)
Revert the commit just from 12.4.x, it is valid for 13.x
I've reverted it as an additional commit, to ensure we can explain our reasoning in the history of commits, instead of completely removing it.

None, this is 12.4.x only.

## Issue tracker
https://www.drupal.org/project/social/issues/3490022

## How to test
Reinstall on 12.4.x with demo content and visit `/user/1/groups`

### DIFF
--- a/modules/social_features/social_group/src/Plugin/views/argument/UserUid.php
+++ b/modules/social_features/social_group/src/Plugin/views/argument/UserUid.php
@@ -60,7 +60,7 @@ class UserUid extends ArgumentPluginBase {
       $subselect2 = $this->database->select('group_content_field_data', 'gc');
       $subselect2->addField('gc', 'gid');
       $subselect2->condition('gc.entity_id', $this->argument);
-      $subselect2->condition('gc.plugin_id', '%' . $this->database->escapeLike('membership') . '%', 'LIKE');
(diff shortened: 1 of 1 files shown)


tbsiqueira (approved):

ronaldtebrake:
Behat tests for 12.4.x are not supported at the moment

PR #4219, opened 2024-11-29
PROD-31065: [IOSA-SUP] Reported content cannot be seen or deleted from the Report Overview Page

kashandarash:
## Problem (for internal)
Reported content cannot be seen or deleted from the Report Overview Page by a Site Admin role. Please see the attachment, there is no option to delete and also when clicked on the view comment it doesn't direct to the comment, the user stays at the page. If we try to delete the reported content from the post itself, it gives the following message and do not delete. {"message":"The "comment" parameter was not converted for the path "/ajax_comments/{comment}/delete" (route name: "ajax_comments.delete")"} Reported contents can only be deleted from content overview page.

## Solution (for internal)
Rework the reports page to add a delete option, filters and fix errors on the view action.
The destination parameter was removed from action links because it breaks a redirect that exists on routes.

The Reports page (/admin/content/reports) was updated, direct link to remove content and filters by type and status were added.

## Issue tracker
https://getopensocial.atlassian.net/browse/PROD-31065

## How to test
- [x] Install Open Social
- [x] Report Content
- [x] Go to /admin/content/reports page
- [x] Filters and action links should work without errors

**There was another PR https://github.com/goalgorilla/open_social/pull/4190 that was declined because the update path was not provided, the only difference now is hook update, so we need to check only this.**

### DIFF
--- a/modules/social_features/social_content_report/config/update/social_content_report_update_130003.yml
+++ b/modules/social_features/social_content_report/config/update/social_content_report_update_130003.yml
@@ -1,0 +1,150 @@
+views.view.report_overview:
+  expected_config: { }
+  update_actions:
+    add:
+      dependencies:
+        module:
+          - better_exposed_filters
+      display:
+        default:
+          display_options:
+            filters:
+              field_status_value:
+                id: field_status_value
+                table: flagging__field_status
+                field: field_status_value
+                relationship: none
+                group_type: group
+                admin_label: ''
+                plugin_id: boolean
+                operator: '='
+                value: '1'
+                group: 1
+                exposed: true
+                expose:
+                  operator_id: ''
+                  label: 'Report Status'
+                  description: ''
+                  use_operator: false
+                  operator: field_status_value_op
+                  operator_limit_selection: false
+                  operator_list: { }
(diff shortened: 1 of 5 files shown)


modules/social_features/social_content_report/config/update/social_content_report_update_130003.yml
```diff
@@ -0,0 +1,150 @@
+views.view.report_overview:
```
tbsiqueira, line 1: We can remove this file now

Orest-Solinskyi:
This PR was tested and [approved](https://getopensocial.atlassian.net/browse/PROD-31065?focusedCommentId=152525) on November 20

v-kovaliks (approved):
LGTM
Approve from my side, hook_update works correctly
Screenshot 2024-12-03 at 14 06 32
Screenshot 2024-12-03 at 14 06 24

robertragas:
I changed the milestone to 13.0.0-alpha19.

If it really needs to be in 12.3.x and 12.4.x, then it needs to be cherry-picked but also update_hook number should be adjusted accordingly to avoid conflicts.

PR #4220, opened 2024-11-29
Issue #3490695: Wrong users number on bulk email sending

nechai:
## Problem
In https://github.com/goalgorilla/open_social/pull/3905 we added the ability to prevent sending emails on bulk operation for groups and events if a user is unsubscribed from these email receiving.

On the group "manage members" page, the group manager can send an email collectively to selected users.

The problem is that if the group has a large number of members (> 100), then choosing the "select all" option, only a part of users will receive the email (this part is always equal to the number of pager).

## Solution
If select all users and send bulk email:
Screenshot 2024-11-29 at 18 23 15

_Before_:
Screenshot 2024-11-29 at 18 27 58

_After_:
Screenshot 2024-11-29 at 18 23 41

_Before_:
Screenshot 2024-11-29 at 18 29 38

_After_:
Screenshot 2024-11-29 at 18 24 14

## Release notes
Fix the wrong users number during bulk email sending

## Issue tracker
- https://www.drupal.org/project/social/issues/3490695
- https://getopensocial.atlassian.net/browse/PROD-31389

## How to test
- [x] Login as admin
- [x] Enable modules: `social_group_flexible_group`, `social_group_gvbo`, `social_email_broadcast`
- [ ] Create a flexible group
- [ ] Add more than 100 members to a group
- [ ] Try to send bulk email to all users on _/group/{gid}/membership_
**EB**: You should see correct numbers during email sending flow

### DIFF
--- a/modules/social_features/social_email_broadcast/social_email_broadcast.services.yml
+++ b/modules/social_features/social_email_broadcast/social_email_broadcast.services.yml
@@ -1,3 +1,9 @@
 services:
   Drupal\social_email_broadcast\SocialEmailBroadcast:
     arguments: ['@database']
+
+  Drupal\social_email_broadcast\ViewsBulkOperationsActionProcessorDecorator:
+    decorates: views_bulk_operations.processor
+    parent: views_bulk_operations.processor
+    decoration_on_invalid: ignore
+    public: false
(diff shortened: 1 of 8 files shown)


ribel (approved):
Changes were tested and works as expected now.

image

PR #4221, opened 2024-12-01
Issue #3490929: Add a redirect for profile canonical links to user information page

navneet0693:
## Problem (for internal)
In https://www.drupal.org/project/social/issues/3044337, we added a user experience in which we added a redirect from canonical user entity urls (/user/uid) to user stream page.

Even though we will hardly see the profile canonical urls being presented to end user on the platform, but we have certain edge cases where these URL may appear. For example, the profile entities are indexed with the profile canonical urls in social_user and social_all indexes. They may appear in searches if someone decides to add some custom search pages. Also, a user can simply visit /profile/profile-id which may appear broken as it is not intended user information page.

## Solution (for internal)
Since, we do not want to alter the canonical profile urls as it may break some default behaviors of profile entity, so we are adding a redirect to user stream page or to the page which set by site manager or admin as default user profile landing page at /admin/config/opensocial/user.

Clicking on the user profile result in search autocomplete results will now land you on user information page.

## Issue tracker
https://www.drupal.org/project/social/issues/3490929

## Theme issue tracker
N.A

## How to test
- [ ] Using latest version of Open Social
- [ ] Go to /profile/profile-id. You can use user ID in this url.
- [ ] You will land on a page which may appear broken
- [ ] Checkout to this branch
- [ ] Clear Cache.
- [ ] Go to /profile/profile-id
- [ ] You should end up on default user profile page

## Change Record
N.A

## Translations
N.A

### DIFF
--- a/modules/social_features/social_user/src/EventSubscriber/RedirectSubscriber.php
+++ b/modules/social_features/social_user/src/EventSubscriber/RedirectSubscriber.php
@@ -2,8 +2,8 @@
 
 namespace Drupal\social_user\EventSubscriber;
 
-use Drupal\Core\Routing\CurrentRouteMatch;
 use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Routing\CurrentRouteMatch;
 use Drupal\Core\Session\AccountProxy;
 use Drupal\Core\Url;
 use Drupal\user\Entity\User;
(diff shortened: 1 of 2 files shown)


Kingdutch:
This pull request has been moved to `migrate/issue/3490929-profile-canonical-redirect` in Cable Car. You can re-open the pull request there using: https://github.com/goalgorilla/cablecar/pull/new/migrate/issue/3490929-profile-canonical-redirect

PR #4226, opened 2024-12-03
Issue #3490982: Usernames in activity streams shouldn't be translatable

navneet0693:
## Problem (for internal)
The usernames of the people in the streams were getting translated when using an external translation tool like Google Translation. They shouldn't be translatable

The output in the stream is coming from message templates text ( https://github.com/goalgorilla/open_social/blob/main/modules/custom/activity_creator/src/ActivityFactory.php#L146  ) which is getting added on activity page in https://github.com/goalgorilla/open_social/blob/main/modules/custom/activity_creator/activity.page.inc#L38 which then gets added in https://git.drupalcode.org/project/socialbase/-/blob/2.6.x/templates/activity/activity.html.twig?ref_type=heads#L48

## Solution (for internal)
We added “no translate” attribute with message output text.

We smashed a bug where our usernames in the activity streams were getting translated when using third party services like Google Translate

## Issue tracker
https://www.drupal.org/project/social/issues/3490982

## Theme issue tracker
N.A

## How to test
- [ ] Install a multilingual Open Social website
- [ ] Create content in different language (other than English)
- [ ] Use module that provides third party translators to convert them in different languages
- [ ] This is tricky that which username or word it picks for translation, but observe the usernames in activity stream
- [ ] Checkout to this branch
- [ ] Run updates
- [ ] Try translating again
- [ ] You should observe that usernames will remain same.

## Change Record
N.A

## Translations
N.A

### DIFF
--- a/modules/social_features/social_activity/config/install/message.template.create_content_in_joined_group.yml
+++ b/modules/social_features/social_activity/config/install/message.template.create_content_in_joined_group.yml
@@ -24,13 +24,13 @@ label: 'Create a post, topic or event in a joined group'
 description: 'A person created a post, event or topic in a group I joined'
 text:
   -
-    value: "<p><a href=\"[message:revision_author:url:absolute]\">[message:revision_author:display-name]</a> created [social_group:created_entity_link_html] in the <a href=\"[message:gurl]\">[message:gtitle]</a> group [message:count_groups_per_node]</p>\r\n"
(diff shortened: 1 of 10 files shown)


Kingdutch:
This pull request has been moved to `migrate/issue/3490982-usernames-translations` in Cable Car. You can re-open the pull request there using: https://github.com/goalgorilla/cablecar/pull/new/migrate/issue/3490982-usernames-translations

@navneet0693 In addition to the move (since I'm looking at this anyway): As an architectural suggestion I would suggest to re-evaluate the scope. I don't think the problem is "Usernames shouldn't be translatable in the activity stream" I think it likely is "Profile data shouldn't be translatable" or at least "User display names shouldn't be translatable". Reframing it in that way may cause you to change the output of the username (and the `author:display-name` token) instead of all of our notifications. That provides a more scalable solution.

PR #4232, opened 2024-12-04
PROD-31497: Fix error while logging out the platform

tregismoreira:
## Problem (for internal)
In social_user_user_logout(), we used User::load() to load the current user entity. It turns out the previously imported class was `Drupal\social_user\Entity\User` instead of `Drupal\user\Entity\User`.

## Solution (for internal)
Properly import both classes.

N/A

## Issue tracker
https://getopensocial.atlassian.net/browse/PROD-31497

## Theme issue tracker
N/A

## How to test
- [ ] Checkout branch `bugfix/PROD-31497-error-logging-out`
- [ ] Login
- [ ] Logout

You should not see the exception.

## Change Record
N/A

## Translations
N/A


ribel (approved):
LGTM. I was ale to login and logout on a test env.

image

ribel:
@tregismoreira Can you also create an issue on Drupal.org for this, so we can reference it in the next release?

PR #4233, opened 2024-12-04
Issue #3491678: Fix UI issue of the add to calendar event popup

Agamich-FE:
## Problem (for internal)
1. If the title is too long, the last characters are ending under the X button and is not looking nice. Ideally the title is cut slightly shorter so that is all visible before the X
2. When I hover on the popup the shadow border disappears and the popup looks weird. ideally the border is always there, regardless if I am hovering or not

## Solution (for internal)
1. Improve popup title behaviour - add three dots in the end of the title if it is very long.
2. Add the same box shadow to the popup on the hover state.

1. The add to calendar popup has a box shadow on the hover state.
2. The title of the add to calendar popup has three dots in the end if it is very long.

## Issue tracker
https://www.drupal.org/project/social/issues/3491678

## How to test
- [ ] Create an Event page.
- [ ] Enroll on the created event page.
- [ ] You will see the popup.
- [ ] Hover on this popup.
- [ ] Add a long title using the inspector elements.


Agamich-FE:
Those changes will be added to the socialbase theme

PR #4236, opened 2024-12-06
PROD-31305: Non verified users are not able to see their profile properly

vcsvinicius:
## Problem (for internal)
Non verified user can't access they own profile information page.

## Solution (for internal)
I created a custom Access View for check the access for User Information page.

Fixed the access for non-verified user on Profile User Information page.

## Issue tracker
[PROD-31305](https://getopensocial.atlassian.net/browse/PROD-31305)

## Theme issue tracker
N/A

## How to test
- [ ] Create a non-verified user
- [ ] Log in with non-verified user
- [ ] Access page: /user/{user-id}/information

## Change Record
N/A

## Translations
N/A

[PROD-31305]: https://getopensocial.atlassian.net/browse/PROD-31305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/social_features/social_profile/src/Plugin/views/access/SocialProfilePagesAccess.php
+++ b/modules/social_features/social_profile/src/Plugin/views/access/SocialProfilePagesAccess.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace Drupal\social_profile\Plugin\views\access;
+
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Session\AccountProxyInterface;
+use Drupal\profile\Entity\ProfileInterface;
+use Drupal\views\Plugin\views\access\AccessPluginBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Routing\Route;
+
+/**
+ * Access plugin that provides access control to user information page.
+ *
+ * @ingroup views_access_plugins
+ *
+ * @ViewsAccess(
+ *   id = "profile_pages_access",
+ *   title = @Translation("Profile pages access"),
+ *   help = @Translation("Access to any profile page."),
+ * )
+ */
+class SocialProfilePagesAccess extends AccessPluginBase {
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected EntityTypeManagerInterface $entityTypeManager;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, EntityTypeManagerInterface $entity_type_manager) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+
+    $this->entityTypeManager = $entity_type_manager;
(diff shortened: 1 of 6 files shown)


modules/social_features/social_profile/src/Plugin/views/access/SocialProfilePagesAccess.php
```diff
@@ -0,0 +1,36 @@
...
+ *   help = @Translation("Access to any profile page."),
+ * )
+ */
+class SocialProfilePagesAccess extends AccessPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function access(AccountInterface $account): bool {
+    return TRUE;
```
Kingdutch, line 26: This is potentially dangerous. I know we have this in a few other places but we shouldn't implement it this way.

In the current situation our route check is enough to deny access to the page. However, this plugin may be used in other ways (like a views generated block). In that case suddenly no access checking is happening anymore because this is hardcoded to `TRUE` and route checking doesn't apply.

We should implement the proper access logic (get the profile from context, then `$profile->access('view', $account);`. If that's not possible for some reason we should default this to `FALSE` with a comment of why we set this default value. Then worst case we have another "Access denied" bug where it shouldn't be denied, rather than having a security issue by allowing access where it shouldn't be allowed :)

modules/social_features/social_profile/src/Service/SocialProfileAccessService.php
```diff
@@ -0,0 +1,93 @@
...
+      return AccessResult::allowedIfHasPermissions($account, ['view blocked user']);
+    }
+
+    // Check if current page is from logged user.
+    if ($account->id() === $user->id()) {
+      return AccessResult::allowedIfHasPermissions($account, ['view own profile profile']);
+    }
+
+    // Check if logged user can view all profiles.
+    return AccessResult::allowedIfHasPermissions($account, ['view any profile profile']);
```
Kingdutch, line 90: How does this logic differ from what would be evaluated if we called `$profile->access('view', $account);`?

What happens if other modules change when a user can be viewed by using Drupal's access APIs?

Kingdutch (changes requested):
Nice find on where the issue was coming from with the custom code in `social_user`.

There is a potential security issue in here and I think we should reconsider why we need `_custom_access` here.

BiaInacio:
Non-verified user is able to access their information:
image

modules/social_features/social_profile/src/Service/SocialProfileAccessService.php
vcsvinicius: I debugged with entity-access instead of custom-access and I found where and why the isn't give access.

The validation happen here: html/core/lib/Drupal/Core/Entity/EntityAccessCheck.php:65
The code try to found a parameter with profile, but the view has user as parameter, so to work we need to add this on view.

image

robertragas:
This pull request has been moved to `[bugfix/PROD-31305/fix-access-to-own-profile-non-verified-user](https://github.com/goalgorilla/cablecar/pull/2411/)` in Cable Car.

PR #4242, opened 2024-12-10
Issue #3476285: Hide path field from Nationality taxonomy

vcsvinicius:
## Problem (for internal)
The path field should be hide after: https://github.com/goalgorilla/open_social/pull/4139, but the Nationality taxonomy continue with this field.

## Solution (for internal)
Create a hook-update to hide path field

## Issue tracker
Improved the user-experience to create taxonomy-terms, some unnecessary fields was hide.

## Theme issue tracker
N/A

## How to test
- [ ] Enable Social Profile Field module
- [ ] Create a term at Nationality taxonomy and check if path field is not available

## Change Record
N/A

## Translations
N/A


denis-getopensocial (approved):

BiaInacio:
URL/path field is not displayed for nationality taxonomy add term page!
image

PR #4248, opened 2024-12-12
Issue #3493522: Album still available in stream post when disabled

volodymyr-sydor:
## Problem (for internal)

When album feature is disabled, the existing and new album is still available when a picture is posted in the general stream, user stream, group stream:

6c4f0710-752d-4d4a-8108-eb8de066b8f2

## Solution (for internal)

- Hide field_album in the post form when feature is disabled on the stream pages (general stream, user stream, group stream)

- Fixed issue with displaying albums in the post when feature is disabled

## Issue tracker

- https://www.drupal.org/project/social/issues/3493522
- https://getopensocial.atlassian.net/browse/PROD-31434

## Theme issue tracker

N/A

## How to test
- [ ] Using the latest version of Open Social with the social_album module enabled
- [ ] As a sitemanager
- [ ] Disable album feature on the `/admin/config/opensocial/album` page
- [ ] Try add a post on the /stream page (see: screen above)
- [ ] When adding a post I expect do not see a list of albums but instead it still available.

## Change Record

N/A
## Translations

N/A

### DIFF
--- a/modules/social_features/social_album/social_album.module
+++ b/modules/social_features/social_album/social_album.module
@@ -363,13 +363,20 @@ function social_album_form_post_form_alter(&$form, FormStateInterface $form_stat
     }
   }
   elseif (isset($form['field_album'])) {
-    $form['field_album']['#states'] = [
-      'visible' => [
-        ':input[name="field_post_image[0][fids]"]' => [
-          'filled' => TRUE,
+    // Hide album select field when feature is disabled on the image post form.
+    $status = \Drupal::config('social_album.settings')->get('status');
+    if (!$status) {
+      unset($form['field_album']);
+    }
+    else {
+      $form['field_album']['#states'] = [
+        'visible' => [
+          ':input[name="field_post_image[0][fids]"]' => [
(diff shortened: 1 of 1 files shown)


modules/social_features/social_album/social_album.module
```diff
@@ -363,13 +363,20 @@ function social_album_form_post_form_alter(&$form, FormStateInterface $form_stat
...
   }
   elseif (isset($form['field_album'])) {
-    $form['field_album']['#states'] = [
-      'visible' => [
-        ':input[name="field_post_image[0][fids]"]' => [
-          'filled' => TRUE,
+    // Hide album select field when feature is disabled on the image post form.
+    $status = \Drupal::config('social_album.settings')->get('status');
+    if (!$status) {
+      unset($form['field_album']);
```
nechai, line 369: In general, when we need to hide form elements from the UI, it is better to use the `hide()` function rather than `unset()` as there could be cases in which we can deal with default values in form submissions, etc.

nechai (approved):
Looks good to me 👍

PR #4255, opened 2024-12-16
Issue #3494301: Event address format should be displayed in a different format

vcsvinicius:
## Problem (for internal)
In the event page we show the event address.
The format is different than the one we show on group pages and we display the Country code instead of the country name.

Correct format and order of the fields should be: Location • Street, zip code city, country (full name)
Note that there is no comma between the city and the zip code when both values are provided

## Solution (for internal)
I created a display-plugin to address type-field and it will format the address as a site-manager user had defined on admin pages.

Two new admin pages was created with different permissions.
1. Page to define the address format;
2. Page to define regex pattern from address (This page is more technical and should be used only with knowledge);

The address field can be controlled how they will be rendered to the end-user in event page and that format can be controlled by site-manager.

## Issue tracker
[PROD-30807](https://getopensocial.atlassian.net/browse/PROD-30807)
[#3494301](https://www.drupal.org/project/social/issues/3494301)

## Theme issue tracker
N/A

## How to test
- [ ] Log in as site-manager
- [ ] Go to Address Format page "/admin/config/opensocial/address-settings/format" and check if exist an template register
- [ ] Create an event
- [ ] Check address field on header, it should follow the template registered on Address Format page
- [ ] Change the Address Format
- [ ] Clear cache
- [ ] Check if format changed on Event page as well
- [ ] If the new template register to address has a new pattern, example a value between parentheses, it should be registered on Regex Pattern page by an admin user

## Change Record
N/A

## Translations
N/A

[PROD-30807]: https://getopensocial.atlassian.net/browse/PROD-30807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cspell.words.txt
+++ b/cspell.words.txt
@@ -25,3 +25,5 @@ strikethrough
 Stringable
 # Address region in behat tests.
 Lvivska
+# Address label from address module
+Neighbourhood
--- a/modules/social_features/social_core/config/install/social_core.address.settings.yml
+++ b/modules/social_features/social_core/config/install/social_core.address.settings.yml
@@ -1,0 +1,2 @@
+format: '@address_line1%, @address_line2%, @address_line3%, @postal_code% @locality%, @country_code%'
+regex_pattern: '/@[\s\S]+?%, |@[\s\S]+?% |@[\s\S]+?% - /'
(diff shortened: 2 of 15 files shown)


vcsvinicius:
I am closing this pull-request and opened a new one to fit with new structure: https://github.com/goalgorilla/cablecar/pull/2560

PR #4326, opened 2025-07-09
Issue #3477972 add primary key

harlor:
See https://www.drupal.org/project/social/issues/3477972


epapaniko:
Just a heads up, I think the `hook_schema()` changes might not take effect on fresh installs. 

- `activity_creator_schema()` and `activity_send_email_schema()` have no "primary key" array key
- `activity_send_schema()` uses "primary keys" (plural)

Also, minor thing, the docblocks on `activity_send_update_13000()` and `activity_send_email_update_13000()` both reference the `activity_notification_status` table when they act on different ones.

modules/custom/activity_creator/activity_creator.install
```diff
@@ -37,6 +39,10 @@ function activity_creator_schema() {
       'ans_uid_aid' => ['uid', 'aid'],
       'ans_status' => ['status'],
     ],
+    [
```
harlor:
```suggestion
    'primary key' => [
```

modules/custom/activity_send/modules/activity_send_email/activity_send_email.install
```diff
@@ -58,6 +60,10 @@ function activity_send_email_schema() {
       'uad_uid' => ['uid'],
       'uad_frequency' => ['frequency'],
     ],
+    [
```
harlor:
```suggestion
    'primary key' => [
```

modules/custom/activity_send/activity_send.install
```diff
@@ -45,6 +47,11 @@ function activity_send_schema() {
       'uas_message_template' => ['message_template'],
       'uas_frequency' => ['frequency'],
     ],
+    'primary keys' => [
```
harlor:
```suggestion
    'primary key' => [
```

harlor:
Oh man I made quite a mess :D Do my suggestions look good?

But the hook_schema should run on new installs AFAIK.

epapaniko:
Yes, what I meant is that even though `hook_schema()` takes effect on new installs, if the keys are missing or different than expected, they will be silently ignored.

Great, looks good now, thanks!

